PR #4, opened 2021-03-18
Docker setup for Django + Postgres

kgodey:


README.md
```diff
@@ -1,3 +1,26 @@
...
+```
+You should now have a web server and database server running. Opening `http://localhost:8000` in your browser will open the application.
+
+You run database migrations like so:
+```
+docker exec -it mathesar_web_1 python manage.py migrate
+```
+You can also open a shell in the container running Django using:
+```
+docker exec -t mathesar_web_1 bash
```
mathemancer, line 21:
```suggestion
docker exec -it mathesar_web_1 bash
```
I don't remember exactly why the `-i` flag is necessary here, but the terminal you get won't show any output to your commands otherwise.

README.md
```diff
@@ -1,3 +1,26 @@
...
+
+From this directory, run:
+```
+docker-compose up
+```
+You should now have a web server and database server running. Opening `http://localhost:8000` in your browser will open the application.
+
+You run database migrations like so:
+```
+docker exec -it mathesar_web_1 python manage.py migrate
```
mathemancer, line 17:
```suggestion
docker exec mathesar_web_1 python manage.py migrate
```
No need for an interactive terminal here, I think.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: "3.9"
+   
+services:
+  db:
+    image: postgres
+    environment:
+      - POSTGRES_DB=mathesar
+      - POSTGRES_USER=mathesar
+      - POSTGRES_PASSWORD=mathesar
```
mathemancer, line 9:
```suggestion
      - POSTGRES_PASSWORD=mathesar
    ports:
      - "5432:5432"
```
This will let psql see the DB.

mathemancer: I think we should also add a volume for data persistence.  I usually prefer one in some out-of-the-way place in the repo, and `.gitignored`, since it makes it easy to manually intervene if you want.

requirements.txt
```diff
@@ -0,0 +1,4 @@
+alembic==1.5.7
+Django==3.1.7
+psycopg2==2.8.6
+SQLAlchemy==1.3.23
```
mathemancer, line 4: Please consider updating these to the newest versions before freezing.  In particular, SQLAlchemy would be better at `1.4.0`+ (some new features were added, which are being used)

mathemancer (changes requested):
I added a couple of fixes in suggestions.

README.md
kgodey: This was a typo, thanks!

docker-compose.yml
kgodey: Done

mathemancer (approved):
Nice!

PR #9, opened 2021-03-21
Add ignore directives for IDE configs

dhruvkb:
This PR adds `.gitignore` directives to ignore config files created by PyCharm (which I use) and VS Code (which is also quite popular).


mathemancer (approved):
Good catch, thanks

PR #15, opened 2021-04-03
Take in collection name as form input

eito-fis:
Add additional fields to `UploadFileForm` for the name of the collection and application to be stored, then pass it to `create_collection_from_csv` for use as the table and schema name.


kgodey:
@eito-fis this looks good, and it runs for me locally.

A couple of things to add here would be:
(1) Allow the user to pick an existing application from a dropdown instead of having to enter the application name every time. We should also allow the user to enter an application name if they want to create a new application.
(2) It would be useful to show which application a collection belonged to in the "View Existing Collections" section on the homepage. For example, you could change it to "View Existing Applications" and have a list of applications, and a list of collections under each application. 

Thanks!

kgodey:
@eito-fis this looks good! I agree with your comments on Matrix; this isn't the nicest UX for the application name, I think either a custom widget or a toggle approach would work, it's up to you what you'd want to pick.

mathesar/imports/csv.py
```diff
@@ -5,23 +5,26 @@
 from mathesar.models import Collection
 
 
+def get_application_name(cleaned_data):
```
kgodey, line 8: I don't think this function makes sense in the `csv.py` file since it's not related to CSV files specifically. I'd put it either in `forms.py` or make a `utils` file/folder.

kgodey (changes requested):

mathesar/imports/csv.py
eito-fis: Removed the function, as it is no longer needed with the `data_list` widget.

eito-fis:
Updated to use a custom widget for application name for a slightly nicer UI. Also made a new `forms` folder to accommodate the new `forms_widgets.py` file, unsure if this or a totally separate `widgets.py` folder would be preferred.

kgodey (changes requested):
This looks great, @eito-fis! I requested a couple of small changes. You're also missing the `__init__.py` inside the `forms` folder, which you'll need to make it a regular Python module.

kgodey (approved):
Looks good!

mathemancer (approved):

PR #16, opened 2021-04-05
Add Email type

mathemancer:
This is a PR that adds a function to create an `email` composite type in the DB, and the corresponding Column type in SQLAlchemy.  The SQLAlchemy type is automatically patched into the engine from `mathesar.database.base`


mathemancer:
The main problem keeping this as a draft is the brittle way that the input/output parsing functions are set up in the custom type.

mathesar/database/types/constants.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+TYPE_SCHEMA = "mathesar_types"
```
kgodey, line 3: I think you should use the `APP_PREFIX` from `base` instead of `mathesar_` here.

mathesar/database/types/email.py
```diff
@@ -0,0 +1,101 @@
+from sqlalchemy import text, create_engine, func
+from sqlalchemy.sql.functions import Function
+from sqlalchemy.types import UserDefinedType
+from mathesar.database.types import constants as c
+
+preparer = create_engine("postgresql://").dialect.identifier_preparer
```
kgodey, line 6: Could you add comments to this file describing what everything does/describes?

mathesar/database/types/constants.py
kgodey: Can't this just go in the `types.py` file? Or maybe in `__init__.py`?

mathesar/database/types/install.py
```diff
@@ -0,0 +1,8 @@
+from sqlalchemy import text
+from mathesar.database.types import constants as c
+
+def create_type_schema(engine):
+    query = f"""CREATE SCHEMA IF NOT EXISTS "{c.TYPE_SCHEMA}";"""
```
kgodey, line 5: We already have `create_schema` function in `mathesar/database/schemas.py`

mathesar/database/base.py
```diff
@@ -2,12 +2,19 @@
...
+from mathesar.database.types import types
 
 APP_PREFIX = mathesar_settings["APP_PREFIX"]
 ID = f"{APP_PREFIX}id"
 
 
-engine = create_engine(
+def create_engine_with_custom_types(*args, **kwargs):
+    engine = create_engine(*args, **kwargs)
+    engine.dialect.ischema_names.update(types.CUSTOM_TYPE_DICT)
```
kgodey, line 13: Can you add a comment to this line describing what it's doing?

mathesar/database/types/types.py
```diff
@@ -0,0 +1,5 @@
+from mathesar.database.types import email
```
kgodey, line 1: I think this would make sense in `__init__.py`

kgodey (commented):
I think writing some tests for this would be good. Also, it would be helpful to document the difference (in the code) between what's needed for any new type to be added and what's specific to the email type's implementation.

mathesar/database/types/constants.py
mathemancer: I think I'll put it in the `__init__.py` .

As for the `APP_PREFIX`, the way that `SQLAlchemy` handles quoting for user defined functions results in problems if the schema needs quoting (as our `APP_PREFIX` does).  I'm hoping to find a way around that, but if I can't it seemed preferable to have the schema name defined near where its used so that the problem isn't spread around (the problem being the need for a no-quotes-needed schema name)

mathesar/database/types/email.py
mathemancer: Sure, I'll also try to think of better variable names

mathesar/database/types/install.py
mathemancer: oops, I'll change that

mathesar/database/base.py
mathemancer: Sure, I can do that.

mathesar/database/types/types.py
mathemancer: Agreed

mathesar/database/types/constants.py
mathemancer: I ended up keeping the constants file (I expect it to grow), but managed to get `APP_PREFIX` working

kgodey (changes requested):
@mathemancer please fix flake8 errors and fix merge conflicts, I'll rereview after.

mathemancer:
> @mathemancer please fix flake8 errors and fix merge conflicts, I'll rereview after.

Some of the flake8 errors came from what appear to be auto-generated Django config files.  I went ahead and modified them to fix the errors, but put that in a separate commit in case we want to revert it and just ignore those files instead for linting.  Later, we might consider ignoring long lines in test files, since that's sometimes convenient for writing tests that make sure the string wrangling to avoid long lines in real files has produced the expected outcome.

mathemancer:
Converted back to draft; I noticed some more tests I want to write.

mathesar/database/schemas.py
```diff
@@ -1,24 +1,28 @@
 from sqlalchemy.schema import CreateSchema
+from sqlalchemy import inspect
 
-from mathesar.database.base import engine, inspector
+from mathesar.database.base import create_mathesar_engine
 
+engine = create_mathesar_engine()
```
kgodey, line 6: Why not import the engine from `base.py`?

mathesar/database/types/types.py
kgodey: I still think the contents of this file would make  sense in `__init__.py` instead of importing it into there.

kgodey (changes requested):
Looks good, @mathemancer. 

- For the tests, I recommend putting `__init__.py` in each folder.
- Re flake8, I'm going to make some configuration changes in https://github.com/centerofci/mathesar/issues/25.

Also, there's still one flake8 error.

mathesar/database/schemas.py
mathemancer: I think eventually, we'll use the function for anything that's dealing with the engine to connect to the user table DB (as opposed to the Django table DB).  I was thinking that the exported engine (exported from `base.py`, that is) would be for the Django tables.  So, I'm planning to have that engine be created via a different function (the one that takes args), and have the one that makes the mathesar engine give the appropriate args for the user table database connection.

kgodey: I don't think we'll be using the engine for the Django tables at all, we'll just use the Django ORM.

mathemancer: In that case, I think I'll leave it this way for the moment and reorganize when I split into separate DBs.  I do think the function version will be more flexible in the long haul, i.e., a given function call should know what connection it needs.  I suppose maybe it's premature while we only have one DB for user tables.

config/settings.py
```diff
@@ -91,16 +90,28 @@
 
 AUTH_PASSWORD_VALIDATORS = [
     {
-        "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
+        "NAME": (
+            "django.contrib.auth.password_validation"
```
kgodey, line 94: Could you put these back in a single string, this is harder to read, and the new flake8 rules shouldn't complain.

kgodey: Also applies to below

requirements-dev.txt
```diff
@@ -1,3 +1,6 @@
 flake8==3.9.0
 ipython==7.21.0
+isort==5.7.0
```
kgodey, line 3: Please remove isort

mathesar/tests/database/types/test_install.py
```diff
@@ -0,0 +1,34 @@
...
+    install.create_type_schema(engine)
+    with engine.connect() as conn:
+        res = conn.execute(text("SELECT * FROM information_schema.schemata"))
+    schemata = {row['schema_name'] for row in res.fetchall()}
+    assert constants.TYPE_SCHEMA in schemata
+
+
+@pytest.mark.django_db
+def test_create_type_schema_when_exists(engine):
+    install.create_type_schema(engine)
```
kgodey, line 33: I don't understand this test, why does it run the same thing twice without an assertion?

kgodey (changes requested):
Looks good overall, some minor changes requested

mathesar/tests/database/types/test_install.py
mathemancer: It's just making sure that the function doesn't throw an error if the type schema already exists.  This way, we can initialize if needed without needing to reflect first.

kgodey (approved):

PR #19, opened 2021-04-09
Wrote up MVP roadmap, changed theme.

kgodey:


docs/conf.py
```diff
@@ -44,7 +44,7 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'alabaster'
+html_theme = 'nature'
```
dhruvkb, line 47: Just out of curiosity, was there any particular reason to switch to 'nature'? Alabaster looked nice.

ghislaineguerin (approved):

ghislaineguerin:
@kgodey Looks good to me, although since it's mostly additions, I'm not sure if a more thorough review was necessary.

docs/conf.py
kgodey: I wanted more color.

docs/product/roadmap.rst
```diff
@@ -1,6 +1,237 @@
...
+- Allow user to change columns to this type
+- Add additional grouping options by value (yes/no)
+
+New Text Types
+--------------
+- Add new types:
+
+  - Email
+  - URL
+- Autodetect these types during import
```
mathemancer, line 90: I think there should be some agreed-upon minimum functionality to have a type be included in the UI.  Perhaps this should be part of that?

docs/product/roadmap.rst
```diff
@@ -1,6 +1,237 @@
...
+  - Date & Time, Date, Time support all grouping options supported by Postgres EXTRACT function.
+  - Duration: Range
+- Allow filtering using natural language for dates (e.g. "next month")
+
+Location Type
+-------------
+- Add new type, using existing PostGIS type where possible:
+
+  - Location
+- Autodetect this type during import
```
mathemancer, line 134: I think we may want to see what we can get by default from PostGIS for this, and extend if necessary.  It may be that not _all_ possible location inferences will be reasonable in the MVP

docs/product/roadmap.rst
```diff
@@ -1,6 +1,237 @@
...
+-----------------
+- Add new type
+ 
+  - Phone Number
+- Autodetect this type during import
+- Allow user to change columns to this type
+- Add grouping options:
+
+  - Country Code
+  - Area Code
```
mathemancer, line 164: We'll want to figure out how to generalize "area code" appropriately for different countries; we'd probably want some third-party package for that since it's quite different in different contexts (for example, German numbers don't all have the same number of digits in different parts).

docs/product/roadmap.rst
```diff
@@ -1,6 +1,237 @@
...
+
+Relationship Type
+-----------------
+Users should be able to:
+
+- Create a column that represents a relationship to another record (e.g. Book --> Author)
+- "Extract" a column from a table into a separate table (change the underlying schemas)
```
mathemancer, line 171: We can do sets of columns with what's already in the prototype.  It's no more difficult (at that layer) to do sets than single columns.  In fact, if I wanted an 'extract a single column' function done ASAP in the back end, it'd just be extract a set of one.

mathemancer (commented):
Regarding the roadmap, I think there's a _lot_ in there, and I'm not sure it's all necessary before getting something into users hands.  I think we should (at least) think hard about whether the PostGIS integration is necessary for the MVP, and also think hard about which visualizations are top priority.  TBH, we could probably have a _minimal_ viable product without any custom types at all.  This is not to say that I don't think all of the features in the roadmap are important, I just really want to get something in users hands ASAP.  

That is, unless we're thinking of having some alpha version in users hands before we reach the end of this roadmap.  In that case, my comment probably doesn't apply

kgodey:
@mathemancer I'm going to go ahead and merge this without making any changes. I anticipate there will be lots of revisions to the roadmap in the upcoming weeks and I'm not sure we can have a useful discussion of the content in the pull request comment format.

PR #20, opened 2021-04-12
Added potential features to docs.

kgodey:
These features are not ordered in any way.


docs/development/potential-features.rst
```diff
@@ -0,0 +1,117 @@
...
+- Excel web import (via API)
+- Apple Numbers file upload
+- Collabora import
+- Airtable
+
+Additional Fields
+-----------------
+- File field (for images, attachments, etc.)
+- IP Address field
+- Formula field (use of spreadsheet like formulas)
```
mathemancer, line 30: An intermediate step might be to give users the ability to make a generated column using SQL if they know how.

mathemancer (approved):
At some point we should consider adding inference and simulation features (from IQL) to the roadmap.  It would be pretty useful in some scientific and business use cases.  I think it makes sense to consider this document a constant WIP, though.

PR #24, opened 2021-04-14
Removed isort and black, updated flake8 workflow, added workflows rel…

kgodey:
…ated to GitHub Projects.


mathemancer (approved):
Do we need to be concerned about the failed action?

kgodey:
@mathemancer I don't think so, it should only run when issues and PRs are created, not for this PR.

silentninja:
Why was this removed, were there any issues with the integration? Wouldn't having isort and black make the linting consistent

kgodey:
@silentninja `black` and `isort` reformats _all_ your code automatically, which we didn't like. The people who work with the code are humans, I think generally other humans know better than machines how to write readable code.  Some of black's formatting ended up with code that was harder to parse than the manually written code it replaced. We decided to rely on `flake8` rules for things we thought made sense and perhaps a Python style guide on the wiki in the future.

PR #27, opened 2021-04-14
Issue and PR templates

kgodey:


ghislaineguerin (approved):
Looks good

mathemancer (approved):
I think these look good.  We may want to add specific feature-request-type templates at some point to reduce boiler plate around issues for, e.g., adding a new type.

PR #45, opened 2021-04-20
Extract mathesar_db library

### LINKED ISSUE #31: Reorganize to clearly separate webapp and database concerns
Django shouldn't really know more about a given user-table database than its connection info, and functions that don't need to know about Django should just take an engine as an argument.  Thus, we'll split functionality that manipulates user-defined tables out of the webapp to its own library which the webapp will import.  This will be easier to maintain in the long run.

mathemancer:
Fixes #31 

Everything that deals directly with the database holding user-defined tables (i.e., everything with SQLAlchemy in it) is now in the `mathesar_db` module.  Nothing in the `mathesar_db` module knows about or needs Django or any of the associated magic.

**Technical details**

`mathesar_db` is set up so that it does not need Django at all to function, and could eventually be extracted to an entirely different project (or used to create a non-webapp interface to a mathesar-enhanced DB).  The only caveat is that the tests use a settings module to pick up an initial DB connection string with super user privileges so that they can create a test db.  This use of the settings is confined to a single `conftest.py` file.

On the other hand, the web app now doesn't import any SQLAlchemy functionality at all (directly).  In particular, much of the table-handling logic of the `Collection` model was extracted so that the model doesn't need to have its own SQLAlchemy table when running.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_db/tables.py
```diff
@@ -0,0 +1,51 @@
...
+from mathesar_db import constants, schemas
+
+DEFAULT_COLUMNS = [
+    Column(constants.ID, Integer, primary_key=True),
+]
+
+
+def create_table(name, schema, column_names, engine):
+    """
+    This method creates a Postgres table corresponding to a collection.
```
kgodey, line 12: Please remove "corresponding to a collection" since we've decided not to use "collection" any longer.

mathesar/database/base.py
```diff
@@ -1,33 +1,12 @@
...
-ID = f"{APP_PREFIX}id"
-
-
-def create_engine_with_custom_types(*args, **kwargs):
-    engine = create_engine(*args, **kwargs)
-    # We need to add our custom types to any engine created for SQLALchemy use
-    # so that they can be used for reflection
-    engine.dialect.ischema_names.update(types.CUSTOM_TYPE_DICT)
-    return engine
+from mathesar_db import engine
```
kgodey, line 2: Could we just name this `db` or `database`? It seems redundant to call it `mathesar_db` when it's inside the `mathesar` project.

I think we should also rename `mathesar` the Django app to `web`, but I can do that when I'm renaming collections, etc.

mathesar/settings.py
```diff
@@ -1,8 +1,6 @@
 from django.conf import settings
 
-DEFAULTS = {
-    "APP_PREFIX": "mathesar_",
-}
+DEFAULTS = {}
```
kgodey, line 3: Remove this file, it's confusing to keep it around if we're not using it for anything.

mathesar_db/tests/conftest.py
```diff
@@ -0,0 +1,52 @@
+"""
```
kgodey, line 1: can we call this `base.py` since we're using `base.py` for similar setup files in other places?

mathesar_db/tests/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+from config.settings import DATABASES
+
+TEST_DB = "mathesar_db_test_database"
+
+
+@pytest.fixture(scope="session")
+def test_db():
+    superuser_engine = _get_superuser_engine()
+    with superuser_engine.connect() as conn:
+        conn.execute(text("COMMIT"))
```
kgodey, line 17: why use raw SQL here?

kgodey (changes requested):

kgodey:
Looks great, @mathemancer.

mathesar/database/base.py
mathemancer: I was trying to avoid confusion with `mathesar.database` .  I'll change it to `db` for the moment.

mathesar_db/tests/conftest.py
mathemancer: Naming it `conftest.py` gives us some `pytest` magic that automatically makes the fixtures defined in that file available to every test.  We'll lose that if we rename it.

mathesar_db/tests/conftest.py
mathemancer: For some reason, `conn.commit()` doesn't end an empty transaction.  The reason that's important is that you can't execute `DROP DATABASE` or `CREATE DATABASE` inside of a transaction block, but `engine.connect` always starts an implied transaction block in SQLAlchemy.  Admittedly, I didn't spend lots of time investigating _why_ `conn.commit()` doesn't seem to end the transaction block when it's empty, since it's just a test file.

mathesar_db/tests/conftest.py
kgodey: ah okay, I didn't realize it was a special `pytest` name, that's fine.

mathesar_db/tests/conftest.py
kgodey: Could you add that as a comment? also maybe a TODO to investigate?

kgodey (approved):
@mathemancer I approved it but could you add the comment I requested before merging?

Issue #38, opened 2021-04-19
API for Schemas


kgodey:
This issue covers a read-only API for list and detail views of schemas.

Previously, this issue was for a CRUID API, but I'm reducing the scope because:
- Updating involves renaming schemas, which involves tables, foreignkeys etc. and is complicated/deserves its own issue.
- Creation and deletion behavior should be driven by UI needs, which are yet to be defined.

kgodey:
This is done.

PR #50, opened 2021-04-22
Read only API for schemas

kgodey:
Related to #38

This pull request tracks schemas in the Django DB (to give them an ID), and creates a read-only API for them. Future PRs will expand on the API and add tests.

**Technical details**
Please note that this will require local DB nuking once merged, I did not want to create a new DB migration since the project is at a very early stage.

**Screenshots**

Screenshot_2021-04-22 Api Root – Django REST framework
Screenshot_2021-04-22 Table List – Django REST framework
Screenshot_2021-04-22 Schema List – Django REST framework

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models.py
```diff
@@ -18,13 +18,17 @@ def __str__(self):
         return f"{self.__class__.__name__}: {self.name}"
 
 
+class Schema(DatabaseObject):
+    database = models.CharField(max_length=63)
```
mathemancer, line 22: We may want more characters here to handle multiple hosts (in the case where the user-defined tables are on a separate host).

mathesar/imports/csv.py
```diff
@@ -22,8 +22,8 @@ def create_db_table_from_csv(name, schema, csv_reader, engine):
 
 def create_table_from_csv(name, schema, csv_file, engine=engine):
     csv_reader = get_csv_reader(csv_file)
-    table = create_db_table_from_csv(name, schema, csv_reader, engine)
-    table, _ = Table.objects.get_or_create(
-        name=table.name, schema=table.schema
-    )
+    db_table = create_db_table_from_csv(name, schema, csv_reader, engine)
+    database = engine.url.database
```
mathemancer, line 26: I think it would be better (if possible) to instead use the key associated with `settings.DATABASES` for the DB, so all the relevant connection info can be retrieved using the data in the model (and `settings.py`).  Or, we could store the whole connection string, but that would have sensitive info (the password).

mathemancer (changes requested):
I think we should sync up on how to handle the DB info for a given schema.  See my more specific comments.

mathesar/imports/csv.py
kgodey: That makes sense, I'll update it.

mathemancer (approved):
Looks great!

PR #84, opened 2021-04-30
Create & Update API for Records

### LINKED ISSUE #39: CRUD API for Records
Records in tables should be able to be created, read, updated, and deleted via the REST API.

kgodey:
Fixes #39 

- Added create record API and related test
- Added (partial) update record API and related test
- Changed the CSV file used in the tests
   - I had assumed the file we were using was copyright-free because it was a government produced file from data.gov, but I decided to check. Apparently only federal-government produced files are copyright-free. I could not find a license for the one we were using, and a different part of Fairfax County data website said their data was all rights reserved, so I decided to switch to a NASA patents CSV file instead.

**Technical details**
- Moved record related functions in `db` into their own file.
- I'm not implementing a `PUT` (full update) for the API since you can accomplish the same thing with `PATCH` and there doesn't seem to be a use case for completely replacing a record's attributes. I can implement it later if needed for the UI.
 
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/records.py
```diff
@@ -0,0 +1,55 @@
...
+def create_records(table, engine, record_data):
+    """
+    records can be a dictionary, tuple, or list of dictionaries or tuples.
+    """
+    id_value = None
+    with engine.begin() as connection:
+        result = connection.execute(table.insert(), record_data)
+        if result.rowcount == 1:
+            id_value = result.inserted_primary_key[0]
+    if id_value:
```
mathemancer, line 36: I'd prefer a check for `if id_value is None` to avoid a problem if the id happens to be zero.  While I realize PostgreSQL starts its integer incrementing at 1 under default settings, this function does allow for the submitted record to contain its own custom (integer) primary key value, which could be zero.

db/records.py
```diff
@@ -0,0 +1,55 @@
...
+        return result[0] if result else None
+
+
+def get_records(table, engine, limit=None, offset=None):
+    query = select(table).limit(limit).offset(offset)
+    with engine.begin() as conn:
+        return conn.execute(query).fetchall()
+
+
+def create_records(table, engine, record_data):
```
mathemancer, line 27: This function (through its call to `get_record` needs two connections (and two transactions) to insert a single record.  Getting those connections from the pool will bog this down under heavy load.

Why not just build the returned `Row` object (or named tuple) using the info in hand?  We'll know the insert was successful once we have the primary key in hand, and that's the only piece that we don't know beforehand.  Or, we might try to figure out how to batch multiple inserts for performance.

db/records.py
```diff
@@ -0,0 +1,55 @@
...
+    with engine.begin() as connection:
+        connection.execute(
+            table.update().where(primary_key_column == id_value).values(record_data)
+        )
+    return get_record(table, engine, id_value)
+
+
+def delete_record(table, engine, id_value):
+    primary_key_column = _get_primary_key_column(table)
+    query = delete(table).where(primary_key_column == id_value)
```
mathemancer, line 53: It occurs to me that this would be safer if `id_value` were globally unique.  Perhaps a future PR could make that so.

mathesar/imports/csv.py
```diff
@@ -28,4 +27,5 @@ def create_table_from_csv(name, schema, database_key, csv_file):
     database = get_database_key(engine)
     schema, _ = Schema.objects.get_or_create(name=db_table.schema, database=database)
     table, _ = Table.objects.get_or_create(name=db_table.name, schema=schema)
+    table.create_records([row for row in csv_reader])
```
mathemancer, line 30: Given my comment about the connections on the `create_records` function, this will be pretty slow for large CSVs.  We'll need to update the function to use the `COPY` function (probably via the `copy_expert` function of `psycopg2`) at some point.

mathemancer (changes requested):
Please address the comment about the `id` check (to be None rather than Falsy), then I'll re-review.

mathemancer:
Forgot to mention:  Good idea on splitting out the `records` module.

db/records.py
kgodey: Good catch.

db/records.py
kgodey: We can't ensure that it's globally unique if we're working with an existing database.

db/records.py
kgodey: This does batch multiple inserts, `get_record` is only called when there's a single insert. I put it under `engine.begin()` so there should only be one transaction/connection now.

mathesar/imports/csv.py
kgodey: We'll only be doing `get_record` if there's a single record being created. I renamed the `create_records` function, restructured it a bit to combine connections, and added some comments to clarify.

mathemancer (approved):
Okay, looks good to me.

PR #85, opened 2021-04-30
Add Function to move columns from Table A to Table B

### LINKED ISSUE #81: Move columns between tables
**Problem**

While data modeling, it may be that a user would want to be able to move a set of columns from Table A to Table B, where these tables are connected by a foreign key relationship.

**Proposed solution**

We should provide this functionality, with the following restrictions (for now):
- The tables involved must be connected by a foreign key.
- The columns being moved cannot be refered to by a foreign key, or be a foreign key column
- The columns being moved cannot contain the primary key of a table.

mathemancer:
Fixes #81

This PR adds a function that, given 2 tables A and B connected by a Foreign Key relationship, moves a list of columns of the users choice from A to B, or from B to A.

**Technical details**

The resulting tables may change in row count, since when moving the column we may need to increase the number of rows in the target table according to the number of distinct tuples defined by the new column set.  Also, the source table may reduce in row count according to the number of distinct rows in the reduced column set remaining.

## Checklist
 
  index.md`).
- [ ] TODO I added tests for the changes I made (if applicable).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/tables.py
+++ b/db/tables.py
@@ -213,10 +213,112 @@ def merge_tables(
             select(merged_columns, distinct=True).select_from(merge_join)
         )
         conn.execute(insert_stmt)
-        if drop_original_tables:
-            table_one.drop()
-            table_two.drop()
+    if drop_original_tables:
+        table_one.drop()
+        table_two.drop()
     return merged_table
+
+
+def move_column_between_related_tables(
+        source_table_name,
+        target_table_name,
+        column_names,
+        schema,
+        engine,
+):
+    TEMP_MERGED_TABLE_NAME = f"{constants.MATHESAR_PREFIX}_temp_merge_table"
+    source_table = reflect_table(source_table_name, schema, engine)
+    target_table = reflect_table(
+        target_table_name, schema, engine, metadata=source_table.metadata
+    )
+    relationship = _find_table_relationship(source_table, target_table)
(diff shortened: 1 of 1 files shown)


db/tables.py
```diff
@@ -213,12 +213,119 @@ def merge_tables(
...
+            table_two.drop(bind=engine)
+        else:
+            table_two.drop(bind=engine)
+            table_one.drop(bind=engine)
+
     return merged_table
 
 
+def move_column_between_related_tables(
+        source_table_name,
```
kgodey, line 229: We might need to change the function signature later, since we will already have reflected table objects. I don't think that needs to be fixed in this PR, though, we can update it when we actually use the function.

kgodey (approved):

db/tables.py
mathemancer: I actually think we should go rather the other way.  These sorts of DDL functions will be safest, and be easiest to chain together if they take only strings (or lists of strings) to refer to tables, columns, schemas, etc.  I mostly actually regret returning the actual table object,  since it'll make these harder to compose.

PR #127, opened 2021-05-14
Controlled casting

### LINKED ISSUE #126: Controlled Casting
**Problem**

In the course of implementing #92 , I've realized that the default casting in PostgreSQL doesn't quite serve our needs.  For example, many different strings cast to booleans, but this loses information.  Another example:  Numeric 1s and 0s _can't_ cast to boolean.  When casting `NUMERIC` to `INTEGER` values are silently rounded.  And so on.

**Proposed solution**

We need to define functions for all supported casting to be used in altering column types and column inference.

**Additional context**

This will probably involve a number of PRs and sub-issues.

mathemancer:
Fixes #126 

This adds a number of DB functions, and replaces the column alteration logic to use them.  These functions give us more precise control over what is done when casting one type to another (mostly to make things safer).  

**Technical details**

This requires knocking out your Mathesar DB and reinstalling, since we still don't have upgrade functionality working.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/types/alteration.py
```diff
@@ -0,0 +1,209 @@
...
+BOOLEAN = "boolean"
+EMAIL = email.QUALIFIED_EMAIL
+INTERVAL = "interval"
+NUMERIC = "numeric"
+TEXT = "text"
+VARCHAR = "varchar"
+
+
+def get_supported_alter_column_types(engine):
+    dialect_types = engine.dialect.ischema_names
```
kgodey, line 13: Datetimes should be fairly simple to add too, right?

db/types/alteration.py
```diff
@@ -0,0 +1,209 @@
...
+
+def install_all_casts(engine):
+    create_boolean_casts(engine)
+    create_email_casts(engine)
+    create_interval_casts(engine)
+    create_numeric_casts(engine)
+    create_varchar_casts(engine)
+
+
+def create_boolean_casts(engine):
```
kgodey, line 59: Could you document what this function (and other similar functions) are doing in comments?

kgodey (approved):
@mathemancer I added some comments but I'm approving to unblock merge. If I'm around when you're done making changes, feel free to ask me to rereview before merge.

db/types/alteration.py
mathemancer: I'd rather leave those for a separate issue / PR, since there'll be some complexity around dates vs times vs datetimes, and TZs (how to decide whether a column of times should be imported with the local TZ, assuming UTC, etc.)

mathemancer: I know we've discussed the plan w.r.t. TZs, but I think having dates/times/datetimes separated out in a well-documented PR might benefit future code archaeologist.

db/types/alteration.py
mathemancer: Good idea.

mathemancer: I made them docstrings to make it visible using the `help` function.  I also considered documenting them in comments in the `PL/pgSQL` functions themselves instead, but I think folks are more likely to be able to (or want to) read the python docstrings than comments in the SQL.  Do you agree?

kgodey: Yes, agreed.

db/types/alteration.py
kgodey: Makes sense.

kgodey (approved):
Looks great, @mathemancer

PR #135, opened 2021-05-19
Infer column type two

### LINKED ISSUE #92: Type Inference 2: Determine a type for a column
**Problem**

Different types in Mathesar will enable different operations; for example, strings could be aggregated by concatenating, but numeric types could be aggregated by summing or multiplying. So far, while we can reflect different types, we have no way to determine the type most appropriate for a column.

**Proposed solution**

Using the function implemented for #91 , we should then test the column against types in an algorithmic way to determine the best type prediction we can for the column

mathemancer:
Fixes #92 

This adds a function to infer the type of the column by attempting to alter the column to various types in a logical order.

**Technical details**

This has the side effect of losing data in the case where the column is altered to a type via some non-injective mapping.  For example, altering the strings `'0123'` and `'123'` to a numeric will result in the value `123` in both cases.  It's not possible to recover the original strings from the resulting value.

We'll add mechanisms to recover the original version of an altered column in future PRs.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/tests/types/test_inference.py
+++ b/db/tests/types/test_inference.py
@@ -1,6 +1,6 @@
 import pytest
 from sqlalchemy import Column, MetaData, Table
-from sqlalchemy import BOOLEAN, String, NUMERIC, VARCHAR
+from sqlalchemy import BOOLEAN, Numeric, NUMERIC, String, VARCHAR
 from sqlalchemy.schema import CreateSchema, DropSchema
 from db.engine import _add_custom_types_to_engine
 from db.types import base, inference, install
@@ -34,6 +34,8 @@ def engine_email_type(temporary_testing_schema):
 
 
 type_data_list = [
+    (Numeric, [0, 2, 1, 0], NUMERIC),
+    (Numeric, [0, 1, 1, 0], BOOLEAN),
     (String, ["t", "false", "true", "f", "f"], BOOLEAN),
     (String, ["t", "false", "2", "0"], VARCHAR),
     (String, ["a", "cat", "mat", "bat"], VARCHAR),
--- a/db/tests/types/test_inference.py
+++ b/db/tests/types/test_inference.py
@@ -1,0 +1,73 @@
+import pytest
+from sqlalchemy import Column, MetaData, Table
+from sqlalchemy import BOOLEAN, String, NUMERIC, VARCHAR
+from sqlalchemy.schema import CreateSchema, DropSchema
+from db.engine import _add_custom_types_to_engine
+from db.types import base, inference, install
+
+TEST_SCHEMA = "test_schema"
+
+
+@pytest.fixture
+def engine_with_types(engine):
+    _add_custom_types_to_engine(engine)
+    return engine
+
+
+@pytest.fixture
+def temporary_testing_schema(engine_with_types):
+    schema = TEST_SCHEMA
+    with engine_with_types.begin() as conn:
+        conn.execute(CreateSchema(schema))
+    yield engine_with_types, schema
+    with engine_with_types.begin() as conn:
+        conn.execute(DropSchema(schema, cascade=True, if_exists=True))
+
+
+@pytest.fixture
+def engine_email_type(temporary_testing_schema):
+    engine, schema = temporary_testing_schema
+    install.install_mathesar_on_database(engine)
+    yield engine, schema
+    with engine.begin() as conn:
+        conn.execute(DropSchema(base.SCHEMA, cascade=True, if_exists=True))
+
+
+type_data_list = [
+    (String, ["t", "false", "true", "f", "f"], BOOLEAN),
+    (String, ["t", "false", "2", "0"], VARCHAR),
+    (String, ["a", "cat", "mat", "bat"], VARCHAR),
+    (String, ["2", "1", "0", "0"], NUMERIC),
+]
+
+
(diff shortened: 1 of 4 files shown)


db/tests/types/test_inference.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.fixture
+def engine_with_types(engine):
+    _add_custom_types_to_engine(engine)
+    return engine
+
+
+@pytest.fixture
+def temporary_testing_schema(engine_with_types):
```
kgodey, line 18: You're using this fixture in multiple files, I recommend putting it in `conftest.py` in the root of this folder so that you can reuse it.

kgodey: Same with other reused fixtures.

db/tests/types/test_inference.py
```diff
@@ -0,0 +1,75 @@
...
+]
+
+
+@pytest.mark.parametrize("type_,value_list,expect_type", type_data_list)
+def test_type_inference(engine_email_type, type_, value_list, expect_type):
+    engine, schema = engine_email_type
+    TEST_TABLE = "test_table"
+    TEST_COLUMN = "test_column"
+    metadata = MetaData(bind=engine)
+    print(value_list)
```
kgodey, line 52: Is this a debug print?

db/types/inference.py
```diff
@@ -0,0 +1,77 @@
...
+                schema,
+                table_name,
+                column_name,
+                engine,
+                depth=depth + 1,
+                type_inference_dag=type_inference_dag,
+            )
+            break
+        except Exception:
+            logger.info(
```
kgodey, line 74: Shouldn't this be `logger.error`?

kgodey (approved):
Approving to unblock, please make requested changes before merging.

db/tests/types/test_inference.py
mathemancer: oops, good eye.

db/tests/types/test_inference.py
mathemancer: My reasoning was that these fixtures use code from mathesar that is tested by test files with _no_ access to these fixtures.  (This makes sure we're not using fixtures in a way that invalidates the code they use.  Do you know of a way to separate those out (or make it difficult/obvious when you _shouldn't_ use a given fixture for testing?  In particular, I'd want to find some way to guarantee that these fixtures won't be used for testing the `db/types/install.py` file at some point in the future.

kgodey: You could organize the files differently so that any tests that can use these fixtures are in a single directory with the appropriate `conftest.py` and files that shouldn't use these fixtures are in a different directory.

kgodey: that seems like the obvious solution to me but I might be missing something.

db/types/inference.py
mathemancer: This exception is an expected part of the control flow (it's how we know when a given type is inappropriate for a given column).  I think `info` is reasonable for that.  I _do_ think I should remove the spot where I set the log level to `INFO` in this file, though (it should be set by the importer if they want a non-default level).  What do you think?

db/tests/types/test_inference.py
mathemancer: We'd either need to break the "test file directory structure mirroring the real file directory structure" convention, or reorganize the real files in a (to me) less logical way, then mirror it.  I'm okay with the former, I guess, but I tend towards letting code repeat in test files over obfuscating what the files are testing (though it's not too hard to figure out at this stage in the project).

mathemancer: We could also put these fixtures in a non-default file (i.e., not a `conftest.py` file), then import it.  It's kind of non-standard, but I think it'd be less confusing than partially mirroring the directory structure of the tested files with the structure of the test files.

mathemancer: Actually, I think that's probably my favorite solution.

mathemancer: For that kind of reorganization, I think it'd merit its own PR to document what was done and why with the restructuring.  @kgodey, what do you think?

kgodey: @mathemancer I think importing a non-standard file makes sense, and doing it in a separate PR also makes sense. I was originally thinking you'd reorganize the real files as well, but I like your solution too.

mathemancer: I'll double-check that the organization of the real files makes sense and adjust accordingly.

db/types/inference.py
kgodey: Can you catch the specific exception you're expecting and add a comment explaining it's the expected flow?

I agree re: removing log level in the file.

mathemancer: Unfortunately, there's a bunch of different exceptions based on which type was casting to which other.  I can try to figure out if they're sub-classes of something appropriate (I think python will catch the parent class in that case, I'll have to check)

mathemancer: Okay, we'll use `sqlalchemy.exc.DatabaseError`.  It's not super-precise, but it at least restricts to "you asked to do something and the DB said no".

PR #151, opened 2021-05-25
add logic to filter records into groups

### LINKED ISSUE #143: Get filtered set of records from table
**Problem**

At the moment, when we request records from a table, it's _all_ of the records or none.  We should be able to filter a way that supports retrieving "groups" of records, where a group is defined as a set of records where some subset of the columns match a set of values, equivalent to the SQL:

```sql
WHERE col1 = val1 AND col2 = val2 AND ... AND coln = valn
```

**Proposed solution**

We should create a function that lets us filter string columns using the pattern above, and another function that retrieves the distinct tuples for a set of columns (i.e., the groups in this context).

**Additional context**

The interesting bit will be figuring out how to paginate the results, but without having to reperform the (costly) filtering query each time.

mathemancer:
Fixes #143 

This PR adds the method `get_distinct_tuple_values` to the `db.records` module that gets distinct tuple values to be used as "groups" from a given table.  It also adds a `filters` argument to the `get_records` method that uses the output of the new method to produce a group of records corresponding to a given distinct tuple value.

**Technical details**

The return type of the new method feeds directly into the new `filters` argument of the `get_records` function (this is the intent).  This means that it's possible to first call `get_distinct_tuple_values`, then use its output directly as input to the `filters` argument of the `get_records` method.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (approved):
Looks good!

PR #174, opened 2021-06-03
Create Table from Data File API endpoint

### LINKED ISSUE #146: API to create a table from a data file
**Problem**

Currently (after #141 is merged), creating a data file automatically creates a table, and that happens synchronously. Uploading a file should be a completely separate operation from converting a file to a table.

**Proposed solution**

A table creation API that takes a `data_file` as input. The flow via the API will then be:
- upload a data file
- create a table using that data file.

We should delete the code that creates a table automatically when a data file is uploaded.

eito-fis:
Fixes #146

Adds an endpoint to the API to create a `Table` from a `Data File`. 

**Technical details**
Adds an additional endpoint at `/api/v0/tables` which takes a POST request with a `file_pk` argument in the request body. `file_pk` is used to look up a data file to create the new table from.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Not entirely sure what validation should look like here. It seems like normally you depend on a `ModelSerializer` to do validation for POST requests, but because our body is unrelated to a model this doesn't seem workable. Should we make a `TableCreateSerializer`? Or would manual validation be preferred?

mathesar/views/api.py
```diff
@@ -21,11 +22,18 @@ def create(self, request):
...
     serializer_class = TableSerializer
     pagination_class = DefaultLimitOffsetPagination
 
+    def create(self, request):
+        file_pk = request.data["file_pk"]
+        data_file = DataFile.objects.get(id=file_pk)
+        table = create_table_from_csv(data_file)
+        serializer = TableSerializer(table, context={'request': request})
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
```
mathemancer, line 36: I suggest moving all logic out to some intermediary function that just takes the `request.data` blob, and then "does the right thing".  We'll want this `create` method to be quite generic and have the ability to handle multiple creation methods (not just from a CSV on the webserver disk).

mathemancer (commented):
I think this makes sense, if we presuppose the idea that we want one `/blah/blah/tables` endpoint that accepts POSTS of a number of different formats that define different table creation methods.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -76,6 +95,18 @@ def test_table_detail(create_table, client):
     check_table_response(response_table, table, table_name)
 
 
+def test_table_create(create_table, client, data_file, test_db_name):
```
kgodey, line 98: Please add a test case for a creation failure as well (or more than one).

mathesar/views/api.py
```diff
@@ -21,11 +22,18 @@ def create(self, request):
...
 
 
-class TableViewSet(viewsets.ReadOnlyModelViewSet):
+class TableViewSet(viewsets.GenericViewSet, ListModelMixin, RetrieveModelMixin):
     queryset = Table.objects.all().order_by('-created_at')
     serializer_class = TableSerializer
     pagination_class = DefaultLimitOffsetPagination
 
+    def create(self, request):
+        file_pk = request.data["file_pk"]
```
kgodey, line 31: Tables have a reverse relation to data files. I think you should use the key that's used for the reverse relation here so that there's some relation to the model.

You might also want to add data file IDs to the default table serializer.

mathesar/views/api.py
kgodey: Agreed

mathesar/views/api.py
```diff
@@ -21,11 +22,18 @@ def create(self, request):
...
+        data_file = DataFile.objects.get(id=file_pk)
```
kgodey, line 32: You'll need to handle the case where there's no data file with that ID.

kgodey (changes requested):
Overall, this looks good. For validation, you might be able to use the `TableSerializer` if you use the reverse relation to the data file, if not, you'll need to implement it manually. If you raise the DRF `ValidationError`, it should provide the appropriate output. See https://www.django-rest-framework.org/api-guide/exceptions/

mathesar/views/api.py
eito-fis: I messed around a bit with validating using the reverse relation, but ended up getting stuck. In particular, when creating a table from a data file (I assume) we don't care about non-`data_file_pk` fields, so the validation logic gets strange. It seems like we could write custom `.validation()` and `.create()` methods that implement the logic we use outside the serializer right now, but I'm not sure that's preferred.

Although, thinking about it now, there might be cases where we want to combine `data_file_pk` with other fields , we could specifying a new name or schema? Unsure what the long term plans for the table creation API is - maybe something for another PR.

kgodey: Now that I think about it, I think it's probably best to specify the name and schema in this API (and remove the schema from the data file model entirely). It made sense to specify the schema in the data file when we were immediately converting it into a table because we needed that information, but now that table creation is a separate step, the name and schema should be specified here.

mathesar/models.py
```diff
@@ -81,7 +81,9 @@ class DataFile(BaseModel):
         upload_to=model_utils.user_directory_path,
         validators=[FileExtensionValidator(allowed_extensions=['csv'])]
     )
-    table_imported_to = models.ForeignKey(Table, blank=True, null=True, on_delete=models.SET_NULL)
+    table_imported_to = models.ForeignKey(Table, related_name='data_file',
+                                          blank=True, null=True,
+                                          on_delete=models.SET_NULL)
     schema = models.ForeignKey(Schema, blank=True, null=True, on_delete=models.SET_NULL)
     user = models.ForeignKey(User, blank=True, null=True, on_delete=models.CASCADE)
```
kgodey, line 89: For some reason, I can't comment on the actual line I want to, but you don't need the `save` method here at all if you remove the `schema` from this model.

mathesar/views/api.py
```diff
@@ -21,11 +22,18 @@ def create(self, request):
...
+class TableViewSet(viewsets.GenericViewSet, ListModelMixin, RetrieveModelMixin,
+                   CreateModelMixin):
     queryset = Table.objects.all().order_by('-created_at')
     serializer_class = TableSerializer
     pagination_class = DefaultLimitOffsetPagination
 
+    def create(self, request):
+        if "data_file_pk" in request.data:
+            return create_table_from_datafile(request)
+        else:
```
kgodey, line 34: I don't think we want to support other types of table creation right now, since it will just create an empty `Table` object in the Django DB that isn't backed by any underlying table in the `mathesar` DB.

kgodey (changes requested):
Please also see my comment about removing the `schema` from the DataFile model and making name and schema required parameters for table creation.

mathesar/views/api.py
eito-fis: Got it - I'll probably add the reverse link back in and use the model serializer for validation then.

eito-fis:
Made changes:
* `DataFile` no longer takes a `schema` argument, no longer stores `table_imported_to` and no longer has a custom `.save()` method. Tests updated to reflect changes.
* `Table` now has an additional, optional `data_file` foreign key field. The goal here is to have a many to one relation, where each table imports from one data file, but each data file can import to many tables.
* Added migration for above model changes.
* The `create_table_from_csv`, `create_db_table_from_data_file` and `create_table_from_datafile` now take name and schema as arguments instead of pulling from data file. Tests updated to reflect changes.

Smaller things:
* `TableSerializer` changed from a `HyperlinkedModelSerializer` -> `ModelSerializer`, as relations are easier to handle when they're just primary keys. Can figure out how to make the hyperlinked version work if the format is important.
* `create_table` function in tests now returns the table it created. Hopefully this doesn't break anything, can revert if it didn't return for a specific reason.

eito-fis:
Side note - is worth merging migrations together? I could clean them up if needed.

kgodey:
@eito-fis don't worry about migrations for now, we'll reset them before we do our first release.

kgodey (approved):
Looks good @eito-fis!

PR #204, opened 2021-06-07
Allow users to upload TSV files

### LINKED ISSUE #62: Allow users to import TSV files
**Problem**
Currently, the user can only import `csv` files into Mathesar. `tsv` files are pretty similar and the user should be able to import them too.

**Describe the solution you'd like**
The user should be able to upload a `tsv` file to the `api/v0/data_files/` API endpoint.

eito-fis:
Fixes #62

Allows for TSV files to be uploaded as data files, and for tables to be made from TSV data files.

**Technical details**
Adds an optional `delimiter` argument to `create_records_from_csv()` that is then passed to inserted into Postgres's `COPY` command. Also modifies `get_csv_reader()` and `create_db_table_from_datafile()` as appropriate to handle arbitrary delimiters. Finally, adds a `get_sv_delimiter()` function that infers a file's delimiter based on the file extension.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Two questions about possible changes to this PR:
1. Would it useful to support arbitrary file delimiters, specified when the data file is created? We could support this with a `delimiter` field on the data file model, but it would complicate the implementation a bit (would probably require custom validation as passing a file with a recognized extension or passing `delimiter` are both valid) .  Happy to add if it would be useful.
2. Is there a proper way to implement parametrized fixtures when we're modifying a test DB? The tests I added aren't very nice in that they duplicate a ton of code, just changing `csv_filename` -> `tsv_filename`, which doesn't seem ideal. I did try parameterizing the `csv_filename` fixture with both the `.csv` and `.tsv` extensions, but this broke all tests that used hardcoded table names (as it would try to build the same table twice). Not sure what the desired solution here would be.

kgodey:
1. For the scope of this PR, just let's limit support to TSVs and CSVs. It's great that the function supports arbitrary delimiters and we can make use of that in the future, but we shouldn't complicate the frontend with offering that option yet since we are trying to keep things simple for the first version.

2. Maybe take a look at the create_table fixture as an example? It takes a name and uses an inner function to create a different table every time. I'm not sure if that's what you're looking for.

eito-fis:
Re: 1, sounds good.
Re: 2, The problem is more that I want to reuse a set of tests, only changing a single input (`patents.csv` -> `patents.tsv`). I think the normal way to do this in pytest is by parametrizing a fixture (in our case, the `csv_filename fixture`), which automatically re-runs the tests multiple times. This seems ideal, but when trying it I get errors due to the table being created for the test already existing from an earlier run of the same test.

I am a bit surprised by this behavior as I can create a table with the same name in two different tests without problem, so I'm not entirely sure why it breaks when using parameters. I'll take a crack at debugging it later today.

eito-fis:
After poking around a bit more, it doesn't seem like there will be an easy way to do what I want given the DB persists over the entire session. I'll go ahead and leave the tests as they are and open for review. We could reduce the duplication of the `test_data_file_create()` and `test_table_create_from_datafile()` tests by adding [a library to parametrize by fixtures](https://pypi.org/project/pytest-lazy-fixture/) (or just straight up parametrizing by hard coded paths) - I'm happy to do either if people think it would be better.

mathemancer:
@eito-fis I use the strategy of burning down the schema (not the DB) between any tests that create tables in the `db` module tests.  I'm not sure how well that will work with the Django tests setup, but it might be worth considering.

mathesar/imports/csv.py
```diff
@@ -35,19 +35,31 @@ def legacy_create_table_from_csv(name, schema, database_key, csv_file):
...
 
 
+def get_sv_delimiter(filename):
+    if filename.endswith('.tsv'):
+        return '\t'
+    elif filename.endswith('.csv'):
+        return ','
+    else:
+        # Should never be hit because we only allow csv and tsv extensions
+        raise NotImplementedError("File extension not supported!")
```
mathemancer, line 45: This will be pretty brittle with real-world files of any sort.  In particular, the extension isn't standardized at all; many CSV-like files are created via excel export, for example, and those get a `.txt` extension by default when they're not comma-separated (comma-separated gets `.csv` as expected).  LibreOffice always uses `.csv` by default for any text-file table with some delimiter.  Google Sheets has `.csv` and `.tsv`, but can't handle any other delimiters.  Etc.

I think we should either infer the file type by actually looking for reasonable delimiters, or take it as input to the API in the request.

mathesar/models.py
```diff
@@ -79,9 +79,8 @@ def delete_record(self, id_value):
 class DataFile(BaseModel):
     file = models.FileField(
         upload_to=model_utils.user_directory_path,
-        validators=[FileExtensionValidator(allowed_extensions=['csv'])]
+        validators=[FileExtensionValidator(allowed_extensions=['csv', 'tsv'])]
```
mathemancer, line 82: See above comment about why this isn't a great test of validity.  It may be that we want to defer solving it to a later PR, but I think it's not too difficult to handle sooner.

mathemancer (changes requested):
I think we either need to come up with a better way to determine CSV vs. TSV, or take which it should be as input.  Alternatively, we can do what Excel (or libreoffice) do and let the user preview with different delimiter choices, then take the one they choose as the delimiter for import.

mathesar/models.py
kgodey: We can add `.txt` here, but I don't think we should allow arbitrary file extensions.

For `.txt` files, we can parse and determine the delimiter or let users know that it's invalid.

mathemancer: But file extensions _are_ arbitrary.  I've also seen:  `.text`, `.TEXT`, `.TSV`, `.dat`, etc. (or just no extension at all) in the wild for tabular text files, and these are all with "clean" data.  We'd at least need to accept any reasonable modification of any of these extensions (and probably others) to avoid people being confused as to why their valid files are being rejected.  In my experience with such files, the extensions are sometimes useful for heuristics (i.e., to direct the algo to try certain file-types preferentially), but not validation, since they're not actually tied to the file contents in any meaningful way.  Much of the software I use basically ignores file extensions, and I think that's a good thing.  (For fun try renaming an `.odt` file to `.pdf` and opening it in LibreOffice.  It opens it happily as an ODT file, and saves edits to the file with a `.pdf` extension)  I think that the actual validation should be whether any of the delimiters we support end up resulting in tabular data (or just asking the user for a delimiter and checking whether that gives tabular data).

It might make sense to blacklist a few extensions (e.g., `.pdf`) that are strongly associated with file types we don't support, but even that seems less robust to me than just testing the content to see if we can make a table from it.

kgodey: Okay, I'm convinced that we can ignore file extensions.

My other concern is that we're reading whatever file people upload and we need to have some way of ensuring that we're safe from malicious file uploads. If we're going to read the file to figure out what the delimiter is, we need to make sure we do it securely. I'm not currently sure what that would entail.

mathemancer: That's a great point about security, and one that I think we should tackle regardless of extension.  I'm not at all convinced that we're secure against SQL injection, or any other type of attack.  (I don't think there's any checking of files imported using `COPY` for injections, and I wouldn't want to roll our own if we're parsing the files in python).

I'm also not sure what that would entail.

kgodey: @eito-fis perhaps you could do some research on how to ensure security with reading user-uploaded files?

eito-fis: Some quick googling makes it seem like just reading files shouldn't be too much of an issue?  [Django's section on security for user uploaded files](https://docs.djangoproject.com/en/3.2/topics/security/#user-uploaded-content-security) mostly talks about how to ensure they're safe to serve to other users, and [stack overflow thinks that it's safe as long as you just open in read mode](https://docs.djangoproject.com/en/3.2/topics/security/#user-uploaded-content-security).

For the SQL injection, it seems like were safe when the actual `COPY` statement is executing ([ref1](https://stackoverflow.com/questions/18156063/how-can-i-prevent-sql-injections-during-csv-uploads), [ref2](https://stackoverflow.com/questions/65328138/postgresql-copy-sql-injection)), but the raw input we use for the headers do seem sketchy. Might be worth sanitizing them?

(Also, sorry if SO isn't appropriate for security considerations - couldn't find anything in the documentation)

eito-fis:
Updated the PR to use python's built in `csv.Sniffer` object to infer relevant file settings. Some notes:
* With this approach, we can't really validate if a file is valid or not since the `Sniffer` will always do its best to find fits for the fields. I imagine if we leave it like this we would have to validate after the table is created on the user side.
```
>>> dialect = csv.Sniffer().sniff("testing")
>>> dialect.delimiter
't'
``` 
* We can constrain the types of delimiters allowed, so if we have an idea of what those might look like I can implement that and handle the error.
```
>>> dialect = csv.Sniffer().sniff("testing", delimiters=", |")
_csv.Error: Could not determine delimiter
```
* I tried the technique  @mathemancer pointed me towards, but ran into some problems with the `legacy_create_table_from_csv()` function that would take some work to fix. If it seems reasonable, can I open an issue to reduce test duplication and tackle it when we get rid of the legacy functions?

kgodey:
@eito-fis I think we should just support a handful of delimiters. [According to Wikipedia](https://en.wikipedia.org/wiki/Delimiter-separated_values), the most common ones are comma, tab, colon, pipe, and space, so let's just do those.

kgodey:
Also, it's fine to do the test de-duplication as a separate issue that we tackle when we get rid of the legacy functions.

eito-fis:
Updated to constrain delimiters and catch errors, but it doesn't work as well as we want it to (as evidenced the `test_create_table_from_datafile_invalid_delimiter` test failing). Primary issue is that the `csv.Sniffer` doesn't seem to check the validity of delimiters across rows and just looks for any of the valid delimiters in a line, so it  picks up on something like a space and runs with it. Then, we error down the line when executing the actual `COPY` command.

Two possible solutions I can think of:
* Write our own inference function (I've talked to @mathemancer a bit about this). Given we have a list of delimiters, just iterating over them and checking if they produce valid tabular data seems relatively straightforward. There might be some weird behavior when multiple delimiters are valid (just have an order of precedence?) but overall it doesn't seem horrible.
* Having user specified values. We could move the appropriate values (`delimiter`, `quotechar`, `escapechar`) into fields on the `DataFile` model, and just have the option to specify them at file creation. We would still have to validate the values work to avoid SQL errors down the line, but it is nice in that it matches the UX of something like LibreOffice, where you specify how you want to import a file. The idea of validating the files at file creation time and not table creation time is also nice to me. I guess the question for this approach is how to expose this to the user, and if we need to do anything with the API to support it. (Ex: LibreOffice shows what the table will look like with your current settings.)

The first approach fits the current framing better so I'll probably start there tomorrow unless someone prefers the second approach.

mathesar/models.py
mathemancer: For the column headers, I think we're covered in the table creation step, but we might have an issue in the `copy_expert` step.  While the current implementation uses double-quotes to delimit the column names (as it should), someone could (perhaps) escape containment by passing an un-escaped double-quote as part of a column name.  I think it would be best to wrap that SQL using the `psycopg2.sql.SQL` object:  https://www.psycopg.org/docs/sql.html#module-psycopg2.sql .

For the table data, I don't see any glaring problems, and I also think we're okay on the actual `COPY` step, but it somehow feels weird to then have a table full of potentially malicious SQL commands in the actual DB.  I want to think more about that.

mathemancer:
Completely agree that support of a limited set of delimiters makes sense.  @eito-fis What were you thinking w.r.t. the checking?  Would you still be using `csv.Sniffer` to sort out quotes and escapes?

kgodey:
I agree with going ahead with option 1. We shouldn't ask the user for delimiters, quote characters, or escape characters. We can maybe do that in a future iteration for advanced users, but we're trying to make the UX work well for non-technical users.

We could use the extension as a clue to figure out which delimiter it is (e.g. if it's `.csv`, try commas first).

eito-fis:
Sounds good, will take a crack at it today.

@mathemancer: For the delimiters I was imagining tabulating the data by the delimiter (just a `.split()`?), and checking to ensure that the number of columns was consistent across the the first `x` rows or so. Not entirely sure about the quotes and escape chars - I'll start with the sniffer and see if I can break it before replacing it.

mathemancer:
Sounds reasonable, but keep in mind that escapes will affect the delimiter count.

eito-fis:
Updated with new process. Now uses [clevercsv](https://github.com/alan-turing-institute/CleverCSV) to do the sniffing, and then validates the dialect that's produced with the `check_dialect()` function. The goal with `check_dialect()` is to validate the file / dialect and throw errors in a way we can handle, before we get to the data layer. Currently only checks for mismatched number of columns between rows and unmatched quotes. Added a set of tests to ensure we have reasonable behavior, and getting much better results than the previous method.

Side-note: It feels weird validating data files at table creation time. Would it be reasonable to move the current validation logic to when the data file is created? (This would involve saving the dialect fields to the data field, so I guess this wouldn't work if we planned to also upload non-*sv files.)

kgodey:
@eito-fis I too think validating on data file creation makes the most sense, and I'm fine with storing extra fields on the DataFile as long as we populate them automatically and don't ask the user. We can worry about how to extend to other file types later.

mathesar/imports/csv.py
```diff
@@ -1,15 +1,141 @@
...
+        i += 1
+
+    if in_quote:
+        return None
+
+    splits.append(current_split)
+    return splits
+
+
+def check_dialect(file, dialect):
```
mathemancer, line 72: Did you consider using the `stream_table` function from `clevercsv`?  

https://clevercsv.readthedocs.io/en/latest/source/clevercsv.html#clevercsv.wrappers.stream_table

It would be easier to maintain than the custom parser, and I'd much prefer that unless there's a good reason to roll our own here.

mathemancer (changes requested):
I'd very much prefer having our own CSV parser to maintain if possible (as much as I appreciate the fun of writing your own CSV parser).  To that end, I'd like to see whether it's possible to put together a pipeline that's something like:
1. detect dialect of CSV using `clevercsv.wrappers.detect_dialect`
2. check detected dialect object against CSV using `clevercsv.wrappers.stream_table`, and checking that the row lengths are identical (i.e., set a var from the first row, then assert that the rest are equal to it).

If this isn't working or has some suboptimal result, please just let me know the problem and I'll reconsider whether we need our own parser.

mathemancer (approved):
Okay, I think this is looking good.  Thanks for all your work on this.

PR #217, opened 2021-06-08
Sync DB objects in service layer

### LINKED ISSUE #61: Ensure database objects are synced with web application
**Is your feature request related to a problem? Please describe.**
Currently, if someone uses a pre-existing database in Mathesar, or updates a database outside of Mathesar, these changes are not synced with the tables and schemas stored in the web application.

**Describe the solution you'd like**
We should ensure that database objects are synced, both during initial database setup and regularly.

mathemancer:
Fixes #61 

This PR adds logic to keep things synced between the service models and the user's data DB.  

**Technical details**

- The DB object models are changed to use the OID rather than name of the object for identification at the DB level.
- Names are now read on the fly from the DB when needed.
  - Name property is cached for 5 minutes using the Django cache functionality
- We set a cache object to avoid re-reflecting DB objects constantly, the current setting is 5 minutes
- If some error occurs when reflecting any db object, all DB objects are re-reflected silently, only errors which persist through the re-reflection are raised.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models.py
```diff
@@ -16,32 +19,56 @@ class Meta:
...
+        try:
+            schema_name = cache.get(cache_key)
+            if schema_name is None:
+                schema_name = schemas.get_schema_name_from_oid(
+                    self.oid, self._sa_engine
+                )
+                cache.set(cache_key, schema_name, NAME_CACHE_INTERVAL)
+            return schema_name
+        except TypeError:
+            return 'MISSING'
```
kgodey, line 51: Can you add a comment about when this would happen?

mathesar/models.py
```diff
@@ -16,32 +19,56 @@ class Meta:
...
-
-    @property
     def _sa_table(self):
-        return tables.reflect_table(self.name, self.schema.name, self._sa_engine)
+        try:
+            table = tables.reflect_table_from_oid(
+                self.oid, self.schema._sa_engine,
+            )
+        except TypeError:
+            table = tables.create_empty_table("MISSING")
```
kgodey, line 66: If there's no table matching the `oid` in the database, we should delete the `Table` object in the Django DB, since we should be reflecting the tables accurately.

Same with schemas.

mathesar/serializers.py
```diff
@@ -9,7 +9,7 @@ class NestedTableSerializer(serializers.HyperlinkedModelSerializer):
 
     class Meta:
         model = Table
-        fields = ['id', 'name', 'url']
+        fields = ['id', 'oid', 'name', 'url']
```
kgodey, line 12: I'm not sure we should expose the `oid` in the API, it seems unnecessary. We can always add it later if it's useful.

mathesar/utils/tables.py
```diff
@@ -0,0 +1,17 @@
...
+        table["oid"]
+        for table in get_table_oids_from_schema(schema.oid, schema._sa_engine)
+    }
+    tables = [
+        Table.objects.get_or_create(oid=oid, schema=schema)
+        for oid in db_table_oids
+    ]
+    for table in Table.objects.all().filter(schema=schema):
+        if table.oid not in db_table_oids:
+            table.delete()
```
kgodey, line 16: I see we are deleting tables, so I'm confused about why we're creating an empty table above.

mathesar/models.py
kgodey: I see below that we are doing this, so I'm not sure why we're creating an empty table here. Can you add a comment to explain?

kgodey (changes requested):
Looks good, @mathemancer! I added a few comments, mostly questions.

mathesar/models.py
mathemancer: This is to handle the case where the user is looking at a specific table, and it's deleted under their feet.  Basically, on whatever the next API call is, the front end would get back a table named `MISSING`, and could (for example) just pop up with "the table was deleted" rather than "table not found", before redirecting back to the list.  Once it's redirected back to the list, the DB Sync function will kick in and delete the table's record.

The side benefit is that it gives us the flexibility to decouple the cadence of the overall DB Syncing from the expiration of the cache with the table's name.

I'll put a shorter version in a code comment.

mathesar/serializers.py
mathemancer: Makes sense.

mathesar/models.py
kgodey: That makes sense. I'll merge it for now, we might want to figure out a better way to handle this case in the future (there might actually be empty tables named `MISSING` if the user creates them).

kgodey (approved):

mathesar/models.py
mathemancer: Agreed.  Another option I considered was to just raise a custom exception, and then pass that via the API to the front end.

kgodey: I like that idea better.

PR #265, opened 2021-06-16
Additional table filtering options

### LINKED ISSUE #162: More filters for table list API
**Problem**

The table list API should allow filtering. For example, we might want to get the list of all tables in a schema to see if the table the user is trying to create already exists in that schema.

**Proposed solution**

The table list endpoint should support filtering by:
- schema
- before/after: created, last updated
- whether the import was verified

**Additional context**

We should use `django-filter` since it integrates with DRF and makes setting up filters easy.

eito-fis:
Fixes #162 

Add more options for filtering to the table list endpoint.

**Technical details**
Adds an `CharIn` filter on the schema name, a `DateTimeRange` filter on `created_at` and `updated_at`, and a `Boolean` filter on `import_verified`.

The `DateTimeRange` filter allows for several types of filtering:
* An exact time (`?created={time}`)
* A min time (`?created_after={time}`)
* A max time (`?created_before={time}`) 
* A range (`?created_after={start}&created_before={end}`).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Sidenotes:
* How should merging this PR work? I branched off @mathemancer's branch to use the changes to filtering - should we hold off on merging until the `db_syncing` branch is merged?
* Is there a case where we would want to filter by `import_verified = None`? [Django-filters doesn't support filtering by `None` by default](https://github.com/carltongibson/django-filter/issues/1012), so we would have to make [some changes](https://github.com/carltongibson/django-filter/issues/522#issuecomment-258171253) to support it.

mathemancer:
@eito-fis How difficult would it be to rework this to branch off of `master`?  It will be easier to merge these changes into my branch than the other way, I think (though it's still going to be a pain).  Where is `import_verified` set to `None`?  We should probably just disallow `null` for that, and set a `False` default value.  This will be more logical to me, and would avoid the problem you mentioned.

kgodey:
@mathemancer I had asked @eito-fis to work off of your branch since you'd made changes to the filtering there, my thought was that we could merge this PR in once your DB syncing PR had been merged in to spare you additional merge conflicts. I'm fine with reworking off of `master` if that's easier, just offering more context.

Also, my thought behind `import_verified` being set to `None` was that if a table was created from scratch (rather than from a data file), we'd set `import_verified` to `None` rather than `False`. This way, we can distinguish between tables that don't need to be verified and tables that need to be verified. So yes, we should support filtering by `null`. Also, adding a comment somewhere that that's how we plan to use `import_verified` would be helpful.

eito-fis:
Updated to add an additional `not_imported` parameter. `?not_imported=true` returns all tables with `import_verified = None`, and `?not_imported=false` returns all tables with `import_verified != none`. Having some issues with `django_property_filters` here, and wasn't able to implement this how I wanted. The initial solution of passing `none`, `true`, or `false` to the `import_verified` parameter doesn't seem to work with the library, and the `exclude` option doesn't work with the library (tested), so we have to use `not_imported` instead of just `imported`. Maybe something we can change if `django_property_fitlers` ends up being unneeded.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -102,6 +102,127 @@ def test_table_list_filter(create_table, client):
...
+
+
+def test_table_list_filter_created(create_table, client):
+    table_name = 'Fitler Created'
+    table = create_table(table_name)
+    query_str = '2020-01-01 8:00'
+
+    response = client.get(f'/api/v0/tables/?created_before={query_str}')
+    response_data = response.json()
+    assert response.status_code == 200
```
kgodey, line 137: Since a lot of these assertions are the same, it might be worth pulling them into a separate method so you don't repeat yourself.

kgodey (approved):
Looks good, added a comment but I don't think it's essential to fix before merge.

I'm not merging since this depends on the `db_syncing` branch being merged first. I'll label as blocked for now.

eito-fis:
Currently blocked by #280

mathesar/models.py
```diff
@@ -28,13 +28,19 @@ def __str__(self):
         return f"{self.__class__.__name__}: {self.oid}"
 
 
+_engine = None
```
kgodey, line 31: Please add a TODO comment that this needs to be removed once we have the proper fix (and also add a note to #280)

kgodey (approved):

PR #268, opened 2021-06-18
Type inference func

### LINKED ISSUE #200: Table creation from data file should infer data types automatically
**Problem**

Currently, we create an all-string table when we create a table from a data file. In order to support our [design for file import](https://wiki.mathesar.org/design/process/documents/table-import-specs), we'll need to show the user the types we recommend for each column.

**Proposed solution**

We should have an API that provides inferred data types for each column in a table. We don't actually want to change the table yet, just infer data types for each column.

**Additional context**

See #92 for DB functions to be used for this.

Related to #199, although I don't think it's actually blocked by it.

eito-fis:
Related to #200 

Adds a function to `db` that takes a table and returns a list of inferred types, corresponding to the columns of the passed table.

**Technical details**
Updates the old `infer_table_column_types` to `update_table_column_types`, and writes a new `infer_table_column_types`. Uses a `CREATE TABLE AS` postgres statement to copy the passed table into a temporary table, which is then passed to `update_table_column_types`. Finally, we extract the types from the temp table, drop the temp table, and return the types.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Will look to add a few more tests before opening for review.

kgodey:
@eito-fis I don't think this entirely resolves #200 so please change your note from "fixes" to "related to" so that #200 isn't auto-closed.

eito-fis:
Should be mostly good for review, but I did have one question - do we have a set way of creating temporary tables? Right now the function literally makes a table called `temp_table`, which doesn't seem safe at all.

db/tables.py
```diff
@@ -358,3 +358,26 @@ def infer_table_column_types(
...
+    temp_full_name = schema + "." + temp_name
+    columns = [c.copy() for c in table.columns]
+    temp_table = Table(temp_name, metadata, *columns)
+
+    create_table = f"""
+    CREATE TABLE {temp_full_name} AS
+    TABLE {table.schema}.{table.name}
+    """
+    with engine.begin() as conn:
+        conn.execute(DDL(create_table))
```
mathemancer, line 377: For creating the temporary table, I suggest using the recipe posted by the SQLAlchemy maintainer here:  https://github.com/sqlalchemy/sqlalchemy/issues/5687 , but modified with the `TEMPORARY` prefix.  This will let you quickly create the table with the data you want to use for inference.

Your goal should be to end up with SQL along the lines of:

```sql
CREATE TEMPORARY TABLE my_temp_table AS SELECT * FROM my_orig_table;
```

The `TEMPORARY` prefix will result in the table being automatically dropped either at the end of the current transaction block, or whenever the connection is closed (when using the context manager, this should be when the outermost context-management block ends).  One caveat:  Temporary tables _cannot_ be created in a schema (they exist in a system schema).  For details about this schema caveat, and how to set the dropping behavior properly, see the docs here:  https://www.postgresql.org/docs/13/sql-createtable.html .

mathemancer: An alternative route would be to try to copy the table in SQLAlchemy (i.e., reflect the table, and then copy the object).  You'd then do an `insert().from_select(...)` sort of statement:  https://docs.sqlalchemy.org/en/14/core/dml.html?highlight=from_select#sqlalchemy.sql.expression.Insert.from_select .  

I suspect that won't be as nice in the end, and it'll be a bit less efficient.  If you want to try, you can use `prefixes=["TEMPORARY"]` in your temp table definition, and then use the `table.create(engine)` method of the SA table object (see `create_mathesar_table` for an example of idiomatic table creation).  Once the table is created, use the `insert().from_select(...)` statement.

mathemancer (changes requested):
Please try to see if you can make it work with an actual DB-level `TEMPORARY` table.  See the novella I wrote in my line-level comment for some ideas about how that might go.

mathemancer:
@eito-fis Regarding `Column.copy`, the deprecated version didn't actually copy the column in the expected way under all circumstances.  So, for now, we need to do copying ourselves.  This is the motivation behind the `MathesarColumn` object.  The point of that is to constrain the parts of the column that we're aware of so that we can make sure _those parts_ get copied correctly.  See the docstring of that object for more.  

If you want to do anything that feels like "copying" a column, I advise first making it a `MathesarColumn`, then creating a copy by calling `MathesarColumn.from_column` on that column.  (or just using `MathesarColumn.from_column` on the actual SA `Column` type).  If you need to handle some property that's not included in `MathesarColumn`, please add it to that class and relevant methods, and then make sure that the `from_column` method has the expected behavior w.r.t. that property.

db/tables.py
eito-fis: I updated to use the `DDLElement` and `MathesarColumns`, but am struggling with getting the proper temporary table implementation working. Two issues so far:
* Problems reflecting the table after the type has been changed. Ideally we should be able to do this with `Table(autoload_with=conn)`, but I'm running into a `a transaction action has already begun` error. Making sure to run the type inference outside the initial `with conn.begin()` block, so not entirely sure whats happening here. As a temporary fix, we just pass the `MetaData` through all the inference functions, but this leads to cycles since the sqlalchemy table is never updated. 
* An error in the middle of the type inference functions seems to prevent the table from being dropped. When the above error occurs and our first test fails, the subsequent tests also fail with a `temp_table already exists` error. Will add a test for this behavior, but not entirely sure how to fix yet.

If there's anything obvious I'm doing wrong, please let me know! Otherwise, I'll keep hacking away at this tomorrow.

eito-fis:
The function should be mostly up and running now, but still running into an issue with the `same_name` test. Weirdly, it only fails when run after the `drop_temp` test. When run on its own, or with `drop_temp` commented out, we don't get a `temp_table already exists` error. Not entirely sure whats going on there, need to look into it.

eito-fis:
Updated, should be passing all relevant tests now (might still have `too many clients` error).

Unfortunately, we might be stuck with having to manually drop the temp table. I did try using `ON COMMIT DROP` instead, but ran into problems.  Doing so means we have to execute all type inference commands in a single transaction. But, we check types by intentionally causing the transaction to error and psycopg2 seems to prevent additional commands from executing if an error has already been thrown with a `current transaction is aborted, commands ignored until end of transaction block` error. Not sure if we can get around this without rewriting the type inference code.

db/tables.py
```diff
@@ -408,4 +413,46 @@ def infer_table_column_types(
...
+def compile(element, compiler, **_):
+    return "CREATE TEMPORARY TABLE %s AS (%s)" % (
+        element.name,
+        compiler.sql_compiler.process(element.selectable, literal_binds=True),
+    )
+
+
+def infer_table_column_types(schema, table_name, engine):
+    table = reflect_table(table_name, schema, engine)
+    temp_name = "temp_table"
```
mathemancer, line 436: I think this should be randomly-chosen on the fly (or random-ish:  Think `temp_table_<uuid>` or something similar).  The reason is that PostgreSQL puts temp tables into a kind of "global" space, and while they're not _accessible_ from other sessions, they block creation of a table with the same name in those session.  This would lead to problems if more than one client is connected, trying to use this function.

db/tables.py
```diff
@@ -408,4 +413,46 @@ def infer_table_column_types(
...
+
+    select_table = select(table)
+    with engine.connect() as conn:
+        with conn.begin():
+            conn.execute(CreateTempTableAs(temp_name, select_table))
+        with conn.begin():
+            temp_table = reflect_table(temp_name, None, conn)
```
mathemancer, line 443: It seems from this that the temp table is persisting over these connections.  In that case, there isn't really a point in making it an actual temporary table.  I.e., we could just use a standard table at that point, which would make some things slightly simpler (and allow for namespacing the created table).  It also avoids confusion for our future selves, trying to figure out why we need to drop a temp table.

db/tables.py
```diff
@@ -408,4 +413,46 @@ def infer_table_column_types(
...
+            with conn.begin():
+                temp_table = reflect_table(temp_name, None, conn)
+                types = [c.type.__class__ for c in temp_table.columns]
+        except Exception as e:
+            # Ensure the temp table is deleted
+            with conn.begin():
+                temp_table.drop()
+            raise e
+        with conn.begin():
+            temp_table.drop()
```
mathemancer, line 457: Same as above.  If we can't get the temp table to work (since SQLAlchemy won't give up connections), we should probably just use a normal table to keep from confusing people.

db/types/alteration.py
```diff
@@ -25,15 +27,20 @@ def get_supported_alter_column_types(engine):
...
     _preparer = engine.dialect.identifier_preparer
     supported_types = get_supported_alter_column_types(engine)
     target_type = supported_types.get(target_type_str.lower())
-    with engine.begin() as conn:
+
+    with ExitStack() as stack:
+        if conn is not None:
+            stack.enter_context(conn.begin())
+        else:
+            conn = stack.enter_context(engine.begin())
```
mathemancer, line 40: For my own information:  Why is this move necessary?

mathemancer (changes requested):
Please consider changing the temp table to a standard table to avoid confusion.  If you think there's still utility to creating it as a _temp_ table, please document why in a comment.  Long term, I think we should try to get that working for safety, but I think we can push it till a future issue.

Also, if we _aren't_ using a temp table, I think it should be possible to clean up some (lots) of the ickier context management that's necessary to get access to the temp table.  Is that correct?  If so, I think it's an even better argument for just giving up on the temp table idea (for the moment) and changing to a standard table.

db/types/alteration.py
eito-fis: To access the temp table, we have to start our transaction using the connection that made the temp table. So here we start the transaction using a connection if there is a connection, and the engine otherwise. (Hopefully will be getting rid of this with the move from temp tables though.)

eito-fis:
Heres a snippet to confirm the issue is with connection pools for future reference:
```python
 with engine.connect() as conn:
    _conn = conn
    with conn.begin():
        conn.execute(CreateTempTableAs(temp_name, select_table))
with engine.connect() as conn:
    print(conn == conn) # True
    with conn.begin(): 
        temp_table = reflect_table(temp_name, None, conn) # No error
```
```python
 with engine.connect() as conn:
    _conn = conn
    with conn.begin():
        conn.execute(CreateTempTableAs(temp_name, select_table))

# Force a set of new connections
engine.pool.dispose()
engine.pool = engine.pool.recreate()

with engine.connect() as conn:
    print(_conn == conn) # False
    with conn.begin(): 
        temp_table = reflect_table(temp_name, None, conn) # sqlalchemy.exc.NoSuchTableError: temp_table
```
The problem being that the temp tables are default tied to a connection, but since connections are re-used the temp table isn't dropped. We could force the connection we used to be recreated, but that ends up being more work than using a non-temp table and making sure to drop it. I'll go ahead and migrate to using non-temp tables.

db/tables.py
eito-fis: Correct me if I'm wrong, but I believe this shows the table persisting across several transactions within a single connection? However, given the issue with the connection pooling (see comment below), I agree that there isn't a point to using temp tables. For namespacing, are there schemas we have reserved for this sort of work?

eito-fis:
Updated to use non-temp files with `{MATHESAR_PREFIX}temp_schema` and `{MATHESAR_PREFIX}temp_table_{current_epoch}` as the schema and table name, respectively. Now also ensures that the table name is unique. Assuming we protect schemas from starting with `MATHESAR_PREFIX`, I think we should be safe from table name collisions.

mathemancer (approved):
I think this looks good for now.  We should reassess using temp tables after the completion of #280 and any other pressing engine or connection modifying work.  Long term, we need a way to make absolutely sure that this function can't leave random tables littered about the DB under any conditions.

PR #276, opened 2021-06-22
Type inference endpoint

eito-fis:
Fixes #200 

Adds an endpoint (`/api/v0/tables/{table.id}/type_suggestions/`) which returns a dictionary mapping column names to inferred types.

**Technical details**
The endpoint calls the `infer_table_column_types()` function from #268, then converts the SQLAlchemy types into strings and returns the values. 

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [ ] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #284, opened 2021-06-25
Ensure pytest-django doesn't make an extra database

### LINKED ISSUE #283: API tests using wrong tables database
**Describe the bug**
The django API tests are running on the `test_mathesar_db_test_database`, which differs from the `mathesar_db_test_database` tables database we should be using. As a result, we don't have a proper reference to the database being used by the API functions, which prevents us from certain operations like installing types for a test. 

**Expected behavior**
We should ensure `pytest-django` doesn't build a separate tables database.

**Additional context**
Currently blocking #276

eito-fis:
Fixes #283 

Updates `settings.py` to ensure that `pytest-django` won't make an additional database.

**Technical details**
Adds a line to `settings.py` that sets `DATABASES[mathesar_db_name]['TEST']['NAME']` equal to `DATABASES[mathesar_db_name]['NAME']` at test time.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -74,6 +74,8 @@ DATABASES = {
     decouple_config('MATHESAR_DATABASE_KEY'): decouple_config('MATHESAR_DATABASE_URL', cast=db_url)
 }
 
+# pytest-django will create a new database named 'test_{DATABASES[table_db]['NAME']}'
+# and use it for our API tests if we don't specify DATABASES[table_db]['TEST']['NAME']
 if decouple_config('TEST', default=False, cast=bool):
     DATABASES[decouple_config('MATHESAR_DATABASE_KEY')]['TEST'] = {
         'NAME': DATABASES[decouple_config('MATHESAR_DATABASE_KEY')]['NAME']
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -8,13 +8,19 @@
 from config.settings import DATABASES
 
 TEST_DB = "mathesar_db_test_database"
```
kgodey, line 10: @eito-fis Once you merge `master` in, I think it would be good to read these from the `python-decouple` instead of defining them in two places.

kgodey (changes requested):
You'll need to merge `master` in, I added a comment about that.

I'm a little confused about the purpose of this PR. We should **never** need an engine for the Django database because we use the Django ORM for anything Django manages. We only use SQLAlchemy and the `engine`object for the database(s) that you see in the Mathesar UI.

This discussion might be helpful: https://github.com/centerofci/mathesar/discussions/23

eito-fis:
@kgodey: Sorry, I think I got a bit turned around on this one - the changes here don't  address the issue. The primary problem is that we currently have two tables just for the mathesar tables, `mathesar_db_test` and `test_mathesar_db_test`. The engine fixture we have connects to `mathesar_db_test`, but `create_mathesar_engine()` connects to `test_mathesar_db_test`. As a result, we can't modify the mathesar tables database that the API is using outside of the API, which is problematic if we want to do something like install types on it.

eito-fis:
Updated to properly address the problem. I've rename the PR and edited the original comment to reflect the changes.

I've also confirmed that this fixes #283 on #276.

kgodey (approved):
Looks good, thanks! Please add a comment to the settings file explaining what you did in the comment before merging.

Issue #69, opened 2021-04-27
Setup API documentation and Postman collection


pavish:
* Having a dedicated API documentation would make things easier for quick reference of our APIs as well as for anyone developing clients for Mathesar. Personal preference: [Redoc](https://github.com/Redocly/redoc)
* [Postman](https://www.postman.com) collections make it easier to store, query, test and share APIs. Having one would make development and debugging easier. (Optional - just an idea)

## Related Discussion
- #875

ManishShah120:
Hi @pavish I too felt the need of this feature while trying to understand all the present API endpoints and its allowable parameters.  May be we can also think of [Swagger](https://swagger.io/)!

seancolsen:
I'm closing this because it seems out of date. In 2023 we had a GSoC project to document the API, and that project made some slightly different decisions from what is specified in this ticket description.

PR #299, opened 2021-06-28
Record filtering

### LINKED ISSUE #157: Allow filtering in record list API
**Problem**

The record list API should allow filtering the list of records returned by values in various columns.

**Proposed solution**

Passing query parameters in the URL of the format `column=value` in the record list API should return a filtered list of records instead of all records. Multiple filters should be supported, and records returned should satisfy all of the filters passed in.

**Additional context**

- Look into using https://django-filter.readthedocs.io/ since it integrates with DRF.
- Filtering records is already implemented on the DB level, see https://github.com/centerofci/mathesar/issues/143

eito-fis:
Fixes #157

Adds the option to filter the values returned from the record list endpoint.

**Technical details**
Modifies `records.get_records()` to use our fork of `sqlalchemy_filters` for filtering, and adds handling for the `filters` url parameter to the record list endpoint.  

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Quick question before I mark this ready for review - how extensive should tests be when using an external library? In this case, I'm  mostly just checking our error handling and a single case of filtering (excluding the old tests I updated) since it feels like comprehensive testing should be done on the `sqlalchemy-filters` end. Is this reasonable? Happy to add more testing if it seems necessary.

kgodey:
@eito-fis I would think about tests as verifying the behavior of our API rather than testing specific functionality. We want to test all the ways we want to use our API so that if something breaks, we notice. For example, we might upgrade the library and it might change behavior, which would be expected for the library but unexpected for our API. I hope that helps.

eito-fis:
Updated with fairly extensive tests for filtering on the `db` side. Tests on the API side are simpler, and primarily check that the proper function is being called on the backend.

db/records.py
```diff
@@ -35,32 +36,22 @@ def get_records(
         order_by: list of SQLAlchemy ColumnElements to order by.  Should
                   usually be either a list of string column names, or a
                   list of columns from the given table.
-        filters:  list of tuples of type (ColumnElement, value), where
-                  ColumnElement is an SQLAlchemy ColumnElement, and value
-                  is a valid value for the associated column (i.e., the
-                  type must be correct)
+        filters:  list of dictionaries, where each dictionary has a 'field' and 'op'
```
kgodey, line 39: Can we document the list of `op` parameters somewhere? (or link to existing documentation)

db/tests/conftest.py
```diff
@@ -8,6 +8,7 @@
 FILE_DIR = os.path.abspath(os.path.dirname(__file__))
 RESOURCES = os.path.join(FILE_DIR, "resources")
 ROSTER_SQL = os.path.join(RESOURCES, "roster_create.sql")
+FILTERSORT_SQL = os.path.join(RESOURCES, "filter_sort_create.sql")
```
kgodey, line 11: Can you call this `FILTER_SORT_SQL`, for clarity?

mathesar/views/api.py
```diff
@@ -85,9 +89,19 @@ class RecordViewSet(viewsets.ViewSet):
 
     def list(self, request, table_pk=None):
         paginator = TableLimitOffsetPagination()
-        records = paginator.paginate_queryset(self.queryset, request, table_pk)
-        serializer = RecordSerializer(records, many=True)
-        return paginator.get_paginated_response(serializer.data)
+        filter_form = RecordListFilterForm(request.GET)
```
kgodey, line 92: Why use a form here?

mathesar/tests/views/api/test_record_api.py
```diff
@@ -42,6 +51,40 @@ def test_record_list(create_table, client):
         assert column_name in record_data
 
 
+def test_record_list_filter(create_table, client):
```
kgodey, line 54: I think it would be good to document somewhere how to use the filters from an API perspective – format, operations supported, etc.

mathesar/views/api.py
```diff
@@ -85,9 +89,19 @@ class RecordViewSet(viewsets.ViewSet):
...
+        if filter_form.is_valid():
+            try:
+                records = paginator.paginate_queryset(
+                    self.queryset, request, table_pk,
+                    filters=filter_form.cleaned_data['filters']
+                )
+            except (FieldNotFound, BadFilterFormat) as e:
```
kgodey, line 99: Could you document why you're using these specific errors here and generic ones later?

mathesar/views/api.py
```diff
@@ -85,9 +89,19 @@ class RecordViewSet(viewsets.ViewSet):
...
+                raise ValidationError({'filters': e})
+            serializer = RecordSerializer(records, many=True)
+            return paginator.get_paginated_response(serializer.data)
+        else:
+            raise ValidationError(filter_form.errors)
```
kgodey, line 104: Wouldn't you want to raise the same error format for all exceptions?

db/tests/resources/filter_sort_create.sql
```diff
@@ -0,0 +1,131 @@
...
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
+ALTER SEQUENCE "filter_sort_mathesar_id_seq" OWNED BY "FilterSort".mathesar_id;
+
+ALTER TABLE ONLY "FilterSort" ALTER COLUMN mathesar_id SET DEFAULT nextval('"filter_sort_mathesar_id_seq"'::regclass);
+
+INSERT INTO "FilterSort" VALUES
```
kgodey, line 21: Couldn't you load from the CSV instead of duplicating the data? If you change the data, you'll need to change it in two places.

kgodey (changes requested):
@eito-fis Looks good overall, I added a few questions and comments.

I'd like @mathemancer to review this as well since it touches `db` heavily.

mathesar/views/api.py
eito-fis: Forms and form fields are nice for the JSON input as they automatically parse and validate the url parameters for us. In this PR it's somewhat weird since there's only a single field, but its a bit more useful with the sorting and grouping fields added in. It is a little un-intuitive though - I'm happy to do it manually if this isn't a pattern we want to use.

mathesar/tests/views/api/test_record_api.py
eito-fis: Do we have somewhere we've put similar things? It is literally the `sqlalchemy-filters` format so I think linking to [those docs](https://github.com/centerofci/sqlalchemy-filters) might be enough, but I'm not sure where to put it.

mathesar/views/api.py
eito-fis: I think the error formats do work out the same, `form.errors` should contain a dictionary with fields as keys and errors as values.

mathesar/views/api.py
eito-fis: This is so we can catch specific sorting errors (`SortFieldNotFound` and `BadSortFormat`) separately down the line in order to return the error in the appropriate field of the return dictionary. It might make more sense in [the context of the sorting PR](https://github.com/centerofci/mathesar/pull/300/files#diff-ed2d8e28f999e57c7acd4c3e92bd4efd27dc832b90ad625913866af4b41a54d3R102)?

db/tests/resources/filter_sort_create.sql
eito-fis: Yep, will remove the CSV as the other file is also referenced as SQL.

mathesar/tests/views/api/test_record_api.py
kgodey: Somewhere in the comments is fine for now, it's on my todo list to set up proper API docs: https://github.com/centerofci/mathesar/issues/69

mathesar/views/api.py
kgodey: That's fine, add a comment explaining that in the code in case anyone else is confused. We can figure out a different pattern later if desired.

mathesar/views/api.py
```diff
@@ -83,9 +87,24 @@ class RecordViewSet(viewsets.ViewSet):
     # where the entire record needs to be replaced, PATCH suffices for updates.
     queryset = Table.objects.all().order_by('-created_at')
 
+    # For fitlers parameter formatting, see:
+    # https://github.com/centerofci/sqlalchemy-filters#filters-format
     def list(self, request, table_pk=None):
         paginator = TableLimitOffsetPagination()
-        records = paginator.paginate_queryset(self.queryset, request, table_pk)
+
+        # Use a form to parse JSON url parameters
```
kgodey, line 95:
```suggestion
        # Use a Django Form to automatically parse JSON URL parameters
```

mathesar/views/api.py
```diff
@@ -87,8 +87,12 @@ class RecordViewSet(viewsets.ViewSet):
     # where the entire record needs to be replaced, PATCH suffices for updates.
     queryset = Table.objects.all().order_by('-created_at')
 
+    # For fitlers parameter formatting, see:
```
kgodey, line 90:
```suggestion
    # For filter parameter formatting, see:
```

kgodey (approved):
Looks good, I made some suggested comment changes.

I'll leave this to @mathemancer to review/merge.

db/tests/resources/filter_sort_create.sql
```diff
@@ -0,0 +1,131 @@
...
+    "array"   integer[]
+);
+
+CREATE SEQUENCE "filter_sort_mathesar_id_seq"
+    AS integer
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
```
mathemancer, line 15: Please see:  https://www.postgresql.org/docs/13/datatype-numeric.html#DATATYPE-SERIAL

Also, are you actually using this sequence?  It seems like there's lots of trouble involved in setting it up, but it's not actually used anywhere.

db/tests/resources/filter_sort_create.sql
```diff
@@ -0,0 +1,131 @@
+CREATE TABLE "FilterSort" (
```
mathemancer, line 1: It doesn't matter much for a testing table, but underscores are preferable to capital letters for denoting words in SQL names (since SQL is non-case-sensitive by default).

mathemancer (approved):
The only (minor) change I have to request is to rename the function for morphing tuples into filters.  Also consider using a `SERIAL` column in the testing table setup SQL, but that's not really a big deal.

PR #315, opened 2021-07-01
Record grouping

### LINKED ISSUE #216: Record API should support grouping results by a given field.
**Problem**

The user should have the option to show records grouped by a given field. We need API support for the UI to show this.

**Proposed solution**

The API should support this use case, in addition to filtering and sorting options. See #157 and #158. Specifically, we need an API that provides group counts of various values (see comments below).

**Additional context**

As part of this issue, we'll need to define what a grouped API response looks like. Please update the [API standards](https://wiki.mathesar.org/engineering/architecture/api-standards) with details and request review from @kgodey and @mathemancer before proceeding.

eito-fis:
Fixes #216

Adds the ability to get group count information from the record list endpoint.

**Technical details**
Adds a `get_group_counts()` function to the `db` module, which takes in a list of fields to group by and returns a dictionary mapping tuples of fields to their counts. Also adds a `group_count_by` parameter to the record list endpoint, where a user can pass a list of fields to group by and get
```python
'group_count': {
    'group_count_by': [fields_used_for_grouping],
    'results': {dictionary_mapping_fields_to_counts},
}
```
including in the return data. The keys for `results` are strings, with composite fields names joined by `,`s. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/pagination.py
+++ b/mathesar/pagination.py
@@ -47,14 +47,16 @@ class TableLimitOffsetPagination(DefaultLimitOffsetPagination):
                 group_count_by, self.limit, self.offset,
                 filters=filters, order_by=order_by
             )
+            # Convert the tuple keys into strings so it can be converted to JSON
+            group_count = {','.join(k): v for k, v in group_count.items()}
             self.group_count = {
-                "group_count_by": group_count_by,
-                "results": group_count,
+                'group_count_by': group_count_by,
+                'results': group_count,
             }
         else:
             self.group_count = {
-                "group_count_by": None,
-                "results": None,
+                'group_count_by': None,
+                'results': None,
             }
 
         return table.get_records(
--- a/mathesar/pagination.py
+++ b/mathesar/pagination.py
@@ -11,7 +11,6 @@ class DefaultLimitOffsetPagination(LimitOffsetPagination):
     def get_paginated_response(self, data):
         return Response(OrderedDict([
             ('count', self.count),
-            ('group_count', self.group_count),
             ('results', data)
         ]))
 
@@ -32,7 +31,7 @@ class ColumnLimitOffsetPagination(DefaultLimitOffsetPagination):
 class TableLimitOffsetPagination(DefaultLimitOffsetPagination):
 
     def paginate_queryset(self, queryset, request, table_id,
-                          filters=[], order_by=[], group_count_by=[]):
+                          filters=[], order_by=[]):
         self.limit = self.get_limit(request)
         if self.limit is None:
             self.limit = self.default_limit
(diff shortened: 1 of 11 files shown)


eito-fis:
Question before I open this for review - do we have a set place to handle errors for the functions were exposing in the API? As of now we don't have much explicit error handling except for some of the stuff that's automatically included in `sqlalchemy-filters`, but it would be nice to check the validity of the format and fields for the `group_count_by` parameter. 

In practice, I think the most principled place for validation would be in the form object that we're using to parse the initial url parameters (This would also remove the awkward error handling in the middle of the `list` view right now). However, if we want to validate using try-catch patterns then handling things in the db layer seems more appropriate.  I'm leaning towards doing the former and re-implementing the validation `sqlalchemy-filters` does on our end, but would appreciate input on what would be proper.

kgodey:
@eito-fis I don't think I have enough information to make an informed recommendation. Questions:

1. What does validating using try-catch patterns involve, and how does it differ from validation in the form object?
2. What validation from `sqlalchemy-filters` would we need to reimplement if we went with the first option?

eito-fis:
@kgodey 
1. I guess by try-catch I mean literally catching formatting errors from whatever we're running in the db layers, in contrast to manually checking whether our inputs are valid. Ex: catching a `BadFilterFormat` error out of `sqlalchemy-filters` or a `ProgrammingError` out of `psycopg2`.
2. The primary ones are around validating that the input is in the proper format (iterable or dictionary), that each dictionary contains the required fields, and that the field being targeted in the operation actually exists. 

Having written this out, re-implementing the validation doesn't seem like a great idea...
I think I'll try and implement group validation in the db layer - in hindsight, having form validation separate from the actual grouping function is a bit weird since there would only be an implicit connection between the two.

kgodey:
@eito-fis Sounds like you've answered your own question, that all makes sense to me.

eito-fis:
This should be reviewed after #300 is merged.

mathesar/pagination.py
```diff
@@ -39,5 +40,42 @@ def paginate_queryset(self, queryset, request, table_id, filters=[], order_by=[]
...
+        )
+
+        table = queryset.get(id=table_id)
+        if group_count_by:
+            group_count = table.get_group_counts(
+                group_count_by, self.limit, self.offset,
+                filters=filters, order_by=order_by
+            )
+            # Convert the tuple keys into strings so it can be converted to JSON
+            group_count = {','.join(k): v for k, v in group_count.items()}
```
mathemancer, line 70: This is a bug if a string rather than a tuple is returned by `table.get_group_counts`, i.e., if we want to group by only one column.  I suggest modifying that function so it only has one return type(ish), a tuple.  This will be easier to deal with, and less prone to causing bugs.  There should also be a test for that.

mathemancer (changes requested):
I like the overall approach, and agree that pushing the validation down to the DB layer makes sense.  Unfortunately, there's a bug occurring when grouping by a single column.  See the comment in the `pagination.py` file.

mathesar/pagination.py
mathemancer: In particular, this results in splitting the value of that column up into its constituent characters in the output.

mathemancer: image

eito-fis: Good catch, thanks

mathemancer (approved):
Looks good, thank you!

PR #317, opened 2021-07-01
Add support for multiple databases

### LINKED ISSUE #310: User should be able to configure multiple databases in settings
**Problem**

Currently, the user can only configure one Mathesar database in the settings. They should be able to configure as many databases to connect to Mathesar as they want.

**Proposed solution**

The user should be able to configure multiple databases in the `.env` file.

**Additional context**

We might want to use `python-decouple`'s [built in CSV helper](https://github.com/henriquebastos/python-decouple/#built-in-csv-helper) for this.

Ideally, the user would be able to associate the database key with the connection information directly using a tuple or something like that.

eito-fis:
Fixes #310

Adds support for multiple mathesar tables databases.

**Technical details**
Updated `settings.py` to handle comma separated strings for the `MATHESAR_DATABASE_KEYS` and `MATHESAR_DATABASE_URLS` env variables.

Note, after merging this PR everyone will have to re-copy their `.env` files.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
What should testing look like for this? / Do we need tests for this?

Given the small size of the change and the fact that everything is still running it seems fine as is to me - let me know if I should add some.

kgodey:
I don't think we need tests for this, although you could add a second database to the testing environment to make sure that it works.

It would be ideal if the settings format was tuples of database keys and URLs so that you could see which key was associated with which URL without having to rely on order. It would also ensure that the number of keys was the same as the number of URLs. You can imagine that the current formatting would get confusing if you were adding 10 databases to Mathesar.

eito-fis:
@kgodey Updated to use a tuple like format. Unfortunately can't use normal tuples since `decouple` naively splits on commas, so instead made tuples with `|` as the delimiter. Also, I didn't change the django db variables since the tuple format seems to imply multiple values, but if we want to keep things consistent I'm happy to convert that too.

kgodey (approved):
Looks good to me. Please add a comment somewhere in the code explaining why you're using a pipe instead of a comma before merging. It'll be useful to have the context in case someone wonders about it a year from now.

kgodey:
I think it's fine to leave the Django DB format as is.

Issue #311, opened 2021-06-30
Create new `/v0/databases/` API


kgodey:
**Problem**

We need a database API that has more information than just the database key. For example, we want to show the types installed in the database.

**Proposed solution**

We should create a new `/v0/databases/` API that includes the database key and types installed/supported by the database. We should also create a `Database` model, which should track whether a database is deleted or not.

**Additional context**

Ideally, we'd just rename and extend the `database_keys` API, but the frontend is using it and we don't want to break that. So we'll do this in a two step process where we create a new API first, wait for the frontend to switch, and delete the old API.

eito-fis:
In this case, would it make sense to treat each database as an object? Something like:
```python
{
  'counts': 2,
  'results': [
    'db_1_name': {
      'key': 'db_1_key',
      'types': ['type_1', 'type_2'],
    },
    'db_2_name': {
      'key': 'db_2_key',
      'types': ['type_2', 'type_3', 'type_4'],
    },
  ]
} 
```
for the response?

kgodey:
Yes, it would.

kgodey:
`types` should be something like `supported_types`, to be clearer.

kgodey:
@eito-fis Please update the issue description to include expanded scope from https://github.com/centerofci/mathesar/pull/320. Thanks!

eito-fis:
I've started working on this, but I have a couple questions, in particular about how much we want to use the new Database model:
* How do we want to specify the database for schemas on the API side of things? I've been trying to preserve the current style where we just pass the name of a database, but we have a lot of options here. In particular, we could make the database endpoint the top level endpoint and nest schemas and tables inside them (although this would be a pretty huge change).
* With the addition of the Database model, we have two sources of truth with regards to what dbs we're handling: the Database objects and the django.settings.DATABASES dictionary. It feels like using the Database objects for everything would be nicest (and would reduce our `create_mathesar_engine()` usage quite a bit), but would also require fairly substantial changes. Alternatively, we can leave most functions as is and just use the Database objects when reflecting from the dbs.

kgodey:
1. We should not change the structure of the schema and table APIs. I think it's best to leave the schema API structure as-is to keep the scope of this issue small. We can make further changes in the schema API (like returning a database ID) in future issues if we think it will be an improvement.

[...]

PR #320, opened 2021-07-02
Test importing from multiple databases

### LINKED ISSUE #205: Automatic import of tables when new database is connected.
**Problem**

When a user sets up an existing database in Mathesar, they'll expect to see their tables. We don't currently do anything to sync existing databases with Mathesar.

**Proposed solution**

We should create instances of the appropriate Django models for tables in existing databases.

**Additional context**

Blocked by #61

eito-fis:
Fixes #205 

Adds `test_multi_db.py` alongside several tests to verify that automatic database importing from multiple databases is working.

**Technical details**
Builds a separate `mathesar_multi_db_test` db, modified it and verifies that the changes are properly reflected in our API. The additional db is scoped to the module so as to not affect other tests.

(Also includes a small bug fix for the reflection code, let me know I have to open an issue for it.)

Note: this should be reviewed and merged after #329.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/utils/schemas.py
+++ b/mathesar/utils/schemas.py
@@ -31,6 +31,6 @@ def reflect_schemas_from_database(database):
         for oid in db_schema_oids
     ]
     for schema in Schema.objects.all():
-        if schema.oid not in db_schema_oids:
+        if schema.database == database and schema.oid not in db_schema_oids:
             schema.delete()
     return schemas
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,20 +4,20 @@ about: Report an issue with Mathesar
 labels: "type: bug"
 ---
 
-**Describe the bug**
+## Description
 <!-- A clear and concise description of what the bug is. -->
 
-**Expected behavior**
+## Expected behavior
 <!-- A clear and concise description of what you expected to happen. -->
 
-**To Reproduce**
+## To Reproduce
 <!-- How can we recreate this bug? Please try to provide a Minimal, Complete, and Verifiable (http://stackoverflow.com/help/mcve) example if code-related. -->
 
-**Environment**
+## Environment
  - OS: (_eg._ macOS 10.14.6; Fedora 32)
  - Browser: (_eg._ Safari; Firefox)
  - Browser Version: (_eg._ 13; 73)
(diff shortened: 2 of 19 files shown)


eito-fis:
I've updated this to test multiple db importing instead of just auto importing, as I think most of auto importing is covered by tests added when we implemented reflecting.

For testing, I haven't added cases reflecting changes and deletions since they feel like they overlap pretty heavily with the original reflection tests for schemas and tables - let me know if we want better coverage on this.

mathesar/utils/schemas.py
```diff
@@ -31,6 +31,6 @@ def reflect_schemas_from_database(database):
         for oid in db_schema_oids
     ]
     for schema in Schema.objects.all():
-        if schema.oid not in db_schema_oids:
```
mathemancer, line 34: This makes me realize that we'll probably have a problem if we remove a database.  Wouldn't those schemas just hang around indefinitely, attached to nothing?

kgodey: We could add a `deleted` field to schemas (and databases) and mark them as deleted if we don't see them connected.

mathemancer: Could we solve it by referencing the database in a foreign key from the Schema model, and using `on_delete=models.CASCADE`?  That way, if we notice a DB is missing on reflection, we can delete the DB entry and let it cascade to the schema.  I'd rather avoid building up "deleted" resources in the models tables.

kgodey: that makes sense. I think it would make sense to have a persistent model for the DBs, especially given the new design spec where we show databases that we don't have a connection to with an error message. I think we should mark the DB as deleted rather than deleting the model, though.

kgodey: I think this makes sense to do in https://github.com/centerofci/mathesar/issues/311, not here.

kgodey (approved):
@eito-fis This looks good, I'm going to merge it. Please expand the scope of #311 to include creating a model for databases and handling marking databases as deleted when the connection is no longer available. You'll also need to update the schema and table APIs to only return schemas and tables from non-deleted databases.

PR #321, opened 2021-07-02
Delete and rename schema functions

### LINKED ISSUE #165: Function to handle renaming schemas
**Problem**

Users might want to rename schemas. We don't currently support this.

**Proposed solution**

A function that handles renaming of schemas in the database, including renaming references to them (foreign keys), etc.

**Additional context**

This should be in the `db` module.

### LINKED ISSUE #167: Function to handle deleting schemas
**Problem**

Users might want to delete schemas. We don't currently support this.

**Proposed solution**

A function that handles deleting of schemas in the database. We should raise an error if there is anything outside of the schema referencing the schema.

**Additional context**

This should be in the `db` module.

eito-fis:
Fixes #165, Fixes #167

Adds `delete_schema()` and `rename_schema()` functions to `db.schemas`.

**Technical details**
`delete_schema()` takes an optional `cascade` parameter, which is default `False`. When set, the schema is dropped with the cascade option, deleting all related objects. The `deletes_schema()` function also returns all tables related to the schema being dropped.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
Schema deletion should optionally take `CASCADE` or `RESTRICT` parameters, the user might want to delete everything the schema is associated with.

It would also be good to return a list of everything the schema references if deleting it with `RESTRICT` fails so that we can show the list in the frontend and ask the user if they want to delete all those objects by using `CASCADE`.

eito-fis:
@kgodey Question about:

>It would also be good to return a list of everything the schema references if deleting it with RESTRICT fails so that we can show the list in the frontend and ask the user if they want to delete all those objects by using CASCADE.

Is it sufficient to return the related tables as part of the error message in the response, or is there a better way of returning this information? Right now the function catches the `DependentObjectsStillExist` error, suppresses it and just returns the related tables, but this seems doesn't seem ideal.

kgodey:
@eito-fis That's fine for now, we can improve later if needed.

db/schemas.py
```diff
@@ -71,10 +75,65 @@ def get_all_schemas(engine):
...
+def compile(element, compiler, **_):
+    return "ALTER SCHEMA %s RENAME TO %s" % (
+        element.schema,
+        element.rename_to
+    )
+
+
+def rename_schema(schema, engine, rename_to):
+    """
+    This method deletes a Postgres schema.
```
kgodey, line 135: This comment is inaccurate. :)

db/schemas.py
```diff
@@ -71,10 +75,65 @@ def get_all_schemas(engine):
...
+    """
+    if schema in get_all_schemas(engine):
+        with engine.begin() as connection:
```
kgodey, line 138: Have you verified that this updates references to the schema from other schemas? (e.g. if there's a foreign key to a table in this schema from a different schema)

kgodey: Please add a test to verify behavior.

kgodey (changes requested):

kgodey (approved):
Looks good to me. Since this is entirely `db` focused, I'd like @mathemancer to look at it quickly before merge.

db/schemas.py
```diff
@@ -71,10 +75,65 @@ def get_all_schemas(engine):
...
+    pg_tables = Table("pg_tables", metadata, autoload_with=engine)
+    sel = (
+        select(pg_tables.c.schemaname, pg_tables.c.tablename).
+        where(pg_tables.c.schemaname == schema)
+    )
+    with engine.begin() as conn:
+        results = conn.execute(sel).fetchall()
+
+    related_tables = [tables.reflect_table(table, schema, engine)
+                      for schema, table in results]
```
mathemancer, line 89: You can get these related tables by initializing the metadata with args:
```python
metadata = Metadata(bind=engine, schema=schema)
```
Then, use:
```python
metadata.reflect()
```
Then, you can get all the tables in that schema via
```python
metadata.tables
```

mathemancer (changes requested):
I'm not sure it's wise to drop a schema with dependent objects with a _partial_ list of objects to be deleted in the warning.  This would potentially give the user false confidence, and they might be confused (enraged) if they end up deleting something via a cascade that wasn't listed in the warning.  Currently, things which aren't listed are:
- types
- functions
- higher-order dependencies (e.g., tables that depend on tables in the schema)
- domains
- views
- constraints (likely not a problem)
- sequences
- indexes

There are likely more things I'm forgetting.

I suggest we avoid trying to list dependencies for the moment, and instead just display something like "some database objects depend on this schema.  Delete anyway?"  We could later add a separate `/api/v0/schemas/<schema_id>/dependencies/` endpoint that will list _all_ dependencies for a schema.  This will allow us to check them at will, and also make things like pagination more natural.  (Note that there could be hundreds or thousands of dependencies for a schema, and we'd want to avoid just returning them all in an error message blob).

kgodey:
@mathemancer That makes sense, and I take your point about a partial list of dependencies being confusing.

I'll create a separate issue for the list of dependencies and corresponding API. For now, @eito-fis, please update the error message to be generic as @mathemancer suggested.

eito-fis:
Updated based on feedback:
* Removed the `related_tables` return from `delete_schema`.
* Updated `delete_schema` to raise a `psycopg2.DependentObjectsStillExistError` instead of returning related tables. (Do we have a preference for custom errors, or are existing errors reasonable?)
* Added the `if_exists` option to `delete_schema`.
* Removed checking from `rename_schema` so it breaks as expected when targeting a schema that doesn't exist.

kgodey:
I think existing errors are best.

kgodey (approved):
Approved, I'll wait for @mathemancer to review/merge.

mathemancer (approved):
I think this looks good.  Thank you!

PR #329, opened 2021-07-06
Prevent pytest-django from handling non-'default' dbs

### LINKED ISSUE #318: Test databases in API tests are dropped twice
**Describe the bug**
The `test_db` fixture in our top level `conftest` and `pytest-django` both attempt to drop our test databases during the API tests in `mathesar/tests`.  This normally doesn't cause problems as using the `engine` fixture causes a connection to open that prevents `pytest-django` from dropping the database, allowing `test_db` to drop it later. However, if we create a test in `mathesar/tests` that takes the engine but doesn't use it, we will get  a `database "mathesar_db_test" does not exist` error when `test_db` tries to drop the test database (as `pytest-django` already has).

**Expected behavior**
We should delegate setup and teardown of our test databases to either fixtures or `pytest-django`, and we shouldn't throw an error in the below case.

**To Reproduce**
Create `mathesar/tests/test_reproduce.py` with the following contents:
```python
def test_reproduce(engine):
    pass
```
Then, running `docker exec mathesar_web_1 pytest mathesar/tests/test_reproduce.py` will produce the `database "mathesar_db_test" does not exist` error.

**Additional context**

Overall is a pretty tiny issue. Some notes:

It turns out that under the hood, `pytest-django` [just makes a `django.test.TestCase`](https://github.com/pytest-dev/pytest-django/blob/cf6b229bce0c717bc590d8fd7cef281f3aa2c1fa/pytest_django/fixtures.py#L168) and manually runs setup and teardown from there. As a result, I don't think there's a clean solution where we can get `pytest-django` to ignore our non-django dbs, as [`TestCase` builds all dbs in the settings](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TransactionTestCase.databases).

Possible solutions:
* Make `table_ db` use `DROP IF EXISTS`. This doesn't really solve the underlying problem, but prevents any errors and is maybe just good enough. 
* Move our `db` fixture to the top level `conftest` and delegate all table creation to `pytest-django`. I'm not entirely sure what Django does to setup dbs so this may or may not be feasible. It seems like setting [`MIGRATE = False` ](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.TransactionTestCase.databases) might work? I believe Django by default re-creates the db for every test, which would force some changes to existing `db` tests.
* Create a subclass of `TestCase` that ignores the non-django dbs, and extend `pytest-django` to use it instead. Would leave everything mostly as is, but is probably too much work.

Finally, it turns out the source of
```
db_1   | 2021-06-25 01:45:01.711 UTC [757] ERROR:  database "django_db_test_database" already exists
db_1   | 2021-06-25 01:45:01.711 UTC [757] STATEMENT:  CREATE DATABASE "django_db_test_database" 
```
was from `pytest-django` trying to create the database our fixtures has already made, and
```
db_1   | 2021-06-25 01:44:28.840 UTC [749] ERROR:  database "django_db_test_database" is being accessed by other users
db_1   | 2021-06-25 01:44:28.840 UTC [749] DETAIL:  There are 3 other sessions using the database.
db_1   | 2021-06-25 01:44:28.840 UTC [749] STATEMENT:  DROP DATABASE "django_db_test_database"
```
was from `pytest-django` trying to drop the database while the engine was still active. Mystery solved :)

eito-fis:
Fixes #318

Prevents pytest-django from handling the building or destruction of non-'default' dbs, allowing us to handle our tables databases as needed.

**Technical details**
Overwrites the pytest-django `django_db_setup` fixture to use pass `aliases=['default']` to `django.test.utils.setup_databases()` (See docstring of new fixture for more information). This restricts the setup to only the `default` db, preventing the multiple teardown issue we were running into.

I've confirmed this prevents the error shown in the original issue, but didn't leave the test in as the problem seems niche enough that I don't we'll run into this again.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/tests/conftest.py
+++ b/mathesar/tests/conftest.py
@@ -5,6 +5,40 @@ This inherits the fixtures in the root conftest.py
 import pytest
 
 
+@pytest.fixture(scope="session")
+def django_db_setup(request, django_db_blocker) -> None:
+    """
+    A stripped down version of pytest-django's original django_db_setup fixture
+    See: https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/fixtures.py#L96
+    Also see: https://pytest-django.readthedocs.io/en/latest/database.html#using-a-template-database-for-tests
+
+    Removes most additional options (use migrations, keep / create db, etc.)
+    Adds 'aliases' to the call to setup_databases() which restrict Django to only
+    building and destroying the default Django db, and not our tables dbs.
+    """
+    from django.test.utils import setup_databases, teardown_databases
+
+    with django_db_blocker.unblock():
+        db_cfg = setup_databases(
+            verbosity=request.config.option.verbose,
+            interactive=False,
+            aliases=["default"],
+        )
+
+    def teardown_database() -> None:
+        with django_db_blocker.unblock():
+            try:
+                teardown_databases(db_cfg, verbosity=request.config.option.verbose)
+            except Exception as exc:
+                request.node.warn(
+                    pytest.PytestWarning(
+                        "Error when trying to teardown test databases: %r" % exc
+                    )
+                )
+
+    request.addfinalizer(teardown_database)
+
+
 @pytest.fixture(autouse=True)
 def enable_db_access_for_all_tests(db):
     pass


mathemancer (approved):
Fantastic, thank you!

PR #347, opened 2021-07-08
Database Endpoint

### LINKED ISSUE #311: Create new `/v0/databases/` API

eito-fis:
Fixes #311

Creates a new `Database` model with the `name`, `deleted` and `supported_types` attributes. Also, updates `reflect_db_objects()` to call `update_databases()`, which compares the current `settings.DATABASES` to the `Database` models and updates the models accordingly. Finally, adds a new database viewset with the `list()` and `detail()` methods. 

 Additionally, updates the `Schema` model to use a `ForeignKey` for it's `database` field, and moves the `_sa_engine` property up to the `Database` model. Note that the `database` field change required the modification of a number of tests unrelated to the PR.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
@kgodey , @mathemancer : Some questions about how types work in Mathesar:
* I'm using `get_supported_alter_column_types()` to get calculate the `supported_types` field for the database models - is the the right way to do it? Or should I write something else?
* Why do we call `_add_custom_types_to_engine()` for the engines created with `create_mathesar_engine()` (via `create_engine_with_future_types()`) ? It seems like we need to call `install_mathesar_on_database()` to actually use the custom types, so I don't understand why we seem to only do part of the installation.

The main problem I'm running into right now is that `get_supported_alter_column_types()` always returns custom types as databases have `_add_custom_types()` called on them when the engine is created. I'm not sure this is proper behavior, so clarification would be apprecaited.

kgodey:
@mathemancer has a better handle on types than me at the moment, so I'll defer to him.

mathesar/database/utils.py
```diff
@@ -21,3 +22,20 @@ def get_database_key(engine):
...
+def update_databases():
+    databases = set(settings.DATABASES)
+
+    # Update deleted databases
+    for database in Database.objects.all():
+        if database.name in databases:
+            databases.remove(database.name)
+        else:
+            database.deleted = True
+            Schema.objects.filter(database=database).delete()
```
kgodey, line 36: I assume this also deletes `Table` objects.

eito-fis: Yep, confirmed in the database api tests.

eito-fis:
Side note: Have we considered using the database endpoint to manage the databases instead of using the env file? With the database model in place I don't think it would be a crazy amount of work - something like:
* Add creation and deletion endpoints for databases
* Replace  calls to `create_mathesar_engine(database_key)` outside the database model with `database._sa_engine` (Most of the work here is probably fixing broken tests)
* Move `settings.DATABASES` attributes to the database model (or don't, and just store the db url), and update `create_mathesar_engine()` accordingly

Far out of scope for this PR and probably not something critical, but might be better for the frontend's interaction with the dbs?

kgodey:
@eito-fis We will need to do something like that eventually to make connecting to databases easier. 

The main issue is storing the database password. We need access to the database password in plaintext to access the database, but we _do not_ want to store the password in plaintext in the database. We need to figure out a way to store the password encrypted but not have the user need to enter it all the time to decrypt it.

It didn't seem like an important enough feature to figure out for the MVP, but if you have ideas on how to handle the password issue, we can consider doing it sooner.

mathemancer:
> @kgodey , @mathemancer : Some questions about how types work in Mathesar:
> 
> * I'm using `get_supported_alter_column_types()` to get calculate the `supported_types` field for the database models - is the the right way to do it? Or should I write something else?
> * Why do we call `_add_custom_types_to_engine()` for the engines created with `create_mathesar_engine()` (via `create_engine_with_future_types()`) ? It seems like we need to call `install_mathesar_on_database()` to actually use the custom types, so I don't understand why we seem to only do part of the installation.
> 
> The main problem I'm running into right now is that `get_supported_alter_column_types()` always returns custom types as databases have `_add_custom_types()` called on them when the engine is created. I'm not sure this is proper behavior, so clarification would be appreciated.

Yes, `get_supported_alter_column_types` is the correct method for now.  It will automatically include or exclude the custom types if they're also included in the given engine. If you need different behavior, please don't alter that function since I'm currently giving it a major overhaul.  For now, just use it or write a wrapper for it if needed.

There are 3 places where custom types need to be dealt with:

1. The actual PostgreSQL types need to be created and installed on the database.  Note that this is database by database, so even if the custom types are installed on a given database on a DB server, they won't automatically be installed on new DBs created on that server.  This is accomplished for the email type by the `create_email_type` function in the `db.types.email` module.
2. The Python class that represents that type in SQLAlchemy needs to be created, and needs to be defined so that it hooks up to the underlying DB type.  This is accomplished by the `Email` class in the above module.
3. The type needs to be hotpatched into any engine that needs to use it.  This associates the Python class with the DB type for reflection.  Note that for creating columns and adding them to tables, you wouldn't need to add the type to the engine, since the class knows what its underlying DB type is.  For reflection, however, the engine needs a way to map an underlying DB type that it sees to some Python class.  Unfortunately, adding the type to an engine can only be done once the engine is instantiated with the `postgresql` dialect.  (Recall that SQLAlchemy engines support multiple underlying DB protocols).  This is because SQLAlchemy engines actually _only_ support adding custom types for reflection for the PostgreSQL dialect, and it's not very well supported.

I hope that clarifies things.

eito-fis:
@kgodey That makes sense, thanks.

@mathemancer This clears up a lot of things, but I'm not entirely sure what to do with the current implementation. As of now, since `_add_custom_types_to_engine` is always called on the engine, `get_supported_alter_column_types` returns that custom types are installed. But, this seems off to me? My understanding is that the rest of the installation steps you outlined also have to be called to support custom types, but we seem to return that custom types are installed regardless of the additional steps. 

I guess my primary question is whether the current behavior, where we'll always report that the `email` type is installed, is correct.

mathemancer:
@eito-fis I think I understand the problem now.

When the functions to determine which types are supported were written, the assumption was that we'd always run the `install.py` script on a DB for which `mathesar` was to be used.  I think we should still use that assumption, since the amount of functionality we'd lose and have to work around would be a big problem if we can't run that script on a given DB prior to using it for Mathesar tables.  I consider the state of the DB w.r.t. the types being installed to be relatively stable, and reliable.

On the other hand, the _engine_ being used at any given moment is quite volatile.  For example, types can be loaded into and removed from an engine after it's instantiated, and this can modify what types are available for that engine.  Thus, double-checking that the types are installed on the engine is more crucial than checking in the DB for those types being installed.

I'm currently wrenching around with _exactly that function_ (and will probably change its increasingly-inappropriate name; beware).  I'll consider adding a check to see what types are installed on the DB as well as in the engine to that function.  My issue with that is that we should really do that check _once_ when a DB is added to Mathesar (or as part of some installation/upgrade of Mathesar process), not every time.  If we fail for some reason because the Mathesar installation has been undone or modified in some incompatible way under the feet of the webservice, I'm okay with that.  Manual intervention from an admin will and should be required at that point (IMO).

@kgodey Any thoughts on that?

kgodey:
@mathemancer I don't think we should assume that Mathesar will only work with databases that have `install.py` run on them. Currently, `install.py` installs the email type. Some users may not need (or be unwilling to) install our custom types, and we should not force it on them. Most of Mathesar's functionality does not depend on custom types being available, custom types just allow them to do more with their data. Let me know if there are additional things that `install.py` does that I'm missing.

We don't have to make changes to the assumption that `install.py` will be used in this PR though, the scope of this PR is already pretty large. Let's create a separate issue for it.

eito-fis:
Where does this leave us with respect to the `supported_types` return value? Am I alright to just remove the test for the database without types installed and open this for review?

kgodey:
@eito-fis Yes, go ahead. We'll make separate issues for modular installation later.

kgodey:
@eito-fis Attempting to run `./manage.py migrate` locally results in:

```
  File "/code/./manage.py", line 22, in <module>
    main()
  File "/code/./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 85, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 243, in handle
    post_migrate_state = executor.migrate(
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 117, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 227, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 124, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/fields.py", line 236, in database_forwards
    schema_editor.alter_field(from_model, from_field, to_field)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 571, in alter_field
    self._alter_field(model, old_field, new_field, old_type, new_type,
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/postgresql/schema.py", line 166, in _alter_field
    super()._alter_field(
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 721, in _alter_field
    self.execute(
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 142, in execute
    cursor.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 98, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.DataError: invalid input syntax for type integer: "mathesar_tables"
```

kgodey:
@eito-fis It happens while applying the first migration. It looks like it might be because you're altering the database field to a foreign key from a string. I found https://code.djangoproject.com/ticket/25012 which seems somewhat related.

I recommend removing the existing migrations and redoing the migrations so that you rename the old field, add the new field, and drop the new field instead. That should avoid tricky alterations.

kgodey (changes requested):
I'll re-review once I can run this locally successfully, which involves fixing migrations.

eito-fis:
@kgodey Just to be clear, is the goal to be able to run everything outside of the docker container? I've only worked in the container so far, but can work on getting stuff setup locally.

kgodey:
@eito-fis Mathesar should not depend on Docker but for the purposes of this issue, I'm working in the Docker container too. When I said "locally", I meant in the `mathesar_web_1` Docker container. Sorry for the confusion.

eito-fis:
@kgodey Re-did the migration to handle existing models, hopefully fixes the issue?

mathesar/serializers.py
```diff
@@ -59,6 +70,15 @@ def to_representation(self, instance):
         return instance._asdict()
 
 
+class DatabaseSerializer(serializers.ModelSerializer):
```
kgodey, line 73: The entire serializer should be read only. Databases should not be able to be edited or updated anywhere but in the settings.

mathesar/views/api.py
```diff
@@ -218,6 +225,14 @@ def list(self, request):
         return Response(get_non_default_database_keys())
 
 
+class DatabaseViewSet(viewsets.GenericViewSet, ListModelMixin, RetrieveModelMixin):
+    def get_queryset(self):
+        reflect_db_objects()
+        return Database.objects.all().order_by('-created_at')
```
kgodey, line 231: The `default` database should not be included since it's Mathesar's internal database, it's not managed by Mathesar.

kgodey (changes requested):
@eito-fis The new migrations work, thanks!

Looks good overall, I added some comments. I think it would also be good to add a filter for `deleted` so that clients can only get active databases in an API call if needed.

kgodey:
@eito-fis Migrating backwards is failing.

```
root@819bd6177ae4:/code# ./manage.py migrate mathesar 0009
Operations to perform:
  Target specific migration: 0009_auto_20210618_1031, from mathesar
Running migrations:
  Rendering model states... DONE
  Unapplying mathesar.0014_auto_20210713_2331... OK
  Unapplying mathesar.0013_remove_schema_database...Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: column "database" of relation "mathesar_schema" contains null values


The above exception was the direct cause of the following exception:

  File "/code/./manage.py", line 22, in <module>
    main()
  File "/code/./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 85, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 243, in handle
    post_migrate_state = executor.migrate(
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 121, in migrate
    state = self._migrate_all_backwards(plan, full_plan, fake=fake)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 196, in _migrate_all_backwards
    self.unapply_migration(states[migration], migration, fake=fake)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 251, in unapply_migration
    state = migration.unapply(state, schema_editor)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 175, in unapply
    operation.database_backwards(self.app_label, schema_editor, from_state, to_state)
  File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/fields.py", line 172, in database_backwards
    schema_editor.add_field(from_model, to_model._meta.get_field(self.name))
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 487, in add_field
    self.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/base/schema.py", line 142, in execute
    cursor.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 98, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.9/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: column "database" of relation "mathesar_schema" contains null values
```

eito-fis:
@kgodey, sorry about that - updated to be able to migrate backwards, but we have to set `mathesar_tables` as the default value since data is lost on the forward pass.

kgodey:
@eito-fis That's fine for now, we're planning to reset migrations before the first release so we just need to make it work for local/staging until then.

eito-fis:
Updated based on feedback:
* All serializer field are read-only.
* `default` table is excluded when creating the initial db models.
* `deleted` filter added to the database list endpoint.

kgodey (approved):
Looks great!

PR #353, opened 2021-07-12
Fix grouping counts

### LINKED ISSUE #352: Undesirable record grouping behaviours
## Description
Record grouping has a set of behaviours, that are not desirable.
* It considers order_by, which leads to formation of incorrect query on the backend, if we don't group by the sorted column.
Screenshot 2021-07-13 at 12 05 01 AM

* It considers limit and offset. These apply on the grouped result itself, and is unrelated to the record limit & offset.
Screenshot 2021-07-13 at 12 02 20 AM
Screenshot 2021-07-13 at 12 02 42 AM

## Expected behavior
* It should not consider order_by.
* It should not consider limit and offset.

We could also probably have a dedicated API for this. It could also obtain the values for columns, to filter the grouped results. Having it as part of records API makes less sense, since the group count is not a reflection of the record results.

eito-fis:
Fixes #352

Updates the record endpoint to have proper grouping behaviors.

**Technical details**
Updates the `get_group_counts()` function to create a subquery with the appropriate parameters applied, before grouping on the result. Also updates the limit offset tests and adds a new filtering + grouping test.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/pagination.py
+++ b/mathesar/pagination.py
@@ -67,7 +67,7 @@ class TableLimitOffsetGroupPagination(TableLimitOffsetPagination):
                 filters=filters, order_by=order_by
             )
             # Convert the tuple keys into strings so it can be converted to JSON
-            group_count = [{"columns": list(cols), "count": count}
+            group_count = [{"values": list(cols), "count": count}
                            for cols, count in group_count.items()]
             self.group_count = {
                 'group_count_by': group_count_by,
--- a/mathesar/pagination.py
+++ b/mathesar/pagination.py
@@ -67,7 +67,8 @@ class TableLimitOffsetGroupPagination(TableLimitOffsetPagination):
                 filters=filters, order_by=order_by
             )
             # Convert the tuple keys into strings so it can be converted to JSON
-            group_count = {','.join(k): v for k, v in group_count.items()}
+            group_count = [{"columns": list(cols), "count": count}
+                           for cols, count in group_count.items()]
             self.group_count = {
                 'group_count_by': group_count_by,
                 'results': group_count,
--- a/db/records.py
+++ b/db/records.py
@@ -96,10 +96,8 @@ def get_group_counts(
         if field_name not in table.c:
             raise GroupFieldNotFound(f"Group field {field} not found in {table}.")
 
-    group_by = _create_col_objects(table, group_by)
     query = (
-        select(*group_by, func.count(table.c[ID]))
-        .group_by(*group_by)
+        select(table)
         .limit(limit)
         .offset(offset)
     )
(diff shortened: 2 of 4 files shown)


kgodey:
@eito-fis Please see the new format suggestion in the issue (comma separation of values could be confusing, since values can have commas in them).

mathesar/pagination.py
```diff
@@ -67,7 +67,8 @@ def paginate_queryset(self, queryset, request, table_id,
                 filters=filters, order_by=order_by
             )
             # Convert the tuple keys into strings so it can be converted to JSON
-            group_count = {','.join(k): v for k, v in group_count.items()}
+            group_count = [{"columns": list(cols), "count": count}
```
kgodey, line 70: @eito-fis On second thought, I think changing this key to `values` instead of `columns` makes more sense, these aren't column names.

kgodey (commented):
@eito-fis I tested this locally by going to `http://localhost:8000/api/v0/tables/2/records/?group_count_by%3D[%22Status%22]` in my browser and I'm getting 

```
    "group_count": {
        "group_count_by": null,
        "results": null
    },
```

It should be returning the group count for the `Status` column, right? Am I accessing the API correctly?

eito-fis:
> @eito-fis I tested this locally by going to `http://localhost:8000/api/v0/tables/2/records/?group_count_by%3D[%22Status%22]` in my browser and I'm getting
> 
> ```
>     "group_count": {
>         "group_count_by": null,
>         "results": null
>     },
> ```
> 
> It should be returning the group count for the `Status` column, right? Am I accessing the API correctly?

That's strange, running `client.get(f'/api/v0/tables/{table.id}/records/?group_count_by=["Status"]')` in a test returns:
```python
'group_count': {
    'group_count_by': ['Status'],
    'results': [
        {'values': ['Issued'], 'count': 48},
        {'values': ['Application'], 'count': 2}
    ]
}
```
Are the tests passing for you?

kgodey:
@eito-fis The tests are passing for me. Do you get the same result when you go the URL in your browser?

kgodey:
@eito-fis I figured out the issue. When I replace the `%3D` with `=`, it works fine. It looks like you'll need to handle URL encoding correctly.

kgodey (changes requested):
@eito-fis Looks good overall, nice work. Please handle URL encoding correctly and also add a test for it so that we don't break it in the future.

eito-fis:
@kgodey After looking around a bit, I think this is desired behavior? it seems like [the `=` character is reserved to separate keys and values in the url](https://developers.google.com/maps/url-encoding), and shouldn't be encoded when being used for that purpose.

kgodey:
@eito-fis I think you're correct. I probably ended up with that URL because I mistyped the query to begin with and my browser got confuse by the `=`. Sorry for the false alarm.

kgodey (approved):

PR #354, opened 2021-07-13
Delete and rename tables

### LINKED ISSUE #183: DB function to rename tables
**Problem**

Users might want to rename tables. We don't currently support this.

**Proposed solution**

A function that handles renaming of tables in the database, including renaming references to them (foreign keys), etc.

**Additional context**

This should be in the `db` module.

### LINKED ISSUE #186: DB function to handle deleting tables
**Problem**

Users might want to delete tables. We don't currently support this.

**Proposed solution**

A function that handles deleting of tables in the database. We should raise an error if there is anything outside of the table referencing the table.

**Additional context**

This should be in the `db` module.

eito-fis:
Fixes #183, Fixes #186

Adds the `delete_table()` and `rename_table()` functions to `db.tables`.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/tables.py
```diff
@@ -52,6 +54,19 @@ def create_mathesar_table(name, schema, columns_, engine, metadata=None):
     return table
 
 
+def delete_table(name, schema, engine):
```
kgodey, line 57: How does this handle references to this table from other tables? e.g. another table might have foreign keys pointing to this one.

db/tables.py
```diff
@@ -52,6 +54,19 @@ def create_mathesar_table(name, schema, columns_, engine, metadata=None):
...
+    table = reflect_table(name, schema, engine)
+    table.drop()
+
+
+def rename_table(name, schema, engine, rename_to):
```
kgodey, line 62: Same question about handling references to this table from other tables.

eito-fis:
Updated tests to handle foreign key cases, and also added the `cascade` and `if_exists` options to the `delete_table` function.

db/tables.py
```diff
@@ -52,6 +55,42 @@ def create_mathesar_table(name, schema, columns_, engine, metadata=None):
...
+@compiler.compiles(DropTableCascade, "postgresql")
+def compile_drop_table(element, compiler, **_):
+    expression = compiler.visit_drop_table(element)
+    if element.cascade:
+        return expression + " CASCADE"
+    else:
+        return expression
+
+
+def delete_table(name, schema, engine, cascade=False, if_exists=False):
```
kgodey, line 74: If `cascade` is `False`, can you return a list of related items just like with schemas?

kgodey (changes requested):
Looks good overall, just one question. Also, since this is entirely `db` focused, I'd like @mathemancer to look at it quickly before merge.

mathemancer (commented):
Similarly to schemata, I think we should have a separate endpoint at something like:
```
/api/v0/tables/<table_id>/dependencies/
```
That would return a paginated set of dependent objects.  There could be many, and I think the pagination would be a good idea.  I suggest trying to use the `pg_depends` table (view?) to crawl the dependency graph.  The reason is that it would be highly undesirable to list some but not all dependent objects, since this could give the user false confidence, and result in accidental deletion of objects they don't want to get rid of.

`CASCADE` is an extremely aggressive flag, and care should be taken here.  I.e., `CASCADE` will follow the dependency graph as far as it goes (and as far as permissions allow).

kgodey:
ll create a separate issue for the list of dependencies and corresponding API. For now, @eito-fis, please update the error message to be generic as @mathemancer suggested in the other issue.

eito-fis:
Similar update to the schema PR, just modified `delete_table` to raise the original `DependentObjectStillExists` error.

kgodey (approved):

kgodey:
I'm going to merge this since @mathemancer didn't request changes.

PR #370, opened 2021-07-14
Allow table deletion and renaming via the API

### LINKED ISSUE #184: Allow users to edit table name via API
**Problem**

Users might want to rename tables. We don't currently support this.

**Proposed solution**

Extend `TableViewSet` to support editing table names.

Blocked by #183

### LINKED ISSUE #187: Allow table deletion via API
**Problem**

Users might want to delete tables. We don't currently support this.

**Proposed solution**

Extend `TableViewSet` to support deleting tables.

**Additional context**

Blocked by #186

eito-fis:
Fixes #184, Fixes #187

Exposes the ability to delete and rename tables via the API.

**Technical details**
Adds the `partial_update` and `delete` methods to the `TableViewset` which rely on the `rename_table` and `delete_table` functions respectively. The partial update method only supports a `name` parameter to update, and the delete method support an additional `cascade` url parameter. The delete method also returns a 400 response with `Cannot delete, dependent database objects exist` if cascade is not set and dependent objects exist.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Some questions:
* Had to change `from mathesar.models import Table` to `import mathesar.models as models` in `mathesar/utils/tables.py` due to a circular import with `mathesar/models.py`. Is there a nicer way to solve this, or is this just something to live with?
* It turns out, serializers work just find for URL parameters - I'd like to go back and change the record endpoint to use a serializer instead of a form for processing the filter params. Can I just open a small PR for this? Or does it need an issue attached?

mathesar/utils/tables.py
```diff
@@ -29,3 +34,15 @@ def get_table_column_types(table):
...
+def update_sa_table(table, validated_data):
+    for arg in validated_data:
+        if arg not in SUPPORTED_TABLE_UPDATE_ARGS:
+            raise ValidationError({arg: f'Updating {arg} for tables is not supported.'})
+    if 'name' in validated_data:
+        rename_table(table.name, table.schema.name, table.schema._sa_engine,
+                     validated_data['name'])
+        # Force the table and nameto reload
+        del table._sa_table
+        del table.name
```
mathemancer, line 48: I think we should either rework this function to only validate, then call an `update_table` in the `db.tables` module, or move the `update_column` from the `db.columns` module to a new `mathesar.utils.columns` module.  My preference is the former, since I think that `update_table` function will be useful outside of the Mathesar webservice.

mathemancer (changes requested):
I think most of the `update_table` function should be moved to the `db.tables` module.  My general rule is that anything that's not specific to the web service itself should be over there, since we may eventually want to ship that module by itself, and the more functionality we put there the better (unless that functionality requires imports of any django bits or pieces).

mathesar/utils/tables.py
kgodey: I agree with choosing the former.

kgodey:
> Had to change `from mathesar.models import Table` to `import mathesar.models as models` in `mathesar/utils/tables.py` due to a circular import with `mathesar/models.py`. Is there a nicer way to solve this, or is this just something to live with?

You need to structure your code so that you're either: 
- not using `utils` in the `models` fiie
- not using `models` in the `utils` file
- create a separate `utils` file for utility functions you'll be using in the `models`, and ensure it doesn't import any `models`

> It turns out, serializers work just find for URL parameters - I'd like to go back and change the record endpoint to use a serializer instead of a form for processing the filter params. Can I just open a small PR for this? Or does it need an issue attached?

Just make a PR for this, no issue is needed.

eito-fis:
Updated in response to feedback:
* Moved most of the logic for `update_sa_table` to `db.tables`.
* Move `update_sa_tables` to the `utils/models.py` file to avoid circular import issues.

mathesar/views/api.py
```diff
@@ -85,6 +85,34 @@ def create(self, request):
...
+        table = self.get_object()
+        table.update_sa_table(serializer.validated_data)
+
+        # Reload the table to avoid cached properties
+        table = self.get_object()
+        serializer = TableSerializer(table, context={'request': request})
+        return Response(serializer.data)
+
+    def destroy(self, request, pk=None):
+        serializer = TableDeleteParameterSerializer(data=request.GET)
```
kgodey, line 104: Using parameters in `DELETE` endpoints is not RESTful. Here's a good [Stack Overflow conversation about it](https://stackoverflow.com/questions/2539394/rest-http-delete-and-parameters).

`DELETE` should not accept parameters and use the `cascade` flag. We should have a separate API flow for getting dependencies and rely on the frontend to do that before calling `DELETE`.

I recommend adding a `has_dependencies`: `True` flag to the `TableSerializer` so that the Table API lists whether a table has dependencies. That should suffice until we add the new `dependencies` endpoint that @mathemancer suggested.

kgodey (changes requested):

mathesar/views/api.py
eito-fis: I didn't know that about `DELETE` - will do, thanks.

What is end use case the the `has_dependency field`? I'm imagining it being `True` or `False` dependent on the dependency data, then the frontend hitting the dependency endpoint for additional details if necessary.

kgodey: Yes, that was the use case I was imagining as well. It's also useful before we have the dependency endpoint, the frontend can at least show a generic warning that says that dependencies exist.

mathemancer (changes requested):
For reasons I can't quite understand, when updating the name of a table via the API, I end up with two copies of the table, one with the old name, and one with the new.  @kgodey @eito-fis can you confirm this bug?

kgodey:
@mathemancer I'm unable to reproduce. I did a `PATCH` request to `/api/v0/tables/2/` with `{"name": "new name"}` and the table name seems to have been successfully updated.

mathesar/models.py
```diff
@@ -112,6 +112,10 @@ def sa_columns(self):
     def sa_column_names(self):
         return self.sa_columns.keys()
 
+    @property
+    def has_dependencies(self):
```
kgodey, line 116: @eito-fis Could you update this to actually check if it has dependencies? If that's going to be a lot of work, we can separate it into a new issue, but we should at least add a `TODO` here that indicates it's a placeholder and needs to be replaced.

kgodey (changes requested):

eito-fis:
@mathemancer I haven't been able to reproduce in tests, only see the renamed table after the API call.

@kgodey I took a shot at this by just trying to delete and seeing if we fail or not. Not sure if this is optimal, but might be good enough for now? I'll add it to the schema PR if this looks good.

kgodey:
@eito-fis For now, let's just add a `TODO` to unblock merge. I'm uncomfortable with the idea of trying to delete it, preserving user data is very important and I don't think having a call to delete is a good thing to do programmatically. There's too much of a risk of bugs being introduced at a later stage even if it works fine for now.

https://stackoverflow.com/questions/4462908/find-dependent-objects-for-a-table-or-view has a query to find dependencies but it seems complicated enough that it's worth putting into a separate issue.

mathemancer:
Try as I might, I can't seem to reproduce my own table duplication problem today.  Perhaps I had some state lying around in my docker containers, and completely rebuilding solved it.

mathemancer:
> @eito-fis For now, let's just add a `TODO` to unblock merge. I'm uncomfortable with the idea of trying to delete it, preserving user data is very important and I don't think having a call to delete is a good thing to do programmatically. There's too much of a risk of bugs being introduced at a later stage even if it works fine for now.
> 
> https://stackoverflow.com/questions/4462908/find-dependent-objects-for-a-table-or-view has a query to find dependencies but it seems complicated enough that it's worth putting into a separate issue.

Beware, that query is from a very old version of PostgreSQL, and doesn't work with the modern information schema spec.  I'll start a GH discussion about how to safely deal with dependent objects.  I have a couple options in mind, but I think we should put off handling it to a separate issue.  As @kgodey noted, we want to be _very_ careful when deleting user data.

mathemancer (approved):
I think this looks good, once the `TODO` requested by @kgodey is added.  Please don't merge until that's done, and she's approved.

kgodey (approved):
The TODO is done.

PR #415, opened 2021-07-17
Create table from paste

### LINKED ISSUE #190: Allow user to create a new table via pasting from spreadsheet
**Problem**

Users may want to create a new table by copy/pasting data from a spreadsheet or CSV.

**Proposed solution**

We should extend our table creation API to take in raw pasted data and create a table.

eito-fis:
Fixes #190

Allows users to upload textual data from copy pasting text from a spreadsheet. 

**Technical details**
Adds the `paste` parameter to the data file endpoint. If both `file` and `paste` or neither are specified, an error is returned.

Google Sheets and Libre Office both export spreadsheets to clipboard as `.tsv` files. As a result, all we have to do to process the pasted text is to save it as a file, then process it as we would any other `.*sv` file.

Other changes:
* ~Table creation logic was moved to the `utils/tables.py` file, where all table creation is now handled. In particular, the viewset calls `create_table_from_data`, which chooses to create a table from data file or an empty table based on the passed parameters.~
* The `test_data_file_create_*` tests in `test_data_file_api.py` have been split into multiple parts. Tests that check if the correct dialect is generated for a file are now in `imports/test_csv.py` and directly test `get_sv_dialect()`. The tests in `test_data_file_api.py` only ensure the data file fields match the dialect.
* Table creation tests were generally updated to also ensure the correct column name and row values were created.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Things that should still happen for this PR:
* The current parsing logic is naive, and could benefit from some additional research into how different spreadsheet software produce the copy paste text. In particular, I'm curious how tabs inside the values are handled, and how empty fields are handled. I've noticed some weird pattern with empty fields for Google Sheets (tab followed by a single space?) but we don't handle it yet. The current test data also doesn't include missing values, so it also needs to be updated.
* `create_records_from_paste` could probably be renamed to something closer to `create_records_from_lists` or  something. Could also use some direct unit tests.
* Need more tests to ensure the paste fails when we expect it to, and works in tricky situations.
* Need tests to ensure we fail when data file + paste are passed, and when neither are passed.

kgodey:
I'm going to leave this assigned to @eito-fis since I don't think it will block anybody by the time he's back from vacation.

eito-fis:
This is almost ready for review - I just have to add handling for a `header` option and it should be good. However, I am realizing that there might be a better way to do this, and was hoping for some feedback.

Currently, we take in the paste data as an argument to the create table endpoint. This is in contrast to creating a table from a csv, which requires the creation of a data file first. It occurred to me that we could handle pastes the same way, where we create a data file from the paste and create the table from that data file instead. This has the added benefits of:
* Giving us handling of `header` for free
* Not requiring additional parsing logic, since paste data is in the same format as `.tsv` files
* Not requiring additional handling on the table creation endpoint

On the other hand, this would require redoing a little more than half of this PR, so I figured I would ask before going ahead. (And we might not want to store every paste a user uploads, depending on what we're planning for the UX of this?)

kgodey:
@eito-fis I think that making the paste a data file makes the most sense, it'll cut down on repeated code (which generally helps prevent bugs). I'd add a boolean argument like `created_from_paste` to the data file model and set it to `True` for pasted data to make the provenance clear.

As far as the UX, we're not currently showing data files in the UI, so there are no UX concerns at the moment. Even when we do show files in the UI, having the `created_from_paste` argument will help us distinguish files that were created by pasting in the UI, which should be helpful.

eito-fis:
Updated to use data files instead of directly passing data to the table creation endpoint - should be good for review now.

mathesar/serializers.py
```diff
@@ -93,16 +95,20 @@ class DataFileSerializer(serializers.ModelSerializer):
         default=serializers.CurrentUserDefault(), read_only=True
     )
     header = serializers.BooleanField(default=True)
+    paste = serializers.CharField(required=False, trim_whitespace=False)
```
kgodey, line 98: I don't see a field on the model that stores how the file was created. You should store something on the model so that you can always tell when using a data file object whether it was created by uploading a file or pasting it. For example, we might one day want to have a page that shows all the files that the user has created, including information on whether it was uploaded or pasted in.

 I'd also recommend naming the field on the model something a little more verbose like `from_paste` or `created_from_paste` if it's a boolean. Alternatively, you could also make it a `creation_method` field with choices of paste or upload (that's extendable to other methods, like getting it from a remote URL).

mathesar/utils/datafiles.py
```diff
@@ -1,41 +1,44 @@
...
-        data_file = data['data_files'][0]
-    else:
-        raise ValidationError({'data_files': 'Multiple data files are unsupported'})
-
-    table = create_table_from_csv(data_file, name, schema)
-    return table
+def create_datafile(request, data):
+    header = data.get('header', True)
 
+    if 'paste' in data and 'file' in data:
```
kgodey, line 17: I should have caught this in previous PRs, but validation like this belongs in the serializer's `validate_<blah>` and `validate` methods. See:
https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation
https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation

mathesar/utils/tables.py
```diff
@@ -16,17 +22,37 @@ def get_table_column_types(table):
...
+    serializer = TableSerializer(table, context={'request': request})
+    return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+def create_table_from_datafile(data_files, name, schema):
+    if len(data_files) == 1:
+        data_file = data_files[0]
+        table = create_table_from_csv(data_file, name, schema)
+    elif len(data_files) > 1:
+        raise ValidationError({'data_files': 'Multiple data files are unsupported.'})
```
kgodey, line 43: This validation should go in the serializer too.

mathesar/utils/tables.py
```diff
@@ -16,17 +22,37 @@ def get_table_column_types(table):
     return col_types
 
 
-def create_empty_table(data):
+def create_table_from_data(request, data):
```
kgodey, line 25: I think this should go directly in the viewset instead of being a separate function/file, since you're using DRF responses.

mathesar/utils/tables.py
```diff
@@ -16,17 +22,37 @@ def get_table_column_types(table):
...
+    name = data['name']
```
kgodey, line 26: I'm not sure that you need separate variables for these, you can just use them directly. It's not like they need to be calculated again if they're reused, you're just accessing a key in a dictionary.

kgodey (changes requested):
@eito-fis Nice work! I requested a few changes.

mathesar/serializers.py
```diff
@@ -67,6 +70,11 @@ def get_records(self, obj):
         else:
             return None
 
+    def validate(self, data):
```
kgodey, line 73: Since this only touches a single field, `data_files`, I think it should be in a `validate_data_files` method.

kgodey (approved):
Looks good, just one minor comment. Please resolve before merge.

PR #476, opened 2021-07-22
Frontend build

### LINKED ISSUE #358: Serve built static files in production
## Problem

Currently, the frontend is only set up for local development and depends on the Vite server running at port 3000 to serve the frontend. This will not work for production, since we'll have pre-built static files.

## Proposed solution

We need to update the Django templates to use the existing setup for development environments and serve pre-built static files for production environments (or if the vite server is not running).

## Additional context

This is required for https://github.com/centerofci/mathesar/issues/225

pavish:
Fixes #358

This PR:
* Sets up client build for running Mathesar in production mode
* Combines ui and web services to a single service
* Runs as non-root user in docker

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


.env.example
```diff
@@ -1,6 +1,7 @@
 ALLOWED_HOSTS=*
 DEBUG=True
+MODE=DEVELOPMENT
 SECRET_KEY=2gr6ud88x=(p855_5nbj_+7^bw-iz&n7ldqv%94mjaecl+b9=4
 DJANGO_DATABASE_KEY=default
 DJANGO_DATABASE_URL=postgres://mathesar:mathesar@db:5432/mathesar_django
-MATHESAR_DATABASES=(mathesar_tables|postgresql://mathesar:mathesar@db:5432/mathesar)
+MATHESAR_DATABASES=(mathesar_tables|postgresql://mathesar:mathesar@db:5432/mathesar)
```
kgodey, line 7: Please add the newline back to the end of the file.

config/settings.py
```diff
@@ -150,9 +152,23 @@ def pipe_delim(pipe_string):
...
 
 STATIC_URL = "/static/"
 
+# When running with DEBUG=False, the webserver needs to serve files from this location
+# python manage.py collectstatic has to be run to collect all static files into this location
+# The files need to served in brotli or gzip compressed format
+STATIC_ROOT = os.path.join(BASE_DIR, 'static/')
+
+# Location of client build
+CLIENT_BUILD_LOCATION = os.path.join(BASE_DIR, 'client_build/')
```
kgodey, line 165: Why not use `STATIC_ROOT` for this? Why have a separate setting?

mathesar/templatetags/client_files.py
```diff
@@ -0,0 +1,122 @@
...
+
+MODE = getattr(settings, "MODE")
+CLIENT_BUILD_LOCATION = getattr(settings, "CLIENT_BUILD_LOCATION")
+MANIFEST = os.path.join(CLIENT_BUILD_LOCATION, "manifest.json")
+
+STATIC_URL = getattr(settings, "STATIC_URL")
+CLIENT_DEV_URL = getattr(settings, "CLIENT_DEV_URL")
+
+
+class ClientAssetHandler:
```
kgodey, line 21: Could you add some comments to this method that explain what it's doing for people that are not familiar with frontend code structure?

.env.example
pavish: It seems like dotenv is reading the newline and throwing a warning saying that the line has invalid env variable configuration. I removed it to avoid the warning. I will add it back for consistency.

config/settings.py
pavish: Django looks at directories in `STATICFILES_DIRS` to find static assets. It also does not allow `STATIC_ROOT` to be placed within `STATICFILES_DIRS`.

The client build location has to be present in `STATICFILES_DIRS` inorder for the assets to be served when DEBUG=False. So I've used a different location.

As an added note: When serving with DEBUG=False, Django does not serve assets on it's own, so we'll have to collate all static files into the STATIC_ROOT folder, and configure the webserver we intend to use, to serve those files.

mathesar/templatetags/client_files.py
pavish: I will add comments and tests for this file.

config/settings.py
kgodey: Sorry, I didn't phrase my question well. I meant, why not build the client into the existing `STATIC_ROOT` folder?

andyharley-cci:
Pushed a change to this branch to get the python tests running successfully. While running this on a forked repo for testing the code is checked out with a uid of `1001`. Our user in the container runs as `1000`, so the mounted volumes are not writable. The two options around this are to do a `chown` across all the files after checkout or set the uid of the mathesar docker user to `1001`.I initially went for the latter but then realised it is probably succeeding locally because our local user ids default to being `1000`. So instead, I've opted for adding a `chown` step in the github action to set the permissions appropriately.

config/settings.py
pavish: Django cannot serve the STATIC_ROOT folder, so if we built the client into the STATIC_ROOT folder, it will not be possible to run it locally with Debug=True, without vite server.

It would work if we use Debug flag to determine the mode as Development or Production, but we are using a separate flag for it.

In short: We will not be able to run Mathesar as Debug=True and Mode=Production.

pavish: This is the current behaviour:

Debug=False, Mode=Development (Vite server runs) [Can be run locally]
Debug=True, Mode=Develoment (Vite server runs) [Can be run locally]
Debug=False, Mode=Production (Client is built. But cannot be served, unless we collate files and configure webserver to serve the files) [**Cannot be run locally**]
Debug=True, Mode=Production (Client is built. Files are served by Django) [Can be run locally]

kgodey: Thanks for the explanation, makes sense.

.env.example
pavish: Added the newline back.

mathesar/templatetags/client_files.py
pavish: Added comments and tests. Please let me if the test approach is okay.

The file loads the settings when the module loads, and the defined class is a singleton which stores the result. So, for each individual test, I've reloaded the entire file, inorder to reload the settings (which change per test), and to try to read the manifest.json file for each test.

mathesar/tests/templatetags/test_client_files.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+import importlib
+import mathesar.templatetags.client_files as client_files
+from django.template import Template, Context
+
+
+def test_client_links_development(settings):
+    settings.MODE = 'DEVELOPMENT'
```
kgodey, line 8: The standard way to use different settings per-test is to use the `@override_settings` decorator.

https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.override_settings

mathesar/templatetags/client_files.py
```diff
@@ -0,0 +1,132 @@
+import logging
```
kgodey, line 1: It looks like you are using this file to generate the HTML and CSS includes to put in `base.html`. I think a more idiomatic Django way to do this would be to build the HTML and CSS includes in `base.html` directly, using the [Django template language](https://docs.djangoproject.com/en/3.2/ref/templates/language/).

You can do the parts where you read the `manifest.json` file etc. in the Django view and pass the variables you need via context https://github.com/centerofci/mathesar/blob/master/mathesar/views/frontend.py#L17. You can then use conditionals, for loops, etc. to build the script tags and stylesheet tags in the HTML. All the HTML would then be in a single place and not spread between the template file and the Python file.

Template tags are usually used when you need to repeat the same macro in multiple places, since you don't need to do that, it doesn't seem like a good use of a template tag. Even if you did need to use it in multiple places, since you're outputting HTML, I would recommend a template that you include in other templates instead.

Sorry, I should have caught this in the first review, but I didn't fully understand this file. I might also be missing some context, so let me know if any of what I wrote doesn't make sense.

kgodey (changes requested):
@pavish I suggested a different approach, please let me know if you have questions.

pavish:
Placing this PR on hold.

kgodey (approved):

pavish (commented):
@kgodey This change expects the client build files to be always present, even when starting in development mode.

This error occurs when running in development mode, without building the client. 

```
Environment:


Request Method: GET
Request URL: http://localhost:8000/mathesar_tables/1/

Django Version: 3.1.7
Python Version: 3.9.6
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'django_filters',
 'django_property_filter',
 'mathesar']
Installed Middleware:



  File "/home/mathesar/.local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
  File "/home/mathesar/code/mathesar/views/frontend.py", line 82, in schema_home
    return render(request, 'mathesar/index.html', {
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/shortcuts.py", line 19, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/template/base.py", line 168, in render
    with context.bind_template(self):
  File "/usr/local/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/home/mathesar/.local/lib/python3.9/site-packages/django/template/context.py", line 244, in bind_template
    updates.update(processor(self.request))
  File "/home/mathesar/code/config/context_processors.py", line 9, in frontend_settings
    'manifest_data': get_manifest_data()
  File "/home/mathesar/code/mathesar/utils/frontend.py", line 9, in get_manifest_data
    with open(settings.MATHESAR_MANIFEST_LOCATION, 'r') as manifest_file:

Exception Type: FileNotFoundError at /mathesar_tables/1/
Exception Value: [Errno 2] No such file or directory: '/home/mathesar/code/mathesar/static/mathesar/manifest.json'

```

pavish:
@kgodey I've added a couple small commits, to fix a link and remove an unnessary package-lock.json file (I assume it was committed by mistake).

However, I have 2 major concerns:
1. Currently, the manifest.json file seems to be read, on every request to the base.html template. Since this file will rarely change during production, I think it should be cached in memory.
2. The error mentioned in https://github.com/centerofci/mathesar/pull/476#pullrequestreview-742542025. I think we need to allow running in development mode, without the need for client to be pre-built. Also, we cannot encourage backend developers to run in production mode locally, because any change made in client will not be reflected until they run a client build, when they merge from master/other branches.

Should we also specify this in the readme: https://github.com/centerofci/mathesar/pull/476#discussion_r676786291?

kgodey:
@pavish I didn't test without running the build, I agree with both your points. I'll make the changes. 

Thanks for removing the `package-lock.json` file. Please also go ahead and add  https://github.com/centerofci/mathesar/pull/476#discussion_r676786291 to the `README`.

kgodey:
@pavish Both your review comments should be handled now. Please update the `README` and merge.

pavish:
@kgodey Thank you! Looks great.

I'm not sure where to place the issue regarding MODE=PRODUCTION and DEBUG=False.

Basically, we need to mention that Django doesn't serve files itself when run with Debug=False, and we need to configure the webserver to do so. However, I do not know all the steps involved.

Should we keep track of this, and create a separate wiki page for production build setup? I don't think this should go in the README for the moment.

kgodey:
@pavish Yes, creating a wiki page for notes on production builds sounds good. It will be useful when we get to the Deployment milestone. I'll go ahead and merge this PR.

PR #486, opened 2021-07-26
Reflect db objects before rendering page

### LINKED ISSUE #365: DB syncing is not triggered when objects are fetched without involving the API
## Description

DB syncing currently is triggered by accessing the `SchemaViewSet`. However, schemas can be fetched without the API being involved. Our initial load of the frontend uses Django templates (and querysets) to fetch schemas and pre-render them without needing a network call. This should trigger DB syncing.

## Expected behavior

The DB should be synced when the Schema queryset is fetched, not when the API is called. This also applies to tables and databases.

## To Reproduce

Wipe your database and load the Mathesar UI. It won't load any schemas.

eito-fis:
Fixes #365

Reflects all db objects before rendering the requested page.

**Technical details**
Adds a call to `reflect_db_objects` in `mathesar/views/frontend.py` before schemas are retrieved, ensuring that the initial data will be up to date. I've confirmed locally that this resolves the issue detailed in #365.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/database/utils.py
+++ b/mathesar/database/utils.py
@@ -1,6 +1,5 @@
 from django.conf import settings
 
-from mathesar.models import Schema, Database
 from db.engine import get_connection_string
 
 
@@ -22,22 +21,3 @@ def get_database_key(engine):
 
 def get_non_default_database_keys():
     return [key for key in settings.DATABASES if key != 'default']
-
-
-def update_databases():
-    databases = set(settings.DATABASES)
-    # We only want to track non-django dbs
-    databases.remove('default')
-
-    # Update deleted databases
-    for database in Database.objects.all():
-        if database.name in databases:
-            databases.remove(database.name)
-        else:
-            database.deleted = True
-            Schema.objects.filter(database=database).delete()
-            database.save()
-
-    # Create databases that aren't models yet
-    for database in databases:
-        Database.objects.create(name=database)
--- a/mathesar/models.py
+++ b/mathesar/models.py
@@ -3,8 +3,9 @@ from django.core.cache import cache
 from django.db import models
 from django.utils.functional import cached_property
 
-from mathesar.database.base import create_mathesar_engine
+from mathesar import reflection
 from mathesar.utils import models as model_utils
+from mathesar.database.base import create_mathesar_engine
 from db import tables, records, schemas, columns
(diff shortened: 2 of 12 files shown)


eito-fis:
Some questions:
* Do we need to worry about tests for this? Given the small scope I'm assuming its fine, but figured I should ask.
* How strict do we want to be about ensuring that db objects are reflected? A more principled approach might be to force the reflection anytime a queryset is generated using a custom [Django Manager](https://docs.djangoproject.com/en/3.2/topics/db/managers/#django.db.models.Manager) for our models. This would guarantee that our Django models are always up to date, but I imagine would also be much less efficient.

kgodey (changes requested):
@eito-fis I think it would make the most sense to override the default queryset used while fetching schemas and tables so that we reflect DB objects whenever we look at schemas or tables. This would allow us to remove the API-specific code and it also means we don't have to keep adding the `reflect_db_objects` method to new views.

A quick Google led to https://stackoverflow.com/questions/19813375/override-what-objects-all-returns-when-using-passthroughmanager-in-django-mode, I'm not sure if that's the best question/answer for this problem, but it should at least lead you in the right direction.

kgodey:
@eito-fis sorry, I missed your questions. I think a custom Django manager would work fine, I believe that the reflection method has a 5-min cache built in, which would help with efficiency.

I think it would be helpful to have tests for the queryset modification since it will be pretty fundamental to how we access objects.

eito-fis:
Made changes:
* Moved all reflection functions into `mathesar/reflection.py` and updated references accordingly.
* Updated to use a custom manager for reflection. Now, db objects are reflected whenever the `.objects` attribute of the `Database`, `Table` or `Schema` model is used. 
* Also maintains a default manager that doesn't reflect db objects with the `.current_objects` attribute. Used for default Django operations and reflection operations.
* Added simple tests to verify that db objects are being reflected when the queryset is referenced.

mathemancer (commented):
I think this looks good; let's wait for @kgodey to verify.

kgodey (approved):
Looks good. I wiped my DB locally and verified that this fixes the issue.

Issue #451, opened 2021-07-20
API for working with foreign keys


kgodey:
## Problem

Users may want to add a column that references a column from another table (i.e. a foreign key). To support this functionality, we need to enable working with foreign keys in our API.

## Proposed solution

We need to make the following changes to our API:

- High priority for front end:

    - [x] #945
    - [x] #972

- Medium priority for front end:

    - [ ] #946 
    - [ ] #944

- Low priority for front end:

    - [x] #942
    - [x] #943

- Not needed for front end

    - [ ] #973
    - [x] API support for deleting FK constraints -- (this already works)

## Additional context

- front-end meta ticket: #263
- #326 
- #243 
- Please ensure that creation parameters do not clash with #452
- Mathesar's [API Standards](https://wiki.mathesar.org/en/engineering/architecture/api-standards)

Issue #440, opened 2021-07-19
Backend support for working with primary keys


kgodey:
## Problem

Users may want to modify primary keys on tables.

## Proposed solution

We need to support the following functions via API:
- Removing primary key constraints from a column
- Removing primary key constraints from a table
- Adding primary key constraints to a column
- Adding primary key constraints to a table
- Ensuring that existing records have their primary key values set to read-only. 

## Additional context

- #439 
- #395 
- #396
- Blocked by #326

kgodey:
I'll close this issue, we can figure out more detailed tickets once design work is done.

PR #488, opened 2021-07-26
API for adding and removing table constraints

### LINKED ISSUE #326: CRUD API for table constraints
**Problem**

The user might want to create a constraint on a table, and we should allow them to do so.

**Proposed solution**

Create a `api/v0/tables/<id/constraints/` endpoint that allows users to create, read, update, or delete table constraints. For now, we should only support `UNIQUE` constraints, but please keep in mind that we may want to extend the API to other constraints in the future.

**Additional context**

- Design issue: https://github.com/centerofci/mathesar/issues/324
- [Postgres documentation on constraints](https://www.postgresql.org/docs/current/ddl-constraints.html)

kgodey:
Fixes #326

This pull request adds an API to add and remove constraints from tables.

## Technical details

Per the design specs for this milestone, we will only support:
- listing constraints
- adding a unique constraint
- deleting a constraint

Adding foreign keys and primary keys will be supported in:
- #440 
- #451 

Support for adding check constraints is not planned for the MVP.

## Refactors
There's some miscellaneous cleanup and refactoring in here.

## TODO
- [x] DB function for adding a unique constraint
- [x] DB function for dropping a constraint
- [x] DB tests for adding a unique constraint
- [x] DB tests for dropping a constraint
- [x] API to read constraints
- [x] API to add a unique constraint
- [x] API to drop a constraint
- [x] Tests for API to list constraints
- [x] Test for API to retrieve a constraint
- [x] Tests for API to add a unique constraint
- [x] Tests for API to drop a constraint
- [x] Test for trying to set a unique constraint for a column with non-unique values
- [x] Test for trying to duplicate constraint name
- [x] Ensure constraint names are idiomatic
- [x] Ensure constraint names are unique
- [x] Add constraints URL to table serializer
- [x] Add constraint ID to serailizer

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/views/api.py
```diff
@@ -290,3 +292,61 @@ def create(self, request):
...
+        except AttributeError:
+            pass
+
+        constraint = table.get_constraint_by_type_and_columns(data['type'], data['columns'])
+        out_serializer = ConstraintSerializer(constraint, context={'request': request})
+        return Response(out_serializer.data, status=status.HTTP_201_CREATED)
+
+    def destroy(self, request, pk=None, table_pk=None):
+        table = Table.objects.get(id=table_pk)
+        table.drop_constraint(pk)
```
eito-fis, line 351: Could catch a 404 here?

eito-fis (approved):
Looks good to me! Just a small 404 nitpick.

mathesar/views/api.py
kgodey: Good catch, I'll fix it.

kgodey:
@eito-fis I added handling of 404 errors for deletion and also for every instance where we're trying to get a table by `pk`. I added tests for both scenarios.

mathemancer (commented):
Since all constraints have OIDs in the Postgres, I think we should have a model for them to hold that mapping for referring to them (similar to tables and schemata), both for consistency with those objects, and convenience.  This would let us notice whether a constraint was dropped or renamed, and also give us more stability when looking up and working with constraints.  

Documentation of the relevant system catalog table:  https://www.postgresql.org/docs/13/catalog-pg-constraint.html

kgodey:
> Since all constraints have OIDs in the Postgres, I think we should have a model for them to hold that mapping for referring to them (similar to tables and schemata), both for consistency with those objects, and convenience. This would let us notice whether a constraint was dropped or renamed, and also give us more stability when looking up and working with constraints.
> 
> Documentation of the relevant system catalog table: https://www.postgresql.org/docs/13/catalog-pg-constraint.html

That makes sense, I'll refactor. I hadn't realized that constraints had OIDs, it definitely makes sense to store them in a model.

powellc (approved):
Excellent work, @kgodey !  I need to get more familiar with SQLAlchemy to be confident reviewing that code, but the DRF and general python code looks great. Just a few comments but none of them would block a merge. 🚀

mathesar/signals.py
```diff
@@ -0,0 +1,22 @@
+from django.db.models.signals import post_save, pre_delete
+from django.dispatch import receiver
+
+from mathesar.models import Table, Constraint
+from mathesar.reflection import reflect_new_table_constraints
+from db.constraints import drop_constraint
+
+
+@receiver(post_save, sender=Table)
+def create_new_constraints(sender, instance, created, **kwargs):
```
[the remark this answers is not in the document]
kgodey: I ended up removing the `pre_delete` signal because I realized that we delete constraint models in other places and we don't always automatically want to drop them. However, I did rename the function to be clearer and added a comment.

mathesar/reflection.py
```diff
@@ -39,33 +40,56 @@ def reflect_databases():
...
+    engine = create_mathesar_engine(database)
+    db_constraints = get_constraints_with_oids(engine)
+    for db_constraint in db_constraints:
+        try:
+            models.Constraint.current_objects.get_or_create(
+                oid=db_constraint['oid'],
+                table=models.Table.current_objects.get(oid=db_constraint['conrelid'])
+            )
+        except models.Table.DoesNotExist:
+            pass
```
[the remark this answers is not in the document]
kgodey: I agree the readability could be improved. Let me know if the version I pushed looks better. There really shouldn't be duplicate OIDs, given that it's the primary identifier of the table in the database, so I deliberately didn't handle that case, I'd rather it throw an error.

kgodey:
Thanks for the review, @powellc!

mathesar/views/api.py
```diff
@@ -292,3 +301,50 @@ def create(self, request):
         serializer = DataFileSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
         return create_datafile(request, serializer.validated_data)
+
+
+class ConstraintViewSet(viewsets.GenericViewSet, ListModelMixin, RetrieveModelMixin, CreateModelMixin, DestroyModelMixin):
```
eito-fis, line 306: Do we need `CreateModelMixin` and `DestroyModelMixin` if we overwrite those methods?

eito-fis (approved):
Some tiny things, but otherwise LGTM :+1: 

Also, unrelated to this PR, but it seems like we're missing 404s for a lot of the records endpoint and most of the pagination classes. Worth opening an issue for?

db/constraints.py
```diff
@@ -0,0 +1,108 @@
...
+        result = conn.execute(query).fetchall()
+    return result
+
+
+def get_constraint_from_oid(oid, engine):
+    metadata = MetaData()
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="Did not recognize type")
+        pg_constraint = Table("pg_constraint", metadata, autoload_with=engine)
+        # We only want to select constraints attached to a table.
```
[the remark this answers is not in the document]
kgodey: It is not, good find, I copied it from the other method and didn't update it.

mathesar/models.py
```diff
@@ -211,6 +215,66 @@ def update_record(self, id_value, record_data):
...
 
+    def get_constraint_by_type_and_columns(self, constraint_type, columns):
+        for constraint in self.sa_constraints:
+            current_constraint_type = constraints.get_constraint_type_from_class(constraint)
+            if current_constraint_type == constraint_type and set(columns) == set([column.name for column in constraint.columns]):
+                return constraint
+        return None
+
+    def add_constraint(self, constraint_data):
+        if 'type' not in constraint_data or constraint_data['type'] != constraints.ConstraintType.UNIQUE.value:
```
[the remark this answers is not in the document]
kgodey: Good catch, agreed.

mathesar/views/api.py
kgodey: No, we don't. :)

kgodey:
@eito-fis Thanks for the review! I think it makes sense to create an issue for the missing 404s. Would you be able to do that?

mathesar/reflection.py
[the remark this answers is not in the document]
mathemancer: The OIDs do roll over at some point, but that would be an _extreme_ edge case, since the user would have needed to create many _many_ more tables than the throughput of our app is probably going to enable in a reasonable timeframe.  I definitely think throwing an error would appropriate.  An amusing (if ancient) thread from the PostgreSQL mailing list:

https://www.postgresql.org/message-id/Pine.GSO.4.02A.10004101348060.8269-100000%40Falk.DoCS.UU.SE

mathesar/views/api.py
```diff
@@ -292,3 +301,50 @@ def create(self, request):
...
+        # If we don't do this, the request.data QueryDict will only return the last column's name
+        # if there are multiple columns.
+        data = request.data.dict()
+        data['columns'] = request.data.getlist('columns')
+        try:
+            constraint = table.add_constraint(data)
+        except ProgrammingError as e:
+            if type(e.orig) == DuplicateTable:
+                raise ValidationError(
+                    'Constraint with the same name already exists'
```
mathemancer, line 326: This will also happen if a table with the desired name already exists (or any Relation).  I suggest just using the generic "Relation" to avoid confusion.

mathesar/views/api.py
```diff
@@ -292,3 +301,50 @@ def create(self, request):
...
+        return Constraint.objects.filter(table__id=self.kwargs['table_pk']).order_by('-created_at')
+
+    def create(self, request, table_pk=None):
+        table = get_table_or_404(table_pk)
+        serializer = ConstraintSerializer(data=request.data, context={'request': request})
+        serializer.is_valid(raise_exception=True)
+        # If we don't do this, the request.data QueryDict will only return the last column's name
+        # if there are multiple columns.
+        data = request.data.dict()
+        data['columns'] = request.data.getlist('columns')
```
mathemancer, line 320: When I try to submit POST requests via the DRF-created web ui, `request.data` seems to come through as a `dict` type here, which breaks these lines.  Am I doing something wrong?  The request I submitted was:
```
{
    "name": "aconstraint",
    "type": "unique",
    "columns": ["Status"]
}
```
The error I got was 
image

Fixing that error just let the next line cause the error (since there's no `getlist` method of python `dict` either).

mathesar/models.py
```diff
@@ -211,6 +215,66 @@ def update_record(self, id_value, record_data):
...
+            del self._sa_table
+        except AttributeError:
+            pass
+        sa_constraint = self.get_constraint_by_type_and_columns(
+            constraint_data['type'],
+            constraint_data['columns']
+        )
+        engine = self.schema.database._sa_engine
+        constraint_oid = constraints.get_constraint_oid_by_name_and_table_oid(sa_constraint.name, self.oid, engine)
+        return Constraint.objects.create(oid=constraint_oid, table=self)
```
mathemancer, line 246: You can have more than one constraint on the same columns of the same type.  This might happen through migrations, user inattention, etc.  This could end up reflecting the wrong one, and running into confusion when trying to add a Constraint object with a previously-reflected OID, rather than the one intended.

mathemancer (changes requested):
Overall, I think the new way of doing this works, and will extend well.  There's a bug when I try to submit `POST` requests (or I'm screwing something up); see the specific comment and screenshot.

mathesar/reflection.py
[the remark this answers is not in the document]
powellc: @kgodey looks great, thank you!

mathesar/views/api.py
kgodey: This should be fixed now. I don't know why it's a `QueryDict` sometimes and a dict other times, but both cases are handled.

mathesar/models.py
kgodey: This is fixed. Since we know the pattern of the name, I generated the name instead of trying to get the constraint by type/columns. The name should be unique.

mathesar/views/api.py
```diff
@@ -350,3 +359,52 @@ def create(self, request):
...
+        table = get_table_or_404(table_pk)
+        serializer = ConstraintSerializer(data=request.data, context={'request': request})
+        serializer.is_valid(raise_exception=True)
+        # If we don't do this, the request.data QueryDict will only return the last column's name
+        # if there are multiple columns.
+        if type(request.data) != dict:
+            data = request.data.dict()
+            data['columns'] = request.data.getlist('columns')
+        try:
+            name = data['name'] if 'name' in data else None
```
mathemancer, line 381: I don't think `data` is defined here.

mathemancer (changes requested):
When submitting the request
```json
{
    "name": "supernewtable_unique2",
    "type": "unique",
    "columns": [
        "mathesar_id"
    ]
}
```
I get
```
web_1  | Internal Server Error: /api/v0/tables/1/constraints/
web_1  | Traceback (most recent call last):
web_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
web_1  |     response = get_response(request)
web_1  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
web_1  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web_1  |   File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
web_1  |     return view_func(*args, **kwargs)
web_1  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
web_1  |     return self.dispatch(request, *args, **kwargs)
web_1  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
web_1  |     response = self.handle_exception(exc)
web_1  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
web_1  |     self.raise_uncaught_exception(exc)
web_1  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
web_1  |     raise exc
web_1  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
web_1  |     response = handler(request, *args, **kwargs)
web_1  |   File "/code/mathesar/views/api.py", line 381, in create
web_1  |     name = data['name'] if 'name' in data else None
web_1  | UnboundLocalError: local variable 'data' referenced before assignment
web_1  | [04/Aug/2021 09:50:05] "POST /api/v0/tables/1/constraints/ HTTP/1.1" 500 86861
```

pavish (changes requested):
Please revert the changes in package-lock.json

The audit fixes are done in https://github.com/centerofci/mathesar/pull/508/commits/81101df0c11bb052a827754e6545fd9acb2a768b.

pavish (commented):
The changes for package-lock.json in 5ff56c0 look good to me.

pavish (approved):

mathesar/views/api.py
kgodey: Ugh, sorry. It's fixed now, I verified through the HTML interface.

mathemancer (approved):
I think it's looking good, and it works on my machine.  Thank you!

PR #502, opened 2021-07-28
Handle exception and a ValidationError when table already exists

### LINKED ISSUE #364: Table creation returns 500 if there's a table with the same name
## Description

Making a request to create a new table from a data file with the name of a table that already exists results in:
```
InvalidRequestError at /api/v0/tables/

Table 'public.Untitled' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.
```

## Expected behavior

We should be validating that there isn't a table with the same name at the API layer and raising DRF's `ValidationError` (which results in a 400, not a 500).

andyharley-cci:
Fixes #364

**Technical details**
Handle exception thrown by sqlalchemy at API layer when attempting to create a table that already exists

**Screenshots**
The UI throws a really nasty looking error when this is returned:
image

Known issue?

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/views/api.py
```diff
@@ -85,11 +85,13 @@ def create(self, request):
...
-            table = create_table_from_datafile(valid_data)
-        else:
-            table = create_empty_table(valid_data)
-
+        try:
+            if 'data_files' in valid_data and valid_data['data_files']:
+                table = create_table_from_datafile(valid_data)
+            else:
+                table = create_empty_table(valid_data)
+        except InvalidRequestError:
```
kgodey, line 93: There might be other reasons for an `InvalidRequestError`, I would double check that it's because the table already exists. For an example of handling a more specific error, see how `DuplicateColumn` is handled in this file.

kgodey (changes requested):
@andyharley-cci Nice work! I requested a small change.

mathesar/views/api.py
andyharley-cci: Excellent point, I will make this more precise,

andyharley-cci: @kgodey i adjusted the check to be more precise, but in doing so needed to remove the `.reflect` call in the underlying function. Do you see any issues with doing that?

When using reflect, sqlacademy handles the exception and wraps it in it's own `InvalidRequestError` and doesn't seem to expose any extra detail beyond a string of what the error is, which gives some unhelpful advice(`sqlalchemy.exc.InvalidRequestError: Table 'public.blerp' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.`)

kgodey: I don't see any issues with it, tagging @mathemancer to double check since he wrote that code.

kgodey (commented):
Looks good to me, please fix linting errors.

mathesar/views/api.py
```diff
@@ -85,11 +85,18 @@ def create(self, request):
...
-
+        try:
+            if 'data_files' in valid_data and valid_data['data_files']:
+                table = create_table_from_datafile(valid_data)
+            else:
+                table = create_empty_table(valid_data)
+        except ProgrammingError as e:
+            if type(e.orig) == DuplicateTable:
+                raise ValidationError(
+                    f"Table {request.data['name']} already exists"
```
mathemancer, line 96: This should use "Relation" instead of table.  The error also will occur if:
- A view of that name already exists, or
- An index of that name already exists
This list is probably not complete.

Also, consider referring to the schema in the error message ("Relation <table> already exists in schema <schema>").  Table names aren't globally unique in a DB.

mathemancer (changes requested):
The only change I have to request is the error message (very minor).  Overall, I think this looks good.  My original thinking (if you're curious) was to avoid trying to create the table at the DB layer if it already exists, rather than finding out it already exists by trying to create it.  However, I think your way will work too, and seems to fit more nicely with the errors at the API layer.

kgodey (approved):

kgodey:
Looks good to me! I'll let @mathemancer review/merge since he also requested changes.

mathemancer (approved):
I think this looks good.  Thank you!

PR #506, opened 2021-07-28
Add table preview endpoint

### LINKED ISSUE #201: API updates for confirm data step in CSV file import.
**Problem**

Our [design for file import](https://wiki.mathesar.org/design/process/documents/table-import-specs) cannot be handled by our existing APIs, specifically the data confirmation step.

**Proposed solution**

We need API updates that allow the user to:
- Change column names (hopefully handled by #199)
- Change column types (hopefully handled by #199)
- Switch whether we use the first row as headers
- Choose which columns to import (drop columns)

**Additional context**

This issue involves a lot of changes to the architecture, so please write a technical spec on the wiki's [Architecture section](https://wiki.mathesar.org/engineering/architecture) and get some feedback from @kgodey and @mathemancer before implementing this issue.

Please note that even though we have a table in the database at this point, we also have the actual data file stored on disk so we can access it if needed. We may want to do so while switching whether to use the first row as a header. 

Blocked by #199, #200

mathemancer:
Fixes #201

This final step in implementing #201 adds a `/api/v0/tables/<table_id>/previews/` endpoint that accepts a `POST` request describing a preview.  

**Technical details**

For now, the request body must be of the format:
```json
{
    "name": "<some_table_name>",
    "columns": [
        {"name": "<some_column_name>", "type": "<desired_type>"},
       ...
    ]
}
```
All columns must be included, with both the desired name and type specified.  

The response to such a POST will be a JSON with the preview information, and 20 records to be used as an example.  The response will be in the same format as the submitted JSON, with the addition of a `records` top-level key that has a list of record dictionaries.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/views/api.py
```diff
@@ -119,6 +119,32 @@ def type_suggestions(self, request, pk=None):
         col_types = get_table_column_types(table)
         return Response(col_types)
 
+    @action(methods=['post'], detail=True)
+    def preview(self, request, pk=None):
```
kgodey, line 123: I would make this `previews` to be consistent with everything else.

mathesar/views/api.py
```diff
@@ -119,6 +119,32 @@ def type_suggestions(self, request, pk=None):
...
+        table = self.get_object()
+        serializer = TablePreviewSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors)
+        else:
+            columns = serializer.validated_data["columns"]
+
+        if not len(columns) == len(set([col["name"] for col in columns])):
+            raise ValidationError("Column names must be distinct")
```
kgodey, line 132: This error might not always mean that column names are distinct, it could be that there are fewer columns in the list. Right?

mathemancer: I think it should work, since very column in `columns` is required to include a `"name"` key.  So, this should just check that we don't lose any `columns` when taking the set of the list of names of each column in `columns`.  I'll rewrite it to be clearer.

mathesar/views/api.py
```diff
@@ -126,6 +126,36 @@ def type_suggestions(self, request, pk=None):
         col_types = get_table_column_types(table)
         return Response(col_types)
 
+    @action(methods=['post'], detail=True)
+    def previews(self, request, pk=None):
+        table = self.get_object()
+        serializer = TablePreviewSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors)
```
kgodey, line 134: You're not returning an error status code here.

mathesar/serializers.py
```diff
@@ -81,6 +81,11 @@ def to_representation(self, instance):
         return instance._asdict()
 
 
+class TablePreviewSerializer(serializers.Serializer):
+    name = serializers.CharField()
```
kgodey, line 85: I don't think it makes sense to require name on input, we already know which table this is. I recommend making this `required=False`.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -256,7 +257,59 @@ def test_table_type_suggestion(client, schema, engine_email_type):
     assert response.status_code == 200
     assert response_table == EXPECTED_TYPES
 
-# Workaround for not being able to parametrize with fixtures
+
+def test_table_previews(client, schema, engine_email_type):
```
kgodey, line 261: Please write some failure/error test cases as well.

kgodey (changes requested):
@mathemancer I was unable to test this locally.

Here's what I ran:

```
data = {
    "name": "HeaderTest",
    "columns": [
        {
            "name": "mathesar_id",
            "type": "INTEGER"
        },
        {
            "name": "Domain Name",
            "type": "VARCHAR"
        },
        {
            "name": "Domain Type",
            "type": "VARCHAR"
        },
        {
            "name": "Agency",
            "type": "VARCHAR"
        },
        {
            "name": "Organization",
            "type": "VARCHAR"
        },
        {
            "name": "City",
            "type": "VARCHAR"
        },
        {
            "name": "State",
            "type": "VARCHAR"
        },
        {
            "name": "Security Contact Email",
            "type": "mathesar_types.email"
        }
    ],
}

response = requests.post('http://localhost:8000/api/v0/tables/6/previews/', data=data)
```

(I was attempting to preview a column as email, the rest of the table definition is unchanged)

```
In [8]: response.status_code
Out[8]: 200

In [6]: response.json()
Out[6]: {'columns': ['This field is required.']}
```

As you can see, I did have a `columns` field, so I'm not sure why this is returning that error. Additionally, the error is returning a status code of 200.

mathesar/views/api.py
eito-fis: This could just be `serializer.is_valid(raise_exception=True)`?

kgodey: Thanks, good to know.

mathesar/serializers.py
mathemancer: What would you think of leaving it out entirely?  I don't think it's really needed at all.

kgodey: I think that's fine.

kgodey (changes requested):
@mathemancer This looks good, but I found a couple more uncaught exceptions in the API.

Trying to set a column to `mathesar_types.email` when the data doesn't fit results in a 500 and:

```
IntegrityError at /api/v0/tables/6/previews/

(psycopg2.errors.CheckViolation) value for domain mathesar_types.email violates check constraint "email_check"
CONTEXT:  PL/pgSQL function mathesar_types.cast_to_email(character varying) line 4 at RETURN
```

Trying to set a column to `INTEGER`, when the data would fit, results in a 500 error and:

```
UnsupportedTypeException at /api/v0/tables/16/previews/

Target Type "INTEGER" is not supported.
```

Both of these should be caught by the API and return an appropriate JSON error.

mathesar/views/api.py
kgodey: @mathemancer You're still not returning a status code here. Actually, you don't need the if statement at all, you can just have `serializer.is_valid(raise_exception=True)` be the statement and then you don't need to return a `Response`.

kgodey (changes requested):

db/types/alteration.py
```diff
@@ -87,6 +93,31 @@ def alter_column_type(
         conn.execute(DDL(alter_stmt))
 
 
+def get_column_cast_expression(column, target_type_str, engine):
+    """
+    Given an SQLAlchemy Column, we get the correct SQL selectable for
```
powellc, line 98: r/an/a :D

db/types/alteration.py
```diff
@@ -87,6 +93,31 @@ def alter_column_type(
         conn.execute(DDL(alter_stmt))
 
 
+def get_column_cast_expression(column, target_type_str, engine):
```
powellc, line 96: I really appreciate `str` in the variable name here. Makes this much easier to read.

mathesar/views/api.py
powellc: You might also just move lines 141-142 above this `if not` statement. That would make sure the number of fields is checked before checking against the cast set of names.

powellc (approved):
Nice work @mathemancer !  Nothing substantial to add from me. But addressing Kriti's concerns would be great :D

kgodey (approved):
Looks great, @mathemancer!

PR #507, opened 2021-07-29
Generate table names

### LINKED ISSUE #449: Table API should support creating tables without names
## Problem

We shouldn't require a name while creating a table via API.

## Proposed solution

If there's no name provided, we should auto-generate one according to some pattern.

## Additional context

Please post the pattern you plan to use to generate names on this issue and get feedback from @kgodey and @mathemancer.

eito-fis:
Fixes #449 

Auto generates names for tables when a name is not passed to the API.

**Technical details**
Names follow two patterns:
* In the case where we are making a table from a data file that was created from a csv (not a paste), we use the original base name of the file. This is a new field stored on the `Data File` model, saved when we first create the data file. We first check if the base name is available, and if it is not, we append and increment and integer until we find an unused name.
* In the case where we don't have a base name (paste or empty table), we use the pattern `Table n`, where `n` starts at `Table.objects.count()` and increments until we find an unused name.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models.py
```diff
@@ -222,6 +222,7 @@ class DataFile(BaseModel):
     created_from = models.CharField(max_length=128, choices=created_from_choices.choices)
     table_imported_to = models.ForeignKey(Table, related_name="data_files", blank=True,
                                           null=True, on_delete=models.SET_NULL)
+    base_name = models.CharField(max_length=128)
```
kgodey, line 225: We need to ensure that any generated names are smaller than 63 bytes since that's how long Postgres identifiers can be.

https://www.postgresql.org/docs/current/limits.html

kgodey (changes requested):
Looks good overall, but we need to make sure table names aren't too long.

mathesar/utils/tables.py
```diff
@@ -17,8 +24,43 @@ def get_table_column_types(table):
...
+            base_name = base_name[:-1]
+            name = base_name + f' {table_num}'
+        table_num += 1
+    return name
+
+
 def create_table_from_datafile(data_files, name, schema):
     data_file = data_files[0]
+
+    if not name:
```
kgodey, line 57: It might be cleanest to put this in a `pre_save` signal or in the `Table` model's `save` method. That way, you don't have to repeat it in every table creation utility.

kgodey (changes requested):
@eito-fis Looks good! I made some suggestions to improve code clarity and non-repeatability.

mathesar/utils/tables.py
eito-fis: We need access to the source data file for naming, so unfortunately I don't think we can use the save method. I was able to move the naming logic to the viewset, which I think has the desired affect.

mathemancer (commented):
Nothing specific to add from my end.  Thank you!

kgodey (approved):
Looks good!

PR #508, opened 2021-07-30
Default ordering for records

### LINKED ISSUE #397: Create canonical ordering for tables and return it by default
## Problem

We should have a canonical ordering for each table whenever possible. This will make infinite scroll easier to deal with.

## Proposed solution

We should order by primary key by default if the table has one. Otherwise, we should use the entire row as a sorting key (it won't be possible to tell the difference if two identical rows "switch order"). We should always return rows in these orders when they are accessed unless the user specifies a different `ORDER BY`.

## Additional context

- Please see #361

eito-fis:
Fixes #397

Ensures we have consistent ordering of records by setting a default order.

**Technical details**
Ordering is as described in the issue:
* If one or more primary keys exist, they are used for ordering. 
* If no primary keys exist, every column is used for ordering.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
I'll leave this for @mathemancer to review.

db/records.py
```diff
@@ -72,6 +72,17 @@ def get_records(
                   field, in addition to an 'value' field if appropriate.
                   See: https://github.com/centerofci/sqlalchemy-filters#filters-format
     """
+    if not order_by:
+        # Set default ordering if none was requested
+        if len(table.primary_key.columns) > 0:
+            # If there are primary keys, order by all primary keys
+            order_by = [{'field': col, 'direction': 'asc'}
+                        for col in table.primary_key.columns]
```
mathemancer, line 80: Under which conditions can `table.primary_key.columns` change order?  Is it only when the user themselves modifies the underlying primary key?  

I'm hoping that SQLAlchemy publishes the columns in the order they occur in the underlying index.  If so, that's great since we'll also get some performance advantages.  Otherwise, we should try to
- use the order of the columns in the tuples of the underlying index, or
- (if we can't find that order) use some canonical (say lexicographical) order for the columns when constructing the `order_by` list.

mathemancer (changes requested):
I think for this ordering to be truly "canonical" for a table, we need to figure out what order SQLAlchemy uses for the columns of a primary key in the relevant attribute.  If the order of the columns changes (horizontally), then the row order will also change when ordering by those columns.

db/records.py
eito-fis: I poked around the sqlalchemy source code a bit, and found [how they reflect the primary keys](https://github.com/sqlalchemy/sqlalchemy/blob/84be0b42e03b0aaf8000637c4b5e348a39333e2f/lib/sqlalchemy/dialects/postgresql/base.py#L3897). It seems to be they order the keys dependent on the order of the columns in the table. Assuming we don't plan on changing the order of columns, this seems like it should be alright? I have added a simple sort to the primary keys, which I'll revert if the default behavior seems reasonable.

mathemancer: > It seems to be they order the keys dependent on the order of the columns in the table. 

This isn't quite correct. They're unnesting the actual index definition to grab the column `attnum`s, then using the `attnum` (the order in the table) to get the column names.  Note the line
```SQL
                ORDER BY k.ord
```
This is the part (along with the earlier call to `generate_subscripts`) that returns the columns in the order they're specified in the _index_, not in the table.

Since that's (IMO) the best order to use (it should be stable unless the table is completely reindexed, and using that order will be most efficient performance-wise), I think we should stick with the default order from SQLAlchemy rather than sorting it ourselves (so please revert that change).  Sorry for sending you on the wild goose-chase.  Thanks for finding the relevant snippet.

mathemancer (approved):
Looks good to me.

Issue #362, opened 2021-07-14
Implement new top navigation and layout


kgodey:
## Problem

The [schemas design spec](https://wiki.mathesar.org/en/design/specs/schemas) introduced a new top navigation and layout.

## Proposed solution

 We need to implement the new top navigation and layout. This involves
- Removing schemas from the sidebar.
- Implementing the "Search or jump to..." bar in the top navigation.
- Updating the frontend code and structure so that there's only one active schema at any given time.

PR #513, opened 2021-08-01
Create the schema list page

### LINKED ISSUE #171: Implement schema list page
**Problem**

Users should be able to create a new schema, edit schema names, and delete schemas.

**Proposed solution**

We should provide a way to do these actions from the UI using the schema list page introduced in the [design spec](https://wiki.mathesar.org/en/design/specs/schemas).

**Additional context**

- #166
- #168 
- #170
- #393

dhruvkb:
Fixes #171

The PR creates the schema list page, including all necessary child components.

**Technical details**
🚧 WIP

**Screenshots**

The PR creates a `SchemaRow` component.

Locked and default state:
image

Unlocked and non-default state:
image

---

The PR creates a `Schemas` page.

No filter state:
image

Active filter state:
image

---

The PR creates a `Modal` component.

Storybook story screenshot:
image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/components/modal/Modal.svelte
```diff
@@ -0,0 +1,26 @@
...
+
+<div class="modal-wrapper">
+  <div class={['modal', classes].join(' ')} {style}>
+    <slot/>
+  </div>
+</div>
+
+<Portal target={overlayTarget}>
+  <div class="overlay"></div>
+</Portal>
```
pavish, line 22: We already have a `portal` action. We do not require the `svelte-portal` package. We could do this:

```suggestion
<div class="overlay" use:portal={{ target: overlayTarget  }}/>
```

mathesar_ui/src/components/index.ts
```diff
@@ -26,3 +26,4 @@ export { default as Pagination } from './pagination/Pagination.svelte';
 export { default as Dropdown } from './dropdown/Dropdown.svelte';
 export { default as Select } from './select/Select.svelte';
 export { default as FileUpload } from './file-upload/FileUpload.svelte';
+export { default as SchemaRow } from './schema-row/SchemaRow.svelte';
```
pavish, line 29: This is a component specific to schemas page. So, it would be better to place it within `pages/schemas`.

The `components` folder is only for re-usable components. The plan is to spin if off as a separate package, as a fully fledged component library, after the MVP.

mathesar_ui/src/pages/schemas/Schemas.svelte
```diff
@@ -0,0 +1,72 @@
...
+    Button,
+    SchemaRow,
+    TextInput,
+  } from '@mathesar-components';
+
+  // Prop
+  export let database: string;
+
+  let filterQuery = '';
+  const schemaList = [...($schemas.schemaMap as SchemaMap).values()];
```
pavish, line 21: `$schemas.data` would return an array of the schemas. It can be used used directly.

The map is created additionally while fetching the schemas as a way of accessing data of particular schemas in O(1) time.

mathesar_ui/src/pages/schemas/Schemas.svelte
```diff
@@ -0,0 +1,72 @@
...
+  import type { SchemaMap } from '@mathesar/stores/schemas';
+  import {
+    Icon,
+    Button,
+    SchemaRow,
+    TextInput,
+  } from '@mathesar-components';
+
+  // Prop
+  export let database: string;
```
pavish, line 18: Our current UI considers only one database, when fetching schemas. This needs to be handled as part of https://github.com/centerofci/mathesar/issues/362.

Mentioning it here, to help keep track. No changes are required for it in this PR.

mathesar_ui/src/pages/schemas/Schemas.svelte
```diff
@@ -0,0 +1,72 @@
...
+    <h2>Schemas ({schemaList.length})</h2>
+    <TextInput
+        placeholder="Find a schema..."
+        bind:value={filterQuery}/>
+    <ul class="schema-list">
+      {#each displayList as schema, i}
+        <li>
+          <SchemaRow
+            schemaName={schema.name}
+            tableCount={schema.children.length}
```
pavish, line 52: As mentioned in the above comment, we can directly use `$schemas.data`.

In order to fetch the table count here, we could do `$schemas.schemaMap.get(schema.id)?.length || 0` in the template.

mathesar_ui/src/components/schema-row/SchemaRow.svelte
```diff
@@ -0,0 +1,46 @@
...
+    <div class="info">
+      {#if isDefault}
+        <strong>Default</strong>
+        &middot;
+      {/if}
+      {tableCount} Tables
+    </div>
+  </div>
+  <div class="controls">
+    <slot/>
```
pavish, line 40: For the controls, instead of a slot, we could keep the controls here, and dispatch events to the parent.

It would be best not to mix templates between parent and child components using slots, unless it is essential/reusable (like the modal).

Also, we could pass the entire schema object to this component, as a prop, instead of spreading it out. This would allow future changes with respect to the SchemaRow, stay only within this component.

pavish (changes requested):

mathesar_ui/src/components/index.ts
dhruvkb: Understood 👍. Fixed in 775f800.

mathesar_ui/src/pages/schemas/Schemas.svelte
dhruvkb: Updated to use `$schemas.data` in 40096f4.

mathesar_ui/src/pages/schemas/Schemas.svelte
dhruvkb: Updated to use `$schemas.data` in 40096f4.

mathesar_ui/src/components/modal/Modal.svelte
dhruvkb: Replaced the package with the portal action in 017fc5b.

Unfortunately Storybook complains when importing from `@mathesar-components` so I had to turn off the story for the modal.

mathesar_ui/src/components/schema-row/SchemaRow.svelte
dhruvkb: Moved the controls to the `SchemaRow` component in b84e1ba. The reason I went with `<slot/>` was precisely that, I expected `SchemaRow` to be a reusable component. With that changed in 775f800, this was no longer needed.

pavish:
@dhruvkb Is this PR still a work in progress, or is it ready for a complete review?

dhruvkb:
@pavish it's still a work in progress. The UI is ready but the interactivity (i.e. API calls to the backend and commits to the store) are not ready yet. We could split the PRs into two parts if you'd prefer to merge the UI first?

pavish:
@dhruvkb I think it's best to split the PRs.

I've created an issue https://github.com/centerofci/mathesar/issues/518 to keep track of the interactivity, and assigned it to you.

mathesar_ui/src/App.svelte
```diff
@@ -19,16 +20,26 @@
     <Button on:click={() => newImport(selectedDb)}>
       Import CSV
     </Button>
+    <a href="schemas/">
```
pavish, line 23: We could change this to 
```suggestion
    <a href="/{selectedDb}/schemas/">
```

Without a preceding slash, this would append `schemas/` to the current url. Clicking on this link within the schemas route will keep appending `schemas/` to the url.

mathesar_ui/src/App.svelte
```diff
@@ -19,16 +20,26 @@
...
 </header>
 
 <section class="content-section">
-  <Route path="/:db">
-    {#key selectedDb}
-      <Base database={selectedDb}/>
-    {/key}
+  <Route path="/:db/*">
+    <Route path="/schemas">
+      {#key `${selectedDb}__schemas`}
```
pavish, line 32: The key tag is used to rerender its children when the value changes. It would suffice to only mention `selectedDb` here.

mathesar_ui/src/pages/schemas/schema-row/SchemaRow.svelte
```diff
@@ -0,0 +1,57 @@
...
+    faPencilAlt,
+  } from '@fortawesome/free-solid-svg-icons';
+  import { Icon, Button } from '@mathesar-components';
+  import type { Schema } from '@mathesar/utils/preloadData';
+
+  // Props
+  export let schema: Schema;
+
+  $: isDefault = schema.name === 'public';
+  $: isLocked = schema.name === 'public';
```
pavish, line 15: Since both the conditions are the same, it would suffice to use only one computed property here.

Currently, since we only are limiting options to the public schema, it's okay to hardcode it on the client. @kgodey It would be better if we can get these from the API.

pavish (changes requested):

mathesar_ui/src/components/modal/Modal.svelte
pavish: @dhruvkb Could you please open a bug report for storybook not accepting aliases?

mathesar_ui/src/pages/schemas/schema-row/SchemaRow.svelte
dhruvkb: When these are supplied by the API, I assume we could have the functionality to lock non-default schemas as well? In that case it'll be easier if we keep the locking and default as two separate attributes of a schema.

mathesar_ui/src/App.svelte
dhruvkb: Committed the suggestion in 4bfe823.

mathesar_ui/src/pages/schemas/schema-row/SchemaRow.svelte
pavish: Yes, in that case, the schema object will have the attributes, and we wouldn't require computed properties.

mathesar_ui/src/App.svelte
dhruvkb: Updated in 23c5be5.

mathesar_ui/src/pages/schemas/schema-row/SchemaRow.svelte
pavish: I guess it's okay to keep it for now. I see no issues in having an additional property.

pavish:
@dhruvkb The PR looks good to me! Thank you!

Please raise a bug in our repo for handing aliases in our storybook stories, as mentioned in https://github.com/centerofci/mathesar/pull/513#discussion_r681536179.

pavish (approved):

pavish:
I want to keep this open, if @kgodey wants to take a look and add any reviews before merging.

kgodey:
Looks beautiful, thanks @dhruvkb!

Since this is a different page/layout, in the future, we might want to set up a Django view that serves this page and pre-loads data that's separate from the `index` view.

mathesar_ui/src/components/modal/Modal.svelte
dhruvkb: Filed bug report #528 to address this.

PR #516, opened 2021-08-02
Create datafile from URL

### LINKED ISSUE #203: API to upload file from URL
**Problem**

We should allow users to upload data files by specifying a URL where they are hosted.

**Proposed solution**

This API should download the data file from the remote URL and save it as a data file. It can then be processed as any other data file.

**Additional context**

See [Add table from file import spec](https://wiki.mathesar.org/design/process/documents/table-import-specs).

eito-fis:
Fixes #203

Allows for the creation of data files from URL.

**Technical details**
Adds the `url` parameter to the data file creation endpoint. The url is checked for valid formatting by Django, and is also checked manually to ensure it is accessible and has the correct content type (`text/csv` or `text/plain`). Once validated, the file is streamed to a local temp file which is then saved as part of a model.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Before I open this to review - what is the proper way to test something that calls an external resource? Am I supposed to mock the entire resource? As of now this is just downloading the patents dataset from the NASA API each time one of the URL tests are run, which I assume isn't ideal.

kgodey:
Yes, please mock it. We don't want our tests to fail if NASA's API is down. Here's a [tutorial](https://realpython.com/testing-third-party-apis-with-mocks/), in case that helps.

mathesar/tests/conftest.py
```diff
@@ -69,3 +69,37 @@ def paste_filename():
 @pytest.fixture(scope='session')
 def headerless_csv_filename():
     return 'mathesar/tests/data/headerless_patents.csv'
+
+
+@pytest.fixture(scope='session')
+def patents_url():
```
kgodey, line 75: Wouldn't these fixtures make more sense in the `test_data_file_api.py` file? I'm not sure that they will be reused elsewhere.

kgodey (approved):
Looks good, @eito-fis. I added a comment, please resolve before merging.

If #488 ends up being merged first, please also review the migrations again because I also added a merge migration there.

mathesar/migrations/0017_auto_20210731_0107.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('mathesar', '0016_auto_20210728_2156'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='datafile',
+            name='created_from',
+            field=models.CharField(choices=[('FILE', 'File'), ('PASTE', 'Paste'), ('URL', 'Url')], max_length=128),
```
powellc, line 16: Is there a way to put these choices in an Enum or another constant and then just expand them here? If you list choices out literally in migration files, every time a choice is added a new migration file is added, when it really doesn't need to.

This is just a suggestiont though, no need to change it if you feel otherwise!

powellc (approved):
Looks great, nice work @eito-fis !

mathesar/tests/conftest.py
eito-fis: The URL and filename is needed for table tests, but I moved the specific data and mock fixtures to `test_data_file_api` :+1:

mathesar/migrations/0017_auto_20210731_0107.py
eito-fis: Didn't realize this was an option, thanks!

PR #522, opened 2021-08-05
Backend support for tables with 0 or more than 1 primary keys

### LINKED ISSUE #395: Backend support for tables without single-column primary keys
## Problem

Currently, our Record API only works with tables with single-column primary keys. We need to explicitly handle tables without primary keys or with multi-column primary keys to prevent unexpected behavior or errors.

## Proposed solution

We need to:
- disable the record detail views (read, edit, delete) for records without primary keys.
- disable the record detail views (read, edit, delete) for records with multi-column primary keys.
- add a field to the `TableSerializer` (and therefore the table API) that indicates the type of primary key that a table has, so that the frontend can handle displaying it properly.

## Additional context
- There's a partial PR with some work on this: #522. Please use this PR as a starting point when working on this.

aaditkamat:
Fixes #395

This PR aims to disable record views for tables with 0 or more than 1 columns as primary keys. It also specifies a field to indicate the type of primary key for the tables.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


aaditkamat:
I've added a field for the primary key type and also blocked views for tables with no primary key or multiple primary keys. I have to fix the tests.

kgodey (changes requested):
@aaditkamat This is a good start. Other than the issues I commented on, you also will need to handle this at the API level by throwing the appropriate exceptions on the retrieve/update/delete methods of the `RecordViewSet`.

kgodey:
@aaditkamat Please re-request review from the GitHub UI uonce this is ready for re-review

db/records.py
```diff
@@ -17,11 +17,17 @@ class GroupFieldNotFound(FieldNotFound):
...
+class NotUniquePrimaryKey(Exception):
+    pass
+
+
+def _get_primary_key_columns(table):
+    return list(inspect(table).primary_key)
+
+
+def _get_primary_key_column(table, idx):
+    primary_key_columns = _get_primary_key_columns(table)
```
kgodey, line 29: Please leave the `assert` in here so that calling this function if there are multiple primary keys fails.

db/records.py
```diff
@@ -47,8 +53,12 @@ def _execute_query(query, engine):
...
 
 def get_record(table, engine, id_value):
-    primary_key_column = _get_primary_key_column(table)
-    query = select(table).where(primary_key_column == id_value)
+    primary_key_columns = _get_primary_key_columns(table)
+
+    if len(primary_key_columns) != 1:
+        raise NotUniquePrimaryKey("This table does not have a unique primary key.")
+
+    query = select(table).where(primary_key_columns[0] == id_value)
```
kgodey, line 61: Please use `_get_primary_key_column` here, that should be the only way we access a singular primary key.

db/records.py
kgodey: Actually, I think it would be better to move the `if len(primary_key_columns) != 1:` exception raising into this function so that you don't have to repeat it in multiple functions.

mathesar/models.py
```diff
@@ -156,6 +156,16 @@ def sa_column_names(self):
     def has_dependencies(self):
         return True
 
+    @property
+    def pk_type(self):
```
kgodey, line 160: I think instead of this property, it would be cleaner to have a `num_primary_keys` property that just returned the number.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -636,11 +644,13 @@ def test_table_create_with_same_name(client, schema):
     body = {
         'name': table_name,
         'schema': schema.id,
+        'pk_type': 'Single'
     }
     client.post('/api/v0/tables/', body)
     response = client.post('/api/v0/tables/', body)
     response_error = response.json()
     assert response.status_code == 400
+    print(response.json())
```
kgodey, line 653: Please remove print statement.

mathesar/views/api.py
```diff
@@ -331,19 +331,31 @@ def create(self, request, table_pk=None):
         table = get_table_or_404(table_pk)
         # We only support adding a single record through the API.
         assert isinstance((request.data), dict)
-        record = table.create_record_or_records(request.data)
-        serializer = RecordSerializer(record)
+        try:
+            record = table.create_record_or_records(request.data)
+            serializer = RecordSerializer(record)
+        except NotUniquePrimaryKey as e:
```
kgodey, line 337: We should be able to create records even if there's no primary key. Creation doesn't depend on a primary key.

mathesar/views/api.py
```diff
@@ -331,19 +331,31 @@ def create(self, request, table_pk=None):
...
         return Response(serializer.data, status=status.HTTP_201_CREATED)
 
     def partial_update(self, request, pk=None, table_pk=None):
         table = get_table_or_404(table_pk)
-        record = table.update_record(pk, request.data)
-        serializer = RecordSerializer(record)
+        try:
+            record = table.update_record(pk, request.data)
+            serializer = RecordSerializer(record)
+        except NotUniquePrimaryKey as e:
```
kgodey, line 347: Please write additional tests for retrieve, deletion, and partial updates that ensure this error is raised (for both no primary keys and multiple primary keys)

mathesar/views/api.py
kgodey: This probably needs to go on the `retrieve` method instead.

kgodey (changes requested):
@aaditkamat This is looking better! I added a few more comments.

kgodey:
@aaditkamat I forgot to mention in my previous review: please also revert your package lock changes. There shouldn't be any need to update frontend packages in this PR.

aaditkamat:
> @aaditkamat I forgot to mention in my previous review: please also revert your package lock changes. There shouldn't be any need to update frontend packages in this PR.

Yup, my bad, I committed it by accident.

aaditkamat:
I wasn't sure about the tests, hope that asserting the error is sufficient.

mathesar/views/api.py
```diff
@@ -321,10 +321,14 @@ def list(self, request, table_pk=None):
...
-            raise NotFound
-        serializer = RecordSerializer(record)
+        try:
+            record = table.get_record(pk)
+            if not record:
+                raise NotFound
+            serializer = RecordSerializer(record)
+        except NotUniquePrimaryKey as e:
+            logger.error(e.message)
+            raise e
```
kgodey, line 331: You should be raising an `APIException` here, not the error directly. This will return the appropriate JSON error in the frontend. Please see how `APIException` is used elsewhere in the code. This also applies to other methods below.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -873,3 +885,75 @@ def test_table_viewset_checks_cache(client):
     with patch.object(reflection, 'reflect_tables_from_schema') as mock_reflect:
         client.get('/api/v0/tables/')
     mock_reflect.assert_called()
+
+
+def test_table_retrieve_no_primary_key(create_table, client):
+    table_name = 'NASA Table Retrieve'
+    new_table_name = 'NASA Table Retrieve New'
+    table = create_table(table_name)
```
kgodey, line 893: You cannot use the existing `create_table` function here, since it creates a table with a primary key. You'll need to create a table without a primary key manually. Also applies to other `no_primary_key` functions below.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -873,3 +885,75 @@ def test_table_viewset_checks_cache(client):
...
+    try:
+        client.patch(f'/api/v0/tables/{table.id}/', body)
+    except NotUniquePrimaryKey as e:
+        assert e
+
+
+def test_table_retrieve_multiple_primary_keys(create_table, client):
+    table_name = 'NASA Table Retrieve'
+    new_table_name = 'NASA Table Retrieve New'
+    table = create_table(table_name)
```
kgodey, line 929: You cannot use the existing `create_table` function here, since it creates a table with a single primary key. You'll need to create a table with a composite primary key manually. Also applies to other `multiple_primary_key` functions below.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -873,3 +885,75 @@ def test_table_viewset_checks_cache(client):
...
         client.get('/api/v0/tables/')
     mock_reflect.assert_called()
+
+
+def test_table_retrieve_no_primary_key(create_table, client):
+    table_name = 'NASA Table Retrieve'
+    new_table_name = 'NASA Table Retrieve New'
+    table = create_table(table_name)
+
+    body = {'name': new_table_name, 'num_primary_keys': 0}
```
kgodey, line 895: `get` requests do not take a body.

Also, you shouldn't be setting the `num_primary_keys` value, it should be read only. You need to verify that the value returned for the table is `0`.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -873,3 +885,75 @@ def test_table_viewset_checks_cache(client):
...
+    try:
+        client.get(f'/api/v0/tables/{table.id}/', body)
```
kgodey, line 897: The detail view URL should `/api/v0/tables/{table.id}/records/<record-id>/`

mathesar/tests/views/api/test_table_api.py
kgodey: Verifying the `num_primary_keys` value should be in a separate test from retrieve/delete/partial update tests.

mathesar/tests/views/api/test_table_api.py
kgodey: Also applies to all other functions.

kgodey: Since you won't have a `record-id`, you can use a random one.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -873,3 +885,75 @@ def test_table_viewset_checks_cache(client):
...
+    except NotUniquePrimaryKey as e:
+        assert e
```
kgodey, line 899: You should verify here that the response's status code is `400` and the error message returned matches the error message raised in the API.

kgodey (changes requested):
@aaditkamat Thanks for your work on this! I've added some comments on the tests, plus one comment on the APIs. In addition to fixing the tests you have, you'll need to also write two more tests to ensure that `num_primary_keys` is returned correctly for tables with a composite primary key or no primary key.

Please let us know here or on Matrix if you have any questions about any of this. Thanks!

mathesar/tests/views/api/test_table_api.py
aaditkamat: How do I know from the table schema whether it will have zero, one or more than one primary keys?

kgodey: I'm not sure what you mean by the "table schema". You'll need to create the table, so you can create it with however many primary keys you like. 

Here's a tutorial on creating tables in SQLAlchemy: https://www.tutorialspoint.com/sqlalchemy/sqlalchemy_core_creating_table.htm. you can see that the `primary_key=True` argument is passed to a column. when creating the table for tests, you would instead either not pass that argument in for any column for testing no primary keys, or you would pass it to multiple columns to test composite primary keys.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -332,13 +339,6 @@ def test_table_type_suggestion(client, schema, engine_email_type):
     assert response_table == EXPECTED_TYPES
 
 
-def _check_columns(actual_column_list, expected_column_list):
```
kgodey, line 335: Why was this deleted?

mathesar/tests/views/api/test_table_api.py
```diff
@@ -377,8 +382,7 @@ def test_table_previews(client, schema, engine_email_type):
         ],
     }
     actual_dict = response.json()
-    assert all([expect_dict[key] == actual_dict[key] for key in expect_dict if key in ['name', 'records']])
```
kgodey, line 380: Why was this changed?

mathesar/tests/views/api/test_table_api.py
```diff
@@ -405,7 +410,11 @@ def test_table_previews_wrong_column_number(client, schema, engine_email_type):
             {"name": "col_6", "type": "NUMERIC"}
         ]
     }
-    response = client.post(f'/api/v0/tables/{table.id}/previews/', data=post_body, format='json')
+    response = client.post(
```
kgodey, line 413: Why was this changed?

kgodey (changes requested):
@aaditkamat It looks like there was an incorrect merge from `master` into this branch, there are a lot of changes from `master` that have been reverted. I'll review once you fix it, otherwise I cannot find your changes easily.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -1264,3 +1268,136 @@ def test_table_patch_columns_invalid_type_with_multiple_changes(create_data_type
...
+    columns = [
+        Column('id', Integer),
+        Column('name', String)
+    ]
+    table = Table(
+        table_name,
+        metadata,
+        columns,
+        schema=schema
+    )
```
kgodey, line 1287: You will need to run `sa_table.create()` after this to actually create the table.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -1264,3 +1268,136 @@ def test_table_patch_columns_invalid_type_with_multiple_changes(create_data_type
...
+def test_table_delete_no_primary_key(create_schema, client):
+    table_name = 'NASA Table Delete'
+    schema = create_schema('Patents')
+    engine = create_mathesar_engine(schema.database.name)
+    metadata = MetaData(bind=engine, schema=schema)
+    columns = [
+        Column('id', Integer),
+        Column('name', String)
+    ]
+    table = Table(
```
kgodey, line 1282: This needs to be a SQLAlchemy table, not `mathesar.models.Table`. You'll need to import it from SQLAlchemy (to avoid a name conflict, you should do something like `from sqlalchemy import Table as SATable)

Then you would change this line to `sa_table = SATable(`

mathesar/tests/views/api/test_table_api.py
kgodey: After this, you'll need to do the following:

`reflect_db_objects()`

This will create the Mathesar model object corresponding to the new table you just created in the database.

`table = Table.objects.get(name=table_name)`

This will get the Mathesar Table object corresponding to the table you've just created. After this, the rest of the test should work fine.

There are multiple types of `Table` objects because we're using a `Table` Django model to represent the underlying SQL Alchemy `Table` object. I know that might be confusing, please let me know if you have further questions about this.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -1264,3 +1268,136 @@ def test_table_patch_columns_invalid_type_with_multiple_changes(create_data_type
     # The table should not have changed
     original_column_data = _get_data_types_column_data()
     _check_columns(current_table_response.json()['columns'], original_column_data)
+
+
+def test_table_delete_no_primary_key(create_schema, client):
+    table_name = 'NASA Table Delete'
```
kgodey, line 1274: Each test should have a unique table name.

mathesar/tests/views/api/test_table_api.py
kgodey: I also recommend putting the table creation logic in a new function so that you don't have to repeat it in all the tests.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -1264,3 +1268,136 @@ def test_table_patch_columns_invalid_type_with_multiple_changes(create_data_type
...
+        columns,
+        schema=schema
+    )
+    assert table.num_primary_keys == 0
+
+    new_table_name = 'NASA Table Delete New'
+    body = {'name': new_table_name}
+    try:
+        record_id = 233
+        client.get(f'/api/v0/tables/{table.id}/record/{record_id}', body)
```
kgodey, line 1294: There should be a slash after `{record_id}`

kgodey (changes requested):
@aaditkamat I reviewed one of the tests. The comments should apply to the rest.

Thanks!

mathesar/tests/views/api/test_table_api.py
aaditkamat: > After this, you'll need to do the following:
> 
> `reflect_db_objects()`
> 

Is this mimicking the behaviour of the `_sa_table` property within the Table model: https://github.com/centerofci/mathesar/blob/839efb5cbf0b9a539490c0a9e125700d4fc2dccb/mathesar/models.py#L150-L162?

kgodey: It's the reverse of it, you're creating the `sa_table` first (which creates the table in the database) and then syncing what's in the database by creating the necessary Django model instance.

You need to do this this way because the default table creation logic in Mathesar will automatically create a primary key field, so you can't use it for these tests.

kgodey:
@aaditkamat let me know if you need any more help here, or you'd like me to finish this up. You've done great work on it so far, but in retrospect, I should not have marked this a "good first issue", it needs a lot of knowledge of the codebase. I can also assign you a different issue if you're interested.

seancolsen:
@kgodey I'm looking over our open PRs and this one stands out as one that we ought to try closing. Would you be interested in finishing up the work here yourself or assigning it to another core team member since we haven't heard from @aaditkamat in over three months? If not, I think we should consider closing it so it doesn't continue to accumulate merge conflicts.

kgodey:
@seancolsen Thanks for following up. I'm going to close the PR, I've added a note in #395 to use this PR as a starting point for whoever takes this up when it's a priority.

PR #526, opened 2021-08-06
Column defaults db and api support

### LINKED ISSUE #494: DB and API work for adding a default value to columns
## Problem

The ["Working with Columns" design spec](https://wiki.mathesar.org/en/design/specs/working-with-columns) includes the ability to set a default value for columns. We need to support this on the backend.

## Proposed solution

Postgres supports assigning default values to columns. We should:
- add functions to the `db` module to get, create, update, and delete default values for columns.
- update the existing Column API to support getting, creating, updating, and deleting default values. This involves
  - `default_value` being added to the column serializer
  - the Column API accepting `PATCH` requests with `default_value` as a key, and creating/updating/deleting the default value as needed.
- ensure that the default value is a valid value based on the column's type.

## Additional context

- Frontend issue: #500
- Design issue: #324 
- Design discussion: #436
- Mathesar's [API Standards](https://wiki.mathesar.org/en/engineering/architecture/api-standards)
- Postgres [documentation on default values](https://www.postgresql.org/docs/current/ddl-default.html)

eito-fis:
Fixes #494

Adds backend db operations to retrieve / add / update / delete default values for columns, and exposes them as part of the column endpoint.

**Technical details**
Backend Changes:
- Adds a `set_column_default` and `get_column_default` to `db.columns`. 
- Adds support for `server_default` to `MathesarColumn`, and adds a `default_value` property to `MathesarColumn`.
- Updates the `get_column_index_from_name` function to account for dropped columns that still appear in the system table.
- Updates `alter_column` to handle `"default"`
- Updates `create_column` to handle `"default"`
  - Also adds `InvalidDefaultError` and `InvalidTypeOptionError` to `create_column`. This was necessary to distinguish between the root cause of errors on the API level.
- Updates `alter_column_type` to handle defaults. Now removes the initial default, casts the column then casts and sets the default.
- Updates `MathesarColumn` to save the original table it came from when we use `.from_column()`. 

Types:
- Adds the `DATE` type to the list of supported types. I made this change while testing how types serialized, I can revert if there's a reason we didn't have it in already.
- Adds the `TEXT` type to the list of supported types. This was changed as an easy fix for casting email type defaults. They're seem to be stored as `::text`, so we need it as a supported type to support email -> email and email -> varchar casting. If we prefer not to add this in yet, I can manually handle the case for now.

Note that the functions all work with SQLAlchemy's `server_default` parameter, and not its `default` parameter. `server_default` is our choice as it is set on the server, and therefore carries to any operations users want to perform outside of Mathesar.

API Changes:
- Adds the `default` field to the column serializer to support retrieval operations. The default field uses a new `TextualSQLField`, which casts all data to a str. This is needed as defaults must be initialized with textual SQL, and `str` does a good enough job converting data types to a form postgres will accept. This field will likely have to be extended as we add additional types. On return, we can retrieve the raw python value and therefore don't want to coerce to string.
- Updating column defaults is supported through the pre-existing `alter_column` function
- Creating column defaults on column creation is supported through the pre-existing `create_column` function.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
Fair number of questions before I open this for review:
* Right now, all default values are passed as strings (or coerced into strings by the serializer) as sqlalchemy's `DefaultClause` takes textual SQL - I assume this isn't ideal. Does anyone know of a way to map python types into textual SQL that could be used in the `DefaultClause`? I've looked around the SQLAlchemy documentation, but haven't been able to find anything. (I'm happy to hardcode some conversions for this, just wanted to make sure I wasn't missing something easy first.)
* Previously, the columns being serialized and returned from the create and update column endpoints were basic columns instead of `MathesarColumn`s, which results in fields (like `column_index`) that depend on `MathesarColumn` properties to be empty. I changed this, but given the original behavior is accounted for in tests, I was wondering if there was a deliberate reason for this.
* I haven't been able to get deleting defaults to work, and was wondering if anyone had a possible solution. We don't have a specific endpoint for defaults, so I assumed we would delete it using the partial update endpoint. However, the only way to specify for a default to be deleted is to pass a null value, which our endpoint doesn't accept. Are there known solutions to this kind of problem?
* The `DefaultClause` can be a SQL expression that is actually executed. Even if we do implement proper type casting as in my first question, I do worry a bit about text defaults and their ability to potentially execute arbitrary SQL. Is this something to worry about? / Is there a way to escape the text we use?
* Do we want to return a value for default on primary key columns? I believe they are handled as a `SERIAL` default on the server, and consequently currently return the next value that would be inserted. This doesn't seem unreasonable to me so I left it in, but could modify it to return `None` instead if desired.
* Smaller question - would we rather the update and create functions error when a table doesn't / does exist? Right now the create default function is an alias of the update default function, and both will always succeed regardless of if a default does or doesn't already exist. I could see checking if a default exists and throwing an error, but am unsure if that's something we want.

eito-fis:
I think this is ready for review, although I would appreciate feedback on the questions about behavior I raised above. I also think this should be merged after #527 as they almost certainly conflict.

kgodey:
> Do we want to return a value for default on primary key columns? I believe they are handled as a `SERIAL` default on the server, and consequently currently return the next value that would be inserted. This doesn't seem unreasonable to me so I left it in, but could modify it to return `None` instead if desired.

I think this is fine for now. We can modify behavior later if needed. Perhaps document it somewhere in the code so we don't forget?

> Smaller question - would we rather the update and create functions error when a table doesn't / does exist? Right now the create default function is an alias of the update default function, and both will always succeed regardless of if a default does or doesn't already exist. I could see checking if a default exists and throwing an error, but am unsure if that's something we want.

Do you mean "when a **default**  doesn't / does exist?" I don't think you need separate functions for create/update/delete, you could just have a `set_column_default` function that does all three.

db/columns.py
```diff
@@ -153,23 +162,42 @@ def get_column_index_from_name(table_oid, column_name, engine):
...
+    # Account for dropped columns that don't appear in the SQLAlchemy tables
+    sel = (
+        select(func.count())
+        .where(and_(
+            pg_attribute.c.attisdropped.is_(True),
+            pg_attribute.c.attnum < result,
+        ))
+    )
+    with engine.begin() as conn:
+        dropped_count = conn.execute(sel).fetchone()[0]
```
mathemancer, line 175: Great catch!

db/columns.py
```diff
@@ -240,15 +269,52 @@ def change_column_nullable(table_oid, column_index, nullable, engine):
...
+
+
+def _decode_server_default(server_default, engine):
+    if server_default is not None:
+        cast_sql_text = server_default.arg.text
+        with engine.begin() as conn:
+            # Defaults are returned as text with SQL casts appended
+            # Ex: "'test default string'::character varying" or "'2020-01-01'::date"
+            # Here, we execute the cast to get the proper python value
+            return conn.execute(select(text(cast_sql_text))).first()[0]
```
mathemancer, line 295: This is creating a problem for any column with a serial default, since it actually pulls one from the underlying sequence.  To see this, reload the column list for some table, and watch the `mathesar_id` increment.  The values are burnt when they're shown, BTW, so inserts won't actually use the displayed value, but rather the one after that.

I suggest completely avoiding dealing with dynamically-generated defaults for the moment (unless they're something we're dealing with ourselves).  Alternatively, we could try to come up with some supported ones (e.g., basic integer sequences, pseudo-random UUIDs, `NOW()` for time and date columns), and create separate stories for each of those.  I think they'll need to be set up with some bespoke code.

Regardless, the behavior with sequences is at best confusing (since any inserted row will actually not get the index that's shown in the API, but rather the next value after that), and should be fixed.  I also think that regardless of the confusing behavior, we should avoid incrementing sequences that Mathesar may not have created (i.e., the user could import a table that has its own sequences), since this burns values from the sequence, which could create issues under some (admittedly rare) circumstances for users.

mathemancer (changes requested):
Overall, I think this looks okay, with a notable caveat:  I strongly suggest we avoid touching columns with dynamically-generated defaults (or at least avoid columns for which that generation burns values) in this PR.  See my line comment for more details.  If we want to handle those types of columns, I suggest putting it off till a future feature addition.

db/columns.py
kgodey: I'm in support of avoiding dealing with them for now. We can figure out if/when we want to support dynamically-generated defaults later.

eito-fis: I've updated to check for the textual SQL starts with `nextval(` and ends with `)`. This seems somewhat brittle, but given all strings should be wrapped by `'`s I think we should be safe. We could instead check the sqlalchemy column is a primary key, is an integer and has `autoincrement` set to true (the conditions for sqlalchemy creating a serial column). But, this would fail on serial columns manually set on the server, so I figured the current solution was best.

mathemancer: I think the most robust ways would be to either:

- check the `attmissingval` field of the `pg_attribute` entry for the column for a non-null value (this should only be non-null if there's a "non-volatile `DEFAULT`, and has the advantage of actually serving up the value), or
- check the `atthasmissing` field of the `pg_attrdef` entry for the column for the prefix `{CONST`, which is part of the expression that actually "generates" the column, or
- check the boolean field `atthasmissing` of the `pg_attribute` entry for the column (I _think_ this should only be true for non-volatile `DEFAULT` values; please double-check that in the docs), and then get the actual default separately.  

Unfortunately, the columns involved in the first two options are of types not supported by SQLAlchemy (`anyarray` and `pg_node_tree` respectively), so you'd need to handle the return value as text.  I don't think that should be too bad, though.  My favorite option is the first bullet point above, since you can use a single query to both verify whether there's a (non-volatile) `DEFAULT`, and also get the value itself with no further checking.  The only issue will be getting the right type for the value, but that should be in your hands (check `Column.type.python_type`) when you're looking up the value anyway.

mathemancer (changes requested):
Please see my response to the thread for some more robust options for checking whether the default is a constant or not.  One problem with the method you have is that it only excludes a subset of the possible set of dynamically-generated default values (I don't think it excludes a `NOW()` default, for example, or user-defined functions which may still burn values).

eito-fis:
@mathemancer Unfortunately, SQLAlchemy seems to implement the default clause such that it always counts as a generated default so we can only pull the default from `pg_attrdef` which only returns textual SQL. Not too sure what to try next - I'm done for the day, but can look into this more tomorrow.

mathemancer:
> @mathemancer Unfortunately, SQLAlchemy seems to implement the default clause such that it always counts as a generated default so we can only pull the default from `pg_attrdef` which only returns textual SQL. Not too sure what to try next - I'm done for the day, but can look into this more tomorrow.

When I tried fiddling with this, I found that the result of executing `select(pg_attrdef)` is something like:
```
[(371107, 371104, 1, '{CONST :consttype 23 :consttypmod -1 :constcollid 0 :constlen 4 :constbyval true :constisnull false :location 53 :constvalue 4 [ 3 0 0 0 0 0 0 0 ]}'),
 (371108, 371104, 3, '{CONST :consttype 23 :consttypmod -1 :constcollid 0 :constlen 4 :constbyval true :constisnull false :location 55 :constvalue 4 [ 3 0 0 0 0 0 0 0 ]}'),
 (371109, 371104, 4, '{FUNCEXPR :funcid 1178 :funcresulttype 1082 :funcretset false :funcvariadic false :funcformat 2 :funccollid 0 :inputcollid 0 :args ({FUNCEXPR :funcid 1299 :funcresulttype 1184 :funcretset false :funcvariadic false :funcformat 0 :funccollid 0 :inputcollid 0 :args <> :location 53}) :location -1}'),
 (371112, 371104, 5, '{FUNCEXPR :funcid 480 :funcresulttype 23 :funcretset false :funcvariadic false :funcformat 2 :funccollid 0 :inputcollid 0 :args ({FUNCEXPR :funcid 15 ... (144 characters truncated) ... d -1 :constcollid 0 :constlen 4 :constbyval true :constisnull false :location -1 :constvalue 4 [ -90 -87 5 0 0 0 0 0 ]}) :location -1}) :location -1}'),
 (371116, 371104, 6, '{CONST :consttype 23 :consttypmod -1 :constcollid 0 :constlen 4 :constbyval true :constisnull false :location 56 :constvalue 4 [ 4 0 0 0 0 0 0 0 ]}'),
 (371117, 371104, 7, '{CONST :consttype 23 :consttypmod -1 :constcollid 0 :constlen 4 :constbyval true :constisnull false :location 55 :constvalue 4 [ -30 16 0 0 0 0 0 0 ]}')]
```
Here, the rows with `adbin` having the prefix `'{CONST` are the actual constant defaults, and the rows with `adbin` having prefix `'{FUNCEXPR` are the dynamic defaults.  The columns were created with a mixture of SQLAlchemy, `psql`, and raw `psycopg2` (just to see).  Unfortunately, recovering the actual default value from the `adbin` column is probably too much of a pain, and we should just get that value through other means.

db/columns.py
```diff
@@ -284,15 +309,50 @@ def get_mathesar_column_with_engine(col, engine):
...
+    return _decode_server_default(column.server_default, engine)
+
+
+def _decode_server_default(server_default, engine):
+    if server_default is not None:
+        cast_sql_text = server_default.arg.text
+        # Ensure that we never evaluate something that increments a query
+        # This is safe as strings will always start with a single quote
+        if cast_sql_text.startswith("nextval(") and cast_sql_text.endswith(")"):
+            return None
```
powellc, line 334: Ooof, this is kinda hard to parse at a glance. I wonder if you could put 'nextval(' in a variable and ')' like:

```suggestion

    INC_TYPE_START='nextval('
    INC_TYPE_END=')'
    is_incrementing_type = cast_sql_text.startswith(INC_TYPE_START) and cast_sql_text.endswith(INC_TYPE_END)
    
    if is_incrementing_type:
       return None
```

powellc (approved):
Looks good to me. Just one nit about readability, but you don't have to change it. Nice work @eito-fis

eito-fis:
@mathemancer, @powellc: I've updated `get_column_default` to check `pg_attrdef` to check if the default is generated or not, with the actual value fetched using the `column.server_default` value as we have before. 

One consequence of this change was the need for the `MathesarColumn.table` attribute in a `@property` that is serialized on return. However, when we create a mathesar column using `.from_column`, we don't (and can't) copy the table property. So, the previous pattern we had of `ColumnSerializer(MathesarColumn.from_column(column))` results in some of our properties that depend on `.table` breaking (in this case, `column_index` and `default_value`).

As a simple fix, I'm currently storing the from column table in a `_table` attribute and using a `table_` property to return `table` or `_table`. This works, but doesn't seem ideal - is there a better way of going about this?

kgodey:
@eito-fis When you're fixing the merge conflicts for this, can you make sure that the bug I found in https://github.com/centerofci/mathesar/pull/542 about column indices returning `null` after `PATCH` requests is resolved? Maybe also write a test for it?

kgodey:
Thanks @eito-fis! I'll leave it to @mathemancer to review this, but some questions:

(1) Do we validate that default values are acceptable values for the column (based on the type of column) when we set the default value? If a user tries to set an invalid default value from the API, they should get an error.

(2) What happens to default values when the type of a column is altered and the default value is no longer compatible with the new type?

eito-fis:
@kgodey 
1. We don't do validation - I'll make sure to add a catch for type errors on the API end.
2. This isn't handled right now, not entirely sure how to do it either. I can try getting the current value, casting with SQL and then setting the default again?

kgodey:
@eito-fis Both of those sound good to me. Brent might have more thoughts too.

mathemancer:
@eito-fis @kgodey I think those solutions both sound good.  @eito-fis Everything is looking good; please let me know once those fixes are in and I'll do the proper review.

BTW, the default can be cast to the new type using the `cast_to_<type>` functions (same as the one in the `USING` clause of the current logic).

eito-fis:
Updated to add better error handling and to support defaults when altering column type. Some questions:
* I included testing the default casting into the `test_alter_column_casts_data_gen` test - is this fine?
* Just want to double check on the addition of the TEXT and DATE types. They make life a bit easier, but I'm not sure if I should be adding them here - can totally revert if they should be added in a later PR.

kgodey:
Thanks @eito-fis! I'll leave the `db` questions to @mathemancer.

kgodey:
@eito-fis Please fix merge conflicts and I'll re-review today since Brent is out.

eito-fis:
@kgodey Fixed the merge conflicts.

kgodey (changes requested):
@eito-fis I was able to set the default for a `VARCHAR` column with no problems, but I ran into some issues with `NUMERIC` and `INTEGER` columns.

Attempting to set the default of an `INTEGER` column to `5` results in:
```
{
    "detail": "Argument 'arg' is expected to be one of type '<class 'str'>' or '<class 'sqlalchemy.sql.elements.ClauseElement'>' or '<class 'sqlalchemy.sql.elements.TextClause'>', got '<class 'int'>'"
}
```

Attempting to set the default of a `NUMERIC` column to `pizza` results in a 500 error:

```
Environment:


Request Method: PATCH
Request URL: http://localhost:8000/api/v0/tables/2/columns/1/

Django Version: 3.1.7
Python Version: 3.9.6
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'django_filters',
 'django_property_filter',
 'mathesar']
Installed Middleware:



  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1770, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 717, in do_execute
    cursor.execute(statement, parameters)

The above exception (invalid input syntax for type numeric: "pizza"
) was the direct cause of the following exception:
  File "/code/mathesar/views/api.py", line 277, in partial_update
    column = table.alter_column(pk, request.data)
  File "/code/mathesar/models.py", line 179, in alter_column
    return columns.alter_column(
  File "/code/db/columns.py", line 294, in alter_column
    return attribute_alter_map[column_def_key](
  File "/code/db/columns.py", line 430, in set_column_default
    op.alter_column(
  File "<string>", line 3, in alter_column
    <source code not available>
  File "/usr/local/lib/python3.9/site-packages/alembic/operations/ops.py", line 1655, in alter_column
    return operations.invoke(alt)
  File "/usr/local/lib/python3.9/site-packages/alembic/operations/base.py", line 354, in invoke
    return fn(self, operation)
  File "/usr/local/lib/python3.9/site-packages/alembic/operations/toimpl.py", line 43, in alter_column
    operations.impl.alter_column(
  File "/usr/local/lib/python3.9/site-packages/alembic/ddl/postgresql.py", line 141, in alter_column
    super(PostgresqlImpl, self).alter_column(
  File "/usr/local/lib/python3.9/site-packages/alembic/ddl/impl.py", line 200, in alter_column
    self._exec(
  File "/usr/local/lib/python3.9/site-packages/alembic/ddl/impl.py", line 146, in _exec
    return conn.execute(construct, multiparams)
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/future/engine.py", line 280, in execute
    return self._execute_20(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1582, in _execute_20
    return meth(self, args_10style, kwargs_10style, execution_options)
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/sql/ddl.py", line 77, in _execute_on_connection
    return connection._execute_ddl(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1352, in _execute_ddl
    ret = self._execute_context(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1813, in _execute_context
    self._handle_dbapi_exception(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1994, in _handle_dbapi_exception
    util.raise_(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/util/compat.py", line 207, in raise_
    raise exception
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1770, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 717, in do_execute
    cursor.execute(statement, parameters)

During handling of the above exception ((psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type numeric: "pizza"

[SQL: ALTER TABLE public."Libraries" ALTER COLUMN "X" SET DEFAULT 'pizza']
(Background on this error at: http://sqlalche.me/e/14/9h9h)), another exception occurred:
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
  File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/code/mathesar/views/api.py", line 293, in partial_update
    f'parameter dict {request.data["type_options"]} is'

Exception Type: KeyError at /api/v0/tables/2/columns/1/
Exception Value: 'type_options'
```

Please also write tests for these cases.

eito-fis:
@kgodey Could I get the request you made for the first error? The current API test for setting the default does exactly what you described, so I would expect it to work. The second error I can definitely handle :+1:

kgodey:
@eito-fis it was a `PATCH` request to a column with body `{"default": 5}`

eito-fis:
Updated to handle inputs to always ensure defaults are strings, and fixed the error handling for the column PATCH endpoint.

Also added support for duplicating column defaults, although it currently only supports static defaults. Are we interested in also duplicating generated defaults? I'm guessing that it'll be a fair amount of work so it might be out of scope for this PR, but I can open a issue if its something we're thinking about.

kgodey:
I'll review this in the morning, thanks! I think it's worth opening an issue to track duplicating generated defaults, we can decide later if/when to prioritize it.

kgodey (approved):
This looks great, thank you!

kgodey:
I'm merging this even though Brent still has changes requested since he's out.

Issue #403, opened 2021-07-16
Mathesar should not depend on custom types being installed.


kgodey:
## Problem

We currently assume that custom types have been installed in the database in various places in the Mathesar codebase. This is a faulty assumption - Mathesar should work well with whatever types are available. Custom types should only provide additional functionality, not be a requirement.

## Proposed solution

We need to ensure that there is no assumption that custom types are installed anywhere in the codebase. We should detect the types that are installed and only use those for:
- type inference during file import
- the API listing types available in the database
- the column API listing types that a  given column can be altered to
- any SQLAlchemy engines that are created in the codebase 

We should also check the codebase for other places that types are used to make sure there's no other places that need to be updated.

## Additional context

- https://github.com/centerofci/mathesar/discussions/355 has more context on current behavior.
- The conversation on #532 may be useful.
- Discussion on #347 inspired this issue:
   - There are a lot of comments on that PR, relevant ones: [one](https://github.com/centerofci/mathesar/pull/347#issuecomment-876871504), [two](https://github.com/centerofci/mathesar/pull/347#issuecomment-877245010), [three](https://github.com/centerofci/mathesar/pull/347#issuecomment-877373807), [four](https://github.com/centerofci/mathesar/pull/347#issuecomment-877754289), [five](https://github.com/centerofci/mathesar/pull/347#issuecomment-878292814).
- #269 may be useful to look at.  
- [Mathesar's Product Principles](https://wiki.mathesar.org/product/principles)
- We'll need to make sure to thoroughly document the different options with custom types when we create installation documentation for Mathesar. That doesn't seem within scope for this issue, since we have no installation documentation yet.

mathemancer:
It may be possible to solve _all_ of these by altering the `get_supported_alter_column_types` method in the `db.types.alteration` module.

mathemancer:
@pavish Not quite. At least inference will break if we don't have all types installed. Potentially other things as well.

kgodey:
Closing this since it's out of date, we'll track this in our roadmap soon.

PR #532, opened 2021-08-09
Type mapping between Mathesar and database types

### LINKED ISSUE #367: Store friendly to DB type mapping in the backend
Please see https://github.com/centerofci/mathesar/issues/269 for the mapping of types and context.

The plan with this issue is to store the mapping of friendly to DB types so that it can be served as part of the API.

### LINKED ISSUE #368: Extend the `v0/databases/`'s types list to show both Mathesar and DB type information.
Please see https://github.com/centerofci/mathesar/issues/269 for the mapping of types and context.

Blocked by #367

### LINKED ISSUE #369: Ensure that the API only accepts DB types as input everywhere we use types.
Please see https://github.com/centerofci/mathesar/issues/269 for the mapping of types and context.

Places to check:
   - Table API
   - Column API
   - Database API
   - Table preview
   - Table updates
   - Column updates 

As part of this issue, we should also ensure that the name of all our types matches the types in #269.

kgodey:
Fixes #367 
Fixes #368 
Fixes #369

## Details
This PR does the following:
- Stores the type mapping between Mathesar types and DB types in the backend.
- Adds a `/api/v0/databases/<id/types/` endpoint that returns a list of supported types.
- Updates the `/api/v0/databases/` endpoint to remove `supported_types` and instead link to the URL of `supported_types`
- Creates enums for Postgres types and Mathesar custom types and updates the code to use them everywhere.
- Makes some miscellaneous style fixes for readability

### TODO (for author's benefit)
- [x] Store database type mapping in the backend
- [x] Update database endpoint `supported_types` to use new mapping
- [x] Create new `types` endpoint and add URL to DB endpoint
- [x] Update tests for database endpoint
- [x] Update all code in `db` module to use new `PostgresType` and `MathesarCustomType` enums.
- [x] Ensure that the API only accepts DB types as input everywhere we use types
- [x] Write tests to ensure database mapping is generated correctly
- [x] Write tests for `types` endpoint
- [x] Improve code readability

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/types/base.py
```diff
@@ -9,3 +75,11 @@
...
 def get_qualified_name(name):
     return ".".join([preparer.quote_schema(SCHEMA), name])
+
+
+def get_installed_types(engine):
+    return engine.dialect.ischema_names
+
+
+def get_full_mathesar_type_name(name):
+    return f'mathesar_types.{name}'
```
mathemancer, line 85: This should be either replaced by the `get_qualified_name` function, or wrap it (or replace it wherever that function is used).  Otherwise, we risk these getting out of sync, leading to bugs.

db/types/base.py
```diff
@@ -9,3 +75,11 @@
 
 def get_qualified_name(name):
     return ".".join([preparer.quote_schema(SCHEMA), name])
+
+
+def get_installed_types(engine):
+    return engine.dialect.ischema_names
```
mathemancer, line 81: This name could lead to confusion, since the  types are not necessarily installed at the DB level, but only need to be available in the engine.  Also, consider using this function inside of `db.types.alteration.get_supported_alter_column_types` (in case we change where we store the info pulled from `ischema_names` in the future).

db/types/base.py
```diff
@@ -1,6 +1,72 @@
...
+
+
+class MathesarCustomType(Enum):
+    """
+    This is a list of custom Mathesar DB types.
+    Keys returned by get_installed_types are of the format 'mathesar_types.VALUE'
+    """
+    EMAIL = 'email'
+    URI = 'uri'
+    MONEY = 'money'
```
mathemancer, line 67: This results in these strings being defined both here, and in the individual types.  How about using this definition in the `db.types.email` module?  Then creating a new custom type would use this, and anywhere that type is used, the name could be updated by updating the type name here (along with the type schema, if desired).

mathesar/models.py
```diff
@@ -62,8 +63,16 @@ def _sa_engine(self):
 
     @property
     def supported_types(self):
```
mathemancer, line 65: Is this the moment to add support for actually checking the DB to see if the type is supported by the current mathesar installation (i.e., do the db_types actually exist in the DB)?

Or should we wait for a future issue for that?

mathemancer (changes requested):
Overall, really great.  It's so pleasant to see this organization happening.  I have a couple of specific requests in the `db.types.base` module to make sure that we don't repeat ourselves too much (mostly just noting where we should be using the new `Enum`s).

I also think we should either handle checking the DB installation to see if the types are actually there, or put a `TODO` and make an issue to handle it in the future (you choose).

db/types/base.py
kgodey: Thanks, fixed. I opted to use `get_qualified_name` everywhere because it's more robust, it looks at the schema.

db/types/base.py
kgodey: Fixed, I updated most places to use this instead of `engine.dialect.ischema_names` directly. I also changed the name to `get_available_types`.

mathesar/models.py
kgodey: I think we should wait for a future issue. There's already one that should cover it:
- https://github.com/centerofci/mathesar/issues/403

db/types/base.py
kgodey: I updated the `db.types.email` module to use the new enum.

kgodey:
@mathemancer This is ready for re-review. 

I also added an unrelated minor change, which is adding the URLs for type suggestions and previews to the table API.

mathemancer (approved):
I think it looks good!

PR #541, opened 2021-08-10
Add Keyboard Accessibility to Select Component

### LINKED ISSUE #483: Select component keyboard accessibility
## Problem
Issue #374 and PR #460 handle the ARIA configuration for Select component. However, the ARIA keyboard based actions mentioned on w3 spec is yet to be handled. 

## Proposed solution
Keyboard based actions need to be handled in Select component, as per the w3 spec.

## W3 spec
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html
https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox

mr-gabe49:
Fixes #483 

- Handles the following keyboard actions: Enter, Escape, Down Arrow, and Up Arrow. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mr-gabe49:
@pavish This current state works, but I may need feedback to make it look professional. In addition, there are two features that are not ideal. First, no list element is focused when the select list is opened. An user would have to press an up/down arrow first, for a focus to appear on a list element. Second, when an user clicks 'Enter', the select list does not disappear, even though the logic is correct. Thanks for your help!

mr-gabe49:
@pavish Take all the time you need. Let me know if I can assist you in any way.

mathesar_ui/src/components/select/Select.svelte
```diff
@@ -24,25 +24,71 @@
...
+        t += 1; 
+      }
+      if(erasedIndex >= 1){
+        let hoveredElement= document.querySelector(".hovered");
+        hoveredElement.classList.remove("hovered");
+      }
+      currentIndex = currentIndex + index; 
+      items[currentIndex].classList.add("hovered");
+      erasedIndex = 1; 
+    }
```
pavish, line 69: The algorithm to hover elements could be improved.

1. Since we are storing currentIndex, it is not essential to manipulate dom directly i.e manipulating `hovered` class.
    - Instead we can update the template to add class hovered, as follows:
      ```
       <li role='option' ... class:hovered={option[idKey] === options[currentIndex]?.[idKey]} ...>
      ```
    - We can remove all direct manipulation of dom logic that involves classList.add and classList.remove.
    - erasedIndex would not be required, when this change is done.
2. On dropdown open, the currentIndex should be set to the index of the selected item. We can dispatch an event during open/close from the dropdown for this.
3. The need for variable `t` can be avoided.

mathesar_ui/src/components/select/Select.svelte
```diff
@@ -24,25 +24,71 @@
...
+        case "ArrowUp":
+          hoveredItem(-1);
+          break;
+        case "Escape":
+          isOpen = false;
+          break;
+        case "Enter":
+          if (options.length == 0) break;
+          setValue(options[currentIndex]);
+          break;
```
pavish, line 85: * The trigger for dropdown's default behaviour for enter is to switch between open/close. i.e when enter key is pressed on the button, it toggles the state.
* Here, isOpen=false is explicitly mentioned within setValue.

When enter is clicked two things happen:
* Firstly, isOpen is set to false from within setValue.
* Then the button's default behaviour is triggered. It sees that the value for isOpen is false and then switches it to isOpen=true. So the dropdown actually closes and reopens.

It is best to not change isOpen=false when the action happens through keyDown. It should still do it for click events.

pavish (changes requested):
@mr-gabe49 
Good work! I've added a few review comments, to help stabilize this.

Apart from the comments, it is also essential to handle scrollbehaviour when scrollbar is present within the dropdown. But this can be spun off into a different issue.

Please feel free to reach me or the team on matrix if you need additional context/help over this.

mathesar_ui/src/components/select/Select.svelte
mr-gabe49: Thanks for the detailed description, it was very helpful for understanding how it works. I was able to figure how to close the dropdown with 'Enter', however, I still used isOpen = false. I also used .preventDefault to manage the scrollbehaviour.  It prevents the page from scrolling while using the arrow keys. Here is the code: 
```
case "Enter":
          e.preventDefault();
          if (options.length == 0) break;
          value = options[currentIndex];
          dispatch('change', {
            value,
          });
          isOpen = false;
          break;
```
I can try and research other ways to prevent the use of 'isOpen = false'. But will take more time. Let me know.

mathesar_ui/src/components/select/Select.svelte
mr-gabe49: I'm researching more about the dispatch event. I have access to the currentIndex and selected item in the Select component.  Couldn't I pass these variables as props down to the dropdown component?

mathesar_ui/src/components/select/Select.svelte
pavish: @mr-gabe49 I think it should be fine, if you got it to work as expected.

Regarding scroll behaviour, there are two things to consider:
1. The whole page scrolls when keydown is pressed. This can be fixed by preventDefault, as you've done.
2. When there is a scrollbar within the dropdown (if the number of elements are more), upon keydown an element which is not in view could become active. It should be made visible to the user.

Handling the 2nd point can be done separately. I would like to get this PR merged early. We can handle each specific requirement incrementally.

mathesar_ui/src/components/select/Select.svelte
pavish: It is not a good idea to pass component specific props to different components, which are independent.

The Dropdown component is designed to be used independently, it is not specific to Select, so passing these props down would an anti-pattern.

Regarding dispatching events, you can refer the Svelte docs. https://svelte.dev/docs#createEventDispatcher

You could also refer the [Modal component](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/src/components/modal/Modal.svelte), in our repo. It dispatches events upon open/close of modal.

mathesar_ui/src/components/select/Select.svelte
mr-gabe49: @pavish I've handled the changes requested. Only thing left would be managing the scrollbar within dropdown, which we decided to do separately. I would like to be assigned that feature. Would raising an issue be necessary?

Issue #402, opened 2021-07-16
Column alteration should not require custom type casting functions.


kgodey:
## Problem

Our `ALTER COLUMN` behavior assumes we have our custom type casting functions installed, and won't alter a column's type without one. Although this is friendlier to novice users, Mathesar should not require database alterations to function properly.

## Proposed solution

`ALTER COLUMN` behavior should use our custom type casting functions if available and use default Postgres behavior if they are not installed.

## Additional context

- https://github.com/centerofci/mathesar/discussions/355 has more context on current behavior.
- [Mathesar's Product Principles](https://wiki.mathesar.org/product/principles)
- We'll need to make sure to thoroughly document the different options with type casting functions when we create installation documentation for Mathesar. That doesn't seem within scope for this issue, since we have no installation documentation yet.

mathemancer:
A couple of questions we should answer in the issue:

Should we warn the user, or somehow require a special config to enable the DB-default `ALTER` behavior?  Part of the reason for custom functions is to keep newbies out of trouble (of the data-loss kind).  I'm concerned that a non-technical user might not get the installation process correct, and the result would be that that non-technical user end up using less-safe `ALTER` behavior, _because_ they're a novice if we fall back to the less-safe behavior automatically.

kgodey:
I like the idea of requiring a configuration needing to be explicitly set to true to enable the DB-default `ALTER` behavior.

mathemancer:
This is still relevant, even with the new architecture.

kgodey:
Closing this since it's out of date, we'll track this in our roadmap soon.

PR #542, opened 2021-08-10
Implement back end Integer types

### LINKED ISSUE #378: Handle `SMALLINT` data type in the backend.
This issue is to ensure that Mathesar can handle the Postgres `SMALLINT` data type.

- Users can use the API to change a column to `SMALLINT` if it's possible to do so.
- Automatic type inference during file import suggests `SMALLINT` when it makes sense to do so. 
  - I'm not sure this actually ever makes sense, I think we should default to `INTEGER` when we detect integers, but I'm open to differing opinions on when this might make sense.

**Additional Context** 

### LINKED ISSUE #379: Handle `INTEGER` data type in the backend.

### LINKED ISSUE #380: Handle `BIGINT` data type in the backend.

mathemancer:
Fixes #378 
Fixes #379 
Fixes #380 

This adds support for `BIGINT`, `INTEGER`, and `SMALLINT` types in the back end.

**Technical details**

One thing to note is that this PR implements custom casting logic that refuses to cast floats, decimals, etc. that would need to be rounded or truncated to cast to integers.  This is different from the default PostgreSQL behavior for casting other number types to integer types, but is the same as the behavior for casting textual types to integer types.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (changes requested):
@mathemancer The code looks good, but when I try to update a column with type `INTEGER`, it says `This type cast is not implemented`. It actually also says that for `FLOAT`, I didn't check that PR locally since I had checked `NUMERIC` and assumed it was similar code.

I think it's worth writing some tests at the API level as well, to make sure that the expected API endpoints always work.

Also, in the future, we might want to support casting integers even if the number needs to be rounded in order to do so. I can imagine a scenario where someone wants to change a column to integer and doesn't want to manually round all of their data before doing so. We can either handle that case now by taking a flag in the `db` function, or we can handle it when we get to #402. Alternatively, we can create a new issue for it. What do you think?

mathemancer:
> The code looks good, but when I try to update a column with type `INTEGER`, it says `This type cast is not implemented`. It actually also says that for `FLOAT`, I didn't check that PR locally since I had checked `NUMERIC` and assumed it was similar code.

@kgodey Sorry, I should have put better instructions:  You need to knock out the `mathesar_types` schema in your database, then re-run the `install.py` script to update anything that involves changes to the available types.  This is because we still don't have an upgrade path setup for that part of things.

mathemancer:
> Also, in the future, we might want to support casting integers even if the number needs to be rounded in order to do so. I can imagine a scenario where someone wants to change a column to integer and doesn't want to manually round all of their data before doing so. We can either handle that case now by taking a flag in the `db` function, or we can handle it when we get to #402. Alternatively, we can create a new issue for it. What do you think?

I think that's definitely a plausible scenario.  Unfortunately, it won't be handled by #402 , since that's not the default behavior.  Unfortunately, casting to `INTEGER` either rounds, truncates, or fails based on the original type with the default `CAST` function:
```
mathesar=# SELECT 50.5::numeric::integer;
 int4
------
   51
(1 row)

mathesar=# SELECT 50.5::float::integer;
 int4
------
   50
(1 row)

mathesar=# SELECT '50.5'::varchar::integer;
ERROR:  invalid input syntax for type integer: "50.5"
```
 This is part of why I just disallowed it entirely for the moment.  The behavior is a bit inconsistent at the DB level.  I think we should make it a separate issue so we can spec out which behavior(s) we want to support.

kgodey (approved):
@mathemancer Thanks! Wiping out the DB worked. I agree with you about the API tests, I think it would be redundant to check each string.

I found one more bug, which is that when you `PATCH` a column, the `index` in the response is `null` instead of the actual column index. Do you want to fix that here or should I set up a separate issue?

I'll create a separate issue for re-evaluating type casting behavior.

mathemancer:
> @mathemancer I also fixed the merge conflict by reverting your `npm` commit and merging in `master`.

Thank you!

mathemancer:
> I found one more bug, which is that when you `PATCH` a column, the `index` in the response is `null` instead of the actual column index. Do you want to fix that here or should I set up a separate issue?

I think a separate issue would be good.  @eito-fis did some work in the relevant in #526 that would conflict badly with the fix (and might actually solve it, actually).

kgodey:
I added a comment in #526, we can resolve it there if it needs resolving.

PR #578, opened 2021-08-18
Update table PATCH endpoint to work with columns

### LINKED ISSUE #562: Tables PATCH endpoint does not update column properties
## Description
As per the [import preview api documentation](https://wiki.mathesar.org/engineering/architecture/import-preview-api), a patch request to Tables endpoint, with different values for column properties (name, type) should update specified properties.

This is not working at the moment. The request results in a 200, but the column properties are not updated.

## Expected behavior
The specified column properties should be updated.

## To Reproduce
Send a patch requests to any table.

kgodey:
Fixes #562 

The table detail API now accepts `columns` as a valid key in a `PATCH` request. It will updates names and types of columns as well as drop columns. It takes the same options as the table previews endpoint.

## Note to reviewers
Please expand the `test_table_api.py` file, GitHub is hiding it by default as a "large diff".

## Technical details

In order to get multiple operations to happen within a single transaction, we're passing `connection` and `table` objects around. This code is a little messy, but I didn't want to hold this up since it's blocking our next milestone so I created a new issue to refactor it. I also created a separate issue to look into whether we can support altering the table's name in the same request as altering columns.
- #592 
- #593

### Work done in this PR
- [x] Fix existing tests (thanks, @eito-fis!)
- [x] Update table's `PATCH` endpoint to handle updating column names and types
- [x] Ensure that all actions happen in a single transaction.
- [x] Updated `PATCH` function to also handle dropping columns
- [x] Checked that type actually needs to be changed before calling `retype` function.
- [x] Improved code readability and minor refactoring
- [x] Investigated if it was possible to easily alter the table's name in the same request (it wasn't)
- [x] Wrote tests that cover:
  - [x] No changes to columns 
  - [x] Changing names only
  - [x] Changing types only
  - [x] Dropping columns only
  - [x] Changing names and types of different columns
  - [x] Changing names and types of the same column
  - [x] Changing names and types of different columns and dropping a column
  - [x] Changing names and types of the same column and dropping a column
  - [x] Changing name + the type to an invalid type and making sure that the name doesn't change
  - [x] Changing type to an invalid type and dropping a column and making sure that the column doesn't get dropped
  - [x] Changing both column data and table name
  - [x] Doing some of the same things at the `db` level

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@eito-fis @powellc and anyone else interested in backend code: I'm working on supporting a bunch of different operations in one transaction in this PR. it seems to work, but the way I got it to work is by updating a bunch of functions to take `connection_to_use` and `table_to_use` objects and use those instead of opening a new connection or re-reflecting tables. I can see it getting hairy to use very quickly, I'd like to come up with a clean framework for when to open a new connection and when to use an existing connection and create a convention we can follow.

Any suggestions/thoughts would be appreciated

eito-fis:
The sqlalchemy documentation has [some suggestion](https://docs.sqlalchemy.org/en/13/core/connections.html#nesting-of-transaction-blocks) for dealing with the nested transaction pattern, which seem reasonable, if not much different from what we have now.

If we're alright with digging into some of the internals, I think we could also try extending the `Engine` and `Connection` classes to handle something like a `.begin_or_continue()` method (or extend the normal `.begin()` method)? In particular, we could write a modification of the [existing context manager](https://github.com/sqlalchemy/sqlalchemy/blob/5c4a74fc13ca28983706a76ab02ffe09d89d1dbf/lib/sqlalchemy/future/engine.py#L356) that checks to see if [`conn._transaction` exists](https://github.com/sqlalchemy/sqlalchemy/blob/5c4a74fc13ca28983706a76ab02ffe09d89d1dbf/lib/sqlalchemy/engine/base.py#L756) and starts a new transaction accordingly. The [`Engine.begin()`](https://github.com/sqlalchemy/sqlalchemy/blob/5c4a74fc13ca28983706a76ab02ffe09d89d1dbf/lib/sqlalchemy/future/engine.py#L371) method is also very simple, so I don't think it would be much work to extend or make a copy of.

Wrt to the table, I don't have a great solution. Maybe just a function that takes a table and OID, either of which could be `None`, and returns a table? That way we could just replace all the reflection calls with a new function call and have minimal change.

kgodey:
Thanks @eito-fis! Those are all good ideas, I'll think about it and figure out an approach.

kgodey:
Please see the following issues for follow up work related to this issue:
- #592 
- #593

kgodey:
@eito-fis @powellc This is blocking @pavish so we're prioritizing merging it. Please review this even if it is merged, I'll make any code review changes in a new PR.

pavish (approved):
Merging this PR, as it is blocking work on the frontend. Review changes, if any, will be taken up as a separate PR.

db/tests/test_columns.py
```diff
@@ -964,3 +965,150 @@ def test_test_columns_covers_MathesarColumn(mathesar_col_arg):
...
+    updated_table = tables.reflect_table(table.name, schema, engine)
+
+    assert len(updated_table.columns) == len(table.columns) - 2
+    for index, column in enumerate(updated_table.columns):
+        new_column_type = get_db_type_name(updated_table.columns[index].type, engine_email_type)
+        assert new_column_type == column_data[index - 2]['plain_type']
+        assert updated_table.columns[index].name == column_data[index - 2]['name']
+
+
+def test_batch_update_column_all_operations(engine_email_type):
```
eito-fis, line 1095: Really like these tests :+1:

mathesar/tests/views/api/test_table_api.py
```diff
@@ -333,6 +332,13 @@ def test_table_type_suggestion(client, schema, engine_email_type):
     assert response_table == EXPECTED_TYPES
 
 
+def _check_columns(actual_column_list, expected_column_list):
+    # Columns will return an extra type_options key in actual_dict
+    # so we need to check equality only for the keys in expect_dict
+    for index, column_dict in enumerate(expected_column_list):
```
eito-fis, line 338: Small nitpick, it might be nicer to iterate over a zip of the two lists.

eito-fis (commented):
Looks great! Just a couple small comments.

db/columns.py
```diff
@@ -330,20 +339,48 @@ def retype_column(table_oid, column_index, new_type, engine, **kwargs):
...
+def _check_type_option_equivalence(type_options_1, type_options_2):
+    NULL_OPTIONS = [None, {}]
+    if type_options_1 in NULL_OPTIONS and type_options_2 in NULL_OPTIONS:
+        return True
+    elif type_options_1 == type_options_2:
+        return True
+    return False
+
+
+def retype_column_in_connection(table, connection, engine, column_index, new_type, type_options={}):
```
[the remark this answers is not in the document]
kgodey: I was trying out different patterns to figure out which one might work better, I'll refactor this to be more consistent with the other functions.

db/types/alteration.py
```diff
@@ -128,14 +135,13 @@ def alter_column_type(
...
+    execute_statement(engine, DDL(alter_stmt), connection_to_use)
 
     if default is not None:
         cast_stmt = f"{cast_function_name}({default_text})"
-        with engine.begin() as conn:
-            new_default = str(conn.execute(select(text(cast_stmt))).first()[0])
-        columns.set_column_default(table_oid, column_index, new_default, engine)
+        default_stmt = select(text(cast_stmt))
+        new_default = str(execute_statement(engine, default_stmt, connection_to_use).first()[0])
+        columns.set_column_default(table_oid, column_index, new_default, engine, connection_to_use)
```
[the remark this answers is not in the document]
kgodey: Nice catch, thanks!

mathesar/tests/views/api/test_table_api.py
kgodey: Agreed, I'll make the change. Thanks!

PR #586, opened 2021-08-19
Implement `DATE` type in the backend

### LINKED ISSUE #425: Handle `DATE` data type in the backend
This issue is to ensure that Mathesar can handle the Postgres `DATE` data type.

- Users can use the API to change a column to `DATE` if it's possible to do so.
- Automatic type inference during file import suggests `DATE` when it makes sense to do so. 

**Additional Context** 
- #250 (parent issue)

eito-fis:
Fixes #425

Implements the `DATE` type for use in column alterations in the backend.

**Technical details**
Uses the default postgres string -> date conversion, see [here](https://www.postgresql.org/docs/13/datatype-datetime.html#DATATYPE-DATETIME-INPUT).

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/tests/types/test_alteration.py
+++ b/db/tests/types/test_alteration.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import timedelta, date
 from decimal import Decimal
 
 import pytest
@@ -31,6 +31,7 @@ INTERVAL = PostgresType.INTERVAL.value.upper()
 NUMERIC = PostgresType.NUMERIC.value.upper()
 REAL = PostgresType.REAL.value.upper()
 SMALLINT = PostgresType.SMALLINT.value.upper()
+DATE = PostgresType.DATE.value.upper()
 VARCHAR = "VARCHAR"
 
 
@@ -268,6 +269,14 @@ MASTER_DB_TYPE_MAP_SPEC = {
             VARCHAR: {VALID: [(3, "3")]},
         }
     },
+    DATE: {
+        ISCHEMA_NAME: PostgresType.DATE.value,
+        REFLECTED_NAME: DATE,
+        TARGET_DICT: {
+            DATE: {VALID: [(date(1999, 1, 18), date(1999, 1, 18))]},
+            VARCHAR: {VALID: [(date(1999, 1, 18), "1999-01-18")]},
+        },
+    },
     VARCHAR: {
         ISCHEMA_NAME: PostgresType.CHARACTER_VARYING.value,
         SUPPORTED_MAP_NAME: "varchar",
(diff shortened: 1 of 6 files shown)


db/tests/types/test_alteration.py
```diff
@@ -327,6 +336,19 @@
...
             },
+            DATE: {
+                VALID: [
+                    ("1999-01-18", date(1999, 1, 18)),
+                    ("1/18/1999", date(1999, 1, 18)),
+                    ("jan-1999-18", date(1999, 1, 18)),
+                    ("19990118", date(1999, 1, 18)),
+                ],
+                INVALID: [
+                    "18/1/1999",
```
kgodey, line 347: This is a valid date in a non-US format (this is how I grew up writing dates). Does Postgres not parse this?

kgodey (commented):
@eito-fis This looks pretty good. I added a question.

Also, I think it would be good to add some tests for inferring dates on import.

powellc (approved):
Nice work, @eito-fis ! I too am curious about the supported date formats and wonder if there are any others that Postgres doesn't handle 🤔  But otherwise I think it's good to merge 🚀

db/tests/types/test_alteration.py
eito-fis: This is dependent on the server [`DateStyle`](https://www.postgresql.org/docs/13/runtime-config-client.html#GUC-DATESTYLE) option, which defaults to MDY. As its a server wide setting, as of now I don't think we can support parsing multiple different formats. If it is something we want, we might be able to do it with a custom type that has a `format` type option? I'm dealing with a similar problem for the `TIME` type right now, so I'll know how hard it'll be once that's done.

kgodey: Okay, let's not worry about it now, we can open a new issue for improving parsing later if needed.

I'll review again once there's a test for ensuring that dates are inferred correctly on import. Thanks!

kgodey: Also please add a comment in the code about the `DateStyle` option.

eito-fis:
Updated the inference tests with an additional case where we convert `STRING` -> `DATE`.

kgodey (approved):

kgodey:
Looks good, thanks! I think we should add a test at some point at the API layer of importing a CSV with all the data types that we infer and make sure they are inferred correctly. Perhaps you could add this in a future PR?

PR #591, opened 2021-08-19
[FIX]: Double quotes in file headers

### LINKED ISSUE #572: Double quotes in file headers break CSV import
## Description
A CSV file with double quotes in its header names will cause the CSV import to break. This is because we directly insert the CSV header names into the SQL statement that creates the new, imported table.

## Expected behavior
Quotes should either be ignored or included in the column name. Regardless, the import should not fail.

## To Reproduce
Make a copy of any testing CSV, and add double quotes ~around~ inside the name of a column. Then, try to upload.

## Additional context
I think sqlalchemy should automatically escape column names? If this is the case, creating `Column` objects then just using their name or str representation should work.

linzhou-zhong:
This PR is created to handle double quotes in file headers

Fixes #572

**Technical details**

I tried these 3 solutions above:

1. `list(map(lambda col: col.replace('"', ""), reader.fieldnames))`

2. `list(map(lambda col: col.strip('\"'), reader.fieldnames))`

3. `list(map(lambda col: eval(col), reader.fieldnames))`

All three solutions can handle our issue **Double quotes in file headers**. But **solution 2** and **solution 3** cannot deal with the specific corner case: `'" column"_one "'` where a double-quote was added in the middle of the header's name.

As a result, I decide to choose **solution 1** as our first feasible solution.

**Screenshots**

There's a CSV file with illegal double quotes in headers.

Screenshot

Eventually, It can remove all illegal double-quotes right now.

Screenshot

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis (approved):
Looks good to me! Happy to merge once the linting errors get cleaned up.

eito-fis:
Thanks for your work on this! Given that it affects UX to a small degree (no quotes included in column names), I'm going to wait on an additional review from @kgodey before merging.

kgodey:
@linzhou-zhong This looks pretty good, but I'd like to find a way to preserve double quotes if possible since they may be relevant to the user, especially if they are in the middle of the text (for example, people might have a column header that says `Dwayne "The Rock" Johnson`). 

Could we instead escape double quotes if they are in the middle (i.e. not at the beginning or end of a string)? I think turning `"` into `\"` should work.

linzhou-zhong:
After testing a CSV file with double-quotes inside of headers, I tried to convert it into `\"` but I still got an error while uploading. I don't think we're allowed to keep any double-quotes inside of headers (and I've never seen this kind of format before, like `column "hello" one` as a name of column). So maybe removing all double-quotes will be feasible.
What do you think ? @kgodey

linzhou-zhong:
[update]

Instead of keep double-quote `"`, we can directly use single-quote `'`.

Screenshot

@kgodey

kgodey:
@linzhou-zhong Could you share what you did to convert the `"` to `\"` and what the error you got was?

I think using a single quote is a better approach than stripping quotes entirely. Here's a related [Stack Overflow answer](https://stackoverflow.com/questions/10195749/postgresql-copy-and-csv-data-w-double-quotes), in case that inspires anything.

linzhou-zhong:
## input

Here is my test CSV with double-quote which covers the entire header's name and also appears in the middle of the header's name.

screenshot_1

## code
`reader.fieldnames = list(map(lambda col: col.strip('"').replace('"', '\"'), reader.fieldnames))`

`strip('"')` allows us to remove double-quote from head and tail, replace('"', '\"') helps us to replace `"` with `\"`

## upload
When I upload this CSV, I got an error.

screenshot_2

## solution

We can directly replace `"` with `'` instead of double-quote. 

### code

reader.fieldnames = list(map(lambda col: col.strip('"').replace('"', "'"), reader.fieldnames))

I've already tested this code and everything goes well.

@kgodey

kgodey:
@linzhou-zhong Could you post the actual backend error that you get when you try to upload the CSV? You can either get that by running the Python test that you wrote or from the UI by using the `Network` tab in your browser's web developer tools to inspect the request that is erroring. You should be able to see a response there.

I think the single quotes are a good option, but I would like to understand why escaping the double quotes is failing before making the final decision.

powellc (approved):

linzhou-zhong:
[UPDATE]

The reason is evident, please see this log returned by the app.

`psycopg2.errors.SyntaxError: syntax error at or near "middle"`
` LINE 1: COPY "public"."double_quota 9" ("left"middle"right-1","left"...`

If one of header's name is `"left"middle"right-1"` which contains double-quotes at the head, tail and also in the middle of the String. When our app tries to read its header's name, it'll be recognized as `"left"` + `middle` + `"right-1"` where **middle** is considered as a variable but unknow.

So I would like to replace `"` with `'` instead of keeping its original value.

@kgodey

linzhou-zhong:
@kgodey Could you please check my latest version?

kgodey (changes requested):
@linzhou-zhong Thanks for your work on this! Your code looks good, but I'd like to try a couple of approaches to preserve the double quotes to see if they work before I merge this. It's important to me that we try and preserve user data intact if we can make it work technically.

From the error log you posted, it looks like the PostgreSQL `COPY` function is failing. There are some additional arguments we could pass to it that would make it deal with double quotes better (e.g. `QUOTE`, `FORCE_QUOTE`).

If you wanted to look into this, we're constructing the `COPY` call in the `create_records_from_csv` function in `db.records`, and here's the relevant documentation on [the Postgres docs](https://www.postgresql.org/docs/current/sql-copy.html) and [`psycopg2` docs](https://www.psycopg.org/docs/cursor.html#cursor.copy_expert). Otherwise, I'll try playing around with it when I have some time and merge your code as-is if I can't get it to work.

linzhou-zhong:
I would like to verify one more time with several different CSV files. I'll update more information below. Please feel free to ping me before merging this PR. 

Thanks!

linzhou-zhong:
[update]

I tried these 2 CSV files above.

- No space between `Center` and `"USA"`

截屏2021-08-20 下午10 39 01

`reader.fieldnames` returns the result shown below:

`['"Center"NASA"USA"', '"Status"', '"Case Number"', '"Patent Number"', 'Application SN', '"Title,Patent Expiration Date"']`

The header is split correctly.

- One space between `Center` and `"USA"`

screenshot

`reader.fieldnames` returns the result shown below:

`['"""Center', '""NASA""', 'USA""","""Status""","""Case', 'Number""","""Patent', 'Number""",Application', 'SN,"""Title,Patent', 'Expiration', 'Date"""']`

The header isn't split correctly.

## Issue

For the corner case where the header's name contains double-quotes at the head, tail, in the middle and also extra space near the double-quote. CleaverCSV cannot really handle it if we don't specify some extra parameters when we read the file.

## Solution

`reader = csv.DictReader(file, quoting=csv.QUOTE_ALL, quotechar='"', delimiter=',')`

we can specify `quoting`, `quotechar` and `delimiter` which will allow us to keep all double-quote and split correctly.

## TODO

I'll continue to find a solution how to keep the original double-quote without crashing the program.

linzhou-zhong:
[update]

So I found the solution to handle the double-quotes inside of the header's name and right now we can keep its original String.

## Issue

in `records.py`, when we apply the function `create_records_from_csv`, we generated our sql_query as a String which will be passed to `cursor.copy_expert(copy_sql, csv_file)`. It will crash the program if the header's name contains any double-quote in the middle.

For example, a header's name ` "Space "NASA" Station"` will be considered as an illegal variable. Here is an example: [Stackoverflow](https://stackoverflow.com/questions/55550458/psycopg2-syntax-error-because-of-single-quotes-in-insert-command/55550518).

## Solution

I was inspired by this post [Stackoverflow](https://stackoverflow.com/questions/67681689/psycopg2-sql-sql-composed-elements-must-be-composable).

I convert our sql_query to `sql.SQL` and header's name to `sql.Identifier`both of them are based on psycopg2.

**Here is my input CSV**

input_csv

**When CSV is uploaded**

upload

**Our table has been created and the header's name is kept in its original format**

result

@kgodey Here is my solution, tell me what you think please :)

db/records.py
```diff
@@ -299,23 +300,30 @@ def create_records_from_csv(
...
+
+            copy_sql = sql.SQL(
+                "COPY {relation} ({formatted_columns}) FROM STDIN CSV {header} {delimiter} {escape} {quote}"
+            ).format(
+                relation=relation,
+                formatted_columns=formatted_columns,
+                header=sql.SQL("HEADER" if header else ""),
+                delimiter=sql.SQL(f"DELIMITER E'{delimiter}'" if delimiter else ""),
+                escape=sql.SQL(f"ESCAPE '{escape}'" if escape else ""),
+                quote=sql.SQL(
```
kgodey, line 321: Could you add a comment explaining the logic here for people who may not be familiar with it?

db/records.py
```diff
@@ -299,23 +300,30 @@ def create_records_from_csv(
...
+            formatted_columns = sql.SQL(",").join(
+                sql.Identifier(column_name) for column_name in column_names
+            )
+
+            copy_sql = sql.SQL(
+                "COPY {relation} ({formatted_columns}) FROM STDIN CSV {header} {delimiter} {escape} {quote}"
+            ).format(
+                relation=relation,
+                formatted_columns=formatted_columns,
+                header=sql.SQL("HEADER" if header else ""),
```
kgodey, line 318: Do we need to call `sql.SQL` on all the format strings since we're already calling it on the main string?

db/records.py
```diff
@@ -299,23 +300,30 @@ def create_records_from_csv(
...
-                copy_sql += " HEADER"
-            if delimiter:
-                copy_sql += f" DELIMITER E'{delimiter}'"
-            if escape:
-                copy_sql += f" ESCAPE '{escape}'"
-            if quote:
-                if quote == "'":
-                    quote = "''"
-                copy_sql += f" QUOTE '{quote}'"
+            relation = sql.SQL(".").join(
```
kgodey, line 306: Do we need to call `sql.SQL` on the `"."`? It's only one character.

db/records.py
```diff
@@ -299,23 +300,30 @@ def create_records_from_csv(
...
+                sql.Identifier(part) for part in (table.schema, table.name)
+            )
+            formatted_columns = sql.SQL(",").join(
```
kgodey, line 309: Do we need to call `sql.SQL` on the `","`? It's only one character.

mathesar/imports/csv.py
```diff
@@ -72,14 +72,16 @@ def get_sv_dialect(file):
 def get_sv_reader(file, header, dialect=None):
     file = TextIOWrapper(file, encoding="utf-8-sig")
     if dialect:
-        reader = csv.DictReader(file, dialect=dialect)
+        reader = csv.DictReader(file, dialect=dialect, quoting=csv.QUOTE_ALL, quotechar='"', delimiter=',')
```
kgodey, line 75: Do we need to pass these options here if we're handling quoting at the SQL level?

mathesar/imports/csv.py
```diff
@@ -72,14 +72,16 @@ def get_sv_dialect(file):
...
     else:
-        reader = csv.DictReader(file)
+        reader = csv.DictReader(file, quoting=csv.QUOTE_ALL, quotechar='"', delimiter=',')
     if not header:
         reader.fieldnames = [
             f"column_{i}" for i in range(len(reader.fieldnames))
         ]
         file.seek(0)
+    else:
```
kgodey, line 83: Do we need to still do this if we're handling the quotes at the SQL level?

mathesar/imports/csv.py
kgodey: I also don't think we should pass a `delimiter` since this function also handles `TSV` files.

kgodey (changes requested):
This approach looks great overall! I had some comments and questions, but they're hopefully small.

I think we need to standardize on how we use/quote raw SQL in our code, we're using SQLAlchemy's `engine.dialect.identifier_preparer` in some other places for a similar purpose. I'll create a separate issue to track that, though, no need to fix here (edit: here's the issue: #597).

kgodey:
@linzhou-zhong Also, please ensure that existing tests pass.

db/records.py
linzhou-zhong: Unfortunately, when we call `sql.Indentifier` which will return a **Identifier** class (based on sql.Composable) instead of a **String**. So we have to convert our `","` to sql.Composable using sql.SQL. 

Otherwise, we'll receive such an error: 
```
 File "/code/db/records.py", line 306, in create_records_from_csv

    relation = ".".join(

TypeError: sequence item 0: expected str instance, Identifier found
```

db/records.py
linzhou-zhong: Please take a look at my comment above.

db/records.py
kgodey: Okay, thanks! It would be useful to add a comment explaining this so that other people reading the code will also know it.

db/records.py
linzhou-zhong: Yes, we need to recall `sql.SQL` for each component, because sql.SQL is not considered as a `String` but a `sql. Composable` which means we're not allowed to merge these two different types.

For example, when I tried to remove sql.SQL from the header, I received this warning, meanwhile, the program will crash.

Screenshot

db/records.py
linzhou-zhong: Copy that. Thank you kgodey !

mathesar/imports/csv.py
linzhou-zhong: I don't think we need to keep this part. So I'll remove it.

mathesar/imports/csv.py
linzhou-zhong: When we upload a "bad format" csv with double-quotes, dialect.delimiter cannot return the correct value.

`dialect SimpleDialect(' ', '', '')`

linzhou-zhong:
I've passed all tests locally, would you mind approving running workflows in order to check all tests one more time ? @kgodey

db/records.py
```diff
@@ -299,23 +300,36 @@ def create_records_from_csv(
...
         with engine.begin() as conn:
             cursor = conn.connection.cursor()
-            relation = '.'.join('"{}"'.format(part) for part in (table.schema, table.name))
-            formatted_columns = '({})'.format(','.join([f'"{column_name}"' for column_name in column_names]))
-
-            copy_sql = f'COPY {relation} {formatted_columns} FROM STDIN CSV'
-            if header:
-                copy_sql += " HEADER"
-            if delimiter:
-                copy_sql += f" DELIMITER E'{delimiter}'"
```
kgodey, line 312: We should preserve the `E` here. See: https://stackoverflow.com/questions/34823158/whats-the-e-before-a-postgres-string

kgodey (changes requested):
@linzhou-zhong Looks great, just one small change requested.

Also a couple of general things for all PRs:
- It would be better not to apply formatting changes to functions that you're not changing since it's harder to see/review the specific changes you've made.
- Please don't resolve your own comments (see our [Code Review Guidelines](https://wiki.mathesar.org/engineering/code-review) under "Authors")

db/records.py
linzhou-zhong: @kgodey Copy that and I'll pay attention to my next PRs. Thanks

kgodey (approved):
Looks great, thanks for your work on this!

PR #610, opened 2021-08-28
Set currentIndex equal to selected item only if selected item is within options

### LINKED ISSUE #609: UpArrow does not work in the group section dropdown when data is grouped.
## Description

- When a user groups data by more than one column, the UpArrow in the 'add new group column' button does not work initially. 

## Expected behavior

- When the dropdown is open and a user clicks the upArrow the last item in the dropdown list should be focused. 

## To Reproduce

- Group data by a specific column
- Click on 'add new group column' button
- Click to open dropdown 
- Press UpArrow

## Environment
 - Browser: Chrome
## Additional context

Screenshot (16)

mr-gabe49:
Fixes #609 

- Adds an if statement to make sure currentIndex is set to selected item only if selected item is present within the dropdown list elements. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/components/select/Select.svelte
+++ b/mathesar_ui/src/components/select/Select.svelte
@@ -45,7 +45,7 @@
   }
 
   function setSelectedItem() {
-    if (options.some((e) => e.label === value[labelKey])) {
+    if (options.some((e) => e[idKey] === value?.[idKey])) {
       currentIndex = options.indexOf(value);
     }
   }
--- a/mathesar_ui/src/components/select/Select.svelte
+++ b/mathesar_ui/src/components/select/Select.svelte
@@ -45,7 +45,9 @@
   }
 
   function setSelectedItem() {
-    currentIndex = options.indexOf(value);
+    if (options.some(e => e.label === value[labelKey])){
+      currentIndex = options.indexOf(value);
+    }
   }
 
   function hoveredItem(index) {
@@ -56,6 +58,7 @@
     } else {
       currentIndex += index;
     }
+    console.log(currentIndex);
   }
 
   function keyAccessibility(e: KeyboardEvent): void {
--- a/mathesar_ui/src/components/select/Select.svelte
+++ b/mathesar_ui/src/components/select/Select.svelte
@@ -45,8 +45,9 @@
   }
 
   function setSelectedItem() {
-    if (options.some((e) => e[idKey] === value?.[idKey])) {
-      currentIndex = options.indexOf(value);
+    const index = options.findIndex((e) => e[idKey] === value?.[idKey]);
+    if (index > -1) {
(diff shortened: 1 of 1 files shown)


mathesar_ui/src/components/select/Select.svelte
```diff
@@ -45,7 +45,9 @@
   }
 
   function setSelectedItem() {
-    currentIndex = options.indexOf(value);
+    if (options.some((e) => e.label === value[labelKey])) {
```
pavish, line 48: The idKey exists for the sole purpose of acting as the identifier for an option, in cases of checking conditions as such.

We can change this to:
```suggestion
    if (options.some((e) => e[idKey] === value?.[idKey])) {
```

pavish (changes requested):

mathesar_ui/src/components/select/Select.svelte
```diff
@@ -45,7 +45,9 @@
   }
 
   function setSelectedItem() {
-    currentIndex = options.indexOf(value);
+    if (options.some((e) => e[idKey] === value?.[idKey])) {
+      currentIndex = options.indexOf(value);
+    }
```
pavish, line 50: @mr-gabe49 Sorry for not catching this earlier.

The value object need not necessarily be equal to the one in options. They may hold the same values of idKey and valueKey, but be two different objects. So, checking indexOf will return -1, even though we would expect it to return a value.

A better way to do this will be to:

```suggestion
   const index = options.findIndex((e) => e[idKey] === value?.[idKey]);
    if (index > -1) {
      currentIndex = index;
    }
```

pavish (changes requested):

mathesar_ui/src/components/select/Select.svelte
mr-gabe49: @pavish Interesting, I'll go ahead and fix it. That's always the case even if the value is set using the options object?

pavish: @mr-gabe49 No. When the value is set using the options object, it's not the case, because the memory reference for the entries would be the same.

But in some cases, like when the page initially loads (when a filter or group is already applied), we do not set the value from the options. We directly set it based on the url. In such cases, the memory references vary, i.e. the entry within options array would be a different object from the value.

pavish (approved):

mathesar_ui/src/components/select/Select.svelte
mr-gabe49: Got it, Thanks!

PR #616, opened 2021-08-31
[FIX]: Uncaught 404s in record viewsets and pagination classes

### LINKED ISSUE #517: Uncaught 404s in record viewsets and pagination classes
## Description
The record viewset, column viewset and paignation classes regularly call `Table.objects.get(id=table_pk)`, which throws a `mathesar.models.Table.DoesNotExist: Table matching query does not exist.` when an invalid table id is passed.

To recreate, run `client.get(f'/api/v0/tables/3000/records/')`.

## Expected behavior
We should ensure that the table exists before querying, or catch the `DoesNotExist` error after querying. We should also include tests for table 404s. 

This is probably best done after  #488 is merged, as it includes a function to do exactly this.

linzhou-zhong:
Fixes #517

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


linzhou-zhong:
@kgodey I provided an alternative solution for our #517 issue where we firstly test if `id` already exists in our `Table`, otherwise we raise an Exception.

Please tell me if I misunderstood this issue.

kgodey (changes requested):
@linzhou-zhong Rewriting the `get_table_or_404` function is unnecessary, it works fine as-is. The issue that needs to be fixed in this pull request is to use that function instead of getting the table by ID directly in functions that support the API.

For example, in `pagination.py`, there are several places in which this line of code appears:
`table = queryset.get(id=table_id)`

We need to replace that with the `get_table_or_404` function so that the API returns a 404 error instead of throwing an exception when a table is not found.

Please also look for other places in `views/api.py` where we get the table directly instead of using the `get_table_or_404` function.

mathesar/tests/views/api/test_table_api.py
```diff
@@ -715,6 +715,10 @@ def test_table_404(client):
     assert response.json()['detail'] == 'Not found.'
 
 
+def test_table_not_exist():
```
kgodey, line 718: Please remove this test, you're just testing Django's default functionality here, not any of our code.

linzhou-zhong: Removed

linzhou-zhong:
@kgodey I tried to import `mathesar.api` in file 'pagination.py' and an error of `(most likely due to a circular import)` was returned. So I just created another function also named `get_table_or_404` in the file 'pagination.py'.

mathesar/pagination.py
```diff
@@ -3,6 +3,17 @@
 from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.response import Response
 
+from mathesar.models import Table
+from rest_framework.exceptions import NotFound
+
+
+def get_table_or_404(queryset, pk):
```
kgodey, line 10: Instead of repeating this code in two places, please create a `mathesar/utils.api.py` file and put this function there. You can import it into both this file and `viewsets.py`. That way, if we ever change this function, we only need to change it in one place.

kgodey (changes requested):
@linzhou-zhong Looks good, one comment on code organization.

I assume you also looked for this pattern in `api.py` and didn't find anything to replace?

mathesar/pagination.py
linzhou-zhong: Eventually, I used `from django.shortcuts import get_object_or_404` to replace `queryset.get(id)`

kgodey (changes requested):
@linzhou-zhong I'm worried that `get_object_or_404` will not return the correct JSON error response since Django returns HTML responses by default. That's why I wrote a separate function for it instead of using it.

Can you write a test for the API that makes sure that a JSON response is returned in this case? If it is, please remove the custom function I wrote and replace all uses of it with the `get_object_or_404` function for consistency.

linzhou-zhong:
@kgodey You're right, a  `<class 'django.http.response.Http404'>` will return.

linzhou-zhong:
I created an independent `get_table_or_404` method for `pagination.py`.

kgodey (changes requested):
@linzhou-zhong Actually, you can use the one in `mathesar/api/utils.py`, you don't need to create a new one.

https://github.com/centerofci/mathesar/blob/master/mathesar/api/utils.py

linzhou-zhong:
@kgodey Yes, `mathesar/api/utils.py` was created by me from this [commit](https://github.com/centerofci/mathesar/pull/616/commits/0ce2dac28a9dcd4211af776238956d9183b469c6). 

And I've applied it to `pagination.py` already.

kgodey:
@linzhou-zhong There's a new file in this PR at `mathesar/utils/api.py` which is being used in your changes. Please delete that and use the one in `mathesar/api/utils.py`.

linzhou-zhong:
@kgodey I tried to add some doc for this method if it's useful.

kgodey (approved):
Looks good @linzhou-zhong, thanks! I made a minor update to the docs for clarity.

PR #617, opened 2021-09-01
Implement adding a new empty table

### LINKED ISSUE #444: Implement adding a new empty table
## Problem

Users may want to create an empty table.

## Proposed solution

The ["Basic Table Operations" design spec](https://wiki.mathesar.org/design/specs/table-operations) has a solution for this, which we need to implement on the frontend. We will be implementing a different design than the one in the spec, which will need to be improvised.

This issue involves the following portions of the spec:
- _User Experience_: 
  - User adds a new table
  - User edits a new table name
  - User adds an empty table
- _Interactions_: 
  - Sequential Table Names
 
Please note that we'll need to generate and display a default name of the table that the user can use. Following the logic in #449 is recommended.

## Additional context

- Backend work: #184 
- Design issue: #185
- #449 is related.

Logs of conversation on Matrix from @pavish:

> For adding an empty table, instead of showing Empty table option within the tab, I think it might be better to show it on the top.
> 
> The Add new table button can be changed to a dropdown with two options. 1. Empty table, 2. Import data.
> 
> On clicking empty table, it would just send a create table request and create the table. Then the user can rename it if they want.
> 
> This will have clean separation of concerns. Because Import data, involves importing from csv, remote file import and then copy/paste from spreadsheet..
> 
> Empty table, would just create a new empty table.

seancolsen:
Fixes #444 by allowing users to create a new empty table

### Before

- POST to `/tables/` without the `data_files` parameter would fail with an internal error.

### After

- POST to `/tables/` only requires the `schema` parameter. Such a request will create a new empty table with an automatically-generated name and no data.
- Code within `TableViewSet.create` is also slightly more succinct and readable.

### Before

- Users can only add a new table if they have some data to upload.

before

### After

- Clicking "New table" and then "Import data" has the same behavior as the "New table" button did previously.
- Clicking "New table" and then "Empty table" adds a new empty table and opens a new tab for that table.
- _This PR does not add or change any code for user to set table names. Rather (as specified in the ticket) the user is expected to create an empty table and then rename it to their liking afterwards._

after

## Notes

- On `mathesar_ui/src/header/Header.svelte:28` I don't really like the inline type for `postAPI`. I'd rather see this whole function call moved into a store somewhere. But I'm not sure of the best place within the rest of the codebase to organize this sort of code. I looked for a type definition for the API response from the `/tables/` endpoint but couldn't immediately find one. I'm happy to make changes here as needed.
- When running the linter locally I was getting a linting error _"Unsafe assignment of an `any` value  @typescript-eslint/no-unsafe-assignment"_ in `mathesar_ui/src/header/Header.svelte:29` which is why I'm using `as number` (which I generally avoid). I didn't expect this error, and I didn't spend the time fo fully diagnose it. I'm happy to make adjustments here if needed. Some things to note:
    - I ran into some issues while installing Mathesar, so I ended up running the linter from outside the Docker container.
    - I noticed that the `currentSchemaId` store has a type annotation which evaluates to `number`, but a value which can be `null`. I'm surprised TS is not complaining about that. Small code-smell. Could be related. I don't know.

## Limits of scope

Since I'm new I wanted to keep this PR as focused as possible. Here are some tasks I did _not_ complete now but which represent the type of work I would would expect to regularly incorporate into my PRs after becoming more familiar with the project:

- Add test to assert that POST to `/tables/` with only `schema` param succeeds.
- After choosing "Empty table", display a loading spinner somewhere (probably in place of the "New table" drop down) to show the user that we're waiting for the API to create the table.
- If the server fails to create the empty table, display an error message to the user.

## Checklist
- [ ] ~I added tests for the changes I made (if applicable).~
- [ ] ~I added or updated documentation (if applicable).~

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/header/Header.svelte
+++ b/mathesar_ui/src/header/Header.svelte
@@ -7,9 +7,9 @@
     faUpload,
     faTable,
   } from '@fortawesome/free-solid-svg-icons';
-  import { postAPI } from '@mathesar/utils/api';
+  import { createTable, refetchTablesForSchema } from '@mathesar/stores/tables';
+  import { currentSchemaId } from '@mathesar/stores/schemas';
   import { currentDBName } from '@mathesar/stores/databases';
-  import { currentSchemaId, refetchSchema } from '@mathesar/stores/schemas';
   import { newImport, importStatuses } from '@mathesar/stores/fileImports';
   import {
     addTab,
@@ -24,15 +24,12 @@
   import SchemaSelector from './schema-selector/SchemaSelector.svelte';
   import ImportIndicator from './import-indicator/ImportIndicator.svelte';
 
-  async function createEmptyTable() {
-    const response = await postAPI<{ id: number, name: string }>('/tables/', {
-      schema: $currentSchemaId as number,
-    });
-    const { id, name } = response;
-    await refetchSchema($currentDBName, $currentSchemaId);
-    addTab($currentDBName, $currentSchemaId, { id, label: name });
+  async function handleCreateEmptyTable() {
+    const table = await createTable($currentSchemaId);
+    await refetchTablesForSchema($currentSchemaId);
+    addTab($currentDBName, $currentSchemaId, { id: table.id, label: table.name });
   }
-
+  
   function beginDataImport() {
     if ($currentDBName && $currentSchemaId) {
       const fileData = get(newImport($currentDBName, $currentSchemaId));
@@ -71,7 +68,7 @@
           </svelte:fragment>
           <svelte:fragment slot="content">
             <div class="new-table-options">
-              <Button on:click={createEmptyTable} appearance="plain">
+              <Button on:click={handleCreateEmptyTable} appearance="plain">
                 <Icon data={faTable} size="0.8em"/>
                 <span>Empty table</span>
               </Button>
--- a/mathesar_ui/src/header/Header.svelte
+++ b/mathesar_ui/src/header/Header.svelte
@@ -3,10 +3,13 @@
   import {
     faDragon,
     faUser,
+    faPlus,
+    faUpload,
+    faTable,
   } from '@fortawesome/free-solid-svg-icons';
-
+  import { postAPI } from '@mathesar/utils/api';
   import { currentDBName } from '@mathesar/stores/databases';
-  import { currentSchemaId } from '@mathesar/stores/schemas';
+  import { currentSchemaId, refetchSchema } from '@mathesar/stores/schemas';
   import { newImport, importStatuses } from '@mathesar/stores/fileImports';
   import {
     addTab,
@@ -15,17 +18,27 @@
   import {
     Icon,
     Button,
+    Dropdown,
   } from '@mathesar-components';
 
   import SchemaSelector from './schema-selector/SchemaSelector.svelte';
   import ImportIndicator from './import-indicator/ImportIndicator.svelte';
 
-  function createNewTable() {
+  async function createEmptyTable() {
+    const response = await postAPI<{ id: number, name: string }>('/tables/', {
+      schema: $currentSchemaId as number,
+    });
+    const { id, name } = response;
+    await refetchSchema($currentDBName, $currentSchemaId);
+    addTab($currentDBName, $currentSchemaId, { id, label: name });
+  }
+
+  function beginDataImport() {
     if ($currentDBName && $currentSchemaId) {
       const fileData = get(newImport($currentDBName, $currentSchemaId));
       addTab($currentDBName, $currentSchemaId, {
         id: fileData.id,
-        label: fileData.name || 'New table',
+        label: fileData.name || 'Import data',
         isNew: true,
       });
     }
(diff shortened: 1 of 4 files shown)


mathesar_ui/src/header/Header.svelte
```diff
@@ -15,17 +18,27 @@
...
   import SchemaSelector from './schema-selector/SchemaSelector.svelte';
   import ImportIndicator from './import-indicator/ImportIndicator.svelte';
 
-  function createNewTable() {
+  async function createEmptyTable() {
+    const response = await postAPI<{ id: number, name: string }>('/tables/', {
+      schema: $currentSchemaId as number,
+    });
+    const { id, name } = response;
+    await refetchSchema($currentDBName, $currentSchemaId);
```
pavish, line 32: This logic can be moved to `tables` store (/mathesar_ui/src/stores/tables.ts).

`refetchSchema` currently only fetches schema related information, and not the tables within the schema. The `refetchTablesForSchema` method within `tables` store can be used to refetch the table list.

As for the type definition for the post response, the type `TableEntry` can be used.

mathesar/views/api.py
```diff
@@ -99,26 +99,15 @@ def create(self, request):
         serializer = TableSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
 
-        if not serializer.validated_data['name']:
```
kgodey, line 102: Thanks for improving this, it's more readable. :)

mathesar_ui/src/header/Header.svelte
```diff
@@ -15,17 +18,27 @@
...
+    Dropdown,
   } from '@mathesar-components';
 
   import SchemaSelector from './schema-selector/SchemaSelector.svelte';
   import ImportIndicator from './import-indicator/ImportIndicator.svelte';
 
-  function createNewTable() {
+  async function createEmptyTable() {
+    const response = await postAPI<{ id: number, name: string }>('/tables/', {
+      schema: $currentSchemaId as number,
```
pavish, line 29: The linter error for `Unsafe assignment of an any value`, is a bug/known limitation in the eslint plugin for svelte, when dealing with stores. Reference: https://github.com/sveltejs/eslint-plugin-svelte3/issues/104

This error is not thrown if we use the get store method i.e., get(currentSchemaId).

This issue only occurs when dealing with .svelte files. When this logic is moved to the store, the assertion will not be required.

pavish (changes requested):
@seancolsen Thanks for the PR!
It looks good to me, I've added some minor comments.

pavish:
@seancolsen Regarding Typescript not complaining about `currentSchemaId` allowing `null`, we currently do not have strict null checks enabled (https://www.typescriptlang.org/tsconfig#strictNullChecks).

It was not enabled intentionally to avoid unnecessary boilerplate conditionals, but seeing that we will have more contributors working on smaller tasks, I think it might be worth it to enable it.

@kgodey I'll create a bug report and handle this on the stability milestone.

seancolsen:
@pavish

- I've rebased to resolve conflicts.
    - If you prefer making merge commits instead of rebasing, I'm happy to do that in the future.
- I've added one additional commit which I hope will address your requested changes.
- I've moved the table-creation logic to a store.
    - The store certainly feels like the right place for this function.
    - And indeed, this change fixed the linting issue. Thanks for explaining the root cause there!
- I've used the `TableEntry` type. Again, much better -- thank you.
- I've changed `refetchSchema`, to `refetchTablesForSchema`.
    - A couple things to note, here...
    - During my initial work on this PR I was using `refetchSchema` and observing the list of tables refresh after adding a new empty table. Then, when I returned to this PR today, I tested my changes with a fresh DB and no longer observed the list of tables refreshing after adding a new empty table. It's strange to me that I observed different behavior in those different situations, though I didn't spend time trying to figure out why. Oh well.
    - After changing to `refetchTablesForSchema`, I _do_ now observe the list of tables refresh, however the list is not in the order I would expect. Before adding a new table, the list appears to be in alphabetical order, which is what I'd expect. After adding a new table, the list appears to be ordered with my new table second, regardless of its name. I have not spent any time attempting to diagnose or fix the list ordering issue. If we can consider that issue separate from this PR, I'll be happy to open a new ticket for it so that we can address it after merging this PR.

mathesar_ui/src/stores/tables.ts
```diff
@@ -120,6 +121,13 @@ export function getTablesStoreForSchema(schemaId: SchemaEntry['id']): Writable<D
   return store;
 }
 
+export function createTable(
+  schema: SchemaEntry['id'],
+  name: string | undefined = undefined,
```
pavish, line 126: An alternate representation could be:

```suggestion
  name?: string,
```

I personally find this cleaner.

We do not have to make any changes for this. I only want to understand if you had reasons for explicitly mentioning `undefined`.

pavish (approved):

pavish:
@seancolsen Looks great! Thank you.

> If you prefer making merge commits instead of rebasing, I'm happy to do that in the future.

Sure. As a team, we prefer merge commits, over squashing or rebasing. Here is our [wiki page](https://wiki.mathesar.org/en/engineering/code-review#why-not-squash-or-rebase-merges), which contains general guidelines. There is a section of links, which give more context on why we prefer merge commits over squash or rebase.

> During my initial work on this PR I was using refetchSchema and observing the list of tables refresh after adding a new empty table.. ... .It's strange to me that I observed different behavior in those different situations, though I didn't spend time trying to figure out why.

This was a recent change, that happened after you raised the PR. When you started on the PR, we had basic table information needed for the left pane, as part of the schema store. We then decided to use the table store directly, and remove the table list from the schema store and endpoint.

> After changing to refetchTablesForSchema, I do now observe the list of tables refresh, however the list is not in the order I would expect.

Please raise a separate issue for this. We would need to query the table endpoint, with an explicitly mentioned order. We can address this in the stability milestones.

mathesar_ui/src/stores/tables.ts
seancolsen: Thanks for pointing that out. I prefer that as well and have opened #633 to implement that change. My head has been in Python and mypy for the past couple months so it just didn't occur to me.

Issue #385, opened 2021-07-15
Implement filtering options for Number types.


kgodey:
## Problem
We need to ensure that records that include columns of all Number types support the following filters via API: 
- between {x} and {y}
- equals {x}
- does not equal {x}
- greater than {x}
- lesser than {x}
- greater than or equal to {x}
- less than or equal to {x}
- is empty
- is not empty

This involves:
- Implementing the filters in the backend
- Updating the `/api/v0/databases/<id>/types/` endpoint to store available filters on this type
  - Filter information should include the number of parameters needing to be passed in (e.g. `between` needs 2 parameters, `is empty` needs 0)

## Additional context
- We're using our fork of `sqlalchemy-filters` to provide filtering. See: https://github.com/centerofci/sqlalchemy-filters.
- #557 provides some context on storing filters.
- This issue is blocked by implementation of Number types: 
  - #377
  - #378 
  - #379 
  - #380 
  - #381 
  - #382 
  - #383 
  - #434

mathemancer:
I suggest making sure we use inclusive `BETWEEN` (i.e., `x BETWEEN y AND Z` should be equivalent to `x >= y AND x <= z` rather than `x > y AND x < z`).  This will make things consistent with the `BETWEEN` operator in PostgreSQL.  

Do we want to use symmetric between, or not?  I.e., should `5 BETWEEN 8 AND 2` be true, or false?  (for details about symmetric between, see:  https://www.postgresql.org/docs/13/functions-comparison.html).

kgodey:
I agree with using inclusive `BETWEEN`. I think using symmetric between also makes sense.

dmos62:
Since "between" is ambiguous about its symmetry and inclusiveness, maybe instead let's have the user create/compose two simpler filters where these properties are explicit: `a >= x AND a <= y` instead of `a BETWEEN x AND y`. I doubt there's a performance concern, but if there were we could use pattern matching to optimize the filter behind the scenes. Seems desirable to not have the user wonder what "between" means.

mathemancer:
I definitely think we should have the former, but I'm not convinced that should be in lieu of the latter.  Maybe initially, though.  @kgodey I note that `sqlalchemy-filters` doesn't include a `between` operator, so we'd have to write one custom for that.  I guess I'm ambivalent w.r.t. whether `BETWEEN` should be prioritized at the moment.

kgodey:
I'm fine with skipping `between`.

[...]

PR #618, opened 2021-09-01
Add first pass at filter docs for API #388

### LINKED ISSUE #388: Implement filtering options for Boolean types
## Problem
We need to ensure that records that include columns of Boolean type support the following filters via API: 
- is {true/false}
- is not {true/false}
- is empty
- is not empty

This may already be done, please verify that it works and add some test cases specifically for boolean data types.

This involves:
  - Filter information should include the number of parameters needing to be passed in (e.g. `is` needs 1 parameter, `is empty` needs 0)

## Additional context
- See also: #387

powellc:
This commit adds a first pass at trying to add options for filtering to
the /api/v0/databases/<id>/types/ output.

The idea here is to make it pretty trivial to add the options for new
types to the endpoint. I'm not totally sold on the this being the best
format per-se, but I like how it gets injected automatically for a given
type and depends only on providing a definition. The definition can
change, but the place they go stays constant.

Fixes #388

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/filters.py
+++ b/mathesar/filters.py
@@ -1,16 +1,23 @@
 from django_filters import BooleanFilter, DateTimeFromToRangeFilter
 from django_property_filter import PropertyFilterSet, PropertyBaseInFilter, PropertyCharFilter
 
+from mathesar.database.types import MathesarTypeIdentifier
 from mathesar.models import Schema, Table, Database
 
-FILTER_OPTIONS_BY_TYPE_NAME = {
-    'Boolean':
-        {
-            'eq': '<boolean>',
-            'ne': '<boolean>',
-            'is_null': None,
-            'is_not_null': None
-        },
+FILTER_OPTIONS_BY_TYPE_IDENTIFIER = {
+    MathesarTypeIdentifier.BOOLEAN.value:
+    {
+        "db_type": "BOOLEAN",
+        "options": [{
+            "op": "eq",
+            "value": {
+                "allowed_types": ["BOOLEAN"],
+            }
+        }, {
+            "op": "is_null",
+            "value": "null",
+        }]
+    }
 }
 
 
--- a/mathesar/filters.py
+++ b/mathesar/filters.py
@@ -3,6 +3,16 @@ from django_property_filter import PropertyFilterSet, PropertyBaseInFilter, Prop
 
 from mathesar.models import Schema, Table, Database
 
+FILTER_OPTIONS_BY_TYPE_NAME = {
+    'Boolean':
+        {
+            'eq': '<boolean>',
+            'ne': '<boolean>',
+            'is_null': None,
+            'is_not_null': None
+        },
+}
+
 
 class CharInFilter(PropertyBaseInFilter, PropertyCharFilter):
     pass
--- a/mathesar/serializers.py
+++ b/mathesar/serializers.py
@@ -6,6 +6,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.fields import empty
 from rest_framework.settings import api_settings
 
+from mathesar.filters import FILTER_OPTIONS_BY_TYPE_NAME
 from mathesar.models import Table, Schema, DataFile, Database, Constraint
 
 
@@ -225,6 +226,10 @@ class TypeSerializer(serializers.Serializer):
     identifier = serializers.CharField()
     name = serializers.CharField()
     db_types = serializers.ListField(child=serializers.CharField())
+    filter_options = serializers.SerializerMethodField()
+
+    def get_filter_options(self, obj):
+        return FILTER_OPTIONS_BY_TYPE_NAME.get(obj.get('name'))
 
 
 class DatabaseSerializer(serializers.ModelSerializer):
--- a/mathesar/serializers.py
+++ b/mathesar/serializers.py
@@ -6,7 +6,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.fields import empty
 from rest_framework.settings import api_settings
 
-from mathesar.filters import FILTER_OPTIONS_BY_TYPE_NAME
+from mathesar.filters import FILTER_OPTIONS_BY_TYPE_IDENTIFIER
 from mathesar.models import Table, Schema, DataFile, Database, Constraint
 
 
@@ -226,10 +226,10 @@ class TypeSerializer(serializers.Serializer):
     identifier = serializers.CharField()
     name = serializers.CharField()
     db_types = serializers.ListField(child=serializers.CharField())
-    filter_options = serializers.SerializerMethodField()
+    filters = serializers.SerializerMethodField()
 
-    def get_filter_options(self, obj):
-        return FILTER_OPTIONS_BY_TYPE_NAME.get(obj.get('name'))
+    def get_filters(self, obj):
+        return FILTER_OPTIONS_BY_TYPE_IDENTIFIER.get(obj.get('identifier'))
 
 
 class DatabaseSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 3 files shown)


mathesar/serializers.py
```diff
@@ -225,6 +226,10 @@ class TypeSerializer(serializers.Serializer):
     identifier = serializers.CharField()
     name = serializers.CharField()
     db_types = serializers.ListField(child=serializers.CharField())
+    filter_options = serializers.SerializerMethodField()
+
+    def get_filter_options(self, obj):
+        return FILTER_OPTIONS_BY_TYPE_NAME.get(obj.get('name'))
```
kgodey, line 232: I would use the `identifier` instead of name here.

mathesar/filters.py
```diff
@@ -3,6 +3,16 @@
 
 from mathesar.models import Schema, Table, Database
 
+FILTER_OPTIONS_BY_TYPE_NAME = {
+    'Boolean':
```
kgodey, line 7: Please use the `MathesarTypeIdentifier` enum from `mathesar/database/types.py` to get the identifier here so we don't depend on a string that could change.

mathesar/filters.py
```diff
@@ -3,6 +3,16 @@
...
+        {
+            'eq': '<boolean>',
```
kgodey, line 9: I think it might be clearer to have a more descriptive format. I was thinking something like:

```
{
  'op': 'eq',
  'allowed_values': ['boolean'],
  'num_values': 1,
  'description': 'This operation checks for exact equality'
}
```

or

```
{
  'op': 'is_null',
  'allowed_values': ['boolean', None],
  'num_values': 1,
  'description': 'This operation checks for null values'
}
```

`allowed_values` will allow us to specify the values that are acceptable. `num_values` will allow us to support filters that require multiple inputs (such as "between"). I think the description will be useful for API consumers to understand the filter.

This format also meets the "The portion of the API response describing to a given resource should always contain the same set of keys." standard from our [API standards](https://wiki.mathesar.org/en/engineering/architecture/api-standards).

mathesar/filters.py
```diff
@@ -3,6 +3,16 @@
...
+            'ne': '<boolean>',
+            'is_null': None,
```
kgodey, line 11: Why is this `None`? Shouldn't it also be a boolean if it's describing the kind of input it takes?

mathesar/filters.py
kgodey: Also, I think in this case it would be useful to have a helper dictionary mapping `description` to `op` so that we don't have to repeat the same description for the same `op` across different data types.

kgodey (changes requested):
@powellc I think this is a good idea. I had some comments about the format and identifiers.

kgodey:
@pavish Please take a look at the filter format and leave feedback on it since you'll be the main consumer.

pavish:
@kgodey @powellc  I think the following format might be better.

```
{
  "identifier": "number",
  "name": "Number",
  "db_types": [
    "INTEGER",
    "FLOAT",
  ],
  "filter_options": [
    {
      "op": "eq",
      "allowed_values": [
        {
          "identifier": "value",
          "type": "number"
        }
      ]
    },
    {
      "op": "range",
      "allowed_values": [
        {
          "identifier": "min",
          "type": "number"
        },
        {
          "identifier": "max",
          "type": "number"
        }
      ]
    },
    {
      "op": "is_null",
      "allowed_values": []
    }
  ]
}
```
However, this also depends on the filter structure.
When applying the filter, we might need an identifier for the compared values. Currently this is the filter structure: 
```
{
  "and" : [{
    field: "columnId",
    op: "eq",
    value: "somevalue"
  }]
}
```
How would something like range work?

kgodey:
I read through the `sqlalchemy-filters` docs and it doesn't look like they have support for operations with multiple values. That means we'd have to implement `between` on the frontend as two filters with an `and` parameter.

Given that, I think we could do:

```
{
  "identifier": "number",
  "name": "Number",
  "db_types": [
    "INTEGER",
    "FLOAT",
  ],
  "filter_options": [{
      "op": "eq",
      "value": {
         "allowed_types": ["BOOLEAN"],
      }
    },
    {
      "op": "is_null",
      "value": {
         "allowed_types": ["BOOLEAN"],
      }
   }]
}
```

As for types, I think they should be the same values as DB types for now (`BOOLEAN`, `INTEGER`, `TEXT`, etc.) That should keep things simple.

@powellc You're also working on implementing filtering for numeric types (#385), it might make sense to document both together.

pavish:
@kgodey Makes sense. We can do that for `between`.

For op `is_null`, why do we need `value`?
```
{
    "op": "is_null",
    "value": {
       "allowed_types": ["BOOLEAN"],
    }
 }
```

Wouldn't just the `op` suffice?
```
{
  "op": "is_null"
}
```

kgodey:
@pavish You are correct, we don't need `"value"` there. I had assumed you need to pass in `{"op": "is_null", "value": "true"}` or `{"op": "is_null", "value": "false"}`, but it looks like those operators don't take a value.

pavish:
@kgodey For types, the frontend has to be aware of them, inorder to show the correct component for the filter. For eg., Number has to show a number input, while boolean has to show a checkbox.

I think it might be better to use `mathesar_types` here instead of `db_types`.

pavish:
@kgodey 
If the frontend needs to be aware of choosing an input based on a type, the type has to be hardcoded on the frontend as well, for checking the condition.

I was hoping to achieve a plugin based model, where users import a type plugin, which reflects everywhere without the need for frontend code to be changed.

I can use the mapping to identify the related mathesar_type, but we'll still need to hardcode the mathesar_type inorder to check the conditions.

I agree that it is not an API concern, but without this, we cannot build a plugin model for types. Any anyone wanting to develop a frontend for Mathesar would always need to hardcode types.

kgodey:
@pavish I can see an argument for that, but the display options would be defined at the type level, not at the individual filter level. The individual filters would use whatever display options are defined for the types allowed for them. 

That would make it beyond the scope of this issue, we should open a separate issue for it.

pavish:
@kgodey Agreed, we can open a separate issue for it.

> In this case, you don't actually agree that it's not an API concern. :) That's fine, but it's a bit confusing to read when you start with "I agree that it is not an API concern"

Sorry, my bad. I meant to say it's not a concern for API only users, but for anyone wanted to build a frontend for Mathesar.

kgodey:
So in conclusion, the filter format should be something like:

```
{
    "identifier": "boolean",
    "name": "Boolean",
    "db_types": [
        "BOOLEAN"
    ],
    "filters": [{
        "db_type": "BOOLEAN",
        "options": [{
            "op": "eq",
            "value": {
                "allowed_types": ["BOOLEAN"],
            }
        }, {
            "op": "is_null",
            "value": null
        }]
    }]
}

```
@pavish Can you confirm that looks good?

pavish:
@kgodey The format looks good to me!

mathesar/filters.py
powellc: Thanks for this input! I wasn't sure exactly what that dictionary should look like.

powellc:
@kgodey and @pavish thank you so much for hashing through all that! I'll update this PR with that format, and also I'll update the Wiki with a page on API filtering documentation.

kgodey (changes requested):
@powellc Looks great, thanks! Can you add a test to the types endpoint to verify that they're there?

kgodey (approved):
Looks good,. thanks!

PR #620, opened 2021-09-01
API code reorganization

kgodey:
Fixes #510 

Moves API code into a single folder for easier discoverability and splits it up into per-object files. This is all moving code around, there were no logic changes made.

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #625, opened 2021-09-03
DB code reorganization #1

### LINKED ISSUE #582: Reorganize code in `db` and set some standards for imports
## Problem

We have a lot of interrelated code in the `db` module that imports from each other and is prone to circular import errors/

## Proposed solution

We need to reorganize the code so that import paths are tree-like and don't result in circular imports. We should clearly document how the code is organized and where new code should go so that new contributors can easily figure out where code should go and what not to import where.

## Additional context

- #581 
  - Some errors were fixed here. e.g. `tables` imported `columns`, `columns` imported `constraints`, and `constraints` imported `tables`.
  - There was also some circular-ness happening between `types.inference` and `tables`.

kgodey:
- Related to #582 
- Addresses code review feedback from #620

This pull request reorganizes the code in `db` to make it easier to find code and to reduce potential circular import errors.
- `tables.py` is split into:
  - the `ddl/` module, with files for `create`, `delete`, `update`, `split`, `merge`, and `move_columns` files (one per operation type)
  - the `utils.py` file, which contains utility functions to work with tables.
  - Tests are also split accordingly.

Other improvements:
- The `ROSTER` table constants are now pytest fixtures.

TODO:
- [x] Organize `tables.py` and accompanying tests
- [ ] Organize `columns.py` and accompanying tests
- [ ] Organize `constraints.py` and accompanying tests
- [ ] Organize `records.py` and accompanying tests
- [ ] Organize `schemas.py` and accompanying tests
- [ ] Review and organize `types/` folder if needed
- [ ] Organize `constants.py` if needed
- [ ] Organize `engine.py` if needed
- [ ] Organize `install.py` if needed
- [ ] Organize `utils.py` if needed
- [ ] Update pull request
- [ ] Update documentation on wiki

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@mathemancer and @eito-fis, I've reorganized the `tables` code so far and I'd appreciate your thoughts on this approach before I proceed with the rest of the code.

I split the code in `tables.py` into `tables/operations.py` and `tables/utils.py`.
- `operations.py` contains code that actually alters tables.
- `utils.py` contains code that makes working with tables easier, but doesn't alter anything.

I think this division will mostly work for the rest of the database objects. Some caveats:
- `columns` will need an additional file to store the `MathesarColumn` class, I was thinking `classes.py`
- `records` has the `get_records` functions, which seem like they belong in `operations`, not `utils`, even though they don't alter data.
- Perhaps the `operations` and `utils` nomenclature could be improved?

I also think it might make sense to sub-divide `operations` into smaller files e.g.
- `operations/create.py`
- `operations/read.py`
- `operations/update.py`
- `operations/delete.py`
- `operations/merge.py` (for tables)
- `operations/split.py` (for table)

This would solve the problem of where "get" operations go for records, and we could save `utils.py` for anything that doesn't neatly fit into an operation (like reflecting tables).

mathemancer:
I think the approach makes sense overall.  I especially like the focus on splitting DB-altering from non-DB-altering.  Consider also reorganizing imports in each file to be one import per line (makes merging when imports are added/removed easier).

> * `operations.py` contains code that actually alters tables.
> * `utils.py` contains code that makes working with tables easier, but doesn't alter anything.

It's not as generalizable, but `ddl.py` would be clearer to me (or maybe even `data_definition.py`) instead of `operations.py`, and that also makes it really clear that this doesn't change data values, but rather table form.  I think `utils.py` makes sense as a catch-all for whatever isn't in a specific file.  In this case, that's anything that isn't DDL.  

> * `columns` will need an additional file to store the `MathesarColumn` class, I was thinking `classes.py`

Consider just using `mathesar_column.py` (or similar).  We may want to make one file per class (plus associated constants/functions if appropriate).

I think the fine-grained splitting makes sense for actually coding, but it'll make using the module cumbersome.  Consider bringing those up a level (by pulling them into the appropriate `__init__.py` file) to avoid that.  It could either be all the way up to the `tables` module, or maybe `tables.operations` (or `tables.ddl` if you use the earlier suggestion).

> * `records` has the `get_records` functions, which seem like they belong in `operations`, not `utils`, even though they don't alter data.

I like the finer subdivisions you listed for solving this; if you want to go coarser, consider combining them into 'reading' and 'writing'.  

> This would solve the problem of where "get" operations go for records, and we could save `utils.py` for anything that doesn't neatly fit into an operation (like reflecting tables).

Reflecting _is_ "get" for tables (at the `db` level).  I think it'd be neet if there were an appropriate synergy between the way things are organized so that the `get` for each object type (record, column, db, table, view, etc.) is in the same (ish) file for that type.  Same with the other CRUD operations.

kgodey:
@mathemancer Thanks for all the feedback! I'll proceed with my current path.

kgodey:
@mathemancer Since this is getting pretty big, I'm marking this ready for review. 

If it's not merged by the time I'm back to coding tomorrow, I'll continue pushing to it, otherwise I'll continue the reorganization in a new PR. This seems like a good place to stop since the `tables.py` file is handled and tests are passing.

mathesar/api/serializers/columns.py
```diff
@@ -6,7 +6,7 @@
 
 class InputValueField(serializers.CharField):
     """
-    Takes in an arbitrary value. Use to emulate our column and record creation and
+    Takes in an arbitrary value. Use to emulate our column creation and
```
eito-fis, line 9: The intention with the original comment was to point out that record creation endpoint took in arbitrary values (un-validated and un-processed `request.data`), and this field replicates that behavior but in a serializer. I'm realizing now that this isn't very clear - maybe we can just remove the second sentence?

eito-fis (approved):
Looks great - really liking the new structure! Did leave a small comment below.

mathemancer (approved):
I think the overall direction looks good, and makes sense for this project.

- tests are passing on my machine
- Checked the API to make sure things are generally working (they are)

mathesar/api/serializers/columns.py
mathemancer: I think we can change the comment at our leisure in a later PR; I'm going to go ahead and merge so we can base work off of this for now.

kgodey: I'm fixing this in the next PR.

PR #628, opened 2021-09-06
Implement `TIME` type in the backend

### LINKED ISSUE #426: Handle `TIME` data type in the backend
This issue is to ensure that Mathesar can handle the Postgres `TIME` data type.

- Users can use the API to change a column to `TIME`(with time zone) if it's possible to do so.
- Users can use the API to change a column to `TIME`(without time zone) if it's possible to do so.
- Users can set and change the `precision` of a given `TIME` column via a `type_options` field in the API.
- Automatic type inference during file import suggests `TIME` when it makes sense to do so. 

**Additional Context** 
- #250 (parent issue)

eito-fis:
Fixes #426

Implements the `TIME` type in the backend, with options for precision and timezone.

**Technical details**
Some oddities due to how sqlalchemy handles `TIME`. In particular, it will compile using the `timezone` and `precision` attributes, but won't take them in at initialization time. As a result, we have use two additional custom types, `TIME_WITH_TIMEZONE` and `TIME_WITHOUT_TIMEZONE`, that simply extend the init method to take `timezone` and `precision`.

Additional notes:
* We only use `TIME_WITH_TIMEZONE` and `TIME_WITHOUT_TIMEZONE` - just `TIME` is not supported. `TIME` has been removed from the type map to reflect this.
* Precision acts strangely - the `time` object the data is reflected as doesn't really support precision. It has `millisecond`, which is equivalent to a precision of 1, but no more. As a result, currently settings precision as anything valid but greater than 1 is the same as setting a precision of 1. (I believe this is why sqlalchemy doesn't support precision.)
* Inference to `TIME_WITH_TIMEZONE` is not supported as `TIME_WITH_TIMEZONE` is primarily intended to be used as support for legacy databases.[ See postgres documentation for details](https://www.postgresql.org/docs/13/datatype-datetime.html#DATATYPE-TIMEZONES).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


eito-fis:
I'm curious how much we care about what type the API returns. Intuitively, I would guess that consistently returning `TIME` with the appropriate type options would be the nicest, but if it would be less work to just return `TIME WITH TIMEZONE`. In fact, if we're alright with doing away with type options entirely, this would be much simpler implemented as just `TIME WITHOUT TIMEZONE` and `TIME WITH TIMEZONE`.

kgodey:
@eito-fis I'm in favor of going with the simpler implementation. Please update the type mapping files in `mathesar/database/` to include both time types in your PR. Thanks!

eito-fis:
@mathemancer Is there a reason we use `get_available_types` over `get_supported_types` in `test_alter_column_casts_data_gen`? (Or in general?) I'm running into issues since I've implemented TIME_WITH_TIME_ZONE and TIME_WITHOUT_TIME_ZONE as simple extensions of the original TIME type. This works fine for `get_supported_types` since I can insert the custom type into the type map, but breaks with `get_available_types` since it just returns the base `TIME` type.

Happy to change the implementation of the current TIME types if there is a better way to go about this.

mathemancer:
> @mathemancer Is there a reason we use `get_available_types` over `get_supported_types` in `test_alter_column_casts_data_gen`? (Or in general?) I'm running into issues since I've implemented TIME_WITH_TIME_ZONE and TIME_WITHOUT_TIME_ZONE as simple extensions of the original TIME type. This works fine for `get_supported_types` since I can insert the custom type into the type map, but breaks with `get_available_types` since it just returns the base `TIME` type.
> 
> Happy to change the implementation of the current TIME types if there is a better way to go about this.

Using separate types for `TIME WITH TIME ZONE` and `TIME WITHOUT TIME ZONE` is the correct move, good work there.  I suggest the `ischema_names` dict items for `time with time zone`, `time without time zone`, and `time` according to the actual DB types, (using the extensions you've already set up).  I.e., you should end up with the sub-dictionary:
```python
{
    "time with time zone": db.types.datetime.TIME_WITH_TIME_ZONE,
    "time without time zone": db.types.datetime.TIME_WITHOUT_TIME_ZONE,
    "time": db.types.datetime.TIME,
}
```

Reflection is going to be a pain with that setup, but it solves a number of other issues.  (See `db/types/__init__.py` for where to update the dict.)  For reflection to work better, it might make sense to use actual `UserDefinedType` implementations, with the `TIME` class as a mix-in.  Not sure, though, without some tinkering.

As for why those functions are used differently, the intent of `get_supported_alter_column_types` is precisely to separate the names we pass around for things from the arbitrary keys in `ischema_names`.  I'm fine with using that function more if it helps.  It might not be necessary if you add the proper types into `ischema_names`, though.

eito-fis:
This should be in an about functional place - the last question is about inference. In particular, both `TIME WITH TIME ZONE` and `TIME WITHOUT TIME ZONE` can parse strings with and without time zones, which means that whichever is tried first when casting a string is used. I see two possible resolutions here:
* We only ever infer one of the two. In this case, I assume it would be best to do `TIME WITH TIME ZONE` to ensure we don't lose information. (Although, if the user can adjust afterwards, then maybe it doesn't matter too much.)
* We update the SQL cast function for string -> `TIME WITHOUT TIME ZONE` to error if there is a time zone, and then infer to either type. This would mean that we couldn't go straight from string with timezone to `TIME WITHOUT TIME ZONE`, and instead would have to have an intermediate step converting to `TIME WITH TIME ZONE`, but maybe that's alright.

Any thoughts on which would be preferred?

kgodey:
@eito-fis I think it's fine to infer`TIME WITH TIME ZONE` for all times. I assume we'd convert times with existing timezones to the corresponding time in UTC before saving. Please make sure to add a test case for that.

eito-fis:
@mathemancer I added a test for the error and handled it with an extended DRF JSON renderer - should be working now. I can clear up the merge if this solution looks good.

One question I have is whether the Mathesar renderer should be a general default, instead of just being attached to the record viewset as it is now. I can't come up with any cases where another endpoint has to serialize custom types so I think its alright as is, but would like to make sure that assumption is correct.

kgodey:
@silentninja I added you as a reviewer here since you're working on `TIMESTAMP`.

silentninja:
> One question I have is whether the Mathesar renderer should be a general default, instead of just being attached to the record viewset as it is now. I can't come up with any cases where another endpoint has to serialize custom types so I think its alright as is, but would like to make sure that assumption is correct.

It would be better if custom renders are listed explicitly as it is meant for rendering data from the user database. If we make it a default It would unnecessarily end up interfering with Mathesar related data(like datafile model, display_options) if we do end up extending the renderer in the future

db/tests/types/operations/test_cast.py
```diff
@@ -36,6 +37,9 @@
 NUMERIC = PostgresType.NUMERIC.value.upper()
 REAL = PostgresType.REAL.value.upper()
 SMALLINT = PostgresType.SMALLINT.value.upper()
+DATE = PostgresType.DATE.value.upper()
```
silentninja, line 40: `Date` constant has already be initialized at [line 31](https://github.com/centerofci/mathesar/blob/7a5a8da293eed084ffd4bb1bc563a6a6aeb159a0/db/tests/types/operations/test_cast.py#L31)

silentninja (changes requested):
Looks good to me, creating individual time types seems to be much better and flexible.

dmos62 (commented):
Looks good. I don't know this code area as well as some of the others, so I'll refrain from giving explicit approval.

kgodey:
I moved the status back to `review`. I'd like to get this merged ASAP, so @mathemancer, could you please re-review and make any last changes that need to be made to merge it?

db/types/datetime.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy.dialects.postgresql.base import TIME as SA_TIME
+
+from db.types import base
+
+TIME_ZONE_DB_TYPE = base.PostgresType.TIME_WITH_TIME_ZONE.value
+WITHOUT_TIME_ZONE_DB_TYPE = base.PostgresType.TIME_WITHOUT_TIME_ZONE.value
+
+
+class TIME_WITHOUT_TIME_ZONE(SA_TIME):
```
silentninja, line 9: Converting the class to a string still returns `Time` rather than `TIME WITHOUT TIME ZONE`. The `str` representation method should be overridden to return the appropriate value

db/types/datetime.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, *args, precision=None, **kwargs):
+        # On reflection timezone is passed as a kwarg, so we need to make sure it isn't
+        # included in the TIME init call twice
+        if "timezone" in kwargs:
+            kwargs.pop("timezone")
+        super().__init__(*args, timezone=False, precision=precision, **kwargs)
+
+
+class TIME_WITH_TIME_ZONE(SA_TIME):
```
silentninja, line 18: Converting the class to a string still returns `Time` rather than `TIME WITH TIME ZONE`. The `str` representation method should be overridden to return the appropriate value

silentninja (changes requested):

mathemancer:
> One question I have is whether the Mathesar renderer should be a general default, instead of just being attached to the record viewset as it is now. I can't come up with any cases where another endpoint has to serialize custom types so I think its alright as is, but would like to make sure that assumption is correct.

I think that's a reasonable place for it.

mathemancer (approved):
Okay, I've made a couple of minor changes as requested by @silentninja , and everything else looks good to me.

silentninja:
> Okay, I've made a couple of minor changes as requested by @silentninja , and everything else looks good to me.

The `__name__` constant is a class property, so it should be `self.__class__.__name__`.

silentninja:
Thank @mathemancer

silentninja (approved):

Issue #417, opened 2021-07-17
Backend work for Mathesar money custom type 


kgodey:
This issue is to create a custom `Money` data type to handle monetary data.  At the DB level, this will refer to a number of types such as `MATHESAR_TYPES.MONEY_USD`, `MATHESAR_TYPES.MONEY_JPY`, etc.  Each underlying type should have a name of the form `MATHESAR_TYPES.MONEY_<currency code>`, where `currency code` is one of the 3-character currency codes defined in [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html).  

As part of this issue, we need to:
- Create a `Money` friendly type name in the service
- Create a number of `MATHESAR_TYPES.MONEY_<currency code>` types in the database that store currency amounts.  The amount should be a `NUMERIC` type.
  - The types should be case-insensitive at the DB layer and uppercase in the API
- We'll create the types automatically, putting things together using some package like https://github.com/arshadkazmi42/currency-symbols or https://github.com/python-babel/babel to provide currency symbol info.
- Users should be able to use the API to change a column to `MATHESAR_TYPES.MONEY_<currency code>` if it's possible to do so.
- Automatic type inference during file import suggests `MATHESAR_TYPES.MONEY_<currency code>` when it makes sense to do so.

**Additional Context** 
- Blocked until #409 is complete, to serve as a model for DB case insensitivity
- See PR #368 for code that calculates which types are allowed for a column.

mathemancer:
See my comment on #409 .  We should create types in the `mathesar_types` schema unless there's a good reason to do otherwise.

kgodey:
@mathemancer Updated the issue. I made it uppercase to match other types, do you have any issues with that?

mathemancer:
I think it should be case-insensitive at the DB layer, and displayed as upper case in the API.  This will be the most consistent with the way the default type names work.

If we go that route, we'll have to either:
- enforce that setup (i.e., case-insensitive naming at the DB layer) for eventual user-created types, or
- have a separate way of handling user-created type names when that's included.

User-defined types are a ways off, though,  so I think we should just enforce case-insensitive naming of types (and the `mathesar_types` schema) on ourselves for now.  Being able to assume case-insensitivity for type names would actually reduce some complexity I had to introduce around quoting names.

kgodey:
@mathemancer Okay, thanks, makes sense. I'll update the existing custom types issues.

mathemancer:
[...]

PR #647, opened 2021-09-08
Add custom `MATHESAR_TYPES.MONEY` type

mathemancer:
Related to #417 

This adds a basic composite type in the DB called `MATHESAR_TYPES.MONEY`.  The composite type is:
```
( value NUMERIC, currency CHAR(3) )
```

**Technical details**

At the API (and Python) level, we use JSON to represent an entry, for example:
```json
{
    "value": 134.56,
    "currency": "USD"
}
```

Note that in the DB layer, `JSON` (and `JSONB`) numbers are in fact `NUMERIC` types, so there shouldn't be any loss of precision in the translation at the DB layer.  All conversion between JSON and the actual composite type is happening via SQL functions in the DB.

The intent is to use the standard ISO 3-character codes to represent real currencies.  If a user wants to use a custom currency, they can either create or replace these codes.

### TODO

- inference is yet to be done
- `server_default` setting doesn't work very smoothly yet.  An issue #648 is in the works to resolve this.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (approved):
@mathemancer Looks good, the implementation is straightforward and easy to read.

I attempted to cast a `VARCHAR` type (the data was numeric) to money via the column `PATCH` API locally and received the following error:
```

[
    "This type cast is not implemented"
]
```

It doesn't seem like we do any validating of currency yet, I'd also add that to your TODOs or create a separate issue for it.

I'm approving to unblock merge. If you want to make changes in a future PR, feel free to merge and do that.

mathemancer:
> I attempted to cast a `VARCHAR` type (the data was numeric) to money via the column `PATCH` API locally and received the following error:
> 
> ```
> 
> [
>     "This type cast is not implemented"
> ]
> ```

I can't replicate this (i.e., when I cast a `VARCHAR` to `MATHESAR_TYPES.MONEY`, via `PATCH` in the endpoint, it works as expected.  Would you try again with a fresh DB volume?  (I just double-checked by knocking mine out and starting fresh).

kgodey:
@mathemancer It works, sorry, I forgot to install the type. 🤦 

We should be able to set the currency using `type_options` in the future as well, please also add that to your TODO or create a new issue. Thanks!

mathemancer:
> We should be able to set the currency using `type_options` in the future as well, please also add that to your TODO or create a new issue. Thanks!

I don't quite understand.  Do you mean setting a default value for the currency part of the composite, or something else?  It's not possible (using standard `DEFAULT` setting in the DB) to set the default for one component of a composite type.  If  we want, we could have some other column types (`USD`, `EUR`, etc.) that are customized to those currencies, implemented as `DOMAIN`s.

mathemancer:
> It doesn't seem like we do any validating of currency yet, I'd also add that to your TODOs or create a separate issue for it.

Given that users are going to be able to define their own currencies by inserting into the column, I'm not sure what validation would entail.  I had given thought to a "validated money" type (associated with the implementation of inference), or even a few specific money types (`USD`, `EUR`, `JPY`, etc.).  We could give those as optional money types.  As I think about it, having the ability to have a `USD` column (or a whole column for any money type from some small list) would probably be useful.  The DB implementation would just be a `DOMAIN` over a `NUMERIC` to give a name so the Front End knows what type it's seeing.

kgodey:
@mathemancer I'm referring to the following requirements (quoted from #417's description):

> Store a list of currencies and decimal precision for users to select from when creating their column. We could use something like https://pypi.org/project/money/

This is to make sure that whatever is inserted into the `currency` attribute is a money type we know how to display on the frontend. The user should be able to define new currency types, but I think we should figure out how to let them do that later. They'll also need to define formatting rules and precision for their currency.

I'm not sure what the best way to get a list of currencies is, but my inclination is to use a Python package (https://github.com/kalaspuff/stockholm seems newer than the one in the issue description) and sync the currencies it provides to the DB on install, that way we can just update the package when we need to and we can also use it in Python for validation etc. I'm open to other ideas, though.

> Users should be able to change the currency and decimal precision of a column of type MATHESAR_TYPES.MONEY via a `type_options` field in the API.

We decided against allowing the user to change the decimal precision in favor of the maximum (this should be documented in the issue, btw, please update the description) but they should still be able to change the currency. If the user wanted to alter their `MATHESAR_TYPES.MONEY` type column from a currency of `USD` to a currency of `GBP`, wouldn't we want them to use `type_options` to do so for consistency?

kgodey:
Also on a completely separate note, I think numeric types should be valid target types for money, we'd just use the numeric component.

mathemancer:
> Also on a completely separate note, I think numeric types should be valid target types for money, we'd just use the numeric component to figure out how to convert the data.

Agreed.

PR #654, opened 2021-09-13
Tabledata store reorganization

### LINKED ISSUE #651: Refactor frontend stores for table-data
## Purpose of this refactor:
1. Our tableData store handles colums, records and meta information needed for parameters and display purposes. It also includes the type definitions for all of them. This makes it cluttered, and there is an overlap of different concerns.
A better approach would be to split it into separate stores, and have a parent store which maintains the rest.
2. Currently, logic of certain requirements such passing parameters to table record request, refetching table data when params change etc., are done on the component. This can be side stepped and done entirely by observing store values. This gives a cleaner view/data model.
3. We need to simplify certain display specific stores such as column position map, to improve readability.
4. General improvements, separate out cells from header and row as different components.
5. Update svelte, vite and related packages.

pavish:
Related to #651 

**Technical details**

This PR:
- Updates svelte and vite to latest versions
- Handles segregation needed for views
- Creates dedicated stores for columns, records, meta, display
- Handles records re-request directly in store bypassing component layer
- Handles column position recalc directly in store bypassing component layer
- Handles record deletion directly from store
- Uses pagination instead of infinite scroll
- Handles grouping on store response instead of component
- Fixes errors reported by svelte-typecheck
- Separates cells from header and row

Breaking changes:
- URL sync is broken for filtering, sorting and grouping. This will be handled incrementally, as it is not a blocker to other PRs.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (approved):
Things seem to work on my machine, and the tests are passing.

### Optional suggestion:

While I think the new organization makes sense on a file level, I find it a little confusing at the directory level.  The `table-data` directory contains files pertaining to not just the data for a table (i.e., the records), but also the metadata, and also some display concerns.  I suggest either renaming that directory, or even further subdividing into three directories roughly along the lines:
- actual data storage
- metadata storage
- display variables storage.

However, I defer to your experience here, @pavish , since I have little knowledge in the idiomatic use of the tools in question.

pavish:
@mathemancer I agree that it would make more sense to come up with a better name for the directory. Splitting it up into smaller directories would not be ideal at this point, since there is only one file per type of data.

I will handle this in the upcoming PR: https://github.com/centerofci/mathesar/pull/671

pavish:
@kgodey @zackkrida @dhruvkb @mr-gabe49  Please feel free to add review comments. I'll go ahead and merge this PR since it's blocking other PRs, and handle the comments in a separate PR.

mathesar_ui/src/stores/table-data/display.ts
```diff
@@ -0,0 +1,88 @@
...
+export type ColumnPositionMap = Map<string, ColumnPosition>;
+
+export const ROW_CONTROL_COLUMN_WIDTH = 70;
+export const GROUP_MARGIN_LEFT = 30;
+export const DEFAULT_ROW_RIGHT_PADDING = 100;
+export const DEFAULT_COLUMN_WIDTH = 160;
+
+function recalculateColumnPositions(columnPositionMap: ColumnPositionMap, columns: TableColumn[]) {
+  let left = ROW_CONTROL_COLUMN_WIDTH;
+  const newColumnPositionMap: ColumnPositionMap = new Map(columnPositionMap);
```
mr-gabe49, line 20: Hey @pavish, fabulous work here! Quick question: What's the reasoning for passing ` columnPositionMap`  inside of `new  Map()` ? It prevents `columnPositionMap` from updating correctly in line 76-78: ```
this.columnPositionMap.update(
        (map) => recalculateColumnPositions(map, columnData.data),
      ); 
``` when calling `columns.fetch() ` after a column is deleted.

Comment (pavish) [reply mathesar_ui/src/stores/table-data/display.ts:20]: @mr-gabe49 When a column has been previously resized, the columnPositionMap would contain a different value for width instead of the default column width. We create a new map using the existing map to retain it.

Good catch, I hadn't considered the effect of deletion.

In your PR for column deletion, we could update lines 30-32 to:
```
  const newColumnPositionMap: ColumnPositionMap = new Map();
  columns.forEach((column) => {
    const columnWidth = columnPositionMap.get(column.name)?.width;
```
Since we have access to the old map, we could use it to get the previous width. This would handle position recalc for both resize and delete.

PR #671, opened 2021-09-19
Adding records

### LINKED ISSUE #197: Implement adding new records
**Problem**

The user should be able to add new records to a table.

**Proposed solution**

We should support this in the frontend. See the design issue. 

**Additional context**

Record addition will use `POST` requests to `/api/v0/records/`

Blocked by design work in #193.

pavish:
Fixes #197 

**Task list**
- [x] Add record action pane button
- [x] Add record placeholder row
- [ ] Send request on adding record
- [ ] Change update behaviour to send post request if record is not created yet
- [ ] Show indication that record is newly added
- [ ] Show text to indicate record may change position

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/sections/table-view/actions-pane/ActionsPane.svelte
+++ b/mathesar_ui/src/sections/table-view/actions-pane/ActionsPane.svelte
@@ -7,20 +7,30 @@
     faTrashAlt,
     faSync,
     faExclamationTriangle,
+    faPlus,
   } from '@fortawesome/free-solid-svg-icons';
   import { States } from '@mathesar/utils/api';
   import { Button, Icon } from '@mathesar-components';
-  import type { TabularDataStore, TabularData } from '@mathesar/stores/table-data/types';
+  import type {
+    TabularDataStore,
+    TabularData,
+    Records,
+    Columns,
+    Meta,
+  } from '@mathesar/stores/table-data/types';
 
   const dispatch = createEventDispatcher();
 
   const tabularData = getContext<TabularDataStore>('tabularData');
+
+  let records: Records;
+  let columns: Columns;
   $: ({
     columns, records, meta,
   } = $tabularData as TabularData);
   $: ({
     filter, sort, group, selectedRecords, combinedModificationState,
-  } = meta as TabularData['meta']);
+  } = meta as Meta);
 
   $: isLoading = $columns.state === States.Loading
     || $records.state === States.Loading;
@@ -32,12 +42,16 @@
   }
 
   function deleteRecords() {
-    void (records as TabularData['records']).deleteSelected();
+    void records.deleteSelected();
+  }
+
+  function addRecord() {
+    records.addRecord();
   }
 
   function refresh() {
-    void (columns as TabularData['columns']).fetch();
-    void (records as TabularData['records']).fetch();
+    void columns.fetch();
+    void records.fetch();
   }
 </script>
 
@@ -69,6 +83,15 @@
       {#if $group?.size > 0}
         ({$group?.size})
       {/if}
+    </span>
+  </Button>
+
+  <div class="divider"/>
+
+  <Button size="small" on:click={addRecord}>
+    <Icon data={faPlus}/>
+    <span>
+      Record
     </span>
   </Button>
 
--- a/mathesar_ui/src/sections/table-view/header/Header.svelte
+++ b/mathesar_ui/src/sections/table-view/header/Header.svelte
@@ -21,7 +21,7 @@
   } = $tabularData as TabularData);
   $: ({ horizontalScrollOffset, rowWidth, columnPositionMap } = display as TabularData['display']);
 
-  $: paddingLeft = $records.groupData ? GROUP_MARGIN_LEFT : 0;
+  $: paddingLeft = $records.groupCounts ? GROUP_MARGIN_LEFT : 0;
 
   let headerRef: HTMLElement;
 
--- a/mathesar_ui/src/sections/table-view/row/GroupHeader.svelte
+++ b/mathesar_ui/src/sections/table-view/row/GroupHeader.svelte
@@ -4,36 +4,38 @@
   } from '@mathesar/stores/table-data';
   import type {
     TableRecord,
-    GroupData,
+    GroupCount,
   } from '@mathesar/stores/table-data/types';
 
   export let row: TableRecord;
-  export let groupData: GroupData;
+  export let groupCounts: GroupCount;
+  export let groupColumns: string[];
 
-  function getCount(_groupData: GroupData, _row: TableRecord): number {
-    const columns = _row?.__groupInfo?.columns;
-    if (columns) {
-      let group = _groupData;
-      const groupLength = columns.length;
+  function getCount(_groupCounts: GroupCount, _row: TableRecord): number {
+    if (groupColumns) {
+      let group = _groupCounts;
+      const groupLength = groupColumns.length;
       for (let i = 0; i < groupLength - 1; i += 1) {
-        const value = _row?.__groupInfo.values[columns[i]] as string;
+        const value = _row?.__groupValues[groupColumns[i]] as string;
         if (!group?.[value]) {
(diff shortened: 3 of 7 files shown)


ghislaineguerin (commented):
@pavish Thanks for the video, a couple of comments:
- The added row element at the top of the table should allow the user to start typing on any of the empty cells, and only then a new row is added. The implementation replaces that with a large horizontal toggle. I can specify the behavior of this component better. 
- The warning for row repositioning should always be visible and part of a frame around the newly added rows.
- I agree with all of your other points re: UX definitions that are missing

pavish:
@ghislaineguerin I have a few questions. 

> The added row element at the top of the table should allow the user to start typing on any of the empty cells, and only then a new row is added. The implementation replaces that with a large horizontal toggle. I can specify the behavior of this component better.

1. How will the user add a new empty row with null values for all columns except primary key? In our example, only the primary key is mandatory and the user cannot edit edit. This is a valid state for the row. If we only create the row if user edits it, how do we create such a row?
2. If the user types on any of the empty cells, when do we show another new empty placeholder row? If it pushes the rows down while user is typing, would it not affect user flow?

> The warning for row repositioning should always be visible and part of a frame around the newly added rows.

1. This would hide other rows below the newly added rows if it's always visible, which would not be a good experience for the user. 
2. If the last new row has an error, how will that be shown alongside this message?

There are a lot of such scenarios that are not well defined and they all occur on specific cases, some of which we may miss. I think the best approach would be to handle as much as possible and keep the implementation and design as simple as possible, so that we can modify it whenever we encounter a scenario we have missed.

ghislaineguerin:
> 1. How will the user add a new empty row with null values for all columns except primary key? In our example, only the primary key is mandatory and the user cannot edit it. This is a valid state for the row. If we only create the row if user edits a cell, how do we create such a row?

This wouldn't change, please see the video below.

> 2. If the user types on any of the empty cells, when do we show another new empty placeholder row? If it pushes the rows down while user is typing, would it not affect user flow?

Please see the following video for an example of what I'm referring to: https://www.loom.com/share/4f416d06ca0f4c85a7df0ca158383baf

I think this allows users to have a more natural flow, where they can just start defining any of the fields in place and not lose context. We can always improve the transition so that the user can understand what's happening.

pavish:
> How will the user add a new empty row with null values for all columns except primary key? In our example, only the primary key is mandatory and the uer cannot edit it. This is a valid state for the row. If we only create the row if user edits a cell, how do we create such a row?

> This wouldn't change, please see the video below.

The video shows that a new row is added only when the user starts typing, which gives a value to a cell. Assume the user wants a row with null values for all columns, how would they do it?

> I think this allows users to have a more natural flow, where they can just start defining any of the fields in place and not lose context. We can always improve the transition so that the user can understand what's happening.

In the video, as the user starts typing the row shifts down, wouldn't that be a jarring effect to the user? If this was at the bottom of all records, and a user starts typing and new empty row appears below that row, then it does not intrude the user. But this is at the top of all the records. If the user is looking at a different row for reference while typing, they might lost context if the row suddenly moves down.

pavish:
@ghislaineguerin I have all these questions because this is relatively harder to implement compared to the approach that's present in the video I've mentioned in this PR. I am trying to understand the thought process and the UX benefit of this approach over the other, before I put in the effort to change it.

kgodey:
@pavish A couple of issues:

(1) It looks like `MATHESAR_TYPES.MONEY` columns aren't being handled correctly, although that can fixed in the data types milestone.
Mathesar_libraries

(2) if you edit a value, the whole row goes green. Shouldn't only newly added columns be green?

(3) if you edit a value and there's an error, the new value still shows up in the row. the value should go back to what it is on the server (or there should be some sort of clear button). We can discuss it when we discuss the general UX.

pavish:
@kgodey 

> It looks like MATHESAR_TYPES.MONEY columns aren't being handled correctly, although that can fixed in the data types milestone.

Type specific handling is not yet done for any of the columns, for both adding & editing. We will handle each type in the respective data types issue. For eg., date/time fields will require a date selector, email field will need a email validation etc.,

> if you edit a value, the whole row goes green. Shouldn't only newly added columns be green?

1. Showing state of each column within a row would lead to storing n x m states (n - number of rows, n - number of columns). As of now, only n states are stored. This was an optimization done to avoid storing a lot of content on the frontend.

2. If a row is edited, currently the frontend sends a patch request with the contents of the whole row. We could change it send the contents cell by cell, as multiple requests.
  * This would allow the user to update a cell, while another does not get updated due to an error. But it seemed more user friendly to show the user that the whole row is not saved as they expected, to grab their attention.
  * If multiple users are editing the same row, and we allow saving the row cell by cell, neither of the users would be viewing the proper state of the row in the db. If we update the row as a whole, the user that last edited the row would be the source of truth.

>  if you edit a value and there's an error, the new value still shows up in the row. the value should go back to what it is on the server (or there should be some sort of clear button). We can discuss it when we discuss the general UX.

If the value goes back to what it was, how would we show an error? If we change value to previous state, the value is no longer cause of an error. Doing that might also affect user actions. eg., User may have started typing but not yet done typing. We debounce and save with an interval of 500ms, in which case an error might revert the user changes.

We could consider a clear button.

pavish:
@kgodey @ghislaineguerin As discussed during the call,

For unblocking the PR:
- [ ] New records should be moved to the bottom of the list instead of the top.
- [ ] Records need be saved cell by cell instead of the complete row at once.
- [ ] New record reposition warning needs to be always shown.
- [ ] New record placeholder should be editable cell by cell, instead of a static row.

Design specifications required:
* Behaviour when records are grouped.
* Clear/Revert action, to revert a erroneous cell value.
* Behaviour w.r.t. pagination.
* Behaviour w.r.t. sorting and filtering.
* Do we persist records that are added but not saved due to an error, when user moves across pages/sorts/filters/groups?
* Alerts when user moves pages, regarding clearing of unsaved records.

kgodey:
@pavish Can you open a separate issue for the designs required?

kgodey:
I played around with this locally and it looks good. Thanks for the video as well @pavish. We can open new issues if any bugs arise.

I'll leave it to @seancolsen to review and merge.

pavish:
@seancolsen  Regarding,

> * Use a better name for the button instead of "Record".

This was followed based on our UX designs. I think we can add this to https://github.com/centerofci/mathesar/issues/722, so that @ghislaineguerin can decide on this.

> * Fix bug when deleting a record near the end of the table while adding a new record

I am aware of this bug, I've mentioned it on the follow up design issue (https://github.com/centerofci/mathesar/issues/722), and also on the latest video. I've not handled it because it requires a bit of UX rethought.

I do agree we need to handle duplicates, however delete is an operation which reloads the table in-place without additional user intervention. This is different from other reloads like moving across pages, sorting etc., which require explicit user action. So, we cannot clear up the newly added records from the view during delete. We will need to think on the best way to handle this.

@ghislaineguerin Let us know if require further input from us. I've already added it as a point on the design issue, I'll also add @seancolsen's comment in the issue.

> Initiate adding when clicking on the "+" icon at the bottom of the table

Good point. I'll update the PR to handle this.

seancolsen:
This may be splitting hairs, but just to clarify...

I think "Row" and "Record" are both acceptable words to use within the button. The change I'm requesting is that we _add a verb_ like "New", "Add", or "Insert" before the noun. Any of those three verbs seem good to me, but **"New"** seems best because it's consistent with the buttons we already have for "New table" and "New Schema". And, in choosing between "Row" and "Record", I like "Row" because it's more concise. That's how I arrived at the suggestion of **"New row"**.

mathemancer:
> I think we probably want different words for tables and views. I like using "Record" for tables because the user can learn that tables contain a single record whereas views might have multiple records in a single row.

This is a great point.  Seconded.

Issue #668, opened 2021-09-15
Deleting all columns in a table results in a NoSuchTableError


mr-gabe49:
## Description

- Deleting all columns in a table results in a NoSuchTableError

## Expected behavior

- There are two possible scenarios I can think of: 
     1.  The table is deleted entirely
     2. The table remains empty

## To Reproduce

- Delete all columns from a table through the API 

## Environment
 - OS: (_eg._ macOS 10.14.6; Fedora 32)
 - Browser: (_eg._ Safari; Firefox)
 - Browser Version: (_eg._ 13; 73)
 - Other info:

## Full Traceback

```
Environment:
Request Method: GET
Request URL: http://localhost:8000/mathesar_tables/8/

Django Version: 3.1.12
Python Version: 3.9.7
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'django_filters',
 'django_property_filter',
 'mathesar']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/code/mathesar/views.py", line 97, in schema_home
    'common_data': get_common_data(request, database, schema)
  File "/code/mathesar/views.py", line 44, in get_common_data
    'tables': get_table_list(request, schema)
  File "/code/mathesar/views.py", line 35, in get_table_list
    return table_serializer.data
  File "/usr/local/lib/python3.9/site-packages/rest_framework/serializers.py", line 745, in data
    ret = super().data
  File "/usr/local/lib/python3.9/site-packages/rest_framework/serializers.py", line 246, in data
    self._data = self.to_representation(self.instance)
[...]

PR #680, opened 2021-09-26
addressing issue #668

bohemia420:
sqlalchemy engine's `reflection` was rather found absurd and thought of to be attempted - sqlalchemy only sets True that table's found while looping through the columns, which is redundant at first, and a False Negative, for tables with 0 columns (a scenario we touch when we delete all its columns).

While penning down a separately handled reflection method is great, it might rather be replicating few scores of similar code for one small handling. Thence, either we publish our own modified pypis / choose to install customized wheels/eggss i.e releases of various py packages, 

or keep a `tweaked-site-packaged` wherein we bring in select files/snippets to be replaced/handled separately. Whichever appeals more on being neat. 

p.s: `tweaked-site-packages/sqlalchemy/engine/reflection.py` diff is only towards:

```
       found_table = True #False
        cols_by_orig_name = {}
        for col_d in self.get_columns(
            table_name, schema, **table.dialect_kwargs
        ):
            found_table = True

            self._reflect_column(
                table,
                col_d,
                include_columns,
                exclude_columns,
                cols_by_orig_name,
            )

        if not found_table:
            raise exc.NoSuchTableError(table.name)
```

wherein `found_table` can be removed out too (finding it redundant). It only has just 3 weird occurences, all in this block here.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,6 @@ services:
     volumes:
       - ./.volumes:/var/lib/postgresql/data
   service:
-    privileged: true
     build: .
     environment:
       - MODE=${MODE-PRODUCTION}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,11 +12,13 @@ services:
     volumes:
       - ./.volumes:/var/lib/postgresql/data
   service:
+    privileged: true
     build: .
     environment:
       - MODE=${MODE-PRODUCTION}
     command: dockerize -wait tcp://db:5432 -timeout 30s ./run.sh
     volumes:
+      - ./tweaked-site-packages/sqlalchemy/engine/reflection.py:/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/reflection.py
       - .:/code/
       - ui_node_modules:/code/mathesar_ui/node_modules/
     ports:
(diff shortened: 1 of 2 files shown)


kgodey (changes requested):
I don't think we should maintain a separate version of a single file from SQLAlchemy in our repository, keeping it in sync will be a pain. I recommend opening a bug report instead; they may fix it (or provide a good reason for why it shouldn't be fixed).

bohemia420:
@kgodey makes sense, in fact was as well, considering writing to SQLAlchemy instead. They should ideally fix it, `found_table` looks all very redundant, otherwise.

kgodey:
@bohemia420 Please link the issue here once you open it so that we can keep track of it.

bohemia420:
Sure @kgodey , have done that, the issue now stands raised to SQLAlchemy [here](https://github.com/sqlalchemy/sqlalchemy/issues/7117) and have also proposed a PR to help this issue, [here](https://github.com/sqlalchemy/sqlalchemy/issues). My apologies, certainly this issue couldn't have been helped with diverging from origin SQLAlchemy to our own maintained and substituting select/cherrypicked files utterly dangerous - have poured in my thoughts with end-to-end analysis/troubleshooting [here](https://github.com/centerofci/mathesar/issues/668#issuecomment-930894085) #668

kgodey:
Thanks @bohemia420! I'll close this PR since the issue will hopefully be fixed upstream.

Issue #682, opened 2021-09-27
Some `server_default` constants are not handled


mathemancer:
## Description

Currently, we should be able to handle any constant `server_default` setting.  This is not happening, since any cast applied to a server default results in our `db.columns.operations.select.get_column_default` function deciding that the default is dynamic (and that it can't be handled).  In particular, there are entire types where the `server_default` value is always stored as a `CAST` of a different type.

## Expected behavior

We should be able to set and get `server_defaults` whenever they're constant.

## To Reproduce

Set a `server_default` for a column of type `CHAR`, watch it disappear.

## Additional context

Handling server-level defaults is quite painful.  The options are to either
- retrieve a string of the SQL that describes the default, or
- retrieve a `pg_node_tree` (an internal representation of a partially-compiled SQL statement; this is how PostgreSQL actually stores the default internally)

One option (that I currently prefer) is to retrieve the SQL string (note this is actually a decompiled `pg_node_tree`), break it into an Abstract Syntax Tree _in python_ using https://github.com/lelit/pglast, use that to determine whether the default is constant or dynamic, determine its value, and use that value for the representation in the API.

The reason to do this in python rather than using the `pg_node_tree` directly from the DB is to get the AST in a more python-friendly form.

Issue #681, opened 2021-09-27
Unknown server_default settings silently deleted


mathemancer:
## Description

When the column-type-altering function doesn't understand a `server_default` value for a column (the DB-level default setting), it silently removes / deletes that default when altering the column type.  This has the potential to lose metadata about a column that we'd be unable to recover.  Further, there would be no warning to the user in this case.

## Expected behavior

If we try to alter a column's type, and cannot deal with the `server_default` setting, we should refuse, or at least throw a warning where the user can see it.

## To Reproduce

Change the type of any column with a dynamic default (e.g., change a `DATE` column with default `NOW()` to `VARCHAR`), and note that the default has disappeared on the DB (this will be invisible in the API as noted in the description above).

## Environment
 - OS: (_eg._ macOS 10.14.6; Fedora 32)
 - Browser: (_eg._ Safari; Firefox)
 - Browser Version: (_eg._ 13; 73)
 - Other info:

PR #686, opened 2021-09-29
Text types

### LINKED ISSUE #404: Handle `VARCHAR` data type in the backend.
This issue is to ensure that Mathesar can handle the Postgres `VARCHAR` data type.

- Users can use the API to change a column to `VARCHAR` if it's possible to do so.
- Users can set and change the `n` (length) of a given `VARCHAR` column via a `type_options` field in the API.
- Automatic type inference during file import suggests `VARCHAR` when it makes sense to do so. 

**Additional Context** 
- Implementation for this is partially complete (specifically the first and third items).

### LINKED ISSUE #405: Handle `TEXT` data type in the backend.
This issue is to ensure that Mathesar can handle the Postgres `TEXT` data type.

- Users can use the API to change a column to `TEXT` if it's possible to do so.

**Additional Context** 
- Nothing is needed for type inference since `VARCHAR` covers the same use: #404

### LINKED ISSUE #407: Handle `CHAR` data type in the backend.
This issue is to ensure that Mathesar can handle the Postgres `CHAR` data type.

- Users can use the API to change a column to `CHAR` if it's possible to do so.
- Users can set and change the `n` (length) of a given `CHAR` column via a `type_options` field in the API.
**Additional Context** 
- No changes are needed for type inference; I don't think we should be suggesting a restrictive type like `CHAR`.

mathemancer:
Fixes #404 
Fixes #405 
Fixes #407 

This PR adds support for `TEXT`, `CHAR`, AND `VARCHAR` types in the back end.  One can now change columns to these types and restrict the length for `CHAR` and `VARCHAR` types using a `length` type option.

**Technical details**

  The server default handling is not quite working for `CHAR` type (and sometimes `VARCHAR`), and will be fixed in #681 and #682 . 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/base.py
+++ b/db/columns/base.py
@@ -144,6 +144,7 @@ class MathesarColumn(Column):
     @property
     def type_options(self):
         full_type_options = {
+            "length": getattr(self.type, "length", None),
             "precision": getattr(self.type, "precision", None),
             "scale": getattr(self.type, "scale", None),
         }
--- a/db/columns/operations/alter.py
+++ b/db/columns/operations/alter.py
@@ -105,8 +105,12 @@ def retype_column(
     new_type = new_type if new_type is not None else column_db_type
     column_type_options = get_type_options(column)
 
-    if (new_type == column_db_type) and _check_type_option_equivalence(type_options, column_type_options):
+    if (
+            (new_type.lower() == column_db_type.lower())
+            and _check_type_option_equivalence(type_options, column_type_options)
+    ):
         return
+
     try:
         alter_column_type(
             table,
--- a/db/columns/operations/alter.py
+++ b/db/columns/operations/alter.py
@@ -26,9 +26,17 @@ def alter_column(engine, table_oid, column_index, column_data):
 
     with engine.begin() as conn:
         if TYPE_KEY in column_data:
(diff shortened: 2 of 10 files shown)


kgodey (approved):
@mathemancer This looks good. 

It seems like I can't change the type options of an existing column without also passing in the `type` parameter. (A request of `{"type_options": {"length": 100}}` to an existing `VARCHAR` column without a length specified doesn't work. This can be addressed in a separate issue if desired.

mathemancer:
> @mathemancer This looks good.
> 
> It seems like I can't change the type options of an existing column without also passing in the `type` parameter. (A request of `{"type_options": {"length": 100}}` to an existing `VARCHAR` column without a length specified doesn't work. This can be addressed in a separate issue if desired.

This behavior is because the `alter_column` function was testing for the `type` key in the passed JSON to decide whether to do _anything_ involving types.  I think you're correct that the behavior of just being able to change the option without also passing the type makes sense, it's now implemented in this PR since it was quick.

PR #690, opened 2021-10-04
Server default setting and getting improvements

### LINKED ISSUE #681: Unknown server_default settings silently deleted

### LINKED ISSUE #682: Some `server_default` constants are not handled

mathemancer:
Fixes #682 
Fixes #681 

This changes the way we handle server defaults so that we
- correctly handle more of the constant values, and 
- avoid silently passing over dynamic values, instead raising a warning and returning the default-defining string when reading.
- refuse to alter a column type if it has a dynamic default value, since we can't currently do that properly.

**Technical details**

We actually don't attempt to specifically identify constant default settings anymore.  Rather, we identify known dynamic default settings which we can't properly handle, and treat them differently.  Unfortunately, the identification of these dynamic defaults depends on the node types in the Abstract Syntax Tree given for the expression using the `pglast` package.  The only good documentation of what those node types are is in a pair of C header files that the `pglast` project uses:
- https://github.com/pganalyze/libpg_query/blob/13-latest/src/postgres/include/nodes/parsenodes.h , and
- https://github.com/pganalyze/libpg_query/blob/13-latest/src/postgres/include/nodes/primnodes.h

I was only able to find two tags that are guaranteed to appear in any dynamic default setting for a column.  It's possible more exist.  These would need to be added to the `db.columns.operations.select._is_default_expr_dynamic` function when found.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/exceptions.py
+++ b/db/columns/exceptions.py
@@ -12,3 +12,7 @@ class InvalidTypeOptionError(Exception):
 
 class DagCycleError(Exception):
     pass
+
+
+class DynamicDefaultWarning(Warning):
+    pass
--- a/db/columns/operations/select.py
+++ b/db/columns/operations/select.py
@@ -3,6 +3,7 @@ import warnings
 from pglast import Node, parse_sql
 from sqlalchemy import Table, MetaData, and_, select, text, func
 
+from db.columns.exceptions import DynamicDefaultWarning
 from db.tables.operations.select import reflect_table_from_oid
 from db.utils import execute_statement
 
@@ -39,7 +40,10 @@ def get_column_default(table_oid, column_index, engine, connection_to_use=None):
     if column.server_default is None:
         return None
     elif _is_default_expr_dynamic(column.server_default):
-        return None
+        warnings.warn(
+            "Dynamic column defaults are not implemented", DynamicDefaultWarning
+        )
+        return str(column.server_default.arg)
 
     default_textual_sql = str(column.server_default.arg)
     # Defaults are stored as text with SQL casts appended
--- a/db/columns/operations/select.py
+++ b/db/columns/operations/select.py
@@ -45,13 +45,13 @@ def get_column_default_dict(table_oid, column_index, engine, connection_to_use=N
         default_dict = {
             "server_default": column.server_default,
             "is_dynamic": _is_default_expr_dynamic(column.server_default),
-            "default_textual_sql": str(column.server_default.arg)
+            "sql_text": str(column.server_default.arg)
         }
     else:
         return
     if default_dict.get("is_dynamic"):
         warnings.warn(
-            "Dynamic column defaults are not implemented", DynamicDefaultWarning
+            "Dynamic column defaults are read only", DynamicDefaultWarning
         )
     else:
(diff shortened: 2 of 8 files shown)


eito-fis (commented):
Overall looks good to me! One question - do we have a test for the behavior in  #681?

kgodey (approved):
@mathemancer This looks good to me. Could you put the technical details you put in the PR somewhere in a code comment or README file so that we have it with the repo?

I'll leave it to you to merge after you've added the documentation and addressed @eito-fis's comment about potentially adding a test.

mathemancer:
> Overall looks good to me! One question - do we have a test for the behavior in #681?

Actually, the tests I put for that don't go up to the API layer; good catch.  I'll actually change the behavior a bit so the API layer shows the actual SQL string for the dynamic default, rather than pulling one of the values.

mathemancer:
> Could you put the technical details you put in the PR somewhere in a code comment or README file so that we have it with the repo?

I added the details to a code comment, since that will be more visible to whoever is working on the relevant bits of code.

PR #691, opened 2021-10-04
fixes #635; changes to api/filters.py for Database, Schema, Table Filters; (Property)OrderingFilter from django_(property)_filters

### LINKED ISSUE #635: Table, Database, & Schema APIs should support sorting by ID & name
## Problem

The table, database, and schema APIs don't currently support sorting results.

## Proposed solution

All three APIs should support sorting by ID and name.

## Additional context

We can use `django-filter` / `django-property-filter` for this.

bohemia420:
w.r.t issue #635, 
For all the specified api endpoints:
```
- /api/v0/tables/
- /api/v0/schemas/
- /api/v0/databases/
```

fixes in this PR seems to sort the entries by either the field i.e "id" or the property i.e "name" (for `schemas` and `tables`), as specified. 
As of now, the capability is basis the assumptions:
- In the Django API, the Label is set to a standard **"Sort By"** followed by drop-down options (`id`, `id(desc)`, `name`, `name(desc)`)
- The `ordering_param` name is being inferred implicitly from the **OrderingFilter variable name**, set to `sort_by`. 
- As of now, ordering has been only accomplished for fields / properties **"id" and "name"**. 

For example,
`http://0.0.0.0:8000/api/v0/schemas/?sort_by=-name`
`http://0.0.0.0:8000/api/v0/tables/?sort_by=name`
`http://0.0.0.0:8000/api/v0/databases/?sort_by=-id`

Which can easily be reached from the "Filters" Modal inside django API `http://0.0.0.0:8000/api/v0/`.


kgodey (changes requested):
@bohemia420 This looks good, but please add some API tests that test this functionality. I know it's default `django-filters` functionality, but since we're mixing Django and SQLAlchemy code in our API, I think it's better to add additional tests so that we don't inadvertently break our API contract.

kgodey:
@bohemia420 Please let us know if you plan to finish this PR. If you're not able to, one of us will take over on it.

bohemia420:
hey Kriti

im afraid I wont have much time to spare and complete this in near future

AFAIR the test cases had to be written for this fix
kindly take this ahead with someone else from the team

apologies

kgodey (changes requested):
@silentninja Looks like you're missing tests for the database API. Otherwise, it looks good, thanks!

kgodey (approved):
Looks good, thanks!

silentninja:
The database tests are not destroying the database after test and the order is flaky

PR #692, opened 2021-10-04
Add custom URI type

### LINKED ISSUE #412: Create `MATHESAR_URI` custom type and handle it in the backend.

mathemancer:
Fixes #412 

This adds a custom URI type to Mathesar.  It also adds functions at the DB layer for getting the parts of a URI.

**Technical details**

The DB layer functions are based on a regular expression given in the RFC that defines a URI:  https://datatracker.ietf.org/doc/html/rfc3986 .  Note that similarly to the situation for the email type, the functions to get the parts of a URI are not yet hooked up to the API.  The parts that can be gotten are:
- scheme
- authority
- path
- query
- fragment
These are all of the pieces defined in the RFC.  The function to get part `X` is called `mathesar_types.uri_<X>`, takes a `TEXT` (or `VARCHAR` or `CHAR`) and returns the same.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (commented):
@mathemancer This looks good to me generally but it isn't recognizing URLs without schemes (e.g. `centerofci.org`). Could we detect those as well and default to scheme HTTP if none is detected?

[domains.csv](https://github.com/centerofci/mathesar/files/7281407/domains.csv) should serve as a good inference test.

mathemancer:
> @mathemancer This looks good to me generally but it isn't recognizing URLs without schemes (e.g. `centerofci.org`). Could we detect those as well and default to scheme HTTP if none is detected?

There is no such thing as a URL without a scheme :wink:.  The problem is that `localhost` (or `abc`, etc.) is a valid URL if you prepend `http://`.  We'd want to use some kind of "obvious" subset of URLs at that point, rather than trying to notice every valid URL, or the net will be far too wide.  Maybe noticing an authority segment with a suffix of `.<tld>`, where `<tld>` is in some abbreviated list of common TLDs?  I'll see what I come up with that doesn't seem to sloppy.  I agree that it would be a nice feature.

kgodey:
@mathemancer I think it's fine if we don't detect `localhost` or any TLD-less domain. it might be simpler to detect URIs in the form the form `<foo>.<bar>`, I'm not sure we need a list of TLDs to check `<bar>` against. We might have some false positives, but that's okay.

kgodey:
@mathemancer Okay, I'm convinced. :)

PR #733, opened 2021-10-15
Add support for nested Dropdown components

### LINKED ISSUE #732: Add support for nested Dropdown components
## Problem

If we want to place a Dropdown component within another Dropdown component, it's problematic because the outer Dropdown will close when the user clicks within the content of the inner Dropdown. 

For example:

Peek 2021-10-15 14-58

### Store-based approach

Use Svelte's [Context API](https://svelte.dev/docs#setContext). The outer Dropdown would instantiate a store and pass it down through context. For example:
    
```typescript
let accompanyingElements = writable<HTMLElement[]>([]);
setContext('dropdownAccompanyingElements', accompanyingElements);
```

Inner Dropdown components would then retrieve that store through context and update it with a reference to the DOM node of their content. To handle more than two levels of nesting, each Dropdown component would need to forward the references it receives up to any higher Dropdown components too.

### DOM-nesting approach

Either for all Dropdown content, or only for nested Dropdown content, insert the content node into the DOM alongside the trigger node (instead of appending it as a direct descendant of `<body>`).

### Class-based approach

Use the context API to make sure the inner components are aware that they are present within another dropdown and use that to append a class or a data value to the dropdown dom element. Upon receiving the callback from cickOffBounds, we would have to do an another check on the dom element returned by the click event to identify if it's on a child dropdown, and then decide if we need to close the parent.

seancolsen:
Fixes #732

## Before

User is unable to interact with the Dropdown content of an inner Dropdown.

Peek 2021-10-15 14-58

## After

Users can click on inner Dropdown content without closing the outer Dropdown

Peek 2021-10-15 16-52

## Notes

- It would be nice to have a test for this new behavior. I can add one if that's a priority.
- It would also be nice to flesh out the `Dropdown.stories.svelte` file. I only did the minimum amount of work there to test out this change. I figure we can create a proper stories file later when focusing more on the component library.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/components/dropdown/Dropdown.scss
+++ b/mathesar_ui/src/components/dropdown/Dropdown.scss
@@ -34,7 +34,7 @@ button.dropdown.trigger {
   background: #fff;
   z-index: 100;
   max-height: 210px;
-  overflow: auto;
+  overflow: visible;
 
   &[data-popper-placement="top-start"] {
     box-shadow: rgba(0, 0, 0, 0.1) 0px -10px 15px -3px,
--- a/mathesar_ui/src/components/dropdown/Dropdown.svelte
+++ b/mathesar_ui/src/components/dropdown/Dropdown.svelte
@@ -14,7 +14,7 @@
   } from '@mathesar-components/types';
   import type { Placement } from '@popperjs/core/lib/enums';
   import {
-    createEventDispatcher,
+    createEventDispatcher, getContext, setContext,
   } from 'svelte';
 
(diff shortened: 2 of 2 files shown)


seancolsen:
I think it's best to abandon this approach.

Further discussion moved to #732

Issue #421, opened 2021-07-17
Handle non-unicode encoded datafiles


sp35:
## Description
Upload CSV feature fails for non `utf-8` encoded CSVs. The exception is unhandled so the server returns 500.
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xae in position 265: invalid start byte`

## Expected behavior
We can ask for the encoding of the CSV file in the UI and handle the errors if any.

## To Reproduce
1. Upload non `utf-8` encoded CSV (e.g.- `unknown-8bit`: https://sample-videos.com/csv/Sample-Spreadsheet-10-rows.csv)
2. Observe error (Internal Server Error)

## Environment
Docker

kgodey:
Thanks for reporting this bug, @sp35!

If possible, I'd like to avoid asking users for the encoding since we're aiming to be friendly to non-technical users. I think we should investigate a solution that try different encoding options automatically.

asharonbaltazar:
I've used a small 2kb sample file I found on the web and it failed with this error: 

`Unable to tabulate data`

Here's the file I used: 

[grades.csv](https://github.com/centerofci/mathesar/files/7000995/grades.csv)

eito-fis:
The grades file seems to cause the `Unable to tabulate data` error due to a missing comma on line 10. However, after fixing that, we still get an error - it seems like having quotes in the column name also breaks things. I'll open a new issue for this.

silentninja:
### Proposal
---
Use encoding detection libraries like [Charset Normalizer](https://pypi.org/project/charset-normalizer/) or [cChardet](https://pypi.org/project/cchardet/). 

### Issues
---
Blocking request - Encoding detection is not instant and must be dealt asynchronously. This is also the case with creating data from csv(will be blocking for huge csv files) and solving it should also solve this problem.
Return null instead of throwing an exception

### Remarks
---
I am using charset-normalizer and haven't had issues so far, so I am leaning towards using that library.
I can send a PR if this sounds good.

kgodey:
@silentninja That sounds good to me. If you're still interested, please feel free to send a PR.

We've put off handling operations asynchronously for now in order to avoid adding additional dependencies such as a task queue, etc. for Mathesar. Let's handle this synchronously for now and create a separate issue for handling it asynchronously.

PR #738, opened 2021-10-19
Adding a table via uploading a data file

### LINKED ISSUE #446: Implement adding a table via uploading a data file from the web
## Problem

Users may want to create a table from a data file hosted on the internet without having to download to their computer first.

## Proposed solution

The ["Basic Table Operations" design spec](https://wiki.mathesar.org/design/specs/table-operations) has a solution for this, which we need to implement on the frontend. This issue involves the following portions of the spec:
- _User Experience_: 
  - User adds a new table
  - User adds a new table from imported data via URL
 
## Additional context

- Backend work: #203 
- Design issue: #185

mr-gabe49:
Fixes #446 

- This PR allows users to upload a file through a URL

**Technical details**

- Created a Radio component
- Test URL files: 
    - https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv
    - https://people.sc.fsu.edu/~jburkardt/data/csv/zillow.csv
 - This site contains many URL files that can be used for testing: https://people.sc.fsu.edu/~jburkardt/data/csv/csv.html
      -  Many of the csv files fail to upload, even after being downloaded to computer and imported straight from there. This might be related to #421 

**Screenshots**

Screenshot (55)
Screenshot (54)

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mr-gabe49:
I'm currently stuck. I could use either @pavish or @seancolsen help when they have time. It is failing one test. I manually commented out different parts of the code to determine what was causing the problem.  Line 65 ` const { id }: { id: number } = uploadPromise;` of the _importUtils.ts_ file seems to be the cause. I'm having trouble understanding why this is the case and how to move forward.

mr-gabe49:
@seancolsen I think I've handled all your comments on the radio component. Let me know your thoughts. 
@pavish I handled the UI issues you mentioned. It works fine, however, code-wise I don't think it's correct. I'll make a comment on the code, so it's easier for you to review.

mathesar_ui/src/sections/import-data/importUtils.ts
```diff
@@ -378,6 +378,31 @@ export function cancelImport(fileImportStore: FileImport): void {
   }
 }
 
+export async function importFromURL(fileImportStore: FileImport, url: string): Promise<void> {
+  setInFileStore(fileImportStore, {
+    previewTableCreationStatus: States.Loading,
+  });
```
mr-gabe49, line 384: @pavish This works in disabling the button and showing the spinner when an user clicks on it. However, we also call this in the `createPreviewTable` function nested inside the `loadPreview` function. It works fine, but seems wrong to be calling it twice.

seancolsen (approved):
@mr-gabe49 you've addressed my review feedback. The changes @pavish requested are great too, so I'd like him to have another look at this PR before we merge it. I feel good about it once he signs off.

mathesar_ui/src/component-library/radio-group/RadioGroup.svelte
```diff
@@ -0,0 +1,27 @@
...
+</script>
+
+<fieldset class="radio-group">
+  <legend>
+    <slot></slot>
+  </legend>
+  <div class:inline={isInline}>
+    {#each options as option (option[idKey])}
+      <label class="radio">
+          <input type="radio" bind:group value={option[idKey]}>
```
seancolsen, line 18: TypeScript is complaining on this line. I'm fine if we address this in a subsequent PR. I'd like an opinion from @pavish on the best approach. 

One fix is to change `RadioGroup.d.ts` to:

```ts
export interface RadioGroupOption {
  [key: string]: string | number | string[]
}
```

I know you were following the pattern that Pavish set in `Select.d.ts`. In the context of the RadioGroup component, I can't immediately see the usefulness of making `RadioGroupOption` generic. Pavish probably had a good reason to make `SelectOption` generic, but I'm not sure if that applies here. 

I would have designed these components with static keys for the label and id. My personal opinion is that it would simplify things a bit, especially the typing.

kgodey:
FYI I merged the latest `master` into this PR to test some automated workflow changes.

mathesar_ui/src/component-library/radio-group/RadioGroup.svelte
pavish: I'd change it to 
```
export interface RadioGroupOption<T = unknown> {
  [key: string]: T
}
```

This will keep it generic, for cases where we need to extend from this, and the default would be unknown.

mathesar_ui/src/sections/import-data/importUtils.ts
pavish: @mr-gabe49 This is because previously we needed to show the loading indicator only when during the preview step. In this case, we need to show it based on `uploadStatus` as well.

Instead of setting `previewTableCreationStatus` in this section, we should be setting `uploadStatus`

pavish (changes requested):

mathesar_ui/src/component-library/radio-group/RadioGroup.svelte
pavish: For radio value, I think we can go with `string | number | string[]` since these are the only options that can be used as value for the input element.

`SelectOption` was made generic because the entire Select component is a custom element which can use objects as id or value.

pavish:
I've added a commit https://github.com/centerofci/mathesar/pull/738/commits/60e7b4ec11c4f90d9650f9c8b1312dea65d016e0 which resolves https://github.com/centerofci/mathesar/pull/738#discussion_r752480390 and  https://github.com/centerofci/mathesar/pull/738#discussion_r748594975

pavish (approved):

PR #760, opened 2021-10-26
Handle deleting table constraints

### LINKED ISSUE #328: Frontend work for viewing and deleting table constraints
## Problem

Users may want to view or delete table constraints.

## Proposed solution

We should provide a way to do these actions from the UI following [the "Working with Columns" design spec](https://wiki.mathesar.org/en/design/specs/working-with-columns).

This covers:
- Displaying table constraints of all types
- Deleting table constraints of all types

## Additional context

- Design issue: #324 
- Design discussion: #436
- Backend issue: #326 
- Backend PR: #488 
- This issue blocks: #531

seancolsen:
Fixes #328

## Before

- No way to delete table constraints

## After

User can delete table constraints through the "Table Constraints" modal.

Peek 2021-10-26 21-26

_(The slow frame rate of the gif doesn't really do justice to the CSS transitions, which I think look really nice in the browser.)_

## Notes

- After dropping the constraint, the constraints store is re-fetched.

- I added a couple new UI components:

    - `Spinner`, so that we can write `<Spinner />` instead of `<Icon data={faSpinner} spin={true}/>`. (One fewer import!)
    
    - `Seesaw` (open to suggestions for a better name here) which switches between a "left" and "right" content using a slide transition.

## Checklist
 
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mr-gabe49:
Hey @seancolsen, what do you use to create the gifs? They're pretty cool.

seancolsen:
@mr-gabe49 I'm on Linux and using a small tool called Peek. It's very nice!

mathesar_ui/src/sections/table-view/constraints/TableConstraint.svelte
```diff
@@ -0,0 +1,103 @@
+<script lang='ts'>
+  import { fly } from 'svelte/transition';
+  import type { FlyParams } from 'svelte/transition';
+  import {
+    Icon,
+    Seesaw,
+    Button,
+    Spinner,
+  } from '@mathesar-components';
+  import type { Constraint } from '@mathesar/stores/table-data/constraints';
```
pavish, line 10: We can import types from the type definition export file to maintain consistency.

```suggestion
  import type { Constraint } from '@mathesar/stores/table-data/types';
```

mathesar_ui/src/components/spinner/Spinner.svelte
```diff
@@ -0,0 +1,6 @@
+<script lang="ts">
+  import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+  import { Icon } from '@mathesar-components';
+</script>
+
+<Icon data={faSpinner} spin={true}/>
```
pavish, line 6: We could also export a `size` prop which can be passed down to the Icon.

mathesar_ui/src/sections/table-view/constraints/TableConstraint.svelte
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+  
+  $: columnSummary = constraint.columns.join(', ');
+  $: transitionDuration = useTransitionOut ? 200 : 0 as number;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  $: transition = { x: 200, duration: transitionDuration } as FlyParams;
+</script>
+
+<div class='table-constraint' class:isConfirmingDrop class:isSubmittingDrop>
```
pavish, line 44: It would be better if we followed consistent css class naming pattern, in our case we've followed kebab-cased css class naming everywhere.

```suggestion
<div class='table-constraint' class:is-confirming-drop={isConfirmingDrop} class:is-submitting-drop={isSubmittingDrop}>
```

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -63,11 +63,17 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
...
+  async fetch({
+    showLoading = true,
+  }: {
+    showLoading?: boolean,
+  } = {}): Promise<ConstraintsData> {
+    if (showLoading) {
+      this.update((existingData) => ({
+        ...existingData,
+        state: States.Loading,
+      }));
```
pavish, line 75: The state property in each store is meant to only reflect the state of the store itself. As of now, the data layer (stores) is completely separate from the view layer (components), and only the view depends on the data, and not the other way around.

Setting the state based on a conditional passed down from the component ties it to the view layer. It would be better to handle this on the component, instead of the store.

In the component, we can show the spinner only if constraints is empty and state is loading.

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -95,6 +101,10 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
     return null;
   }
 
+  drop(constraintId: Constraint['id']): CancellablePromise<undefined> {
+    return deleteAPI(`/tables/${this.parentId}/constraints/${constraintId}`);
+  }
```
pavish, line 106: This function only performs an API call and does not update the store itself, which makes it inconsistent. We can follow the pattern we discussed to encapsulate the api calls into an object. The drop function can then call fetch within it, after the constraint is deleted.

pavish: This would also remove the need for manipulating the loading state from within the component.

pavish (changes requested):
@seancolsen I love the subtle animation and the simplicity of the confirmation view.

I do have a few comments which I'd like to be addressed before we merge this PR.

pavish:
An additional minor comment: When all constraints are deleted, only an empty modal with a title is shown. It would be better if we showed additional text mentioning that there are no constraints available.

seancolsen:
> When all constraints are deleted, only an empty modal with a title is shown.

- Good call. In d384c4a1a5847270fc60a3ece4827d9e35eb61ea changed the modal to display as follows:

    image

mathesar_ui/src/sections/table-view/constraints/TableConstraint.svelte
seancolsen: Addressed in 284a5a099123040c2b7e75c100a947033ed24ead

mathesar_ui/src/sections/table-view/constraints/TableConstraint.svelte
seancolsen: - Yeah this is that code I mentioned I had already written when we chatted on the phone about this.
- Turns out I wasn't using the `isConfirmingDrop` class. So I removed it.
- I fixed the `isSubmittingDrop` class in af011153ce1ed7b90025307728c08da8c9865796

mathesar_ui/src/stores/table-data/constraints.ts
seancolsen: Addressed in 087452ba0941b2690a0d20bc81d5025c0713a65c

mathesar_ui/src/stores/table-data/constraints.ts
seancolsen: Addressed in 087452ba0941b2690a0d20bc81d5025c0713a65c

seancolsen:
@pavish Ready for re-review

pavish (approved):
@seancolsen The changes look good to me!

PR #776, opened 2021-10-29
Add "Allow Duplicates" toggle option to column header menu

### LINKED ISSUE #772: Add "Allow Duplicates" toggle option to column header menu
- [Specs](https://wiki.mathesar.org/en/design/specs/working-with-columns)

- [Mockup](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=2750%3A17660&node-id=2846%3A20398&viewport=2205%2C368%2C0.5616371035575867&scaling=contain&starting-point-node-id=2816%3A18654&show-proto-sidebar=1)

    image

Additional details:

- Column header menu should display the "Allow Duplicates" state for every column
- Column header menu should let the user alter the "Allow Duplicates" state for columns that don't have a constraint of type `'primary'`.
- Loading spinner should display within menu while making request
- Code comment should provide a placeholder for using the toast system (yet to be built) for displaying error if the request to alter the constraint failed. No other failure handling at for now.

seancolsen:
Fixes #772

[Demo video with comments](https://www.loom.com/share/6894521173f34997894a8a2da874bcb6)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
I watched the demo, it looks good to me. Nice work!

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
```diff
@@ -3,8 +3,10 @@
     faSortAmountDown,
     faSortAmountDownAlt,
     faThList,
+    faSpinner,
   } from '@fortawesome/free-solid-svg-icons';
-  import { Dropdown, Icon } from '@mathesar-components';
+  import { Checkbox, Dropdown, Icon } from '@mathesar-components';
+  import type { ConstraintsDataStore } from '@mathesar/stores/table-data/constraints';
```
pavish, line 9: We can import the types from the type definition file.

```suggestion
  import type { ConstraintsDataStore } from '@mathesar/stores/table-data/types';
```

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
```diff
@@ -35,6 +49,22 @@
...
     }
+    closeMenu();
+  }
+
+  async function toggleAllowDuplicates() {
+    isRequestingToggleAllowDuplicates = true;
+    try {
+      const isUnique = await constraintsDataStore.updateUniquenessOfColumn(column, (u) => !u);
+      const msg = `Column "${column.name}" will ${isUnique ? 'no longer ' : ''}allow duplicates.`;
+      console.log(msg); // TODO display success toast message: msg
```
pavish, line 60: We could add a eslint ignore before the console.log line to avoid eslint warnings.

mathesar_ui/src/stores/table-data/columns.ts
```diff
@@ -41,6 +46,20 @@ function preprocessColumns(response?: Column[]): Column[] {
...
+    },
+    add(columnDetails: Partial<Column>) {
+      return postAPI<Partial<Column>>(url, columnDetails);
+    },
+    remove(index: Column['index']) {
+      return deleteAPI(`${url}${index}`);
+    },
+  };
+}
+
```
pavish, line 62: This is neat! :)

mathesar_ui/src/stores/table-data/columns.ts
```diff
@@ -41,6 +46,20 @@ function preprocessColumns(response?: Column[]): Column[] {
...
+function api(url: string) {
+  return {
+    get() {
+      return getAPI<PaginatedResponse<Column>>(`${url}?limit=500`);
+    },
+    add(columnDetails: Partial<Column>) {
+      return postAPI<Partial<Column>>(url, columnDetails);
+    },
+    remove(index: Column['index']) {
+      return deleteAPI(`${url}${index}`);
```
pavish, line 58: We should probably end all urls with a trailing slash.

The server seems to do a redirect by adding a trailing slash, if we didn't specify any. We can avoid a roundtrip by adding it here.

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -23,13 +32,29 @@ export interface ConstraintsData {
...
+function api(url: string) {
+  return {
+    get() {
+      return getAPI<PaginatedResponse<Constraint>>(`${url}?limit=500`);
+    },
+    add(constraintDetails: Partial<Constraint>) {
+      return postAPI<Partial<Constraint>>(url, constraintDetails);
+    },
+    remove(constraintId: Constraint['id']) {
+      return deleteAPI(`${url}${constraintId}`);
```
pavish, line 44: Same as my previous comment. We can add a trailing slash here.

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
```diff
@@ -16,17 +18,29 @@
...
+  export let constraintsDataStore: ConstraintsDataStore;
+  
+  let isRequestingToggleAllowDuplicates = false;
+  let menuIsOpen = false;
+  
   $: ({ sort, group } = meta);
   $: sortDirection = ($sort as SortOption)?.get(column.name);
   $: hasGrouping = ($group as GroupOption)?.has(column.name);
 
+  $: allowsDuplicatesStore = constraintsDataStore.columnAllowsDuplicates(column);
```
pavish, line 30: This would create a new store for each component instance of HeaderCell. It would also create new stores whenever the user switches tabs.

We could probably optimize this to avoid creation of multiple stores.

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -95,6 +120,92 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
...
+    return derived(
+      this.columnHasUniqueConstraint(column),
+      (hasUniqueConstraint) => {
+        if (column.primary_key) {
+          return false;
+        }
+        return !hasUniqueConstraint;
+      },
+    );
+  }
```
pavish, line 173: As mentioned in the previous comments, creating a new store and returning it here is not optimal.

My suggestion would be to:
* Create just one derived store as a property of ConstraintsDataStore, called duplicatesAllowingColumns or similar. We can think of a better name.

  ```
   type DuplicatesAllowingColumnsSubstance = Set<Column['name']>
   type DuplicatesAllowingColumnsStore = Readable<DuplicatesAllowingColumnsSubstance>
   ```
* This store would only contain columns that allow duplicates. This will include non-unique, non-primarykey columns.
* Within the component, we'd just do `$duplicatesAllowingColumns.has(column.name)`

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -95,6 +120,92 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
...
+
+  async updateUniquenessOfColumn(
+    column: Column,
+    updater: (isUnique: boolean) => boolean,
```
pavish, line 177: What is the reasoning behind this updater argument?

Since this function only requires a boolean to identify whether or not to update the uniqueness, wouldn't a boolean suffice here?

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -1,19 +1,28 @@
...
+import type { Column } from './columns';
+
+export type ConstraintType = 'foreignkey' | 'primary' | 'unique' | 'check' | 'exclude';
 
 export interface Constraint {
   id: number,
   name: string,
-  type: string,
+  type: ConstraintType,
   columns: string[],
```
pavish, line 26: It would be more optimal to make columns a set, we seem to be require finding if a column exists, in several places.

```suggestion
  columns: Set<Column['name']>,
```

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -95,6 +120,92 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
...
+  }
+
+  /**
+   * A constraint only matches if the set of its columns strictly equals the set
+   * of columns supplied here. For example, if a constraint is set on three
+   * columns, two of which are passed to this function, that constraint will
+   * _not_ be returned.
+   */
+  constraintsThatMatchSetOfColumns(columns: Column[]): Readable<Constraint[]> {
+    const columnsNames = columns.map((c) => c.name);
```
pavish, line 141: We can optimize this block further by using a Set for `columnsNames`, or even the argument `columns`.

mathesar_ui/src/stores/table-data/constraints.ts
```diff
@@ -95,6 +120,92 @@ export class ConstraintsDataStore implements Writable<ConstraintsData> {
...
+    function isMatch(constraint: Constraint) {
+      if (constraint.columns.length !== columnsNames.length) {
+        return false;
+      }
+      return constraint.columns.every(
+        (constraintColumn) => columnsNames.includes(constraintColumn),
+      );
```
pavish, line 148: We can use the `intersection` method from '`utils/language.ts`', or even have a new common method in that util file.

pavish (changes requested):
@seancolsen I played around with the feature and it looks pretty good to me!

I have a few comments, most of them are minor but I do have a major comment on optimizing the stores, which I think we need to handle before merging this in.

mathesar_ui/src/stores/table-data/constraints.ts
seancolsen: Addressed in 7d3ba58d38a291cabed95322d59bbd8e26532c8b

mathesar_ui/src/stores/table-data/columns.ts
seancolsen: Addressed in 7d3ba58d38a291cabed95322d59bbd8e26532c8b

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
seancolsen: Addressed in 7d3ba58d38a291cabed95322d59bbd8e26532c8b

mathesar_ui/src/stores/table-data/constraints.ts
seancolsen: Yeah good question. I remember having a reason for using an updater -- but I can't remember what the reason was! I'm guessing that some other changes I made along the way have now rendered my original rationale moot. In any case, I've made a change in bebb4e9372515fcb6b17a7f2ac0edb88bb9f3180 that simplified this function to use a value instead of an updater. Thanks for pointing this out.

seancolsen:

## `string[]` vs `Set<string>` for `columns`

- I had initially considered a `Set` as well but decided against it for the following reasons:
    - It adds complexity.
    - The performance gains are unlikely to be significant because the number of values in `columns` will rarely be large. I'd guess that at least 95% of constraints will contain only one value in `columns`. And a constraint for more than 5 columns would be _super_ rare. Right?

- Personally, I'd still lean towards using an Array here for simplicity's sake, but I don't feel strongly. I went ahead and made a [commit](https://github.com/centerofci/mathesar/commit/974a050855f632032078d0c2b62c668be97fbf3c) to convert `columns` to `Set<string>`, but I put it in the `772_allow_duplicates_toggle__columns_set` branch just in case you feel persuaded by the above argument. Merge this PR as-is to keep `string[]`, or merge `772_allow_duplicates_toggle__columns_set` into `772_allow_duplicates_toggle` first to use `Set<string>`. Your choice.

- The commit I made demonstrates this additional complexity, some of which stems from the fact that the `Constraint` we're storing is no longer immediately serializable as JSON. If you want to propose ways to reduce that complexity or increase readability within that new code, I'm open to making further adjustments.


## Using distinct stores used for component instances

@pavish [said](https://github.com/centerofci/mathesar/pull/776#discussion_r743401030):

> creating a new store and returning it here is not optimal

I'm confused by this. Yes, each component instance will get a separate store. Can you help me to see the ways in which this practice it not optimal? Is it a memory issue? CPU? Other? Have you actually experienced performance problems from this practice? Read about them somewhere? Or is it just a hunch?

My (very rudimentary) understanding of the Svelte internals is that any time we use `$` to access a store or declare a reactive value, Svelte just desurgars that `$` syntax to produce a JS object very similar to a store. So calling `$duplicatesAllowingColumns` (as you propose) would have similar performance characteristics to calling `columnAllowsDuplicates(column)` (as I have).

Some of these thoughts are a little hard to articulate, without being sure where the difference in our understanding lies -- so if we're to continue this discussion, it might be most efficient to hop on a call.

I'd like to get clear on this though because the `return derived` pattern is one that I've used a _lot_ in my work with Svelte. I basically treat it like a more verbose equivalent of reactive assignments. And with reactive assignments only being possible in `.svelte` files, I use a lot of `return derived` because I like to shift as much of the business logic to `.ts` files as I can. If `return derived` is not a good practice, I'd like to understand more about why.

pavish:
@seancolsen 

* Let me first cover 'Using distinct stores for component instances'

  > Can you help me to see the ways in which this practice it not optimal? Is it a memory issue? CPU? Other? Have you actually experienced performance problems from this practice? Read about them somewhere? Or is it just a hunch?

  >  calling $duplicatesAllowingColumns (as you propose) would have similar performance characteristics to calling columnAllowsDuplicates(column) (as I have).

  There are two separate aspects to this:
    1. The store assignment within the Svelte component
    2. The store itself - creation, subscription and unsubscription

  Your argument holds true for (1). Both your approach and mine would have similar characteristics with respect to `$:`.

  However, my point is that in your approach _we are creating new stores_ for each component instance as well as whenever the arguments change, and this causes a major difference in both performance and memory allocation.

  Let me try to illustrate with an example: [REPL link](https://svelte.dev/repl/e052f88b84e64131822d98c983bdfa91?version=3.44.1)
  Please refer the Component.svelte file, and comment/uncomment lines to switch between approaches.

  **Whenever the component is hidden/shown, or when the numbers are reset:**

  ### Approach 1: Approach currently implemented
  There are '2n' console logs (one for the callback being called and 1 for the unsubscriber). This leads to an `O(n)` _time and space_ complexity.

  Each time, there are n new stores created which are then subscribed to, and older stores are unsubscribed to. This approach has the _included cost of additional memory and time taken to create the new stores_.

  Considering that in our code we have multiple functions each returning a derived store 'm' (3 in our case) and dependent on one another, we will have m * n calls, each including store creation, subscription and unsubscription, leading to a `O(m * n)` time and space complexity. Also, all these newly created stores hold similar information, which is highly unnecessary.

  ### Approach 2: Suggested change
  There are 2 console logs (1 for callback and 1 for unsubscriber). This leads to an `O(1)` time and space complexity.

  Each time, we only subscribe and unsubscribe to the same store.

  ### My concern:
  Approach 1 might seem harmless now but it may end up creating bad user experience in the long run. The effects aren't visible because our current codebase is small, but if we follow this approach and the codebase grows, it will be harder to optimize later. 

  Here we have two approaches one that's `O(m * n)` and another `O(1)` solving the same problem. I would like us to go with the most optimal one. 

* Regarding,
  > any time we use $ to access a store or declare a reactive value, Svelte just desurgars that $ syntax to produce a JS object very similar to a store

  >  I basically treat it (return derived) like a more verbose equivalent of reactive assignments

  Reactive assignments behave very differently to stores. The following points are as per my understanding of svelte internals which are based on how the compiled code looks.
  * When the compiler encounters the `$:` syntax,
    *  It tries to optimize internal calls based on how the reactive variables are used. For general values which are not stores, it just handles invalidation and updating during each tick, and it batches all the reactive values into a single call in the same tick. 
    * For stores that are imported directly and just referenced from the reactive variable  (Approach 2 does this), it handles the invalidation logic and calls a `component_subscribe` method which wraps the subscription and unsubscription logic. This has a slightly reduced output size.
    * For stores that are returned from a function and referenced from the reactive variable (Approach 1 does this), all the subscription and unsubscription logic is present within the compiled code of the component. This is similar to point 2, except it has a slightly higher output size.
  * When stores are referenced by `$`,
     * This is just a syntactic sugar for store operations such as store.subscribe, store.set and store.get. It also lets the compiler know that the store needs to automatically be unsubscribed when component is destroyed.

   The invalidation logic of reactive assignments are different to the listener/subscription model of stores.

pavish:
@seancolsen 

> string[] vs Set<string> for columns

This is a case of analyzing tradeoffs and finding the most common usecase. As I see it, we need to be able to:
* Find if a column is part of a constraint
  - Using a Set would be more beneficial here
* Send requests to update constraints and receive constraint data
  - Using an array would be more beneficial here.
  - However, as we move ahead, we may have to store additional metadata specific for the frontend, which will result in us adding a custom serialization/deserialization logic. We already do that for columns and records. So, the serialization/deserialization step might be essential either way.
* List all constraints with columns
  - Using an array would be more beneficial here

I assume the most common usecase as we move ahead will be to find if a column is part of a constraint, since we need to display this information on the column header. My concern is that the most common usecases must have the most optimized time complexity.

If you feel that we can optimize this when we get there, I'm onboard with that. I would just like to avoid repeated array.find, array.every, array.any calls on the same dataset.

seancolsen:
@pavish Wow! Thanks for your incredibly detailed and educational comments! I'll refactor the derived store to follow your approach there.

For the Array vs Set decision:

- You said using a Set would be more beneficial when finding if a column is part of a constraint. Would that truth still hold in a hypothetical case where all constraints were single column constraints? To be specific, I'm assuming that testing a single-element Array for a value has similar performance to testing a single element Set for a value. But maybe I'm wrong. Also, I'd expect the vast majority of constraints to be set for only one column. If I'm correct about both of the above assumptions, then we have a situation which mitigates the performance drawbacks of that most common use case, right? Or is there something else I'm missing?

pavish:
@seancolsen 

> Thanks for your incredibly detailed and educational comments!

You're welcome. :)

I like the Svelte REPL because it shows the compiled JS output. You can play around with it and see the difference in compiled code for between reactive statements and stores.

For the Array vs Set decision,

> Would that truth still hold in a hypothetical case where all constraints were single column constraints?

If the array has a single element, it wouldn't make any difference. 

> Also, I'd expect the vast majority of constraints to be set for only one column.

I also think this holds true.

I am satisfied with these points and I think we can go with an array for `constraints.columns`

 For cases like the `$duplicatesAllowingColumns` (store mentioned in approach 2), I still think we will need a Set, since the number of columns cannot be determined here.

seancolsen:
> I like the Svelte REPL because it shows the compiled JS output.

Oh that's a good point. I've not explored that enough. Thanks for pointing that out!

> For cases like the $duplicatesAllowingColumns (store mentioned in approach 2), I still think we will need a Set, since the number of columns cannot be determined here.

Agreed

seancolsen:
@pavish Ready for re-review now

- I made the store changes in e46e9c442dae9cbb2a15b1cf889e7d00590f55d2
- Since we were on the fence about Array vs Set, I also added a brief comment linking to the discussion here in case future devs have the same question.

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
pavish: @seancolsen This is still present in some files like Header.svelte, but I think this can be handled in one of your later PRs.

pavish (approved):

pavish:
@seancolsen I've approved this PR, and set the base to be the `fix_database_ordering_test` branch. It will automatically change to master when it gets marged.

mathesar_ui/src/sections/table-view/header/HeaderCell.svelte
seancolsen: Whoops, sorry about that. I misunderstood and thought you were only pointing out to one file that needed to be fixed so I didn't look elsewhere. I've fixed others in b92be5be7ff0358847e93a405672b4d6511fa30d

PR #815, opened 2021-11-10
Add Django column model

### LINKED ISSUE #658: Set up `Column` Django model with support for display options
## Problem

Some of our data types have display options. We need to store these display options associated with columns of those data types.

## Proposed solution

(1) Create a `Column` Django model with a `display_options` JSON field to store display options.
  - Update `/api/v0/columns/` to use the Django model, including using the auto-generated `id` for the ID.
  - Implement automatic DB reflection of columns.
 
(2) While reflecting columns from the database:
  - If name or index of the column changes, we destroy the Column object and create a new one with a new `id` (we never update Column objects). 
  - We will copy display options from an old column object to the new object based on these rules:
    - if we see a column at the same index with the same data type but the name is different from the current column, we'll copy the display options over from the current column.
    - if we see fewer columns than we expect from previous reflection, we'll use the names/types of the columns to figure out which ones have been deleted and copy over display options between columns with the same names and types.

(3) If we are updating column names/types from within Mathesar, we will update the existing Column object.

(4) For the purposes of this issue, do not allow display options to be set. We will implement setting display options in each specific data type issue.

## Additional context

- This implementation means that when a table's columns are deleted or renamed outside of Mathesar, we might lose some display options. This seems unavoidable. I investigated the idea of storing display options in Postgres column comments, but columns are identified by the table's `OID` and column index, which is what we'd be storing anyway. This way, we don't have to touch the database (people may already be using column comments for other things).
- #392 has more discussion.

silentninja:
Change table column serializer to use the Django column instead of sql alchemy columns
Reflect column when a table is created

Fixes #658 

Add Django column model replacing the reflected database column. This allows us to store some additional data(like display options) which would be used by Mathesar without affecting the user database 

**Technical details**
Adds column model which acts as a wrapper around SQL alchemy columns.

This PR introduces some breaking changes to tables and columns API due to the change to the column model the API refers to(from direct database column representation to Mathesar custom column)

The breaking changes are:

Column response model in tables API and columns API adds a new field called 

```

id:
  type: integer
  description: Identifier for the column model

```
The `id` field should be used to refer to the columns instead of the column index used previously.

The affected API's are:

`/tables/`: The column response model for the field `columns` returns `id` field
`/columns/`: The column response model returns `id` field
All the `/columns/index` -> `columns/id: Changed from column index to id, 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models.py
```diff
@@ -45,14 +46,23 @@ def get_queryset(self):
         return super().get_queryset()
 
 
-class DatabaseObject(BaseModel):
-    oid = models.IntegerField()
+class ReflectingObject(BaseModel):
```
kgodey, line 49: I would call this `DatabaseObject` and the current `DatabaseObject` something like `ObjectWithOID`.

mathesar/models.py
```diff
@@ -309,6 +319,35 @@ def add_constraint(self, constraint_type, columns, name=None):
         return Constraint.objects.create(oid=constraint_oid, table=self)
 
 
+class Column(ReflectingObject):
+    table = models.ForeignKey('Table', on_delete=models.CASCADE, related_name='columns')
```
kgodey, line 323: I assume you'll add `attnum` and `display_options`?

mathesar/models.py
silentninja: I was planning to go with that first and then realised the Database model would be using it too and using DatabaseObject as a base model will be confusing, will come back to this.

mathesar/models.py
silentninja: Yes, I pushed the changes I was sure of.

mathesar/models.py
silentninja: Mixins should be a better fit. Have to check if Django plays well with mixins that add fields

silentninja:
@pavish @seancolsen There are a few breaking changes to the API. I have included the API changes to the description of the PR. Let me know if it is the frontend fix can be included in this PR or if you would like to handle it in a separate PR

pavish:
@silentninja I can handle the changes in this PR

pavish:
@silentninja The frontend changes are added in https://github.com/centerofci/mathesar/pull/815/commits/a160a8d4b7180d27ad5f3d6914c17dfba7db975b

silentninja:
I haven't made the necessary changes when a column is moved by mathesar using the `move_columns_between_related_tables`, since it is part of the db module and not exposed via an API.

kgodey:
> I haven't made the necessary changes when a column is moved by mathesar using the move_columns_between_related_tables, since it is part of the db module and not exposed via an API.

@silentninja We will be exposing it via the API in the [Data Modeling](https://github.com/centerofci/mathesar/milestone/24) milestone. It's up to you whether you want to update it now or wait until then.

mathesar/tests/api/test_column_api.py
```diff
@@ -103,65 +107,7 @@ def test_column_list(column_test_table, client):
             'default': None,
         }
     ]
-    assert response_data['results'] == expect_results
-
-
-@pytest.mark.parametrize(
```
silentninja, line 109: Removed this as it is pretty much testing Django rest framework ModelViewset

mathesar/models.py
kgodey: Makes sense. I think `ReflectedObject` would be a bit clearer than `ReflectingObject`.

mathesar/api/serializers/columns.py
```diff
@@ -35,13 +37,23 @@ def run_validation(self, data=empty):
...
+        model = Column
+        fields = ('id', 'name', 'type', 'type_options')
     name = serializers.CharField()
     type = serializers.CharField(source='plain_type')
     type_options = TypeOptionSerializer(required=False, allow_null=True)
 
 
 class ColumnSerializer(SimpleColumnSerializer):
+    class Meta(SimpleColumnSerializer.Meta):
+        fields = SimpleColumnSerializer.Meta.fields + ('nullable', 'primary_key', 'source_column', 'copy_source_data',
```
kgodey, line 51: I find lists and tuples easier to read when they're one item per line. I would do something like this:

```python
fields = SimpleColumnSerializer.Meta.fields + (
   'nullable',
   'primary_key',
   'source_column',
   'copy_source_data',
   etc...
)
```

mathesar/models.py
```diff
@@ -45,14 +49,23 @@ def get_queryset(self):
...
     objects = DatabaseObjectManager()
 
+    class Meta:
+        abstract = True
+
+    def __str__(self):
+        return f"{self.__class__.__name__}"
+
+
+class DatabaseObject(ReflectingObject):
```
kgodey, line 66: I think `ReflectedObject` and `ReflectedObjectWithOID` would be clearer names here.

kgodey (changes requested):
Looks good, just some style/nomenclature suggestions.

I also have a question. The original issue has the following requirements:

> 
> (2) While reflecting columns from the database:
>   - If name or index of the column changes, we destroy the Column object and create a new one with a new `id` (we never update Column objects). 
>   - We will copy display options from an old column object to the new object based on these rules:
>     - if we see a column at the same index with the same data type but the name is different from the current column, we'll copy the display options over from the current column.
>     - if we see fewer columns than we expect from previous reflection, we'll use the names/types of the columns to figure out which ones have been deleted and copy over display options between columns with the same names and types.

I don't see any logic for this. Is this unnecessary because we're using `attnum` or is there another reason for omitting this? My concern is that someone could modify the DB outside of Mathesar and we will lose the values of display options.

silentninja:
> Looks good, just some style/nomenclature suggestions.
> 
> I also have a question. The original issue has the following requirements:
> 
> > (2) While reflecting columns from the database:
> > 
> > * If name or index of the column changes, we destroy the Column object and create a new one with a new `id` (we never update Column objects).
> > * We will copy display options from an old column object to the new object based on these rules:
> >   
> >   * if we see a column at the same index with the same data type but the name is different from the current column, we'll copy the display options over from the current column.
> >   * if we see fewer columns than we expect from previous reflection, we'll use the names/types of the columns to figure out which ones have been deleted and copy over display options between columns with the same names and types.
> 
> I don't see any logic for this. Is this unnecessary because we're using `attnum` or is there another reason for omitting this? My concern is that someone could modify the DB outside of Mathesar and we will lose the values of display options.

Yes, that's right we don't actually need that. The [test_column_update_name](https://github.com/centerofci/mathesar/blob/eb243b4f9bfab4c7c49ef36407c7fa13dfe3a78b/mathesar/tests/api/test_column_api.py) should assures that both columns are same after a rename

db/columns/operations/select.py
```diff
@@ -17,17 +17,24 @@
...
             pg_attribute.c.attrelid == table_oid,
-            pg_attribute.c.attname == column_name
+            pg_attribute.c.attname.in_(column_names)
         )
     )
-    result = execute_statement(engine, sel, connection_to_use).fetchone()[0]
+    return execute_statement(engine, sel, connection_to_use).fetchall()
+
+
+def get_column_index_from_name(table_oid, column_name, engine, connection_to_use=None):
```
mathemancer, line 33: I don't know if we want to do it on this PR, but we need to modify every place that the column index is used for identification and replace it with the `attnum`, or look up the name from `attnum` if needed.  The column index idea was a mistake, and is the source of numerous previous and a few visible current bugs.  For example, take a look at the `db.constraints.operations.select.get_column_constraints` function.  That will break if columns have been deleted and what not.

db/tables/operations/alter.py
```diff
@@ -19,9 +18,9 @@ def rename_table(name, schema, engine, rename_to):
...
 
 def alter_table(table_name, table_oid, schema, engine, update_data):
-    if 'name' in update_data and 'sa_columns' in update_data:
+    if 'name' in update_data and 'columns' in update_data:
         raise ValueError('Only name or columns can be passed in, not both.')
     if 'name' in update_data:
         rename_table(table_name, schema, engine, update_data['name'])
-    if 'sa_columns' in update_data:
-        batch_update_columns(table_oid, engine, update_data['sa_columns'])
+    if 'columns' in update_data:
```
mathemancer, line 25: Why change away from `sa_columns`?

mathesar/models.py
```diff
@@ -309,6 +322,31 @@ def add_constraint(self, constraint_type, columns, name=None):
...
+
+    def __getattribute__(self, name: str) -> Any:
+        try:
+            return super().__getattribute__(name)
+        except AttributeError:
+            return getattr(self._sa_column, name)
+
+    @cached_property
+    def _sa_column(self):
+        return self.table.sa_columns[self.name]
```
mathemancer, line 341: Won't this error if the underlying DB table is deleted?

mathesar/models.py
```diff
@@ -309,6 +322,31 @@ def add_constraint(self, constraint_type, columns, name=None):
...
+
+    @property
+    def name(self):
+        return get_column_name_from_attnum(
+            self.table.oid, self.attnum, self.table.schema._sa_engine
+        )
```
mathemancer, line 347: Same

mathemancer (changes requested):
Looks great overall, but I have a couple of notes:
- Please consider whether we can remove the column index concept entirely at this juncture.  If it's not too difficult to do, it'd be great to get that out of the way.
- I think we'll need logic to re-reflect the DB objects if we go looking for a column and it's missing, or the associated table or schema is missing in the DB layer.  (Perhaps this logic is there and I missed it; if that's the case, let me know.)

db/tables/operations/alter.py
silentninja: DRF serializers use the underlying source field as the key name. Since the serializer `columns` field refers to the `columns` field of the table model and not the `sa_column` field of the model, this change is needed.

db/columns/operations/select.py
silentninja: I would prefer to do that in a different PR, once we get the interface right refactoring should be easier.

kgodey:
> Please consider whether we can remove the column index concept entirely

@mathemancer what do you mean by removing it entirely? I think we need to return the index in the API so that the frontend knows what order to display columns in, among other reasons.

silentninja:
> * I think we'll need logic to re-reflect the DB objects if we go looking for a column and it's missing, or the associated table or schema is missing in the DB layer.  (Perhaps this logic is there and I missed it; if that's the case, let me know.)

@mathemancer [Columns are reflected along with tables](https://github.com/centerofci/mathesar/blob/eb243b4f9bfab4c7c49ef36407c7fa13dfe3a78b/mathesar/reflection.py#L74) and only the database reflected columns are passed to the queryset . I didn't account for missing tables, I assume you are referring to [this error](https://github.com/centerofci/mathesar/blob/eb243b4f9bfab4c7c49ef36407c7fa13dfe3a78b/mathesar/models.py#L189)

silentninja:
>  ~~I didn't account for missing tables, I assume you are referring to [this error](https://github.com/centerofci/mathesar/blob/eb243b4f9bfab4c7c49ef36407c7fa13dfe3a78b/mathesar/models.py#L189)~~

Since the Django columns are fetched using SQL alchemy columns, it should not introduce any synchronisation problems which didn't exist before. But I can see a lot of edge cases that would still be present irrespective of the column model(for example when a column is deleted when a request is still being processed and the column reference is still passed around)

mathemancer:
> @mathemancer what do you mean by removing it entirely? I think we need to return the index in the API so that the frontend knows what order to display columns in, among other reasons.

We could still return the index in the API if needed, but I'd prefer ordering the columns by `attnum`, from least to greatest.  As for other places where we use the index (e.g., for finding a column in the SA table), it would be more robust to find it via name calculated from `attnum` using the `pg_attribute` table.  I suppose I'm not against returning it to the front end, but we already have bugs in the current code base due to the fact that the index is unstable and calculated from `attnum` through somewhat arcane means.

kgodey:
@silentninja please re-request review from Brent and myself when this is ready for re-review.

db/columns/operations/select.py
kgodey: Please open a separate issue for it if you're doing it in a different PR (just so we don't forget about it).

mathesar/models.py
kgodey: @silentninja This question appears unaddressed.

kgodey (approved):
@silentninja This looks good to me. I'll leave it to @mathemancer to review and merge.

mathesar/models.py
mathemancer: I can't seem to get the actual API to error in this case, so I'll mark it resolved.

mathemancer (approved):
I think this looks good.

PR #816, opened 2021-11-11
Fix url sync and reorganize Tab store

### LINKED ISSUE #651: Refactor frontend stores for table-data

pavish:
Fixes #651 

**Technical details**
* Fixes broken sync between open tabs and URL. Copying and pasting the URL will retain state for below configurations:
    - Open and active tabs
    - Pagination
    - Filters
    - Sort
    - Groups
* Handles url sync only from the tab context and not the table context
* Reogranizes tabular data base exports with a TabularData class and a manager
* Reorganizes Tab store
  - Adds tabularData as part of tab
* Creates an EventHandler base class
  - Columns and TabularData classes extend this

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/table-data/columns.ts
```diff
@@ -67,7 +68,7 @@ function api(url: string) {
   };
 }
 
-export class ColumnsDataStore implements Writable<ColumnsData> {
+export class ColumnsDataStore extends EventHandler implements Writable<ColumnsData> {
```
seancolsen, line 71: Small note...

Generally I lean towards using composition over inheritance as much as possible. So for a situation like this I'd be inclined to add an `eventHandler` property on `ColumnsDataStore` instead of having `ColumnsDataStore` inherit from `EventHandler`.  That said, I recognize there are situations where inheritance is a better fit. I don't know this code deeply enough to have a well-formed opinion of whether inheritance is best for this case but just wanted to mention this opinion generally. We could always refactor this to use composition later on if needed.

seancolsen (approved):
This looks really good. I'm happy to see all the store refactoring here.

I played with these changes quite a bit and couldn't find a way to break them. My reading of the diff is a bit more cursory, given that a lot of the code changed is code I still haven't yet familiarized myself with. But I feel confident-enough in these changes that I think it's ready to merge when the tests pass.

mathesar_ui/src/stores/table-data/columns.ts
pavish: > Generally I lean towards using composition over inheritance as much as possible

This is also my preferred approach when there is a clear separation of concerns.

> I'd be inclined to add an `eventHandler` property on `ColumnsDataStore`

Composition would not be a good idea in this particular case.

If we have an `eventHandler` property, we would have to expose it. So we'd do `columnsDataStore.eventHandler.on('event', value);`. However, this also leads to exposing properties we wouldn't want to expose, such as `eventHandler.dispatch` and `eventHandler.destroy`.

The other possible way is to make `eventHandler` private, and have a wrapper method for `eventHander.on` on the `ColumnsDataStore` class. This would be boilerplate code we'd have to add in all the classes have need an event handler.

Inheritance works perfectly for such situations. We can decide this on a case by case basis and I don't think we should always lean towards a particular approach.

PR #821, opened 2021-11-12
Set up toast system

### LINKED ISSUE #758: Create toast system
We need a way for the front end to display toast messages.

For example, if an AJAX API request fails, throwing a toast message with info about the failure is often the easiest way to support graceful failure. We might want to display success toast messages in some cases too.

Here's a library that seems to be popular: [svelte-toast](https://github.com/zerodevx/svelte-toast). I'd lean towards rolling our own implementation, but drawing on one or two popular libraries for inspiration.

seancolsen:
Fixes #758

## Demos

- [Toast system in Mathesar](https://www.loom.com/share/9ec323c9235b433d9762c504238128d3)
- [Toast system in Storybook](https://www.loom.com/share/7cc062415e204fdf82a6eb6b426eba97) -- _Loom cut me off at 5 minutes, but I was about done anyway_
- [Code overview - controller](https://www.loom.com/share/3084cdac8b7a4887b108646055f8d56e)

## Features

- Each message:
    - [x] has an optional icon
    - [x] has separate `title` and `message` props, both optional
    - [x] has a progress indicator
      - [x] On by default but can be turned off.
      - [x] Shows a countdown by default but can be manually controlled too.
      - [x] Pauses when the user hovers on the toast message. Pause can be disabled.
    - [x] a close button, on by default but can be turned off
    - [x] can be closed via code from inside and outside the message

- Support for:
    - [x] toast message as a progress indicator
    - [x] toast message as a loading spinner
    - [x] updating toast content after display
    - [x] defining a custom component for all toast items
    
- Customization:
    - [x] ToastController can be imported or passed down through context.
    - [x] Default toast message props can be set for all messages.
    - [x] Rich text in toast messages through components. (No `{@html}` within the toast message)
    - [x] The `makeToast` function can be re-implemented to support custom toast message types.

## Review tips

- Play with "Systems/Toast" in Storybook.
- Read `interface ToastEntryProps` to get a sense of the DX for displaying a toast message.
- See `DefaultOptions.svelte` for the code location within Mathesar that now has toast.

## Notes

- This PR only integrates the toast system into the app to the extent necessary to easily test it (by toggling the "Allow Duplicates" setting for a column.) I'll integrate it more deeply in a subsequent PR.
- I copied some of the code from [svelte-toast](https://github.com/zerodevx/svelte-toast) with attribution in `src/component-library/toast/README.md`. Their license is ISC, which is [GPL compatible](http://www.gnu.org/licenses/license-list.html#ISC).

## Issues

- Within the app, the line height in the toast message is a bit short for my taste. I didn't spend time attempting to address it because it's coming from CSS defined on `body`. I think we should address that separately along with a global style/typography spruce-up.
- Documentation could be better. I'm not sure how we're planning to document our component library, so I didn't do much here. The TS types and Storybook stories are the main documentation at this point.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
@seancolsen I took a high level look at the PR since it's still a draft, and I have a few opinions regarding the requirements.
1. The id for each toast message should be allowed to be passed as an optional parameter. The developer can then use the id to update the same toast instead of creating a new one. toast.show should return the id.
    - We may want to show loading and success/failure message on the same toast. We wouldn't want to destroy the old one and create new or show both at the same time.
2. The toast system should have an option to toggle autoHide.
    - We will be using it to show loading states, which should not hide automatically.
3. The ToastPresenter should allow a slot which can be used to customize the toast item. We could also one have within the toast item for further customization.
4. The styles can be rewritten in scss in dedicated files. This would be in the component library and we would want to build styles separate from the component itself.
5.  The ToastController show allow manually removing a toast by using the id.

seancolsen:
@pavish 

> update the same toast instead of creating a new one

Yeah that's a good point. I'll make sure to build a good way to handle that use-case.

> toggle autoHide

Yeah, that _is_ possible with the code currently, though it's slightly cumbersome. I'm planning to improve this a bit.

> ToastPresenter should allow a slot which can be used to customize the toast item

I like this idea. I'll work on that too.

> The styles can be rewritten in scss

Yep, I just haven't done that yet. CSS is almost entirely copy-paste from svelte-toast

seancolsen:
@pavish This is ready for review now.

One follow-up note from our previous exchange: for the `ToastPresenter` component, I ended up using a prop for the toast component instead of a slot. It felt a bit simpler and more intuitive that way. I can change it to a slot if you prefer.

pavish:
> I copied some of the code from svelte-toast with attribution in src/component-library/toast/README.md. Their license is ISC, which is GPL compatible.

@seancolsen Please include the license of svelte-toast within our /LICENCES folder.

So far we've included the attribution within each individual derivative file. I think we can follow the same and not have this README file, but I am okay with it either way. @kgodey Any thoughts around this?

mathesar_ui/src/App.svelte
```diff
@@ -3,8 +3,12 @@
   import Base from '@mathesar/sections/Base.svelte';
   import Schemas from '@mathesar/pages/schemas/Schemas.svelte';
   import Header from './header/Header.svelte';
+  import { ToastPresenter } from './component-library';
```
pavish, line 6: It would be better to import this using the alias.

mathesar_ui/src/component-library/common/utils/pauseableTweened.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  function set(value: number, opts?: TweenedOptions<number>): Promise<void> {
+    targetValue = value;
+    duration = opts?.duration ?? fullOptions.duration;
+    const promise = makePromise();
+    void store.update((_targetValue, _currentValue) => {
+      startingValue = _currentValue;
+      return value;
+    }, opts).then(promiseResolve);
+    return promise;
```
pavish, line 54: Is there a reason why we create a new promise that sets `promiseResolve` instead of directly returning the promise which is returned from `store.update`?

mathesar_ui/src/component-library/toast/ToastController.ts
```diff
@@ -0,0 +1,230 @@
...
+  faSpinner,
+} from '@fortawesome/free-solid-svg-icons';
+import type { IconDefinition } from '@fortawesome/free-solid-svg-icons';
+import type { SvelteComponent } from 'svelte';
+import { linear } from 'svelte/easing';
+import type { Writable, Readable } from 'svelte/store';
+import { writable, derived } from 'svelte/store';
+import type { PauseableTweened } from '../common/utils/pauseableTweened';
+import { pauseableTweened } from '../common/utils/pauseableTweened';
+import type { IconFlip, IconRotate } from '../icon/Icon.d';
```
pavish, line 13: The imports need some cleanup with respect to using alias. We need to consistently use it for the component-library and for any relative path that backtracks to a parent directory.

mathesar_ui/src/component-library/toast/ToastController.ts
```diff
@@ -0,0 +1,230 @@
...
+  message?: Readable<string> | string,
+  /**
+   * If provided, will be used in place of `title` and `message`.
+   */
+  contentComponent?: typeof SvelteComponent,
+  contentComponentProps?: Readable<unknown> | unknown,
+  icon?: Readable<Icon> | Icon,
+  backgroundColor: Readable<string> | string,
+  textColor: Readable<string> | string,
+  progressColor: Readable<string> | string,
```
pavish, line 47: Can we remove these and replace them with a prop that sets css classes on the wrapper element?

I am not a fan of programmatically styling elements unless warranted. I would also like to avoid inline styling as much as possible. In this case, using css classes would enable us to customize css separately.

It would be essential for us to be able to create a separate css library, for theming and for reusing styles in other components such as inline alerts.

mathesar_ui/src/component-library/toast/ToastItem.svelte
```diff
@@ -0,0 +1,70 @@
...
+  
+  export let entry: ToastEntry;
+
+  $: ({ props, controller } = entry);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  $: ({ progress, dismiss } = controller);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  $: ({ pause, resume } = progress);
+
+  $: readableTitle = ensureReadable(props.title);
```
pavish, line 15: It would be better to replace `ensureReadable` with something like `getFromReadable` (we need a better name).

`ensureReadable` creates a store if it's not a store already. I like having this util method but I don't see value in using it here. It only adds time & space complexity of creating of multiple stores for a usecase which only requires the value of the stores.

Instead, `getFromReadable` would return the value of the store if a store is passed, or would return the value directly if the passed argument is not a store.

mathesar_ui/src/component-library/toast/ToastPresenter.svelte
```diff
@@ -0,0 +1,23 @@
...
+  import ToastItem from './ToastItem.svelte';
+
+  export let entries: Readable<ToastEntry[]>;
+  export let toastItemComponent: typeof SvelteComponent = ToastItem;
+</script>
+  
+<ul class="toast-presenter">
+  {#each $entries as entry (entry.id)}
+    <li in:fly={{ x: 256 }} out:fade animate:flip={{ duration: 200 }}>
+      <svelte:component this={toastItemComponent} {entry} />
```
pavish, line 16: I'm okay with passing down a component instead of slot.

Users may want to provide additional props for the custom component. It might be better to pass them down with:
 ```suggestion
       <svelte:component this={toastItemComponent} {...$$restProps} {entry}/>
```

restProps isn't recommended unless we do not know what the props could be, which is true in this case.

**Review by pavish — CHANGES_REQUESTED** [review]
@seancolsen This PR looks neat! The toast system looks great and the stories are well detailed :)

I have a few comments. Once resolved, we can merge this PR.

File: mathesar_ui/src/component-library/toast/ToastController.ts (hunk 3 above)
Comment (seancolsen) [reply mathesar_ui/src/component-library/toast/ToastController.ts:13]: For line 12, is this what you had in-mind?

```ts
import { pauseableTweened } from '@mathesar-component-library';
```

With that import I get the following linting error

> Dependency cycle detected

What's the proper way to resolve this error?

mathesar_ui/src/component-library/common/utils/pauseableTweened.ts
seancolsen: Yeah good question.

This `PauseableTweened` store was a little tricky to create. The code would probably be more straightforward if I were to write it from scratch. Instead, I built it on top of Svelte's `Tweened` store because that seemed faster, if albeit somewhat more hacky.

In `ToastController.show` we create a `PauseableTweened` store for `controller.progress` and then call:

```ts
controller.progress.set(props.finalProgress, { duration: props.lifetime })
  .then(controller.dismiss);
```

For that code to work correctly, the Promise retuned by `PauseableTweened.set` needs to resolve when the timer runs out -- even if the user has paused and resumed the timer. The problem is that, with the way `Tweened` works, the only mechanism I have to pause the timer is to issue second `Tweened.set` call (in addition to the first `Tweened.set` that initiated the timer). Onece a second `Tweened.set` is issued to pause the timer, the Promise returned from the first `Tweened.set` will never resolve, meaning that the `controller.dismiss` function will never run.

Let's say we changed `PauseableTweened.set` to the following (as I gather you're suggesting):

```ts
async function set(value: number, opts?: TweenedOptions<number>): Promise<void> {
  targetValue = value;
  duration = opts?.duration ?? fullOptions.duration;
  return store.update((_targetValue, _currentValue) => {
    startingValue = _currentValue;
    return value;
  }, opts);
}
```

With that code, the toast message would work fine in most cases, but if the user hovers on the message to pause it, then the toast message will not automatically close itself when the timer runs out afterwards.

If you can think of a better way to structure this code, I'm all for it. Or I could add some code comments atop `PauseableTweened.set` which explain this a bit.

mathesar_ui/src/component-library/toast/ToastPresenter.svelte
seancolsen: Addressed in 1035b68b9d52ac22e47d6070e3dc8e8acbe30628

mathesar_ui/src/component-library/toast/ToastItem.svelte
seancolsen: I'm not really seeing how your `getFromReadable` function could work. I made the changes that I _think_ you're describing, and I put them in the branch `758_toast__getFromReadable`. Now when I go to Storybook > Systems > Toast > "High Level" > "Interactive" > "Dynamic content", I'm no longer able to update the toast content. That makes sense because the `props.message` store _object_ is not changing so the reactive statement `$: readableMessage = getFromReadable(props.message);` will not get triggered. The store _value_ is changing, and we can react to those changes via a subscription. But we can't subscribe to it unless we know it's already a store. If you're seeing a way to accomplish this, then I'd love to learn.

Further, I don't think the toast system warrants the level of performance optimization that you're seeking here. From what you've already taught me about Svelte stores, I understand that creating a new store incurs some performance costs. Here we are creating 7 additional stores per toast message. How much would you predict that those 7 additional stores would increase the total time to display one toast message? My guess is that it's not more than 2x -- meaning that in the worst case scenario, the additional _could_ double the time it would take to display one toast message, but not any more. And I can't imagine a scenario in which we'd ever be creating more than a few toast messages per second. So in a pragmatic sense, I don't see a performance issue here. Am I missing something?

It's worth noting that we _could_ eliminate the "Dynamic content" feature altogether. We'd still be able to update the content of a toast message, we'd just need to throw up a whole new toast message each time we make an update (using the same `id` of course). That's more cumbersome but still totally possible, and "dynamic content" is probably a rare use-case anyway. I'm fine with ripping this part out of the toast system. We wouldn't have the extra stores, and it would simplify things a bit.

All that said, I still really am curious if there's actually a way to implement the `getFromReadable` function you mentioned. Even if we eliminate the "Dynamic content" feature, I'd still like to know, just for the purpose of learning.

mathesar_ui/src/component-library/toast/ToastController.ts
seancolsen: I'd like to continue discussing this conversation once we've decided on an approach to the `getFromReadable` conversation below.

mathesar_ui/src/component-library/toast/ToastItem.svelte
pavish: I didn't think it through for Dynamic content. It would be essential to go with `ensureReadable` for it to work, since the compiler can only ensure store reactivity within the component based on appended the `$` symbol.

mathesar_ui/src/component-library/toast/ToastController.ts
pavish: This error would only occur when there is a circular dependency link between two files, either directly or indirectly. eg., File A depends on File B, which inturn depends on File A.

In this case, I assume it's because there is a chain cause by one of the imports, probably caused by using the:
```
import { ... } from '@mathesar-component-library'
```
alias somewhere. This is an alias to an index.ts file which inturn imports all the components and utilities from within the library. So using this alias within the library is certain cause for this error which would also result in build failure.

Instead, all the files within the library folder should import each other by using the:
```
import File from '@mathesar-component-library-dir'
```
alias.

The `mathesar-component-library-dir` alias is only for use within the component library. `mathesar-component-library` alias is only for use outside the component library.

mathesar_ui/src/component-library/common/utils/pauseableTweened.ts
pavish: Sounds good. I took a second look with the reason in mind and the usecase makes sense. I think it would be better to add some comments.

mathesar_ui/src/component-library/toast/ToastController.ts
seancolsen: @pavish I put some changes in the `758_toast__imports` branch that I think are in line with what you're requesting. However, with those changes I'm unable to start Storybook. Can you take a look and help me resolve the errors? 

It's also worth noting that I see lots of files inside `src/component-library` which contain imports using the `@mathesar-component-library` alias, thereby creating a circular import. It seems to me that the linter just doesn't notice because those imports are in `.svelte` files.

```
grep -Pr '@mathesar-component-library(?!-)' src/component-library
```

mathesar_ui/src/component-library/common/utils/pauseableTweened.ts
seancolsen: I added comments in e4c0fbf63d4249d3fcca7ad9c724bc3ebd63bbf7 that I hope should resolve this

mathesar_ui/src/component-library/toast/ToastController.ts
pavish: @seancolsen This commit https://github.com/centerofci/mathesar/pull/821/commits/2dc2113b688ebec0504f1c7b3c2a57f93d0a02de should fix the alias resolution.

Storybook should now work when we import with `@mathesar-component-library-dir`

seancolsen: I cherry-picked that commit onto this branch now as b65f6053f9f56ef74323df47be69c6ec14478444, which I think should resolve this issue

mathesar_ui/src/component-library/toast/ToastController.ts
pavish: As discussed over private chat, this can be taken up in a subsequent PR

pavish (approved):

PR #837, opened 2021-11-22
Backend filtering refactor

### LINKED ISSUE #385: Implement filtering options for Number types.

### LINKED ISSUE #846: Duplicates filter should work with other filters
## Problem

## Proposed solution

## Additional context

dmos62:
Fixes #385, fixes #846

Sets up the backend infrastructure for filtering.

**Technical details**

Uses our fork of [`sqlalchemy-filters`](https://github.com/centerofci/sqlalchemy-filters).

## Checklist
 
  index.md`).
  visible errors.

## My TODO
- [x] Use Python classes to model filters
- [x] Have a way to convert Python filter representation to SQLAlchemy filter specification
- [x] Have a way to generate filter specification for the frontend
- [x] Find out if/how possible predicates should change depending on column's type
- [x] Expose possible predicates through the REST API
- [x] Start discussion about new filter API
- [x] Accept a filter specification through REST API
- [x] Apply the requested filter, via SQLAlchemy filters, to the table query
- [x] Integrate any input from the [discussion](https://github.com/centerofci/mathesar/discussions/848)
- [x] Resolve conflict with the duplicate-only filter
- [x] In-code documentation
- [x] Describe breaking-changes (as relevant to frontend)
- [x] Reintroduce duplicate-only filter as a dedicated parameter query
- [ ] Port the recent grouping PR's grouping tests to this PR (this PR invalidates many grouping tests)

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@dmos62 I see that you're working off of a fork still, you have write access to the repo now so I recommend switching to a branch instead.

dmos62:
At the moment possible predicates are exposed in the REST API like this:

`http://localhost:8000/api/v0/databases/1/types/`:

```
[
    {
        "identifier": "boolean",
        "name": "Boolean",
        "db_types": [
            "BOOLEAN"
        ],
        "filters": [
            {
                "superType": "leaf",
                "type": "equal",
                "parameterCount": "single",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "leaf",
                "type": "greater",
                "parameterCount": "single",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "leaf",
                "type": "greater_or_equal",
                "parameterCount": "single",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "leaf",
                "type": "lesser",
                "parameterCount": "single",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "leaf",
                "type": "lesser_or_equal",
                "parameterCount": "single",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "leaf",
                "type": "empty",
                "parameterCount": "none"
            },
            {
                "superType": "leaf",
                "type": "in",
                "parameterCount": "multi",
                "parameterMathesarType": "boolean"
            },
            {
                "superType": "branch",
                "type": "not",
                "parameterCount": "single"
            },
            {
                "superType": "branch",
                "type": "and",
                "parameterCount": "multi"
            },
            {
                "superType": "branch",
                "type": "or",
                "parameterCount": "multi"
            }
        ]
    },
    ...
]
```

kgodey:
Couple of quick comments:
- I don't think we should mix snake case and camel case in key names. I think we should stick with snake case since that's what the rest of the API uses.
- I'm not sure what `superType` means.

dmos62:
@kgodey thanks for noticing the casing conflict. I'm using the tree abstraction for predicates. The `empty` predicate's `superType` is `leaf`, because it's always a leaf node on the predicate tree (has height zero). `branch` predicates are never leaves (have height that's non-zero), like `and`, `or` or `not`.

kgodey:
>  I'm using the tree abstraction for predicates. The `empty` predicate's `superType` is `leaf`, because it's always a leaf node on the predicate tree (has height zero). `branch` predicates are never leaves (have height that's non-zero), like `and`, `or` or `not`.

I figured that out, I meant specifically that the "super type" nomenclature is a little confusing, if I was just paying attention to the key names, seems like it's a superset of the "type" key somehow (which it's not). Is there a more obvious name for it?

db/filters/base.py
```diff
@@ -169,4 +158,47 @@ def getSAFilterSpecFromPredicate(pred: Predicate) -> dict:
...
+
+def takesParameterThatsAMathesarType(predicateSubClass: Type[Predicate]) -> bool:
+    return (
+        issubclass(predicateSubClass, Leaf)
+        and not issubclass(predicateSubClass, NoParameter)
+    )
+
+def not_empty(l): return len(l) > 0
+
+def assertPredicateCorrect(predicate):
```
kgodey, line 171: @dmos62 we use snake case for Python functions and variables too. We only use camel case (capitalized) for class names.

dmos62:
`superType` is a superset in that every Predicate subclass (every `type` in other words) is also a subclass of one of the `superType`s. Or do you mean something else?

I'm open to suggestions. Calling it a parent type would have a similar meaning. That's pretty much talking about the underlying class/mixin hierarchy. We could also have nomenclature that talks about predicate names (e.g. empty, greater, and) and positions in predicate trees (leaf or branch).

kgodey:
>  Calling it a parent type would have a similar meaning. That's pretty much talking about the underlying class/mixin hierarchy. 

API users will probably not know or care about how it's implemented, I think the nomenclature should prioritize API readability.

> We could also have nomenclature that talks about predicate names (e.g. empty, greater, and) and positions in predicate trees (leaf or branch).

I like this. How about `name` instead of `type` and `position` instead of `super_type`?

db/filters/base.py
dmos62: Fixed it. Thanks for pointing that out.

dmos62:
Currently blocked by the duplicate-only filter's conflict with the new API https://github.com/centerofci/mathesar/issues/846#issuecomment-981124435

dmos62:
Current progress and dicussion: https://github.com/centerofci/mathesar/discussions/848

dmos62:
Made significant changes:

- Introduced dedicated endpoint for database-specific filtering options (discussed in #851);
- Improved user-facing nomenclature (discussed in #848);
- Included `not-` variants of `leaf` predicates (to make it easier on frontend developers);
- Wrote and rewrote a bunch of tests.

This PR is currently blocked by @mathemancer's grouping PR (#862), since our test suites have some overlap between testing filtering and testing grouping.

dmos62:
PR is ready for review. Failing tests will not be fixed at this time (see end of this comment https://github.com/centerofci/mathesar/pull/837#issuecomment-985593221).

kgodey:
Assigned everyone on the backend team to review since this is a major change. cc @mathemancer @silentninja

dmos62:
This PR introduces breaking changes to how frontend specifies filtering (the filtering spec): the [deserialization test](https://github.com/centerofci/mathesar/blob/0d5a4c83f9280de935113e39de82ef690e63a687/db/tests/filters/operations/test_deserialize.py) should provide a good illustration:

```python
ma_filter_spec = {"and": [
    {"not":
        {"empty": {"column": "col1"}}},
    {"or": [
        {"equal": {"column": "col2", "parameter": 15}},
        {"in": {"column": "col3", "parameters": [1, 2, 3]}},
    ]}]}

expected_predicate = And([
    Not(
        Empty(column="col1")
    ),
    Or([
        Equal(column="col2", parameter=15),
        In(column="col3", parameters=[1, 2, 3]),
    ]),
])
```

Notice that the new specification format is strict. For example, `Not` can only accept a single value, not a list; `Or`, `And`, `In`, `NotIn` can only accept a list. If it's a leaf predicate, and it accepts a list of parameters, the relevant JSON field will always be `parameters` (plural), or `parameter` (singular) otherwise.

Field name constraints are enforced by the underlying dataclass hierarchy. The rest of the constraints are enforced on instantiation by this method: https://github.com/centerofci/mathesar/blob/0d5a4c83f9280de935113e39de82ef690e63a687/db/filters/base.py#L266

It checks things like whether a parameter is a predicate (in the case of branch predicates), whether the parameters are unique, whether parameters are not `None`, etc.

seancolsen:
@dmos62  This is very cool!

- Filtering spec looks great
- `filters` endpoint
    - It's great having this endpoint!
    - I like the "leaf" and "branch" distinction a lot!
    - It strikes me that the following filter types are extraneous: `not_equal`, `greater_or_equal`, `lesser_or_equal`, `not_empty`, `not_in`. We should be able to achieve their behavior by using combinations of other more primitive filters. For example, a `not_equal` is a combination of `not` and `equal`. I would consider removing these extraneous filters and keeping only the most essential ones at the API layer. If the purpose of having these filter types is primarily a UX consideration (i.e. we want to give the user a quick option to set a "not equal"), then I think it would be better to satisfy that need at the UI layer.
- Front end work
    - I'm happy to implement the front end changes for this PR. But it would be nice to check in with Pavish first if possible since he's been working on some similar stuff. He's out this week, so if we're in a hurry to get this PR merged then I can forge ahead and we'll sort out any of the front end confusion after Pavish returns.

dmos62:
@seancolsen  

> if we're in a hurry to get this PR merged then I can forge ahead

I don't get the feeling that this is very urgent, especially since this is waiting on the recent grouping PR to be finalized.

> the following filter types are extraneous: not_equal, greater_or_equal, lesser_or_equal, not_empty, not_in.

I completely agree and I initially omitted them. I put them back in because I was concerned that them being left out would require more frontend and UX changes. I think that in the long run we should focus on making composition of primitives easy. But, for the moment we can go either way, whatever makes more sense for the frontend and UX.

kgodey:
@dmos62 I see your todo list is not yet complete. Please mark your PR as a draft until it is so that it doesn't get accidentally merged.

dmos62:
Since @mathemancer is still making changes on the dependency PR (#862), I'll hold off on merging it into this one.

db/filters/base.py
```diff
@@ -0,0 +1,302 @@
...
+    def saId(self) -> str:
+        return get_SA_id_from_predicate_type(self.type)
+
+    def __post_init__(self):
+        assert_predicate_correct(self)
+
+
+@frozen_dataclass
+class Leaf(Predicate):
+    super_type: PredicateSuperType = static(PredicateSuperType.LEAF)
```
kgodey, line 105: Can we rename `super_type` and `type` here to match the API? `name` and `position`

kgodey (commented):
I'll review this more detail later.

db/filters/base.py
dmos62: Sure. It's actually `identifier` and `position`. I'd prefer to make the change in the latest filtering PR. Does that work?

kgodey: Sure, that's fine. It shouldn't affect the API, right?

silentninja:
Have we decided on using dataclasses and typing in our codebase?

mathemancer (commented):
First pass review.  I'll be more precise once this is more stabilized.  However, I have some broader comments to make.  Overall, I really like the tree concept.  I do think there's a false dichotomy introduced; there _are_ branches and leaves, but you can't assume that an "and" is a branch (since it could be an "and" between two boolean columns).  I.e., the dichotomy isn't between types exactly.

My biggest concern is the specificity.  I really think we need to take a try and see what happens approach to some of these things, rather than trying to check everything beforehand.  Let the DB tell you if a given proposition makes no sense, and handle the error.  This will be more flexible, and avoid having to define things in multiple places.  Long run, I really think it'll be easier to maintain that way.  For example, avoiding specifying branch vs. leaf is more flexible for composition.  I acknowledge that it would sometimes run into issues, but that can be handled by really good feedback and errors.

Final note for this round:  we need to come to some kind of team-wide agreement about type hints (and by implication dataclasses).  I'm ambivalent on these issues, but I suspect I'm the only one.

kgodey:
> Have we decided on using dataclasses and typing in our codebase?
>
> Final note for this round: we need to come to some kind of team-wide agreement about type hints (and by implication dataclasses). I'm ambivalent on these issues, but I suspect I'm the only one.

I think we should take the discussion about dataclasses and type hints to a GitHub discussion so as to not clutter up this PR. @dmos62, I think it would make sense for you to start this.

kgodey:
@dmos62 This is a large PR. I think it would help me review the code if you could do a write up of the changes. Topics I think would be useful:
- Explaining the code structure and why you chose it + any alternatives you considered.
- Extensibility - how to add new filters and new data types
- The benefits of dataclasses in this particular application

I'm having a hard time grokking the code because it doesn't seem Pythonic somehow. That's probably not useful but I can't articulate any more specifics, I'm hoping reading through your explanation will help me either understand the code or articulate why it's hard for me to understand.

mathesar/api/serializers/records.py
```diff
@@ -2,11 +2,25 @@
...
 class RecordListParameterSerializer(serializers.Serializer):
-    filters = serializers.JSONField(required=False, default=[])
+    filters = serializers.JSONField(required=False, default=None)
     order_by = serializers.JSONField(required=False, default=[])
-    group_count_by = serializers.JSONField(required=False, default=[])
+    duplicate_only = serializers.JSONField(required=False, default=None)
+    grouping = serializers.JSONField(required=False, default={})
+
+    def validate_duplicate_only(self, duplicate_only):
+        is_none = duplicate_only is None
```
kgodey, line 11: Why is this a separate variable?

mathesar/api/serializers/records.py
```diff
@@ -2,11 +2,25 @@
...
+        if not is_none:
```
kgodey, line 12:
```suggestion
        if duplicate_only is not None:
```

mathesar/api/serializers/records.py
```diff
@@ -2,11 +2,25 @@
...
+            is_a_list = type(duplicate_only) is list
+            if not is_a_list:
```
kgodey, line 14:
```suggestion
            if type(duplicate_only) is not list:
```

mathesar/api/serializers/records.py
kgodey, line 12:
```suggestion
        if duplicate_only is not None:
```

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+        for item2 in xs:
+            if item1 == item2:
+                times_seen += 1
+            # A non-duplicate will be seen only once.
+            if times_seen == 2:
+                return False
+    return True
+
+
+def assert_predicate_correct(predicate):
```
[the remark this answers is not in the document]
dmos62: These can be broken up and put on the class definitions. That would however distribute the correctness definition across the file. The reason methods like this on or the deserialization method are large and nested is because they use control flow to traverse the predicate class hierarchy. I'm realising that that's a put off. As I said, the alternative is to distribute the correctness/deserialization definition (which I'm seriously considering).

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+
+@frozen_dataclass
+class ReliesOnComparability:
+    """Some predicates require the data types they are being applied to to be comparable
+    (lesser, greater, etc.)."""
+    pass
+
+
+def relies_on_comparability(predicate_subclass: Type[Predicate]) -> bool:
+    return issubclass(predicate_subclass, ReliesOnComparability)
```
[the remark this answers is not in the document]
dmos62: If we know declare what type properties a predicate relies on and what properties a type has, we can check if they are compatible. Check [mathesar/database/types.py::is_ma_type_supported_by_predicate](https://github.com/centerofci/mathesar/pull/837/files#diff-619474cc459b0bbd2b079fd2d241913a6b74e77be0493a674a84ce3e8776824a).

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+@frozen_dataclass
+class MultiParameter:
+    parameter_count: ParameterCount = static(ParameterCount.MULTI)
+    parameters: List[Any]
+
+    @property
+    def parameter_list(self):
+        return self.parameters
```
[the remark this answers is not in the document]
dmos62: Yes. Goal is declarativity and strictness too.

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+    EQUAL = "equal"
+    NOT_EQUAL = "not_equal"
+    GREATER = "greater"
+    GREATER_OR_EQUAL = "greater_or_equal"
+    LESSER = "lesser"
+    LESSER_OR_EQUAL = "lesser_or_equal"
+    EMPTY = "empty"
+    NOT_EMPTY = "not_empty"
+    IN = "in"
+    NOT_IN = "not_in"
```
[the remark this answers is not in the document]
dmos62: I'm not sure what you mean by "specify each leaf every time". Do you mean when introducing a new predicate? A central premise in this structure is that we want to declare what correctness is as early as possible. In your final comment you mentioned that you're disputing that. Is this what you're referring to here?

db/filters/operations/deserialize.py
```diff
@@ -0,0 +1,40 @@
+from typing import Any
+from db.filters.base import (
+    Predicate, Leaf, SingleParameter, MultiParameter, NoParameter, Branch, get_predicate_subclass_by_type_str, BadFilterFormat
+)
+
+
+def get_predicate_from_MA_filter_spec(spec: dict) -> Predicate:
```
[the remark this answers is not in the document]
dmos62: I addressed this comment on a related comment: https://github.com/centerofci/mathesar/pull/837#discussion_r772892728

mathesar/api/serializers/records.py
dmos62: This change taxes the reader since he has to parse the boolean expression to tell what it checks. Are you optimizing for terseness?

dmos62: By the way, I tend to not declare negative booleans, since negation is a discreet cognitive step and treating it as a discreet modifier is more congruent with that, hence `not is_a_list` instead of `is_not_a_list`.

dmos62:
## Writeup

### Features

Notice that I might use the terms filter specification and predicate interchangeably. Some things this new predicate data structure does is:

- Declares (predicate) correctness (to a high degree)
    - An illegal predicate cannot be instantiated; an instantiated predicate is always legal
        - An exception
            - A predicate referencing unexisting columns can be instantiated
                - Referenced column existance is checked right before applying the filter specification to a query
- Declares what predicates the frontend can use under what circumstances and how
    - It can tell if a predicate is compatible with some column type based on its properties (like comparability, compatibility with `LIKE`, compatibility with URI-type-specific SQL functions)
    - It can tell the frontend how many parameters a predicate takes (e.g. `empty`, `equal` and `in` take different number of params) and it can tell it what options it accepts (e.g. should `starts_with` be case insensitive)
    - It can tell the frontend how to compose predicates: leaf vs branch predicates: parameters on branches are other predicates
- It supports SQL functions
    - You can use SQL functions to, for example, destructure a URI and filter based on that
       - Though this extension is/will be in a newer PR
       - Was not possible with `sqlalchemy-filters`

What it does not (currently) do:
- Does not support using other columns as parameters
    - Can't do `{equal: {column: x, parameter: {column: y}}}`
    - This is an oversight
    - Does not seem difficult to implement if/when there's interest

### Technical details

I organized the relevant pieces in the predicate data structure into a mixin hierarchy and also used this PR as a testbed for frozen dataclasses. Some of my objectives with the basic structure were:

- Immutability
  - Used properties where I could
     - Didn't use properties where an SQLAlchemy filter is returned, since its mutability is uncertain
- Many small classes
   - Use the mixin/type system to capture information
       - A class that is a predicate that takes one parameter and relies on `LIKE` will directly or indirectly mix in `ReliesOnLike`, `SingleParameter`, `Leaf` and `Predicate`
   - Some logic I offloaded to and hence centralized in static methods: correctness is declared on `db/filters/base::assert_predicate_correct`: and, what the JSON filter specification is is declared in `db/filters/operations/deserialize`
      - Has the drawback that the control flow in these methods can seem daunting, since it walks itself through the mixin hierarchy
          - These methods are not complicated, but big: each branch is simple: it's just that there's many of them
          - Might fan this logic out into the mixin definitions
- Used `parameter` where a singular parameter is expected and `parameters` where a sequence is expected
    - I have reservations about this
      - It makes the specifics of single/multi-parameter requirements obvious
      - Procedural instantiation is more verbose (like when testing), since you have to change the argument name depending on circumstance
          - But that can be solved with an auxillary constructor that's only for utilities

### How to extend with new predicate

1. Introduce the appropriate class; for example, `Greater`; this includes defining the new Predicate's properties through mixins: `ReliesOnComparability, SingleParameter, Leaf` in this case (note that mixin order matters, see Python docs);

```python
@frozen_dataclass
class Greater(ReliesOnComparability, SingleParameter, Leaf):
    type: LeafPredicateType = static(LeafPredicateType.GREATER)
    name: str = static("Greater")

    def to_sa_filter(self):
        return column(self.column) > self.parameter
```

2. Introduce the predicate type enum: `LeafPredicateType.GREATER` in this case; it's what the JSON filter spec will use to identify a predicate;

3. Tell it how to generate an equivalent SQLAlchemy filter by implementing the abstract `Predicate::to_sa_filter` (as seen above);
4. Update correctness definition (`db/filters/base::assert_predicate_correct`), if needed; currently this involves finding the spot in the method's control flow tree that corresponds to this new predicate, adding a new type check, etc.
5. Add new predicate to the `all_predicates` set;
6. Update `mathesar.database.types::is_ma_type_supported_by_predicate`, if needed; this would involve declaring what types have the properties the new predicate depends on: in this PR it's:
```python
def _is_ma_type_comparable(ma_type: MathesarTypeIdentifier) -> bool:
    return ma_type in comparable_mathesar_types

def is_ma_type_supported_by_predicate(ma_type: MathesarTypeIdentifier, predicate: Type[Predicate]):
    if relies_on_comparability(predicate):
        return _is_ma_type_comparable(ma_type)
    else:
        return True
```

### `dataclasses` and `typing`

As Kriti suggested, I'll start a dedicated discussion on the use of `dataclasses` and `typing`.

But, to summarize:
- `dataclasses` eliminated custom boilerplate; I got immutability, `_post_init` and defaults without writing a single constructor; I think this is great since it's essentially standardized boilerplate;
- `typing` got me partial typing, which is great:
  - in combination with my LSP server/client ([`pyright`](https://github.com/microsoft/pyright)): catches a lot of mistakes and conflicts
  - in that I can annotate when I think it's useful and not when I don't (_partial_ typing)
  - for readability, for example the above method `is_ma_type_supported_by_predicate` operates on uninstantiated predicate classes, not instances, so to express that I can just say `predicate: Type[Predicate])` instead of `predicate_subclass`
  - doesn't have a cost; another developer can just omit type or write `Any` if he prefers

mathesar/api/serializers/records.py
kgodey: I actually find that it's easier to read/parse this way. I don't have to keep track of a variable that's not used anywhere else and it's clear what the conditional is actually checking. The variable name `is_a_list` doesn't let me know _what_ is a list or not. 

One of my favorite things about Python specifically is that the syntax easy to read in terms of English sentences. You can see that with `if type(duplicate_only) is not list:` --> "if the type of duplicate_only is not a list, do blah". You lose that readability with having a separate variable.

dmos62: I can tell that our styles are clashing ⚔️ I'll try to make a strong case for my style, since I find that it suits me very well.

> The variable name is_a_list doesn't let me know what is a list or not.

A central idea behind doing things this way is that it should not. What a list is should be defined separately. For the sake of readability and reuse. It's about being declarative. We're offloading the burden of knowing what a list is (it's something that has the `type` equal to `list`) to a dedicated variable, and a dedicated cognitive step. This would be easier to illustrate with something more complicated, but if we were to change the definition of a list, we'd change the piece of code labeled `is_a_list`, while the `if` block would stay the same. My objective is to split every cognitive step up into discreet code objects. If it were something more complicated, like `type(x) is Sequence and x.__name__ == "dict" and len(x.keys) > 0` (pseudo code), this question would be less nuanced; but, even with a subtle case like `type(duplicate_only) is not list` the same principle applies: we shouldn't care what a list is unless it's relevant. If it's always relevant, then there's conflicting definitions for what a list is and that can be looked at.

This "split every cognitive step up into discreet code objects" principle improves my code reading/reviewing bandwidth significantly.

> `if type(duplicate_only) is not list`: --> "if the type of duplicate_only is not a list, do blah".

```
type(x) is not list:
```
reads as "type of x is not a list", but why do we care that that's `type(x)`'s value? Because, it means that it's a list. That's a subtle cognitive step that's left implicit.

```
not is_a_list
```
reads as "is not a list", and it leaves nothing implicit. Again, this would be more illustrative if the example was more complicated.

`not is_a_list` is a less specific statement. In this case, the type of `x` is an implementation detail that we don't care about. If the way to check whether it's a list were different nothing would change in the context of this method, only the declaration for what a list is, which `type(x) is not list` leaves implicit, and `not is_a_list` (together with declaration of `is_a_list`) makes explicit.

The principle smoothly extends into `not is_something(x)`, for when you want a higher-scope definition for what `something` is.

By the way, I'm surprised that you don't find that `not is_a_list` rolls off the tongue. I too find that being able to easily read code aloud is desirable.

PR #841, opened 2021-11-24
DRF test client request parameters

### LINKED ISSUE #813: Use Django rest framework client instead of django client when testing
## Problem
The test cases are using Django client which is much suited for emulating a browser, so the default request body is priority content type is multipart instead of json which is what we are using mostly on the frontend

## Proposed solution
DRF [test client](https://www.django-rest-framework.org/api-guide/testing/) should be used instead of django client in order to emulate an API call rather than a browser form, this avoids unnecessary json dumps when calling an API during testing

andruwwwka:
Fixes #813.

Use correct django-rest-framework test client parameters when sending json body payload

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@silentninja Assigning this to you for review, please feel free to pull in others if needed.

silentninja:
@andruwwwka Thanks for the PR. I think it would be better to specify `json` as the default request format and use `format` option for the requests that are multipart. @kgodey I would appreciate your comments on this

kgodey:
@silentninja That sounds fine to me, it would reduce boilerplate code.

andruwwwka:
@silentninja Hi! I update the PR.

mathesar/tests/api/test_record_api.py
```diff
@@ -427,10 +429,11 @@ def test_record_list_sort_exceptions(create_table, client, exception):
 def test_record_list_group_exceptions(create_table, client, exception):
     table_name = f"NASA Record List {exception.__name__}"
     table = create_table(table_name)
-    group_by = json.dumps(["Center"])
+    data = {"group_count_by": ["Center"]}
```
silentninja, line 432: This should be a json string.

silentninja (changes requested):
`get` data objects should be JSON string rather than the string representation from `f` string conversion as the backend expects a JSON object and `f` string outputs a urlencoded string which is not a valid JSON object. Moreover, test cases could fail if the object has nested objects as URL encoded objects won't have a clean conversion, so I think it is better to stick with json encoded string

andruwwwka:
@silentninja I have corrected the code according to your comment

silentninja (approved):

kgodey:
Thanks for your work on this, @andruwwwka!

PR #844, opened 2021-11-26
Initial implementation of frontend filtering by duplicates

### LINKED ISSUE #770: Add UI to filter on the presence of duplicate values in one column
## Problem

User wants to add a filter condition that shows duplicate values for one column.

## Proposed solution

image

## Additional context

- Backend issue: #501 

## Out of scope

- filtering on duplicates across multiple columns

yankevn:
Fixes #770 

The initial implementation of creating a frontend filter to filter by duplicates has been implemented here.
Because of the filtering by duplicates being a special case, the user is not allowed to have any filters past the filter by duplicates filter.
The code has also been updated to follow this behavior. Because this filtering is a special case, it has special handling in meta.ts which specifically checks for this filtering behavior.

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/table-data/meta.ts
```diff
@@ -218,9 +218,15 @@ export class Meta {
...
-          );
+          if ('and' in filter && filter.and.length > 0 && filter.and[0].op === 'get_duplicates') {
+            params.push(
+              `filters=${encodeURIComponent(JSON.stringify([filter.and[0]]))}`,
+            );
+          } else {
+            params.push(
+              `filters=${encodeURIComponent(JSON.stringify(filter))}`,
+            );
+          }
```
dmos62, line 229: Some in-code comments about why this is needed might be useful in the future.

pavish, line 229: This is the only region where I'd like the special handling to be present. Once the bug in the backend is fixed, it would be easier for us to revert this change.

pavish (changes requested):
@yankevn Thank you so much for the PR.

You've meticulously handled a series of cases that ensures the frontend works well when `get_duplicates` filter is added, which is amazing. :)

But considering that it is a bug on the backend, I'd rather not handle all these cases on the frontend and wait for the bug to be fixed.

Most of my comments are to revert some of the special handling done for this filter.

mathesar_ui/src/sections/table-view/display-options/FilterEntry.svelte
```diff
@@ -54,9 +56,11 @@
     <Select options={conditions} bind:value={condition}
       on:change={() => dispatch('reload')}/>
   </td>
-  <td class="value">
-    <TextInput bind:value={inputValue}/>
-  </td>
+  {#if !filterByDuplicates}
```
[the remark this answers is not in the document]
yankevn: I believe this change should be kept, as it keeps previous filter entries that chose get_duplicates as the operation consistent. Otherwise, the text box for the value will show up again.

dmos62:
We are making big changes to the `get_duplicates` filter on the backend (we started calling it `duplicates_only`, btw). We haven't finalized anything, but it's quite certain that we won't be using the `filters` query parameter for it. We might drop it entirely, otherwise it will use a dedicated query param `duplicates_only=[col1,col2,col3]`. [I would be partial to dropping it](https://github.com/centerofci/mathesar/issues/846#issuecomment-982840977), but since that would mean discarding these recent PRs, I think I'll implement that dedicated query param.

pavish:
@dmos62 I'm fine with either decision. Using a separate param will require us to rethink the UX entirely since it cannot be part of filters.

We'd still be merging this PR for now and only take the changes up when everything is concrete.

kgodey:
We can continue the discussion on #846, I agree that we should merge this PR and handle any changes based on backend changes separately.

silentninja:
Hey @yankevn! thanks making the requested changes to the PR. @pavish  is on leave until next week and won't be able to review the PR until then, let me know if you would like to contribute to some other issue in the meantime.

yankevn:
Thanks for letting me know. Currently, my build is passing everything except for the "shows mathesar default text when rendered" test. Is this test failing as a result of my code? I'm not sure what might be going wrong.

mathesar_ui/src/stores/table-data/meta.ts
```diff
@@ -211,16 +211,35 @@ export class Meta {
...
           }));
           filter[$filter.combination.id as string] = terms;
-          params.push(
-            `filters=${encodeURIComponent(JSON.stringify(filter))}`,
-          );
+          if ('and' in filter && filter.and.length === 1 && filter.and[0].op === 'get_duplicates') {
+            // Special handling for the get_duplicates operation due to backend bug
+            // The backend can only handle the get_duplicates operation by itself
+            // The "and" must be removed and the value must be in an array
+            filter.and[0].value = [filter.and[0].field];
```
silentninja, line 233: @yankevn This line seems to be causing the issue. `value` is of type `string` and not an `array of string`. @seancolsen Can you look into this?

yankevn: That makes sense. The backend currently requires that the value field is an `array of string` rather than `string`, which is why that is there. If I remove the brackets, the API call fails.

silentninja: Can the [`value` field](https://github.com/centerofci/mathesar/pull/844/files/653eef286526205f5aa791f1bf673a8001d442c2#diff-7bb52cec499a124f38598455ace9ea0fd1bde7e2861a44f3052d5dd94b9b40efR219) be an `array of string`?

yankevn: I didn't expect that to work, but it actually works just fine with our existing filter implementation. I'll push that right now!

mathesar_ui/src/sections/table-view/display-options/FilterEntry.svelte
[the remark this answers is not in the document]
pavish: Sounds good!

mathesar_ui/src/stores/table-data/meta.ts
pavish: This section is hacky but I don't think we can do anything about it until we resolve https://github.com/centerofci/mathesar/issues/846. So, I'm proceeding with merging these changes as is.

pavish (approved):
@yankevn Thanks for your work on this PR :)

I'm approving it and setting it to auto-merge.

PR #852, opened 2021-11-30
Add UI to rename tables

### LINKED ISSUE #447: Implement editing a table's name
## Problem

Users may want to rename tables.

## Proposed solution

The ["Basic Table Operations" design spec](https://wiki.mathesar.org/design/specs/table-operations) has a solution for this, which we need to implement on the frontend. This issue involves the following portions of the spec:
- _User Experience_: 
  - User edits an existing table name
 
## Additional context

- Backend work: #184
- Design issue: #185

seancolsen:
Fixes #447

## Before

No way to rename a table.

## After

Rename table via a modal accessible from the table actions menu.

## Demo

Peek 2021-12-02 15-41

## Additional features

- client side validation for simple error scenarios like a blank name and a name conflict
- auto focus and select, plus save on Enter key -- all to minimize keystrokes and mouse usage
- toast message on error (and modal remains open)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish if you have a minute, I'd love some help troubleshooting this failing test. The app works for me locally. And I can see the test failure locally too. I haven't used Testing Library before. Is there a way I can poke at it a bit to get some more info about why the test is failing when the app seems to load?

kgodey:
@seancolsen sorry, I clicked the wrong button, I didn't mean to change the draft status.

kgodey:
@seancolsen Sorry, I didn't mean to change the draft status, I clicked the wrong button. I'm not sure how to reverse it.

mathesar_ui/src/sections/table-view/actions-pane/RenameTableModal.svelte
```diff
@@ -0,0 +1,97 @@
...
+  export let tabularData: TabularData;
+
+  let allowClose = true;
+  let inputElement: HTMLInputElement;
+  let originalName = '';
+  let name = '';
+  let proceed: () => Promise<void>;
+
+  function schemaContainsTableName(_name: string): boolean {
+    return [...$tables.data.values()].map((t) => t.name).includes(_name);
```
seancolsen, line 22: This is the offending line that prompted me to set `'@typescript-eslint/no-unsafe-return': 'off',`

I'm a little surprised that eslint struggled here. But in general I want to move in the direction of turning off linting rules when they're getting in our way.

Here it seems like `svelte-check` should catch problems. For example, I tried changing the return statement to `return 7` and observed the following error from `svelte-check`:

> Error: Type 'number' is not assignable to type 'boolean'. (ts)

To me, that result from `svelte-check` says that we don't need eslint for this particular rule.

mathesar_ui/src/component-library/text-input/TextInput.svelte
```diff
@@ -25,9 +29,22 @@
       element?.focus();
     }
   }
+
+  function handleKeypress(e: KeyboardEvent) {
+    if (e.key === 'Enter') {
+      dispatch('enter');
+    }
+  }
```
pavish, line 37: I think it would be better to bubble up the keypress event directly rather than handling it within the TextInput component.

mathesar_ui/src/sections/table-view/actions-pane/RenameTableModal.svelte
pavish: I'm not sure what could have caused eslint to fail here, I assume it has something to do with recognizing the type of `$tables`.

I'm okay with disabling problematic rules for now.

pavish (approved):
@seancolsen This PR looks good to me!

mathesar_ui/src/component-library/text-input/TextInput.svelte
pavish: We can handle this on a subsequent PR.

PR #860, opened 2021-12-01
Sort tables list client-side

### LINKED ISSUE #631: Ensure the Tables list is sorted after adding a new empty table
## Description

The list of tables on the left should be sorted alphabetically, even after adding a new empty table.

## To reproduce

1. Begin with a list of tables such as `Table 0` and `Table 1`.
1. Choose "New table", then "Empty table".
1. Expect the "Tables" list to read `Table 0`, `Table 1`, `Table 2`.
1. Observe it to be `Table 0`, `Table 2`, `Table 1`.
1. Refreshing the page will reload the tables list, sorting it correctly.

## Environment

- OS: Ubuntu 20.04
- Browser: Firefox
- Browser Version: 91

## Additional context
- Blocked by backend work, see #635

seancolsen:
Fixes #631 by ensuring that the list of tables in the left pane is always sorted alphabetically.

### Reasons to sort on the front end

The ticket recommended sorting on the back-end. In this PR, I've done the sorting on the front end. Here's why:

- There are two ways that the front end obtains the list of tables, each with their own minor problem:
    1. via [get_table_list](https://github.com/centerofci/mathesar/blob/7e4807ede41fe5b51d747a0cdc8dff6d3d013858/mathesar/views.py#L27-L35) which runs when preloading data for SSR.

        The problem here is that this list isn't already sorted by name. I imagine it would be trivial to change this code so that it sorts by name, but I wasn't able to immediately figure out how. I tried changing `Table.objects.filter(schema=schema)` to `Table.objects.filter(schema=schema).order_by('name')` but that didn't work and I wasn't able to easily figure out why.

    2. via the `tables` API which [now accepts](https://github.com/centerofci/mathesar/pull/691) a `sort_by=name` param.

        The problem here is that the sorting is by byte value instead of unicode collation. For example, in python `sorted(['Ä', 'b', 'C'])` will give `['C', 'b', 'Ä']`.

        If we do want to fix that sorting, it seems like it will require a bit more code. Here are some resources I found towards that end: [1](https://stackoverflow.com/questions/31862420/do-case-insensitive-ordering-with-django-filter), [2](https://stackoverflow.com/questions/1097908/how-do-i-sort-unicode-strings-alphabetically-in-python/1098160#1098160).

- The fact that there _are_ two ways the front end obtains the tables list makes me somewhat leery of setting the sorting in two very separate code locations. I'd rather set it in one place for the sake of maintain consistency as our codebase grows.

- All-things-considered, this client-side approach seemed easiest to implement, and I don't see any serious downsides.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
Regardless of frontend usage, I do think it makes sense to fix the ordering in the API. Could you create a separate ticket for it?

pavish:
i agree with @kgodey on this that we need to handle it at the API, and it would be better if we handled it while preloading data as well (Ideally the response from the API and the preloaded data should be the same).

But I see no harm in handling it on the frontend additionally, so we can go ahead and merge this PR.

pavish (approved):

PR #865, opened 2021-12-03
mathesar-424 Handle `TIMESTAMP` data type in the backend

### LINKED ISSUE #424: Handle `TIMESTAMP` data type in the backend
This issue is to ensure that Mathesar can handle the Postgres `TIMESTAMP` data type.

- Users can use the API to change a column to `TIMESTAMP`(with time zone) if it's possible to do so.
- Users can use the API to change a column to `TIMESTAMP`(without time zone) if it's possible to do so.
- Users can set and change the `precision` of a given `TIMESTAMP` column via a `type_options` field in the API.
- Automatic type inference during file import suggests `TIMESTAMP` when it makes sense to do so. 

**Additional Context** 
- #250 (parent issue)

silentninja:
Fixes #424 Handle `Timestamp` data type in the backend.

This PR introduces Timestamp data type(with and without timezone) and relevant functions to convert text data types that contain a date-time string into a `timestamp` data type.

The order by which the data type is inferred will change to the following
1. If a text contains a date string without any time information `date` type will be inferred.
2. If a text contains a date string along with time information but does not include a `timezone` information `timestamp without timezone` data type will be inferred.
3. If a text contains a date string along with time and timezone information, `timestamp with timezone` data type will be introduced. This type will be inferred if only one  or more rows contains the necessary information as other data type will end up with losing the information 

**Technical details**
Adds TImestamp with timezone and Timestamp without timezone data type and  Relevant casting function that restricts the data that can be converted to the data type.
The casting function of `date` data type has been limited to accept only `text` with `date` element that does not contain `time` information
**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

## TaskList
 
- [x] Create a timestamp column(db/tests/columns/operations/test_create.py)
- [X] Automatic type inference during file import suggests (with time zone or without timezone) when it makes sense to do so (db/tests/tables/operations/test_infer_types.py:31)
- [x] Users can use the API to change a column to TIMESTAMP(with time zone or without timezone) if it's possible to do so (db/tests/types/operations/test_cast.py:169)
- [x] Users can set and change the precision of a given TIMESTAMP column via a type_options field in the API(db/tests/types/operations/test_cast.py:825)
- [x] Fetching a timestamp record(mathesar/tests/api/test_record_api.py)

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/tests/types/operations/test_cast.py
```diff
@@ -37,9 +38,10 @@
 NUMERIC = PostgresType.NUMERIC.value.upper()
 REAL = PostgresType.REAL.value.upper()
 SMALLINT = PostgresType.SMALLINT.value.upper()
-DATE = PostgresType.DATE.value.upper()
```
silentninja, line 40: `Date` was repeated, so deleted this duplicate variable

db/tests/types/operations/test_cast.py
```diff
@@ -447,6 +465,48 @@
             },
         },
     },
+    TIMESTAMP_WITHOUT_TIME_ZONE: {
+        ISCHEMA_NAME: PostgresType.TIMESTAMP_WITHOUT_TIME_ZONE.value,
+        REFLECTED_NAME: TIMESTAMP_WITHOUT_TIME_ZONE,
+        TARGET_DICT: {
+            CHAR: {VALID: []},
```
dmos62, line 472: I presume a timestamp should be castable to any text/character type, including `CHAR`.

silentninja: It is possible to cast a `timestamp` to `CHAR`, but `CHAR` returns exact number of characters specified and this causes problems with testing as the `type_options` has to be specified with each test case, which is not possible right now. I have created a [issue](https://github.com/centerofci/mathesar/issues/869) that should help with testing such cases

kgodey (commented):
I'll leave this to @mathemancer to review/merge. I took a quick glance and don't see any issues.

kgodey:
Updated status to blocked since it's blocked by #628

db/types/operations/cast.py
```diff
@@ -558,10 +622,26 @@ def _get_date_type_body_map():
     # Note that default postgres conversion for dates depends on the
     # `DateStyle` option set on the server, which can be one of DMY, MDY,
     # or YMD. Defaults to MDY.
-    default_behavior_source_types = frozenset([DATE]) | TEXT_TYPES
-    return _get_default_type_body_map(
-        default_behavior_source_types, DATE,
+    source_text_types = TEXT_TYPES
+    default_behavior_source_types = frozenset([DATE])
```
mathemancer, line 626: I think `TIMESTAMP` types should also be valid sources for dates (in case there's a timestamp value at midnight so the cast would be lossless).

mathemancer (changes requested):
Overall, pretty good.  I have a few requests, though.

Please see the PostgreSQL docs about `RETURNS NULL ON NULL INPUT`:  https://www.postgresql.org/docs/13/sql-createfunction.html

This will let you take lots of checks for `NULL` out of your SQL function code (and also make things _much_ more efficient).

Also, I think we should be able to cast `TIMESTAMP` types to `DATE` whenever the cast would be lossless.  I'd kind of expect to be able to do such a cast if I had a bunch of values in my column at midnight.

See some in-line comments for other requests.

mathemancer:
@silentninja Have you noticed the requested changes on this PR?

silentninja:
> @silentninja Have you noticed the requested changes on this PR?

Yes, I did notice it, sorry for not changing the status. I got caught up with other issues, will be working on it today.

db/types/operations/cast.py
```diff
@@ -492,6 +504,58 @@ def _get_time_type_body_map(target_type):
...
+
+def _get_timestamp_without_timezone_type_body_map():
+    source_text_types = TEXT_TYPES
+    default_behavior_source_types = frozenset([TIMESTAMP_WITHOUT_TIME_ZONE, DATE])
+
+    not_timestamp_without_tz_exception_str = f"RAISE EXCEPTION '% is not a {TIMESTAMP_WITHOUT_TIME_ZONE}', $1;"
+    # Check if the value is missing timezone by casting it to a timestamp with timezone
+    # and comparing if the value is equal to a timestamp without timezone
+    timestamp_without_tz_condition_str = f"""
+            IF ($1 IS NULL) OR (timestamp_value_with_tz = timestamp_value AND timestamp_value_with_tz <> date_value) THEN
```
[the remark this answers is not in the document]
silentninja: I wanted to have a strict inference and without relying on the order of the datatype in `INFERENCE_DAG` but missed out on this edge case. If we end up removing the comparison with date value, we would also be allowing timestamp without time `May-2007-29`.

silentninja:
> see the PostgreSQL docs about RETURNS NULL ON NULL INPUT: https://www.postgresql.org/docs/13/sql-createfunction.html

I didn't want to modify the `db.types.operations.cast.assemble_function_creation_sql` as it is being used by other casting functions. 

Moreover, I am not sure if this [special case where performance drops](https://stackoverflow.com/questions/8455177/function-executes-faster-without-strict-modifier) applies in our situation as it mostly applies to table functions but wanted to give it the benefit of doubt

db/types/operations/cast.py
```diff
@@ -492,6 +504,58 @@ def _get_time_type_body_map(target_type):
...
 
 
+def get_text_to_datetime_cast_str(type_condition, exception_string):
+    return f"""
+    DECLARE
+    timestamp_value_with_tz NUMERIC;
+    timestamp_value NUMERIC;
+    date_value NUMERIC;
+    BEGIN
+        SET TIME ZONE 'UTC';
```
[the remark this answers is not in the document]
silentninja: Thanks for pointing this out, felt like I was missing something. As for wrapping it in a transaction, I think you mean the query that is calling this function like the `ALTER TABLE`, I might have to mention that the casting function should be in its own transaction block somewhere. Spent quite some time looking for a better alternative but unfortunately had to end up with this.

mathemancer (approved):
I think this should work, and the choices w.r.t. what casts to what else seem logical.

About setting the functions to `RETURNS NULL ON NULL INPUT`, that should be safe since these are all casting functions that definitely _ought_ to return null on null input.  Also, many of them can't be inlined for various reasons.  I'll make a separate issue to explore that, though, since some of them _can_ be inlined, and these casting functions get run for every row of some table or another quite often.  For now, it's a moot point, since the null checks all got removed (unless we do see some improvement overall in performance).

silentninja:
@mathemancer Thanks! I will change the strict keyword to `RETURNS NULL ON NULL INPUT` as it seems be descriptive compared to `STRICT`

silentninja:
@kgodey We need to specify that the `datetime casting functions` are not transaction-safe and can affect timezones when run in the same transaction, where should this be documented?

kgodey:
@silentninja Please document it in the code and in the wiki (write a short spec that's linked to from "Technical Specs" in https://wiki.mathesar.org/en/engineering/architecture). Please also explain what "can affect timezones" means in greater detail.

kgodey:
@silentninja Feel free to merge this PR when you're ready.

PR #870, opened 2021-12-08
Implement renaming a column

### LINKED ISSUE #497: Implement renaming a column
## Problem

Users might want to rename an existing column on their table.

## Proposed solution

The ["Working with Columns" design spec](https://wiki.mathesar.org/en/design/specs/working-with-columns) has a solution for this, which we need to implement on the frontend.

## Additional context

- Backend work: #199 
- Design issue: #324 
- Design discussion: #436

seancolsen:
Fixes #497

## Demos and explanation

- [Basic column rename](https://www.loom.com/share/8b813dc217184b0899db4e08cc42b551)
- [Error scenarios when renaming a column](https://www.loom.com/share/05d1f2c76a1c47bcb19c03d62a5250bf)
- [Usability of renaming a column](https://www.loom.com/share/5e9ae39de5564b1c94bf8d3950c7035a)

## Issues

- The records within the column appear blank for a short time during the process of renaming. That's while we're awaiting `this.recordsData.fetch()` within `TabularData.ts`. I'm re-fetching the records because that's the easiest way to associate them with the new column name. There are some alternate approaches we could consider to improve this behavior. We could mutate the records store with the new column name. That seems somewhat messy though. We could modify the records API to associate cell values with column _id_ values instead of column names. For now though I think this behavior is acceptable.

- `ColumnsDataStore.rename` is inconsistent with other methods like `add` because it does not re-fetch the columns data after it renames. I don't like this inconsistency, but I'm not sure of the best way to resolve it.

    - After we rename a column, we _do_ need to re-fetch the columns data, but we also need to re-fetch the records data and the constraints data. We need to re-fetch the whole TabularData structure, and we don't (and shouldn't) have a way to do that from within the `Columns` class. So instead of re-fetching within `ColumnsDataStore.rename`, I'm re-fetching within `HeaderCell.svelte`.

    - Alternate approaches:

        - Re-fetch within `ColumnsDataStore.rename` and also call `TabularData.refresh` within `HeaderCell.svelte`. That's duplicate fetches. Not great for performance.
        - Re-fetch within `ColumnsDataStore.rename` and call `RecordsData.fetch` and `ConstraintsDataStore.fetch` from within `HeaderCell.svelte`. Messy because if we change the implementation of `TabularData.refresh` later on, it won't be apparent that we might also need to also change `HeaderCell.svelte`.

    - Several weeks ago @pavish and I chatted about the best re-fetch pattern to establish. I initially argued that store functions should _never_ re-fetch and that we should always do re-fetching at the view layer (as I've done in this PR) to give ourselves maximum granularity and control. We discussed several other potential patterns to employ and ultimately agreed to consistently re-fetch within store functions that update data and avoid re-fetching at the view layer. I think we might need to re-evaluate that pattern in light of this new example. As our codebase grows, I predict that we'll continue to need more granular control over the re-fetching behavior than our current pattern will allow.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish I see. Thanks for explaining.

My initial reaction to your comment was "**Ohhhh!** 💡". The event system which you had previously described to me as "something we'll use eventually" came into focus more clearly. I really like it!

But I think it might need some refactoring before I can leverage it here. Or I need to better understand it.

While looking at the code you pointed out, I realized that it's not just the `ConstraintsDataStore` and `RecordsData` that depend on the column name (and thus need to be updated after renaming a column). `Meta` depends on the column name too. For example, this PR currently has a bug. If I set my table to sort by a column and then I rename that column, I'll have a problem. In following your pattern to fix this bug, I _think_ we'd need to listen to a `'columnRename'` event from within `Meta`. Right? Since `Meta` is client-side only (with nothing to fetch), I imagine we'd want to mutate that store, updating `Meta.sort` to use the new column name. There are two problems though:

- For `Meta` to listen to `ColumnsDataStore`, `Meta` must receive a `ColumnsDataStore` instance during instantiation (at least, if we're to follow the pattern implemented in `RecordsData`). But currently `ColumnsDataStore` needs to receive a `Meta` instance during instantiation, creating somewhat of a circular problem. One way to resolve this might be to create a centralized message bus that we pass into `Meta`, `ColumnsDataStore`, `ConstraintsDataStore`, `RecordsData`, and `Display`.
- Another issue is that to fix the bug above, we need to ensure that `Meta` is updated _before_ `RecordsData`, otherwise `RecordsData.fetch` will pass an incorrect column name for `order_by`. How would we use the event system to guarantee that the events run in the correct sequence (and not in parallel)?

## An alternate approach

Maybe whenever we have a store operation that requires other stores to update, we should run the operation on the entity containing all the relevant stores.

So in this case, within `HeaderCell.svelte`, instead of calling `columnsDataStore.rename`, we would call `$tabularData.renameColumn`. Then `TabularData` would have access to everything it needs to run the operations in sequence.


## API design

I'm sure we'll continue to have situations where updating one store requires that we update other stores too, so all of the above thoughts are still relevant. But for _this_ case of renaming a column, it would be a lot simpler if we could rely on the column _id_ instead of name everywhere. We wouldn't need to update `Meta` or `Records` or `Constraints` if all of those stores held references to the column by its id instead of its name. Is it worth considering changing the API to go that direction?

kgodey:
> Is it worth considering changing the API to go that direction?

What change would you need in the API? I thought we'd already updated the API to use IDs instead of indices/names in #658.

pavish:
> What change would you need in the API? I thought we'd already updated the API to use IDs instead of indices/names in #658.

@kgodey 
#658 updates the PATCH and DELETE endpoint of columns to use id.

Other endpoints for records & constraints still make use of column names. I believe @seancolsen is referring to updating those endpoints to use column id instead, which would certainly make it easier for the client. I am in agreement.

kgodey:
Yes, we should do that. @silentninja, didn't you open a separate issue for that?

seancolsen:
@kgodey and @silentninja I opened #896 and #898 to specify that behavior that I'd like to have from the API for the purpose of simplifying the process of renaming a column. You can close them and integrate that info into any other tickets that are already open for that purpose. And if you have opinions about the usefulness or implementation details within those topics, we can continue the discussion within those tickets. Pavish and I will figure out a way to make the front end work just fine even if we keep the API as-is. The changes in those tickets would just make it a bit easier for the front end.

kgodey:
@seancolsen Do this tickets block this issue or can you refactor once we make those changes? I'm trying to figure out how to prioritize work.

seancolsen:
@kgodey 

> Do this tickets block this issue

Good question. No, I don't think so.

Were those tickets to be closed, then the discussion @pavish and I are having above would be moot and I could make some straightforward changes to this PR to make it merge-worthy. But that would only circumvent the need for our discussion temporarily, as we'd likely run into another situation soon that would rekindle it. I think it's best that Pavish and I figure out how to deal with this situation with the API as-is. Once we figure that out, I can update this PR to reflect our decision and work with the current API. Then if/when the API changes, we'll update the front end again.

kgodey:
Great, thanks @seancolsen. I won't prioritize assigning them yet.

seancolsen:
@pavish What's your opinion on how we should handle the re-fetching of data across multiple stores? What do you think about the "alternate approach" I [proposed](https://github.com/centerofci/mathesar/pull/870#issuecomment-995174178) above?

seancolsen:
This PR diverges so significantly from `master` at this point that it's going to be easier to re-create it from scratch in a new branch. I'm closing it and will open a new one which references this one.

pavish:
@seancolsen It's okay to create a new branch & PR for this, but I'd like us to hold it until our next UI sync call. I think it would be better to establish a pattern for inter-store dependencies.

`Meta` was supposed to be a store that does not depend on other stores, however storing filters, sort etc., in Meta makes it indirectly dependent on the `Columns` store. I would like to restructure this a bit before moving ahead.

PR #885, opened 2021-12-13
Datetime display options

### LINKED ISSUE #428: Implement display options for date & time types
## Problem
The Date & Time Mathesar type supports the following display options:
- Date format
- Time format

We will combine this into one `format` store.

## Solution
(1) We should store these display options in the following format in the `display_options` field of the corresponding column.
```
{
  "format": "YYYY-MM-DD hh:mm A",
}
```

(2) We should also validate these so that:
- Only columns of this type can have these display options. They should not be able to be set if the column is of a different type.
- `format` should be a valid JavaScript date/time format.
- We should validate format based on the type of the column. `DATE` columns should only include date formats, `TIME` column should only include time formats, `TIMESTAMP` columns can have either or both.

(3) If the column type is changed, the display options should be deleted.

(4) We should add supported display options to the `types` endpoint.

## Additional Context
- [Design of Date/Time type options on Figma](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=4260%3A37440&node-id=4270%3A41050&viewport=324%2C48%2C0.29&scaling=contain&starting-point-node-id=4270%3A41050&show-proto-sidebar=1)
- Blocked by #658
- #392

silentninja:
Fixes #428

Accept display options for datetime column types. 

**Technical details**
This PR adds a serializer to accept and validate datetime column `display_options`. It uses [arrow library](https://github.com/arrow-py/arrow) to validate `format` field values to accept only valid datetime object and restrict the format string based on types(date column can't accept format string with time formats).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
Blocked by #865

mathesar/api/serializers/shared_serializers.py
```diff
@@ -106,9 +108,106 @@ class NumberDisplayOptionSerializer(OverrideRootPartialMixin, serializers.Serial
...
+        if 'z' in display_format.lower():
+            raise serializers.ValidationError(
+                "Timestamp without timezone column cannot contain timezone display format")
+
+
+class DateFormatValidator(AbstractDateTimeFormatValidator):
+
+    def validate(self, datetime_obj, display_format):
+        date_obj = arrow.get('2013-09-30')
+        if datetime_obj.time() != date_obj:
```
mathemancer, line 153: Is this correct?  Shouldn't it be
```suggestion
        if datetime_obj.time() != date_obj.time():
```

Maybe I'm missing something.

mathemancer (changes requested):
Overall, this looks pretty good.  Some comments:
- The validation is spread over two functions called in different ways, which is a bit confusing.  Couldn't one call the other?
- One of the validations was missing a piece (I think; see the specific comment)
- Please make `vulture` happy so the pipeline passes.

mathesar/api/serializers/shared_serializers.py
```diff
@@ -106,9 +108,106 @@ class NumberDisplayOptionSerializer(OverrideRootPartialMixin, serializers.Serial
...
+            timestamp_with_tz_obj = arrow.get('2013-09-30T15:34:00.000-07:00')
+            parsed_datetime_str = timestamp_with_tz_obj.format(value)
+            datetime_object = arrow.get(parsed_datetime_str, value)
+        except ValueError:
+            raise serializers.ValidationError(f"{value} is not a valid format used for parsing a datetime.")
+        else:
+            self.validate(datetime_object, value)
+
+    @abstractmethod
+    def validate(self, datetime_obj, display_format):
```
silentninja, line 131: > The validation is spread over two functions called in different ways, which is a bit confusing. Couldn't one call the other?

@mathemancer Do you mean this validation function?

mathemancer: Whoops.  I completely missed the `else`.  My fault.  disregard that.

mathesar/api/serializers/shared_serializers.py
silentninja: Yes, it should be checking the time part of the object. I have made the necessary changes, thanks!

mathemancer (approved):
Looks great to me; thanks!

PR #868, opened 2021-12-07
Import via paste

seancolsen:
Fixes #445

## Demo

Peek 2021-12-06 16-44

## Additional changes

- Don't require user to press "Continue" button after uploading a file.
- Don't display "Continue" button while preview is loading.
- Use icons for continue and cancel buttons.

## Review tips

- Play with all import workflows, as I've refactored some code the touches all of them.

## To handle in subsequent PRs

- Add more features to the TextArea component. It's pretty plain right now. We can add more features as needed.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

Issue #531, opened 2021-08-09
Frontend work for adding multi-column unique constraints


kgodey:
## Problem

Users may want to create unique constraints on tables.

## Proposed solution

We should provide a way to create unique constraints from the UI following [the "Working with Columns" design spec](https://wiki.mathesar.org/en/design/specs/working-with-columns).

This covers allowing the addition of a new unique constraint, including auto-complete for column names.

Adding other types of constraints will not yet be supported.

## Additional context

- Design issue: #324 
- Design discussion: #436
- Backend issue: #326 
- Backend PR: #488 
- Blocking frontend issue: #328

kgodey:
This is ready for work now that #328 is done.

seancolsen:
@ghislaineguerin and I chatted about the design for this and agreed on the following:

- Build on the current Table Constraints modal.
- Insert an "Add" button above the list of constraints.
- When you click "Add", the whole contents of the modal slide to the left and are replaced by a form to add a new table constraint.
    - The form has "Cancel" and "Save" buttons at the bottom.
    - If you click "Cancel", the modal slides back to view the list of constraints.
    - If you fill out the form and click "Add", the modal slides back to view the list of constraints and the newly-added constraint is briefly highlighted.

PR #904, opened 2021-12-17
Radios and checkboxes

### LINKED ISSUE #824: Create Label component
## Problem

We don't have a great way to associate form labels with our input components.

Nesting them like this

```svelte
<label>
  <span>Foo</span>
  <Select options={choices} bind:value={foo} />
</label>
```

will work but the svelte a11y linter complains.

Also W3C [says](https://www.w3.org/WAI/tutorials/forms/labels/):

> Generally, explicit labels are better supported by assistive technology.

## Proposed solution

- Create a utility function

    ```ts
    getAppWideUniqueId();
    ```

    It will return a string containing an auto-incrementing intger each time the function is called.

    Then we can use that function within components to generate unique id values to use for form inputs.

- Create a `Label` component

    - If it recieves a `for` prop, it uses that on the DOM element.
    
        Otherwise, it uses a value from `getAppWideUniqueId()` and also passes that value down through context using a key of `'formInputId'` so that nested form input components can read that context value.

- Modify input components to work with the `Label` component.

    - If they find a `'formInputId'` context value they should use that for their `id`, (unles passed an `id` through props).

seancolsen:
Fixes #824 and adds a slurry of related form component improvements that I need in order to build a relatively simple form for #531.

## Done

- Improve **Checkbox** component
    - Remove the labelling functionality from this component, keeping only a simple checkbox that more closely mirrors the native `<input>` element. To label a checkbox, put one inside a `Label` or LabeledInput component. If this proves too verbose, we can create another component called LabelledCheckbox or CheckboxField. But for now I think it's fine.
    - Center the checkmark within the box.
    - Activate the hover CSS for the _entire_ box (previously it was only activated when hovering over a slightly smaller region of the box).
    - Simplify markup and CSS. Don't use extra elements when a single `<input>` will do.
- Add **Radio** component
    - Just for one radio on its own. I don't imagine we'll ever use it outside the RadioGroup component, but I think it's nice to have on its own because it lines up well with the Checkbox component making it somewhat easier to synchronize stylistic changes between the two components.
    - Style the radio buttons to match the checkboxes (instead of relying on the browser to style the radio buttons).
- Add **Label** component
    - This takes care of synchronizing the `id` value between the label, and also the `disabled` state so that the label can style itself if its associated input becomes disabled.
    - Make the Radio and Checkbox components work well with the Label component automatically when nested.
- Add **LabeledInput** component
    - This is a very small wrapper which only handles layout and CSS.
- Improve **RadioGroup** component
    - Allow entire options to be disabled
    - Improve style, especially with labels that contain wrapping text
- Add **CheckboxGroup** component
    - Compose all the same features used within RadioGroup, bus use a checkbox instead of a radio. You can bind to an CheckboxGroup's `values` prop to get an array of values.
- Misc
    - Change the text "Clipboard" to "Copy and Paste Text" on the import dialog, as [suggested](https://github.com/centerofci/mathesar/pull/868#discussion_r769553314) by @pavish.

## Changes visible in the Mathesar UI

- The only changes this PR _should_ be making to the Mathesar UI are barely noticeable.
- The checkboxes and radio buttons look a _tiny_ bit better.

## Changes visible in Storybook

- When reviewing this PR, make sure to play with all the components mentioned in the "Done" section above. They each have a story.

## Notes

- For the Label component, I diverged a bit from our plans in #824. In the ticket we identified the need to provide some coupling between the Label component and various input components so that the input's `id` attribute matches the label's `for` attribute. I also ended up wanting the label to be aware of the `disabled` state on the input. Instead of sharing two values between the components, I chose to make a new object, LabelController, to contain the `inputId` and the `disabled` values. I also decided that it should be the responsibility of the _input_ to set its own `id` and pass it to the label (instead of the _label_ creating an `inputId` and handing it to the input, as we had spec'd in the ticket) because one input can have multiple labels but one label cannot have multiple inputs (semantically speaking).
- In the RadioGroup component, I abandoned the pattern of using unstructured objects for options.

    Previously, each option looked like this

    ```ts
    export interface RadioGroupOption {
      [key: string]: string | number | string[]
    }
    ```

    and it was up to the consuming component to pass `idKey` and `labelKey` props (or leave them as the default 'id' and 'label' values respectively).

    Now each option looks like this

    ```ts
    export interface Option {
      value: string | number,
      label?: string,
      labelComponent?: typeof SvelteComponent,
      labelComponentProps?: unknown,
      disabled?: boolean,
    }
    ```

    The option has a more rigid structure. I like this a lot better. Within the RadioGroup component it makes the typings a lot simpler. And I feel that it provides clearer documentation to people using RadioGroup on how they need to structure their options. The consuming component might need to do a little restructuring if it already has an array of options. But I still like it better this way. We can chat further on this topic if there were stronger reasons for going with the less-structured approach initially.

    I also named it `Option` instead of `RadioGroupOption` because I'd like to eventually move towards using this pattern for other components which share the same behavior of choosing things from a set (e.g. Select).

## To handle in subsequent PRs

We should create tickets for these upon merging this PR.

- Update other input components (e.g. `TextInput`) to make them work with the new `Label` component, just like `Checkbox` and `Radio` do. We should probably create a separate component to encapsulate that logic use include _that_ component within all of the input components.

## Issues

- When adding the Label component , I would have like to set `class="label"` to follow the class-naming pattern we've been using in other components, but instead I set `class="label-component"` because we already had some `label` classes throughout the codebase and I didn't want to collide with them, given that all our CSS is global. I'm looking forward to a point where we have our CSS system more figured out and don't risk these kinds of naming collisions.
- There is some code duplication across `Checkbox.scss` and `Radio.scss`. I can think of a few ways to reduce this duplication, but overall our CSS setup seems a bit ad-hoc at the moment so I'd kind of prefer to get this ironed out once we improve the overall CSS setup. I added some code comments pointing out the duplication.
- In `RadioGroup.stories.svelte` see `richOptions` for an example of rich text in a radio option label. You'll notice this is somewhat cumbersome. Ideally we'd put that sort of code in the template, not the script. I have thoughts about how to handle that more gracefully with a little bit more work. But the need isn't quite there yet to justify cleaning that up at this point.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish After our conversation today I gave some though to the best way we might create an abstraction to simplify the process of making all our input components compatible with our label component.

These are the 7 line of code that I think we'd want in every input component.

```ts
import { getLabelControllerFromContainingLabel } from '@mathesar-component-library-dir/label/LabelController';
import { getGloballyUniqueId } from '@mathesar-component-library-dir/common/utils/domUtils';

export let disabled = false;
export let id = getGloballyUniqueId();
export let labelController = getLabelControllerFromContainingLabel();

$: labelController?.disabled.set(disabled);
$: labelController?.inputId.set(id);
```

This PR adds that code to `Checkbox` and `Radio` but doesn't add it to others (e.g. `TextArea`).

Can we avoid duplicating that code across all our input components? I'm still not sure how to create an abstraction that handles it. In React I might wrap all our input components in a HOC. Svelte doesn't quite have the same construct.

It's not a ton of duplicated code so at this point I'm leaning towards just continuing to duplicate it. But if you have ideas for how we could meaningfully reduce those 7 lines, I'd be interested to hear.

pavish:
@seancolsen 

> These are the 7 line of code that I think we'd want in every input component.
> Can we avoid duplicating that code across all our input components?

We would not be able to avoid this code duplication since Svelte doesn't allow any form of inheritance. But we could possibly abstract these away in a separate component and include it in all our input components. We could reduce it to:

```javascript
import BaseInputComponent from '...';

export let disabled = false;
export let id = undefined;

<BaseInputComponent {disabled} {id} {...$$restProps} />
```

The BaseInputComponent would be a headless component which would only handle the labelController for the moment, but I'm sure we'll have additional shared functionality we might want to build in the future.

I see that in the PR description you've suggested the same.

mathesar_ui/src/component-library/index.ts
```diff
@@ -5,38 +5,43 @@ export { default as EventHandler } from './common/utils/EventHandler';
...
+export { default as FileUpload } from './file-upload/FileUpload.svelte';
 export { default as Notification } from './notification/Notification.svelte';
-export { default as TabContainer } from './tabs/TabContainer.svelte';
-export { default as Tree } from './tree/Tree.svelte';
 export { default as Pagination } from './pagination/Pagination.svelte';
-export { default as Dropdown } from './dropdown/Dropdown.svelte';
 export { default as Select } from './select/Select.svelte';
-export { default as FileUpload } from './file-upload/FileUpload.svelte';
+export { default as TabContainer } from './tabs/TabContainer.svelte';
+export { default as Tree } from './tree/Tree.svelte';
```
pavish, line 42: Ordering here is maintained based on components relying on one another. For eg., Select relies on Dropdown, so Dropdown would be placed first and Select would be placed after. This would help us easily identify compound components with high level of inter-dependencies.

Do you feel alphabetical ordering would be more advantageous here?

mathesar_ui/src/component-library/labeled-input/LabeledInput.svelte
```diff
@@ -0,0 +1,27 @@
...
+  
+  export let label: string | undefined = undefined;
+  export let layout: LabeledInputLayout = 'inline';
+</script>
+
+<div
+  class="labeled-input"
+  class:layout-stacked={layout === 'stacked'}
+  class:layout-inline={layout === 'inline'}
+  class:layout-inline-input-first={layout === 'inline-input-first'}
```
pavish, line 13: We could probably replace this with:

```suggestion
  class="labeled-input layout-{layout}"
```

mathesar_ui/src/component-library/types.d.ts
```diff
@@ -1,9 +1,18 @@
...
 export type Size = 'small' | 'medium' | 'large';
 
+export interface Option {
+  value: string | number,
+  label?: string,
+  labelComponent?: typeof SvelteComponent,
+  labelComponentProps?: unknown,
+  disabled?: boolean,
+}
+
```
pavish, line 13: I find it very useful in several circumstances having a flexible option structure for Select as well as any component that relies on an array which contains a value and a label (It is fine to have a rigid structure for other properties like disabled and labelComponent since they directly correspond to presentation part of things).

Moreoever, this is a feature almost every popular component library provides.

I understand you like a rigid structure better because it helps with better typing. As a good solution, we can provide a sensible BaseOption type and an Option type. The BaseOption type would be less rigid while Option extends it and would be more rigid.

pavish (approved):
@seancolsen I like the direction this PR is heading towards the structuring of form components. I had a few minor comments.

Regarding styling, we can improve sharing of styles across components after we have the form builder PR merged, which already has a basic implementation.

I am going to go ahead and approve this PR and any change required can be taken up in upcoming PRs.

Issue #883, opened 2021-12-12
Implement new error structure in backend.


kgodey:
## Problem

We've agreed on a standard error structure for the API in #560. Now we need to implement it.

## Proposed solution

I'll format these in terms of desired outcomes. Whoever works on this can feel free to split these up into separate issues if that's helpful.
- All current API error responses should follow our error structure.
- All error codes and their corresponding `details` fields are documented somewhere.
- We have a standard way of generating new error codes.

## Additional context

- #560 
- [API Standards wiki page](https://wiki.mathesar.org/en/engineering/architecture/api-standards)

dmos62:
This seems up my valley. If @silentninja and @mathemancer don't mind, I'm taking this.

silentninja:
I am planning to set up the necessary **infrastructure** so that it is easier to capture and define the error structure for the captured error. Which it be then followed by  each error getting then own error structure(sub-task list or individual issue), where #859 will be the priority

kgodey:
@silentninja That's fine. please make small and incremental PRs so that changes are easy to review and merge. Getting #859 done to unblock the frontend soon should be a priority, so please organize your work so that we get to that ASAP.

kgodey:
Thanks @silentninja

PR #909, opened 2021-12-20
Handle error consistently on the backend using custom drf error handler

### LINKED ISSUE #623: Handle validation, errors, and business logic in the API consistently
## Problem

Currently, we have validation, error handling, and business logic spread across the `db` module, models, viewsets, and serializers to make the API work. This is confusing to work with.

## Proposed solution

- Organize the code with a clear separation of concerns.
- Document how the code is organized for reference.

## Additional context

Related to:
- #582 
- #560 
- #823

silentninja:
Fixes #623 

This PR introduces custom `Api Exception` and custom `Error Handler` to handle non `ApiException`. 
1. Custom Api Exception - Exception are captured and reraised with equivalent `ApiException` subclass which contains the `error code`, reusable logic for parsing the exception. It is responsible for converting the exception into a proper exception body according to the spec.
2. Custom error handler - Used as a last resort, it captures any exceptions that do not conform to the api spec and use a default map to get the equivalent `ApiException`, if an equivalent `ApiException` does not exist, it converts it to a `UnclassifiedException`(error code: 49XX), with the message as the string representation of the exception that is thrown. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/errors.py
```diff
@@ -1,2 +1,53 @@
...
+from contextlib import contextmanager
+
+from rest_framework.exceptions import APIException
+
+
 class InvalidTableError(Exception):
     pass
+
+
+ExceptionTransformerDetail = namedtuple(
```
kgodey, line 11: @silentninja I updated the whitespace here to make it more easily readable (in my opinion) and to match the rest of the code style. I figured updating it directly would be easier.

mathesar/api/viewsets/tables.py
```diff
@@ -105,18 +107,21 @@ def previews(self, request, pk=None):
             raise ValidationError("Incorrect number of columns in request.")
 
         table_data = TableSerializer(table, context={"request": request}).data
-        try:
+        with exception_transformer({**dict.fromkeys([DataError, IntegrityError], ExceptionTransformerDetail(400,
```
kgodey, line 110: @silentninja This function call for `exception_transformer` is a little hard to follow. It would be clearer if we could declare the dictionaries in a separate line and pass them here.

kgodey: I also think it would be clearer to instantiate the named tuple using keyword arguments (e.g. `status_code=400`, etc.) so that we don't have to remember the order in which to pass arguments and it's clearer what's going on without having to look at the original function signature.

mathesar/api/viewsets/tables.py
```diff
@@ -105,18 +107,21 @@ def previews(self, request, pk=None):
...
+                                                                                                            4001,
```
kgodey, line 111: How do we generate numbers for error codes and make sure they are unique?

I think it might make sense to have a mapping of exceptions and error codes somewhere central and use it here.

mathesar/api/viewsets/tables.py
```diff
@@ -105,18 +107,21 @@ def previews(self, request, pk=None):
...
+                                    UnsupportedTypeException: ExceptionTransformerDetail(400,
+                                                                                         4001,
+                                                                                         None,
+                                                                                         None,
+                                                                                         columns_field_key,
+                                                                                         None)
+                                    }):
             preview_records = table.get_preview(columns)
-        except (DataError, IntegrityError) as e:
-            if type(e.orig) == InvalidTextRepresentation or type(e.orig) == CheckViolation:
```
kgodey, line 111: It seems like we've lost this additional `if` check in the refactor. I don't think all `DataError` and `IntegrityError` instances are indicative of an invalid type cast.

kgodey (changes requested):
@silentninja Please see code comments.

I'm also confused about why we're not using DRF's custom exception handler. https://www.django-rest-framework.org/api-guide/exceptions/. You mentioned in the PR description that you'd have to do it on a per-view basis, but I'm not sure why that's the case. You can create an exception handler in a single place and then update our settings file to use it as the default exception handler. This seems cleanest to me - we can then use standard Python/DRF exceptions without as much boilerplate.

See [this Stack Overflow answer](https://stackoverflow.com/questions/35457287/how-to-add-custom-error-codes-to-django-rest-framework) for some more discussion on this topic.

silentninja:
> You mentioned in the PR description that you'd have to do it on a per-view basis, but I'm not sure why that's the case. 
It wouldn't be possible to have different behaviour for an exception based on the route. For example, the below exception handling wouldn't be possible.
```
except (DataError, IntegrityError) as e:
            if type(e.orig) == InvalidTextRepresentation or type(e.orig) == CheckViolation:
```
If we use are using a global DRF error handler, we might have to raise a custom APIException in order to assign a proper error code. 
While in the case of context managers, a parser that handles 
```  if type(e.orig) == InvalidTextRepresentation or type(e.orig) == CheckViolation: ``` can be used, making the exception handling reusable. This does hide away details at the cost of reusability, making it hard to explicitly know about the error.

File: mathesar/api/viewsets/tables.py (hunk 3 above)
Comment (silentninja) [reply mathesar/api/viewsets/tables.py:111]: I will replace the error codes with Enum

**Comment by kgodey** [comment]
I think it's fine to have to raise custom `APIException` errors to raise a proper error code, it'll be more explicit (and generally follow [the Zen of Python](https://www.python.org/dev/peps/pep-0020/)'s guidance on errors better). Even if we do have to catch a lot of DRF errors at the serializer or viewset layer, that's fine. I think it'll be easier for new contributors to the codebase as well.

**Comment by silentninja** [comment]
@kgodey I have replaced context managers with drf error handlers, let me know if your thoughts on this. I find this to be better as it seems to fit the codebase. @dmos62 @mathemancer I would love to hear your opinions too if you have any

**Comment by silentninja** [comment]
I went down the rabbit hole trying to replicate golang syle error handling using context managers as it looked more dry, but it was not verbose and made the code less explicit to the readers.

@dmos62 Based on your [comment](https://github.com/centerofci/mathesar/issues/889#issuecomment-995624672) I made the error codes centralised with option for the caller to supply their own

File: mathesar/exceptions/exceptions.py [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+from rest_framework import status
+from rest_framework.exceptions import APIException
+
+from mathesar.exceptions.error_codes import ErrorCodes
+
+ExceptionBody = namedtuple('ExceptionBody',
+                           [
+                               'code',
+                               'message',
+                               'field',
```
dmos62, line 14: Maybe name this `input_field` or `json_field`? Meaning would be a bit clearer, since SQLAlchemy uses the term `field` a lot (synonymous with our `column`).

silentninja: I was trying to make it work close with the drf tools, changing the name would lead to replicating a lot of existing code. I will see if I can change it to without breaking drf

kgodey: It should follow the [spec](https://wiki.mathesar.org/en/engineering/architecture/api-standards), which means the key should be `field`. API users should be unaware of SQLAlchemy and its nomenclature.

mathesar/exceptions/exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+
+from django.utils.encoding import force_str
+from rest_framework import status
+from rest_framework.exceptions import APIException
+
+from mathesar.exceptions.error_codes import ErrorCodes
+
+ExceptionBody = namedtuple('ExceptionBody',
+                           [
+                               'code',
```
kgodey, line 12: Per the [spec](https://wiki.mathesar.org/en/engineering/architecture/api-standards), this should be `error_code`.

mathesar/exceptions/exceptions.py
```diff
@@ -0,0 +1,49 @@
+from collections import namedtuple
+from typing import List
+
+from django.utils.encoding import force_str
+from rest_framework import status
+from rest_framework.exceptions import APIException
+
+from mathesar.exceptions.error_codes import ErrorCodes
+
+ExceptionBody = namedtuple('ExceptionBody',
```
kgodey, line 10: I think this will be most readable in one of these two options:

```python
ExceptionBody = namedtuple('ExceptionBody', ['code', 'message', 'field', 'details'], defaults=[None, None])
```

or

```python
ExceptionBody = namedtuple(
    'ExceptionBody',
    ['code', 'message', 'field', 'details'],
    defaults=[None, None]
)
```

mathesar/api/viewsets/tables.py
```diff
@@ -96,27 +99,36 @@ def previews(self, request, pk=None):
...
         serializer = TablePreviewSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        columns = serializer.data["columns"]
+        columns_field_key = "columns"
+        columns = serializer.data[columns_field_key]
 
         column_names = [col["name"] for col in columns]
         if not len(column_names) == len(set(column_names)):
-            raise ValidationError("Column names must be distinct")
+            raise CustomValidationError([ExceptionBody(ErrorCodes.DistinctColumnNameRequired.value,
```
kgodey, line 107: This is more involved than I think is necessary. It should be possible to just say something like:

```
raise DistinctColumnsRequiredException(message="Column names must be distinct")
```

The error code should be associated with the `DistinctColumnsRequiredException` in the `exceptions` module, it shouldn't need to be defined here.

This validation logic should ideally be in the `validate_columns` method of the serializer, not here ([see docs](https://www.django-rest-framework.org/api-guide/serializers/#validation)). Then it will automatically associate it with the `columns` field and we don't have to do it manually.

mathesar/exceptions/error_codes.py
```diff
@@ -0,0 +1,13 @@
+from enum import Enum, unique
```
kgodey, line 1: I think this should all live within `mathessar/api`, since these are only API related exceptions.

kgodey (changes requested):
@silentninja Added some feedback.

mathesar/exceptions/exceptions.py
silentninja: Made an intentional change as `drf` seems to be using `code` rather than `error_code`. Moreover `code` seems apt as it is already inside an `error object`

kgodey: Okay, that's fine with me, but please update the spec too.

mathesar/api/viewsets/tables.py
silentninja: > The error code should be associated with the DistinctColumnsRequiredException in the exceptions module, it shouldn't need to be defined here.

Yes that is the idea, I haven't added other custom Exceptions yet as it will be dealt with [in this issue](https://github.com/centerofci/mathesar/issues/883). But do you think it will be better to create a new Exception for one-off validations like this? I was looking at using the `message` or `error code` to match with the relevant information, but found that similar there are places where a different message is needed for a similar error code.

silentninja: > This validation logic should ideally be in the validate_columns method of the serializer, not here

I am aware of it, been itching to move it for a long it, but didn't want to pollute the PR, but it does seem like a good time to move them into serializers

kgodey: > Do you think it will be better to create a new Exception for one-off validations like this?

I think it will be easiest to have a single pattern to follow for tying error codes to exceptions, even for one-off exceptions.

It's fine to pass different `message` values to the same exception, you can do this for Python exceptions as well so it's a common pattern.

> it does seem like a good time to move them into serializers

Since you're already editing that particular code, I think it's fine to do it now. You don't have to do it for all of the code in this PR.

dmos62:
Together with Kriti's comments, this makes sense to me.

mathemancer:
@silentninja I don't have any major suggestions to add; I'm just glad you're taking up this heroic adventure.

silentninja:
Replaced by PR #1016

PR #921, opened 2021-12-22
Backend filtering refactor

### LINKED ISSUE #385: Implement filtering options for Number types.

### LINKED ISSUE #846: Duplicates filter should work with other filters

dmos62:
Fixes #385, fixes #846

Sets up the backend infrastructure for filtering.

### Features

Notice that I might use the terms filter specification and predicate interchangeably.

Some of the things this new predicate data structure does is:

- Declares (predicate) correctness to a high degree
    - An illegal predicate cannot be instantiated; an instantiated predicate is always legal
        - Notable exception
            - A predicate referencing unexisting columns can be instantiated
                - Referenced column existance is checked right before applying the filter specification to a query
- Declares what predicates the frontend can use under what circumstances and how
    - It can tell if a predicate is compatible with some column type based on its properties (like comparability, compatibility with `LIKE`, compatibility with URI-type-specific SQL functions)
    - It can tell the frontend how many parameters a predicate takes (e.g. `empty`, `equal` and `in` take different number of params) and it can tell it what options it accepts (e.g. should `starts_with` be case insensitive)
    - It can tell the frontend how to compose predicates: leaf vs branch predicates: parameters on branches are other predicates
- It supports SQL functions
    - You can use SQL functions to, for example, destructure a URI and filter based on that
       - Though this extension is/will be in a newer PR
       - Was not possible with `sqlalchemy-filters`

What it does not (currently) do:
- Does not support using other columns as parameters
    - Can't do `{equal: {column: x, parameter: {column: y}}}`
    - This is an oversight
    - Does not seem difficult to implement if/when there's interest

### Technical details

I organized the relevant pieces in the predicate data structure into a mixin hierarchy and also used this PR as a testbed for frozen dataclasses. Some of my objectives with the basic structure were:

- Immutability
  - Used properties where I could
     - Didn't use properties where an SQLAlchemy filter is returned, since its mutability is uncertain
- Many small classes
   - Use the mixin/type system to capture information
       - A class that is a predicate that takes one parameter and relies on `LIKE` will directly or indirectly mix in `ReliesOnLike`, `SingleParameter`, `Leaf` and `Predicate`
   - Some logic I offloaded to and hence centralized in static methods: correctness is declared on `db/filters/base::assert_predicate_correct`: and, what the JSON filter specification is is declared in `db/filters/operations/deserialize`
      - Has the drawback that the control flow in these methods can seem daunting, since it walks itself through the mixin hierarchy
          - These methods are not complicated, but big: each branch is simple: it's just that there's many of them
          - Might fan this logic out into the mixin definitions
- Used `parameter` where a singular parameter is expected and `parameters` where a sequence is expected
    - I have reservations about this
      - It makes the specifics of single/multi-parameter requirements obvious
      - Procedural instantiation is more verbose (like when testing), since you have to change the argument name depending on circumstance
          - But that can be solved with an auxillary constructor that's only for utilities

### How to extend with new predicate

1. Introduce the appropriate class; for example, `Greater`; this includes defining the new Predicate's properties through mixins: `ReliesOnComparability, SingleParameter, Leaf` in this case (note that mixin order matters, see Python docs);

```python
@frozen_dataclass
class Greater(ReliesOnComparability, SingleParameter, Leaf):
    type: LeafPredicateType = static(LeafPredicateType.GREATER)
    name: str = static("Greater")

    def to_sa_filter(self):
        return column(self.column) > self.parameter
```

2. Introduce the predicate type enum: `LeafPredicateType.GREATER` in this case; it's what the JSON filter spec will use to identify a predicate;

3. Tell it how to generate an equivalent SQLAlchemy filter by implementing the abstract `Predicate::to_sa_filter` (as seen above);
4. Update correctness definition (`db/filters/base::assert_predicate_correct`), if needed; currently this involves finding the spot in the method's control flow tree that corresponds to this new predicate, adding a new type check, etc.
5. Add new predicate to the `all_predicates` set;
6. Update `mathesar.database.types::is_ma_type_supported_by_predicate`, if needed; this would involve declaring what types have the properties the new predicate depends on: in this PR it's:
```python
def _is_ma_type_comparable(ma_type: MathesarTypeIdentifier) -> bool:
    return ma_type in comparable_mathesar_types

def is_ma_type_supported_by_predicate(ma_type: MathesarTypeIdentifier, predicate: Type[Predicate]):
    if relies_on_comparability(predicate):
        return _is_ma_type_comparable(ma_type)
    else:
        return True
```

Notice that `relies_on_comparability` is an auxiliary function returning true when a predicate is a subclass of `ReliesOnComparability`.

### `dataclasses` and `typing`

As Kriti suggested, I'll start a [dedicated discussion](https://github.com/centerofci/mathesar/discussions/913) on the use of `dataclasses` and `typing`.

But, to summarize:
- `dataclasses` eliminated custom boilerplate; I got immutability, `_post_init` and defaults without writing a single constructor; I think this is great since it's essentially standardized boilerplate;
- `typing` got me partial typing, which is great:
  - in combination with my LSP server/client ([`pyright`](https://github.com/microsoft/pyright)): catches a lot of mistakes and conflicts
  - in that I can annotate when I think it's useful and not when I don't (_partial_ typing)
  - for readability, for example the above method `is_ma_type_supported_by_predicate` operates on uninstantiated predicate classes, not instances, so to express that I can just say `predicate: Type[Predicate])` instead of `predicate_subclass`
  - doesn't have a cost; another developer can just omit type or write `Any` if he prefers

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
Resubmission of #837, which was closed automatically (?) when its base branch's PR was merged.

See this writeup from previous thread: https://github.com/centerofci/mathesar/pull/837#issuecomment-998720493

dmos62:
Copy pasted from previous thread (#837):

------

## Writeup

### Features

Notice that I might use the terms filter specification and predicate interchangeably.

Some of the things this new predicate data structure does is:

- Declares (predicate) correctness to a high degree
    - An illegal predicate cannot be instantiated; an instantiated predicate is always legal
        - Notable exception
            - A predicate referencing unexisting columns can be instantiated
                - Referenced column existance is checked right before applying the filter specification to a query
- Declares what predicates the frontend can use under what circumstances and how
    - It can tell if a predicate is compatible with some column type based on its properties (like comparability, compatibility with `LIKE`, compatibility with URI-type-specific SQL functions)
    - It can tell the frontend how many parameters a predicate takes (e.g. `empty`, `equal` and `in` take different number of params) and it can tell it what options it accepts (e.g. should `starts_with` be case insensitive)
    - It can tell the frontend how to compose predicates: leaf vs branch predicates: parameters on branches are other predicates
- It supports SQL functions
    - You can use SQL functions to, for example, destructure a URI and filter based on that
       - Though this extension is/will be in a newer PR
       - Was not possible with `sqlalchemy-filters`

What it does not (currently) do:
- Does not support using other columns as parameters
    - Can't do `{equal: {column: x, parameter: {column: y}}}`
    - This is an oversight
    - Does not seem difficult to implement if/when there's interest

### Technical details

I organized the relevant pieces in the predicate data structure into a mixin hierarchy and also used this PR as a testbed for frozen dataclasses. Some of my objectives with the basic structure were:

- Immutability
  - Used properties where I could
     - Didn't use properties where an SQLAlchemy filter is returned, since its mutability is uncertain
- Many small classes
   - Use the mixin/type system to capture information
       - A class that is a predicate that takes one parameter and relies on `LIKE` will directly or indirectly mix in `ReliesOnLike`, `SingleParameter`, `Leaf` and `Predicate`
   - Some logic I offloaded to and hence centralized in static methods: correctness is declared on `db/filters/base::assert_predicate_correct`: and, what the JSON filter specification is is declared in `db/filters/operations/deserialize`
      - Has the drawback that the control flow in these methods can seem daunting, since it walks itself through the mixin hierarchy
          - These methods are not complicated, but big: each branch is simple: it's just that there's many of them
          - Might fan this logic out into the mixin definitions
- Used `parameter` where a singular parameter is expected and `parameters` where a sequence is expected
    - I have reservations about this
      - It makes the specifics of single/multi-parameter requirements obvious
      - Procedural instantiation is more verbose (like when testing), since you have to change the argument name depending on circumstance
          - But that can be solved with an auxillary constructor that's only for utilities

### How to extend with new predicate

1. Introduce the appropriate class; for example, `Greater`; this includes defining the new Predicate's properties through mixins: `ReliesOnComparability, SingleParameter, Leaf` in this case (note that mixin order matters, see Python docs);

```python
@frozen_dataclass
class Greater(ReliesOnComparability, SingleParameter, Leaf):
    type: LeafPredicateType = static(LeafPredicateType.GREATER)
    name: str = static("Greater")

    def to_sa_filter(self):
        return column(self.column) > self.parameter
```

2. Introduce the predicate type enum: `LeafPredicateType.GREATER` in this case; it's what the JSON filter spec will use to identify a predicate;

3. Tell it how to generate an equivalent SQLAlchemy filter by implementing the abstract `Predicate::to_sa_filter` (as seen above);
4. Update correctness definition (`db/filters/base::assert_predicate_correct`), if needed; currently this involves finding the spot in the method's control flow tree that corresponds to this new predicate, adding a new type check, etc.
5. Add new predicate to the `all_predicates` set;
6. Update `mathesar.database.types::is_ma_type_supported_by_predicate`, if needed; this would involve declaring what types have the properties the new predicate depends on: in this PR it's:
```python
def _is_ma_type_comparable(ma_type: MathesarTypeIdentifier) -> bool:
    return ma_type in comparable_mathesar_types

def is_ma_type_supported_by_predicate(ma_type: MathesarTypeIdentifier, predicate: Type[Predicate]):
    if relies_on_comparability(predicate):
        return _is_ma_type_comparable(ma_type)
    else:
        return True
```

Notice that `relies_on_comparability` is an auxiliary function returning true when a predicate is a subclass of `ReliesOnComparability`.

### `dataclasses` and `typing`

As Kriti suggested, I'll start a dedicated discussion on the use of `dataclasses` and `typing`.

But, to summarize:
- `dataclasses` eliminated custom boilerplate; I got immutability, `_post_init` and defaults without writing a single constructor; I think this is great since it's essentially standardized boilerplate;
- `typing` got me partial typing, which is great:
  - in combination with my LSP server/client ([`pyright`](https://github.com/microsoft/pyright)): catches a lot of mistakes and conflicts
  - in that I can annotate when I think it's useful and not when I don't (_partial_ typing)
  - for readability, for example the above method `is_ma_type_supported_by_predicate` operates on uninstantiated predicate classes, not instances, so to express that I can just say `predicate: Type[Predicate])` instead of `predicate_subclass`
  - doesn't have a cost; another developer can just omit type or write `Any` if he prefers

mathemancer (commented):
Warning:  I've been thinking a _lot_ about this PR.

@dmos62 Did you look at the implementation in the `sqlalchemy-filters` library?  If so, I'm curious why not just extend that?  Especially since if we want to get away from using that library, we'll need to extend it with any other filters we want anyway.  

I'm really concerned about the process described in the writeup for adding a new predicate.  I think adding a filter should be a "good first issue" sort of task, but the way it's set up, the implementer would need to touch lots of the code base, understand the whole mixin system, etc.  In comparison, to add a predicate to the `sqlalchemy-filters` library, I just need to add a lambda function to the `filters.Operator.OPERATORS` dictionary.  I could see how to do this immediately within a couple minutes of clicking around in that repo.  Note that any checking done in that library (e.g., for number of arguments for an operator) is determined by _inspecting the implementation_.  This is a great quality, since it guarantees that the checking is synced with the implementation.

I do recognize that the setup in this PR has more functionality for giving info to the UI via some endpoint, but I think that should all live in the service layer section of the code, not the database library.  And (again, sorry to be repetitive) I think the description should be derived from the implementation.

I feel like this is a super crucial part of the codebase, both for the product and for potential contributors, so I'm being fussier than I might otherwise.

dmos62:
@mathemancer I didn't extend `sqlalchemy-filters` because I was aiming for those features I listed. `sqlalchemy-filters` would have been easy to extend to support custom SQL functions, but I didn't see a way to have its predicates be queryable/self-describing (and correctness checking comes from that too).

If you were to drop or not use this implementations ability to describe its predicate classes or to check correctness of their instances, you'd have the same process for adding a new filter/predicate as `sqlalchemy-filters`:

```
@frozen_dataclass
class Greater(Predicate):
    type: PredicateType = static(PredicateType.GREATER)
    name: str = static("Greater")

    def to_sa_filter(self):
        return column(self.column) > self.parameter
```

But, I think that these self-describing classes are useful, largely because it gets you procedural API.

It also didn't occur to me until now that adding a new predicate can be complicated. This is the simplest way I saw to get these features. To be fair, if you're adding something that's not very novel, then steps 4 (updating correctness definition) and 6 (updating type definitions) would be skipped, so you're just left with constructing the SA filter and describing the predicate with mixins. And the mixin system can be further optimized for accessibility.

Again, I think procedural/self-describing API is worth it.

dmos62:
I'll add that, when comparing this implementation and `sqlalchemy-filters`, for the full picture it would be good to look at how burdensome it would be to maintain a filtering API declaration with `sqlalchemy-filters`, especially one that lets the frontend not hardcode anything.

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+    """Enforces constraints on predicate instances."""
+    try:
+        if isinstance(predicate, Leaf):
+            column = predicate.column
+            column_name_valid = column is not None and type(column) is str and column != ""
+            assert column_name_valid, f"Column name invalid: {column}. It must be a non-empty string."
+
+        if isinstance(predicate, SingleParameter):
+            parameter = predicate.parameter
+            assert parameter is not None, "A parameter must not be None."
```
silentninja, line 356: Can (Custom)Exceptions be raised instead of assertions error? assert seem to indicate an internal error(something that should not happen), but this exception seems to be a mistake from the client

db/filters/operations/deserialize.py
```diff
@@ -0,0 +1,40 @@
...
+        if issubclass(predicate_subclass, Leaf):
+            columnName = predicate_body['column']
+            if issubclass(predicate_subclass, SingleParameter):
+                return predicate_subclass(column=columnName, parameter=predicate_body['parameter'])
+            elif issubclass(predicate_subclass, MultiParameter):
+                return predicate_subclass(column=columnName, parameters=predicate_body['parameters'])
+            elif issubclass(predicate_subclass, NoParameter):
+                return predicate_subclass(column=columnName)
+            else:
+                raise Exception("This should never happen.")
```
silentninja, line 25: Assertion errors are better than Exception for these error cases

mathemancer:
> I'll add that, when comparing this implementation and `sqlalchemy-filters`, for the full picture it would be good to look at how burdensome it would be to maintain a filtering API declaration with `sqlalchemy-filters`, especially one that lets the frontend not hardcode anything.

Point taken.  I completely agree that we should avoid the front end needing to know how to put together different filtering leaves.  

I do think that (at least) trying to focus on reflecting descriptive information from the actual implementation would reduce the number of places a given filter needs to be defined while achieving the goal.  If you look at the `sqlalchemy-filters` logic around checking the number of arguments (see `filters.Operator.arity`).  It's really slick.  Using that idea would remove the need for the mixins for that part of the description.  (On a side note, I recall you mentioned somewhere being uncomfortable with having both `parameter` and `parameters`.  I agree, and think this should probably always be `parameters`.)

It would remain to deal with the types and comparability parts of things, and I think those should simply wait until the actual DB call.  The reason is, we want this to be as unconstrained and hackable as possible, and this way we avoid imposing artificial limits on what filters can be constructed.  We're not going for a standard web app, we're going for a DB client.  If the user wants to try doing something that works for some reason we don't foresee, I think that should be allowed.  I think this would also help the user understand the differences between different types, and improve their data model.

kgodey:
@dmos62 Thanks for the writeup again and all the comments, I think I wrapped my head around the approach better now.

I don't think I understand why we need the current predicate system to create a self-describing API. I also looked through [the `sqlalchemy-filters` implementation](https://github.com/juliotrigo/sqlalchemy-filters/blob/master/sqlalchemy_filters/filters.py) and it seems like you should be able to create a self-describing API by inspecting the `Operator` class and building the API based what's defined there. 

I'm also not sure what the benefits of predicates being queryable and correctness checking are for the filter system. It seems much easier/less work to rely on the DB to do that - the filter system should be mainly focused on generating the right SQL query in a composable way and then trying that query.

dmos62:
The problem of having the predicate definition in multiple places is trivial to solve by moving the correctness declaration into the Predicate classes. I did it this way to have the correctness information all in one place. We can also drop it, though I like the idea of knowing whether a filter is correct to some degree early on: isn't parsing database's failure to use an invalid filter specification difficult?

There's only so much we can deduce from inspecting the SA filter function (e.g. the lambdas in the [`OPERATORS` dict](https://github.com/juliotrigo/sqlalchemy-filters/blob/master/sqlalchemy_filters/filters.py#L39)). For example, you can't tell that `IN` expects a sequence, or that `LIKE` only works on strings.

silentninja (changes requested):

kgodey:
Summary of ideas from the call we just had (@silentninja @mathemancer and @dmos62 will add their own summaries shortly):
- It would be great to have adding new filters be as easy as possible. I'm hoping we can get it down to "write a function and register it somewhere". I don't think we should do exactly the same thing as `sqlalchemy-filters` given that we need to be more expressive but they do have this down.
- Filter definitions should go in the `mathesar` directory, not `db` since they're related to the API.
- We shouldn't worry about correctness or predicates being queryable at this layer - the DB will handle it and we should delegate as much as possible. This has the advantages of Mathesar being more hackable and us not having to maintain the set of what filters are and aren't allowed.
- The aim should be to make the implementation more reflective rather than declarative.
- Functions should be smaller where possible.

dmos62:
Summary of ideas from the call we just had:

- Make adding new filters accessible to user developers
- Deprioritize correctness checking
    - Minimize constraints on what kind of filter specifcations a user can submit
    - Let the database decide what's valid
- Find ways to guide the user and suggest safe options like applying a `starts_with` predicate on a text column, but also allow the user to apply it to anything he'd like
- At the same time, support procedurally generating the API, so that the frontend does not need to hardcode things to provide a full filter specification UI

dmos62:
@kgodey @silentninja @mathemancer I would find it useful if you could post some of the most advanced or complex filtering scenarios that you can think of and that we'd like to support. I'll work backwards from that.

Some complex stuff that is on my radar at the moment: optional options (as distinct from parameters), like whether a string matching predicate should be case-sensitive; API-specified SQL functions (maybe preprocessing `starts_with` input to be lowercase); using value of some column when applying predicate to another column (or, to say it differently, having predicates apply to rows not columns).

db/filters/base.py
dmos62: I'd disagree. I use an `assert` as an abstraction over "if not true then throw generic exception with this information". I then catch them and wrap them in something useful. This is pretty much a shortcut over using a custom helper like `assertNotBadFilterFormat(parameter is not None, "A parameter must not be None.")`.

db/filters/base.py
```diff
@@ -0,0 +1,381 @@
...
+    BRANCH = "branch"
+
+
+class BranchPredicateType(Enum):
+    NOT = "not"
+    OR = "or"
+    AND = "and"
+
+
+class LeafPredicateType(Enum):
```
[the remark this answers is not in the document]
dmos62: It's for enumerating all leaf predicate types. It's not strictly needed, since there's a list of available predicate classes which we can query. An enum is convenient when you want to check if a predicate type string is valid, or whether it's a leaf or a branch.

mathesar/api/viewsets/records.py
```diff
@@ -36,9 +38,10 @@
                 self.get_queryset(), request, table_pk,
                 filters=serializer.validated_data['filters'],
                 order_by=serializer.validated_data['order_by'],
+                duplicate_only=serializer.validated_data['duplicate_only'],
                 grouping=serializer.validated_data['grouping'],
             )
-        except (BadFilterFormat, FilterFieldNotFound) as e:
```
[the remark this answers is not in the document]
dmos62: I agree. It's not immediately clear to me what a better error structure would be. Note that `ReferencedColumnsDontExist` returns a set of columns that were not found on the selectable object.

dmos62:
I sketched out what the re-refactored filter expression interface might look like.

### Note

I used `dataclasses` (and thus type annotations), because I found it faster to sketch that way. The use of `dataclasses` and `typing` is a sketch implementation detail and not meant to represent how the actual data structures will be implemented.

### Refactor goals and features

- Correctness constraints lifted
    - Replaced with suggestions
        - Those suggestions don't have to be followed
- Simplified extension by user developers
    - To add a new expression you subclass `Expression`, which is pretty much defining `to_sa_expression`, and you add it to the list `supported_expressions`
        - Optionally, you can add suggestions about how the expression can be used
- A way to give suggestions to the frontend and the user about how to use/assemble expressions
    - See the static field `suggestions` on `Expressions` subclasses
- Switch from operating on columns to operating on rows
    - You can do things like "is this column greater than this column"
- Predicates, column references and SQL functions (including custom ones) are just `Expression` subclasses. 

```python
@frozen_dataclass
class Expression(ABC):
    id: str
    name: str
    suggestions: Set = static(None)
    parameters: List

    @staticmethod
    @abstractmethod
    def to_sa_expression():
        return None


@frozen_dataclass
class ColumnReference(Expression):
    id: str = static("column_reference")
    name: str = static("Column Reference")
    suggestions: Set = static({
        suggestions.parameter_count(1),
        suggestions.parameter(1, suggestions.Column)
    })

    @staticmethod
    def to_sa_expression(p):
        return column(p)


@frozen_dataclass
class Empty(Expression):
    id: str = static("empty")
    name: str = static("Empty")
    suggestions: Set = static({
        suggestions.returns(suggestions.Boolean)
        suggestions.parameter_count(1),
    })

    @staticmethod
    def to_sa_expression(p):
        return p.is_(None)


@frozen_dataclass
class Greater(Expression):
    id: str = static("greater")
    name: str = static("Greater")
    suggestions: Set = static({
        suggestions.returns(suggestions.Boolean),
        suggestions.parameter_count(2),
        suggestions.all_parameters(suggestions.Comparable)
    })

    @staticmethod
    def to_sa_expression(p1, p2):
        return p1.gt(p2)


@frozen_dataclass
class In(Expression):
    id: str = static("in")
    name: str = static("In")
    suggestions: Set = static({
        suggestions.returns(suggestions.Boolean),
        suggestions.parameter_count(2),
        suggestions.parameter(2, suggestions.Array)
    })

    @staticmethod
    def to_sa_expression(p1, p2):
        return p1.in_(p2)


@frozen_dataclass
class And(Expression):
    id: str = static("and")
    name: str = static("And")
    suggestions: Set = static({
        suggestions.returns(suggestions.Boolean)
    })

    @staticmethod
    def to_sa_expression(*ps):
        return and_(*ps)


@frozen_dataclass
class StartsWith(Expression):
    id: str = static("starts_with")
    name: str = static("Starts With")
    suggestions: Set = static({
        suggestions.returns(suggestions.Boolean)
        suggestions.parameter_count(2),
        suggestions.all_parameters(suggestions.StringLike)
    })

    @staticmethod
    def to_sa_expression(p1, p2):
        return p1.like(f"{p2}%")


@frozen_dataclass
class ToLowercase(Expression):
    id: str = static("to_lowercase")
    name: str = static("To Lowercase")
    suggestions: Set = static({
        suggestions.parameter_count(1),
        suggestions.all_parameters(suggestions.StringLike)
    })

    @staticmethod
    def to_sa_expression(p1):
        return func.lower(p1)


@frozen_dataclass
class ExtractURIAuthority(Expression):
    id: str = static("extract_uri_authority")
    name: str = static("Extract URI Authority")
    suggestions: Set = static({
        suggestions.parameter_count(1)
    })

    @staticmethod
    def to_sa_expression(p1):
        return func.getattr(URIFunction.AUTHORITY)(p1)


# Enumeration of supported Expression subclasses; useful when parsing.
supported_expressions = tuple(
    [
        ColumnReference,
        Empty,
        Greater,
        In,
        And,
        StartsWith,
        ToLowercase,
        ExtractURIAuthority,
    ]
)
```

This is what a filter expression tree would look like. Notice that the constructor argument is always a sequence.

```
# Sample filter expression tree
And([
    StartsWith([
        ExtractURIAuthority([
            ColumnReference("uri_col")
        ]),
        "google"
    ]),
    Greater([
        ColumnReference("some_col"),
        ColumnReference("some_other_col"),
    ]),
])
```

kgodey:
@dmos62 This structure looks a lot better to me, nice work. I have some thoughts.

It would be ideal to be able to have different types of objects for the following types of things to help differentiate them clearly:
- what's being operated on (`ColumnReference`)
   - I assume this is not extensible since everything operates on a column.
- predicates that involve joining other operations (and/or/not)
   - This probably doesn't need to be extensible either, they're just basic logical operations.
- functions that involve pre-processing data (e.g. `ExtractURLAuthority`) - this is anything that transforms data
   - It should be clear how to add new pre-processing functions
   - I'm not sure there's any benefit of structuring this as a class, it seems like it just contains a function, it seems cleaner to let it be a function.
- functions that involve filtering data - this is anything that compares data and returns a boolean
   - It should be clear how to add new filtering functions and I think it should involve registering it in a different place than pre-processing data, since it's a different type of function.
   - I'm not sure there's any benefit of structuring this as a class, it seems like it just contains a function, it seems cleaner to let it be a function.

db/filters/base.py
silentninja: I am assuming these are like validation errors rather than something which shouldn't have occurred. If that is the case I think we should be using generic `Exception` or some other subclass of `Exception`. Assertion errors by definition are used when an unexpected state or situation occurs due to fault of a developer. `ValueError` seems to a much better options

silentninja:
> I'm not sure there's any benefit of structuring this as a class, it seems like it just contains a function, it seems cleaner to let it be a function.

How would the suggestions work in case of a function

silentninja:
@dmos62 Can the number parameter be derived or reflected rather than being hardcoded, similar to `sqlalchemy-filters`

dmos62:
@kgodey structuring these expressions as classes let's us define the identifier, display name and suggestions together with the function. It's not really about a hierarchy of classes like it was in the previous revision. It's just a convenient way to define these things together.

It's a good observation that there's a few different kinds of expressions here and putting them all in the same place is confusing. We could split the expression definitions into different files (namespaces). At the same time, note that user developers won't import or instantiate these classes, they'll just define new ones, and the API will pick them up.

@silentninja yes, this can be automated. I wrote it in a way that parameter count suggestions could be done automatically. But, then I realized that automating that doesn't really make sense, since you'd still write other suggestions yourself and writing the parameter count suggestions is trivial. In other words, automating this single suggestion would introduce magic and make the mechanism less transparent, without helping the user. Or, maybe it would be helpful, but is the additional code required justified? I'm leaning towards leaving it to be manual, like the rest of the suggestions.

dmos62:
To anyone reviewing at this point:

- the `db.filters` namespace (including `./base`, `./suggestions` and `./operations/*`) is pretty mature; the `dataclasses` usage is still not final; I can envision reimplementing `db.filters.base` in named tuples;
- tests, and the `mathesar` namespace is not up to date at all (should not be reviewed).

dmos62:
> a function in the db layer that takes a DB function name as input, and returns the signature(s) for that name

@mathemancer that's awesome; wasn't aware of the possibility. Yeah, go for it. Feel free to be as involved as you like. I definitely could use your help figuring out the SQL mechanics for that.

db/filters/base.py
dmos62: @silentninja thanks, I wasn't aware that `assert` calls expand to an `if __debug__:` branch 🤦 I was thinking of an `assert`ion as a general purpose assertion.

dmos62:
> - I think the "preprocessing" operations should be separated from the binary-returning direct filters, and not thought of as preprocessing.
>    - These are just formatting and other operations, and we'll be using them to generate user-visible output as much or more as we'll use them in filters eventually.

I've been thinking of all of these as just functions (or expressions, from the POV of assembling a filter expression) and it's pretty powerful, since instead of having a filter that's `URIAuthorityStartsWith`, you can have the functions `ExtractURIAuthority` and `StartsWith`. Even things like a list (as in the right-hand parameter to `IN`) or a column reference are expressions. This uniformity is pretty cool.

Putting all that in the filtering namespace is not accurate, since these can be used in other contexts, but I really like thinking of all these SQL functions (they are SQL functions at least in the abstract sense) as fundamentally the same. Their major difference is just the return type. I wonder what an elegant and future-conscious solution would be.

silentninja:
> It's possible to reflect the parameter information, and I think we should do so. The classes would provide a map from the API name and label to a function name (or operator, but see below).

@mathemancer  Were you trying to mention the Postgres [function to get function information](https://www.postgresql.org/docs/9.2/functions-info.html)?. They can be used to get the required details about a function like the parameters accepted along with the type and return type of the function.

Instead of building filters for each datatype, we can infer the type and show the applicable function based on the datatype of the previous resulting function. These could also be helpful for getting user-created functions, instead of having the user tinker with the code to create a custom filter.

The exhaustive list might be a problem though

mathemancer:
> If it's a lot, then it might be worth it. By a lot, I mean could it give us information that's a bit taxing to manually input, like, not only the number of parameters, but what kind of parameters the function takes?

Remember that functions can be (and often are) overloaded in PostgreSQL.  You can query (using a combinations of joins and a few of the system information functions that @silentninja mentioned) with all information about a function(s) that you could want:
```
********* QUERY **********
SELECT n.nspname as "Schema",
  p.proname as "Name",
  pg_catalog.pg_get_function_result(p.oid) as "Result data type",
  pg_catalog.pg_get_function_arguments(p.oid) as "Argument data types",
 CASE p.prokind
  WHEN 'a' THEN 'agg'
  WHEN 'w' THEN 'window'
  WHEN 'p' THEN 'proc'
  ELSE 'func'
 END as "Type"
FROM pg_catalog.pg_proc p
     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = p.pronamespace
WHERE p.proname OPERATOR(pg_catalog.~) '^(cast_to_boolean)$' COLLATE pg_catalog.default
  AND n.nspname OPERATOR(pg_catalog.~) '^(mathesar_types)$' COLLATE pg_catalog.default
ORDER BY 1, 2, 4;
**************************

                                List of functions
     Schema     |      Name       | Result data type | Argument data types | Type
----------------+-----------------+------------------+---------------------+------
 mathesar_types | cast_to_boolean | boolean          | bigint              | func
 mathesar_types | cast_to_boolean | boolean          | boolean             | func
 mathesar_types | cast_to_boolean | boolean          | character           | func
 mathesar_types | cast_to_boolean | boolean          | character varying   | func
 mathesar_types | cast_to_boolean | boolean          | double precision    | func
 mathesar_types | cast_to_boolean | boolean          | integer             | func
 mathesar_types | cast_to_boolean | boolean          | numeric             | func
 mathesar_types | cast_to_boolean | boolean          | real                | func
 mathesar_types | cast_to_boolean | boolean          | smallint            | func
 mathesar_types | cast_to_boolean | boolean          | text                | func
(10 rows)
```
There's lots more info we can get, but that's the basics.  Other stuff that's not too hard:
- function volatility
- whether the function is safe for parallel processing
- the function owner and permission info
- the language of the function, and
- the actual function body definition.

silentninja:
> wrapping idea (e.g., make a LESS_THAN function that returns whether a < b)

Wouldn't this mean users would have to search for the function name rather than using the `<` symbol or will this be replaced by an alias. I  prefer using function name rather than remembering the symbol especially when dealing with geometry or array operators

dmos62:
> I like this idea. I also agree with the conception of filters as "just functions". Maybe I was misunderstanding: I just wanted to avoid tossing random other functions in with the filters. I'm okay with the other way around.

I'll then be going ahead with the `db.functions` idea. That is, moving all expressions/functions out of `db.filters` into `db.functions` (and/or `mathesar.functions`). We don't consider them as specific to filtering anymore.

> What did you think of the wrapping idea

I think it's essential to get reflection working.

As we discussed in Matrix with Brent, if everyone agrees, I would like to put automatic reflection (automatic suggestion generation through Postgres reflection) off until later and use manual suggestions for now. Manual as in manually adding suggestions for parameter counts and parameter types, etc. to the `Expression` subclasses as seen in the latest code examples.

The rationale is that manual suggestions should be very easy to implement, compared to automatic suggestions, and not require that much manual effort when adding a new function. Also, automatic suggestions can be implemented later, if manual suggestions do not pan out. The way I see automatic suggestions working, it should be possible to "bolt" them on at any time in the future.

The specific disadvantages of automatic suggestions I currently see are:

- Significantly more work to implement than manual suggestions (on the backend; frontend shouldn't care);
- Requires initialization of suggestions at least once per sessions (i.e. automatically reflect SQL function signatures, which suggestions depend on), which introduces more state into Mathesar;
- Less transparent, less simple: you couldn't tell what the suggestions would be for a function just by looking at its class definition: as opposed to manual suggestions.

>  >    wrapping idea (e.g., make a LESS_THAN function that returns whether a < b)
>
> Wouldn't this mean users would have to search for the function name rather than using the < symbol or will this be replaced by an alias. I prefer using function name rather than remembering the symbol especially when dealing with geometry or array operators

I think that what Brent meant is to wrap SQL operator "functions" like `x IN ys` into actual SQL functions like `WRAPPED_IN(x, ys)`, because the former's signature can't be reflected (via Postgres), while the latter's can.

--------

@mathemancer @silentninja @kgodey do above decisions make sense? If you don't have specific comments, thumbs up is enough.

kgodey:
Sounds fine to me. Please open a separate ticket for implementing automatic suggestions once you're done with this one so that we don't forget about it. We can decide if/when to prioritize it later.

If you're discussing code on Matrix, please try to keep those discussions in public channels for visibility. We can set up a backend channel if it's helpful.

silentninja:
I am okay with a manual suggestion as of now, provided it automatic suggestion can be a drop in replacement or can be replaced with minimal changes to the interface

dmos62:
@kgodey I'll open a ticket for automatic suggestions after I flesh out this refactor some more. I'll then be able to be more specific. It's on my todo list.

dmos62:
@mathemancer could you expand on this comment:

> The classes should live in the mathesar namespace, since they're providing service-level metadata

I'm not sure I'm aware of all the reasons to have this in the service layer.

I've renamed `Expression` to `Function` (instances referred to as `db_function`) and moved to `db.functions` (for the time being).

It's not clear to me that we should have these under the `mathesar` namespace. There's some information in these classes (the display name property `name`) that does allude to a service layer (though maybe not necessarily), but the rest seems primitive enough to be in the `db` layer.

Thoughts?

mathemancer:
> It's not clear to me that we should have these under the `mathesar` namespace. There's some information in these classes (the display name property `name`) that does allude to a service layer (though maybe not necessarily), but the rest seems primitive enough to be in the `db` layer.

The eventual goal is to publish the `db` layer as a separate library.  Things there should make sense in the context of a library you'd import to, e.g., write a script that automatically splits a table into 1NF or 2NF.  Or maybe (in the case of filters) a script that automatically generates a report from your data once per day.  Perhaps I was distracted by the `name` property, but these classes (or at least major parts of them) seem focused on representing filters as nested objects (so they can be then represented via JSON).  At a lower level, these should be functions, not objects (since a filter is a function in the end).  So, the apparatus to represent them as objects feels like a service concern to me.  I'm amenable to discussion on this though (as on most things).  

One option is to put the parts that _do_ make sense in a non-service library in the `db` layer, and then call those from the service layer.  See `mathesar/database` for some other things where this pattern made sense to us at some point.

silentninja:
@mathemancer In the near future, I assume we would be moving towards reflecting function information straight from database, so we would need some way to represent the details extracted from the text information returned by the reflection queries, in that case, wouldn't it be better if they are represented as classes

mathemancer:
> @mathemancer In the near future, I assume we would be moving towards reflecting function information straight from database, so we would need some way to represent the details extracted from the text information returned by the reflection queries, in that case, wouldn't it be better if they are represented as classes. I think representing reflected data should be a db concern

How would that be useful in a script?  It seems like it would mostly be useful for building an interactive service?

dmos62:
@mathemancer I'm not sure it's a good idea to represent an SQL function solely with a plain Python function (that returns an SQLAlchemy expression), simply because you probably want to include some metadata, like details about it's signature.

Hints and display name are pretty service-layery. I think I'll move these to the `mathesar` namespace.

dmos62:
Request for feedback:

Currently the `DbFunction` (previously known as `Predicate` and `Expression`) classes look like this:

```python
class DbFunction(ABC):
    id = None
    name = None
    hints = None

    def __init__(self, parameters):
        if self.id is None:
            raise ValueError('DbFunction subclasses must define an ID.')
        if self.name is None:
            raise ValueError('DbFunction subclasses must define a name.')
        self.parameters = parameters

    @property
    def referenced_columns(self):
        """Walks the expression tree, collecting referenced columns.
        Useful when checking if all referenced columns are present in the queried relation."""
        columns = set([])
        for parameter in self.parameters:
            if isinstance(parameter, ColumnReference):
                columns.add(parameter.column)
            elif isinstance(parameter, DbFunction):
                columns.update(parameter.referenced_columns)
        return columns

    @staticmethod
    @abstractmethod
    def to_sa_expression():
        return None


class ColumnReference(DbFunction):
    id = 'column_reference'
    name = 'Column Reference'
    hints = tuple([
        hints.parameter_count(1),
        hints.parameter(1, hints.column),
    ])

    @property
    def column(self):
        return self.parameters[0]

    @staticmethod
    def to_sa_expression(p):
        return column(p)


class List(DbFunction):
    id = 'list'
    name = 'List'

    @staticmethod
    def to_sa_expression(*ps):
        return list(ps)


class Empty(DbFunction):
    id = 'empty'
    name = 'Empty'
    hints = tuple([
        hints.returns(hints.boolean),
        hints.parameter_count(1),
    ])

    @staticmethod
    def to_sa_expression(p):
        return p.is_(None)
```

`DbFunction` subclasses are exposed via the new `/api/v0/functions/` endpoint:

```json
[
    {
        "id": "and",
        "name": "And",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            }
        ]
    },
    {
        "id": "column_reference",
        "name": "Column Reference",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "parameter",
                "index": 1,
                "hints": [
                    {
                        "id": "column"
                    }
                ]
            }
        ]
    },
    {
        "id": "empty",
        "name": "Empty",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 1
            }
        ]
    },
    {
        "id": "equal",
        "name": "Equal",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            }
        ]
    },
    {
        "id": "extract_uri_authority",
        "name": "Extract URI Authority",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "parameter",
                "index": 1,
                "hints": [
                    {
                        "id": "uri"
                    }
                ]
            }
        ]
    },
    {
        "id": "greater",
        "name": "Greater",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "comparable"
                    }
                ]
            }
        ]
    },
    {
        "id": "in",
        "name": "In",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "parameter",
                "index": 2,
                "hints": [
                    {
                        "id": "array"
                    }
                ]
            }
        ]
    },
    {
        "id": "lesser",
        "name": "Lesser",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "comparable"
                    }
                ]
            }
        ]
    },
    {
        "id": "list",
        "name": "List",
        "hints": null
    },
    {
        "id": "not",
        "name": "Not",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 1
            }
        ]
    },
    {
        "id": "or",
        "name": "Or",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            }
        ]
    },
    {
        "id": "starts_with",
        "name": "Starts With",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "string_like"
                    }
                ]
            }
        ]
    },
    {
        "id": "to_lowercase",
        "name": "To Lowercase",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "string_like"
                    }
                ]
            }
        ]
    }
]
```

The frontend will look at the available functions via above endpoint; it will look at the hints defined on those functions and decide how to suggest/constrict user's choices when constructing a database expression.

Suppose the user wants to define a filter. The frontend might make sure that the top-level function used returns a boolean (a filter is an expression that resolves to a boolean), then that the nested expressions return types compatible with that top-level function's parameter hints. An example filter expression in JSON might look like this:

```json
{"in": [
  {"column_reference": ["column1"]}, 
  {"list": [
    {"column_reference": ["column2"]}, 
    {"to_lowercase": [
      {"column_reference": ["column3"]}, 
    ]},
    {"literal": ["foo"]},
    {"literal": ["bar"]},
  ]}, 
]}
```

This would be equivalent to:

```python
In([
  ColumnReference(["column1"]),
  List([
    ColumnReference(["column2"]),
    ToLowercase([ColumnReference(["column3"])),
    Literal(["foo"]),
    Literal(["bar"]),
  ]),
])
```

Which would be equivalent to following SQL:

```sql
column1 IN (column2, LOWER(column3), "foo", "bar")
```

On the backend, database function expressions will receive a minimum of checks. The aim is to constrict the user as little as possible. The frontend will be able to guide the user's choices by using the hint system (hints were previously called suggestions; using hints because shorter). The hint system is very simple, but expressive. Here's the definition of hints on the backend:

```
from frozendict import frozendict


def _make_hint(id, **rest):
    return frozendict({"id": id, **rest})


def parameter_count(count):
    return _make_hint("parameter_count", count=count)


def parameter(index, *hints):
    return _make_hint("parameter", index=index, hints=hints)


def all_parameters(*hints):
    return _make_hint("all_parameters", hints=hints)


def returns(*hints):
    return _make_hint("returns", hints=hints)


boolean = _make_hint("boolean")


comparable = _make_hint("comparable")


column = _make_hint("column")


array = _make_hint("array")


string_like = _make_hint("string_like")


uri = _make_hint("uri")
```

This allows the backend to specify the number of parameters expected (where appropriate), to define hints for all parameters and/or for individual parameters, and to define hints for the return type. New hints are trivial to add.

Previous major revision of this refactor didn't require the frontend to know anything about the filter constructs. This revision does not require knowing anything about the functions, but it does require knowing about the hint system. That is the price of the simplicity and flexibility offered. It will have to be hardcoded in the frontend that a filter expression is a nesting of database functions where the top-level one has the hint `{"returns": ["boolean"]}`.

When it will come to mixing and matching database functions to data types, we'll define relevant hints on the data types. So the frontend, having looked up database functions and data types, will see that a `BOOLEAN` Postgres type has the `{"id": "boolean"}` hint (equivalent of `db.functions.hints.boolean`) and so will know that it should be a safe input to a function where the parameter has the same hint (`{"id": "boolean"}`).

@kgodey @mathemancer @silentninja does this make sense?

kgodey:
I think this all makes sense to me, I like how extensible and simple it is.

The only things that I'd like to add is:
- There should be some way for the frontend to know which filters we're supporting in the UI for each data type. I'd recommend putting a list of filters each type supports in the the `/api/v0/databases/<id>/types/` endpoint (which should actually just be `/api/v0/types/` per discussion in #974, now that I think of it, but that can be a separate issue). This list can just be a pointer to the relevant function in `/api/v0/functions/`.
- Please add documentation on how this system is structured and how to add new filters in the [Architecture](https://wiki.mathesar.org/en/engineering/architecture) section of the wiki once this is merged.

mathemancer:
Overall, I think this makes sense.  Looking at this, I'm convinced that it does make sense to have the hints, etc. in with the `to_sa_expression` part.  Since this lets users construct fairly general make sure to wrap any string (or data) that we're passing to `to_sa_expression` in a `text` blob.  It would be good if that was somehow handled generally to avoid relying on every future implementer getting it right for every such class.

@kgodey I thought we agreed (in some other spot on GH) that filters aren't (in general) subordinate to types, but rather the other way around.  For example, I could filter on whether a string in column A is longer than the integer in column B.  The associated function would take a string-like and integer-like input, and return a boolean, so it wouldn't make sense to put it under either the `TEXT` or `INTEGER` types.  Rather, it makes sense to decide you want to do some sort of function like that, and then choose the appropriate columns for arguments.  Maybe a compromise would be to be able to filter the list of functions to restrict to those that involve a given type as at least one argument?

The same issues apply to groupings for similar reasons.  I think it makes more sense to keep the filters and other functions under some `/.../functions/` endpoint.

silentninja:
I agree with @mathemancer suggestion to have it under `/function` endpoint and not in `types`.

@dmos62 Looks good, I got a few questions:
1. How would overloaded functions work?
2. what does `all_parameter` mean, is it used as a blanket to say any `comparable` or any `string`?

dmos62:
@kgodey

> There should be some way for the frontend to know which filters we're supporting in the UI for each data type. 

Yeah, definitely. Hints fill that role. To say that a function can take strings as input, you give it a hint along the lines of `hints.parameter(1, hints.string_like)` (could be read "I'm hinting that the first parameter is a string"). Then we give the string-like DB types the same hint (that is not yet implemented, but it's straightforward, only a mapping). Imagine:

`GET /api/v0/db_types`:

```json
[
    {
        "id": "character varying",
        "name": "Character varying",
        "hints": [
            {
                "id": "string_like"
            }
        ]
    },
    [...]
]
```

So, when frontend queries both the `/functions` and the `/db_types` endpoints, it has all the information necessary to tell which functions should work with which data types.

@silentninja 

> How would overloaded functions work?

This only supports a single signature per function. If there's a database function with multiple signatures and we'd like to expose them, we'd do best to wrap them in separate `DbFunction` subclasses with different hint sets each.

Note that this mechanism does not actually enforce following the hints. As far as `db.functions` is concerned you can attempt to call and compose `DbFunction` subclasses however you like and it will attempt to not fail or at least fail as late as possible (preferrably during actual evaluation by Postgres). This is in the spirit of allowing the user as much freedom as possible.

This is also a pain point of this concept, in my mind, since it might be difficult to fail gracefully or to show nice error messages.

> what does all_parameter mean [..]?

`hints.parameter(1, *some_set_of_hints)` says parameter at index 1 has the hints in `some_set_of_hints` applied. And, `hints.all_parameters(*some_set_of_hints)` says all parameters have the hints in `some_set_of_hints` applied. When both these hints are provided at the same time, I'd say that the more specific `hints.parameter` should take precedent for the specific parameter it is targeting, but we might want to look into this specific scenario some more if it comes up in a practical setting. For example, CSS does inheritance and then provides things like `hints.unset(hints.some_hint_that_i_dont_want_to_inherit)` in the more specific rule to specify when to not inherit.

kgodey:
> So, when frontend queries both the `/functions` and the `/db_types` endpoints, it has all the information necessary to tell which functions should work with which data types.

@dmos62 Okay, that makes sense. Please open a separate issue for the frontend to switch to the new APIs for filters, we can use it to discuss the API spec with the frontend team and make sure it works for them.

dmos62:
Closing this PR. Splitting it into two PRs. First, there will be a PR that introduces the new function API (borne out of iterating over this PR). Then, there will be a PR that replaces the old filtering API with the new function API.

PR #941, opened 2022-01-05
Add more column default details

### LINKED ISSUE #933: Default value of column should have additional information to identify if default value is auto-generated
## Description
When the default value is auto-generated, the returned value is a string representing the procedure call.

Eg., for all PKs for the default id column, the value for `default` is `"nextval('\"public\".data_types_id_seq'::regclass)"`.

We need additional information to help the client differentiate this from a normal string value.

## Expected behavior

It would be better if default returned an object, something similar to:
```
default: {
   type: 'constant' // auto-generated or constant
   value: 'some default value'
}
```

For auto-generated values, we could send additional information based on how we intend to support formulas, procedures and functions in the future. For now, we can just send the same value and on the frontend the input can be disabled.

mathemancer:
Fixes #933 

Now, the form for the "default" value for columns is as follows:

```
...
"default": {
    "default_value": <some val>,
    "is_dynamic": <true|false>
}
...
```

**Technical details**

This change also applies for setting the default when creating or updating columns.  The "is_dynamic" field is read-only and can't be set.  The "default_value" field is required whenever "default" is non-null.

Because this is a new form for the default value for columns, I expect this to be a breaking change.

Old form:

```
...
"default": <some_val>
...
```

The new form is above.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/base.py
+++ b/db/columns/base.py
@@ -1,7 +1,9 @@
 from sqlalchemy import Column, ForeignKey, inspect
 
 from db.columns.defaults import TYPE, PRIMARY_KEY, NULLABLE, DEFAULT_COLUMNS
-from db.columns.operations.select import get_column_default, get_column_index_from_name
+from db.columns.operations.select import (
+    get_column_default, get_column_default_dict, get_column_index_from_name
+)
 from db.tables.operations.select import get_oid_from_table
 from db.types.operations.cast import get_full_cast_map
 
@@ -80,6 +82,16 @@ class MathesarColumn(Column):
         return None
 
     @property
+    def table_oid(self):
+        if self.table_ is not None:
+            oid = get_oid_from_table(
+                self.table_.name, self.table_.schema, self.engine
+            )
+        else:
+            oid = None
+        return oid
+
+    @property
     def is_default(self):
         default_def = DEFAULT_COLUMNS.get(self.name, False)
         return (
@@ -120,22 +132,29 @@ class MathesarColumn(Column):
                 and self.table_ is not None
                 and inspect(self.engine).has_table(self.table_.name, schema=self.table_.schema)
         ):
-            table_oid = get_oid_from_table(
-                self.table_.name, self.table_.schema, self.engine
-            )
             return get_column_index_from_name(
-                table_oid,
+                self.table_oid,
                 self.name,
                 self.engine
             )
 
     @property
+    def column_default_dict(self):
+        if self.table_ is None:
+            return
+        default_dict = get_column_default_dict(
+            self.table_oid, self.column_index, self.engine
+        )
+        if default_dict:
+            return {
+                'is_dynamic': default_dict['is_dynamic'],
+                'default_value': default_dict['default_value']
+            }
+
+    @property
     def default_value(self):
         if self.table_ is not None:
-            table_oid = get_oid_from_table(
-                self.table_.name, self.table_.schema, self.engine
-            )
-            return get_column_default(table_oid, self.column_index, self.engine)
+            return get_column_default(self.table_oid, self.column_index, self.engine)
 
     @property
     def plain_type(self):
--- a/db/columns/base.py
+++ b/db/columns/base.py
@@ -148,7 +148,7 @@ class MathesarColumn(Column):
         if default_dict:
             return {
                 'is_dynamic': default_dict['is_dynamic'],
-                'default_value': default_dict['default_value']
+                'value': default_dict['value']
             }
 
     @property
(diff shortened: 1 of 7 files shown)


db/columns/base.py
```diff
@@ -120,22 +132,29 @@ def column_index(self):
...
+    def column_default_dict(self):
+        if self.table_ is None:
+            return
+        default_dict = get_column_default_dict(
+            self.table_oid, self.column_index, self.engine
+        )
+        if default_dict:
+            return {
+                'is_dynamic': default_dict['is_dynamic'],
+                'default_value': default_dict['default_value']
```
kgodey, line 151: I would name this `value`, since it's already in a `default` key.

kgodey (changes requested):
@mathemancer In local testing, I was able to update the `id` column with a primary key constraint to use a static default value.

`PATCH` request:
```
{
    "default": {
        "default_value": 3,
    }
}
```

I don't think we should be able to change dynamic defaults to be static via the API.

---

Also, it might make sense to return an error if the client attempts to a changed value for `is_dynamic`, since silently ignoring it could be confusing.

mathemancer:
> I don't think we should be able to change dynamic defaults to be static via the API.

Agreed, but this is the previous behavior, so I think it should be a separate bug.  I'd like to do it separately to keep this PR minimal if possible.

mathemancer:
> Also, it might make sense to return an error if the client attempts to a changed value for is_dynamic, since silently ignoring it could be confusing.

I did some reading, and this seems to be the intended behavior for `read_only` fields.  Wouldn't this imply changing such behavior for every `read_only` field in the API?  For example, the same thing happens if you try to `PATCH` "valid_target_types" to add/remove some (I could actually envision a user attempting this).  I think it should be consistent for our API, and if we want to change that for this field, we should probably create a separate (meta) issue to track all the spots we'll need to change.  From a quick look, we'll need to deal with both the columns endpoint and the data files endpoint.  There might be other spots with the same issue.  @kgodey let me know what you think.

dmos62 (approved):

db/columns/operations/select.py
```diff
@@ -84,43 +84,35 @@ def get_column_name_from_attnum(table_oid, attnum, engine, connection_to_use=Non
...
-            executed_constant=execute_statement(
-                engine,
-                select(text(default_dict['sql_text'])),
-                connection_to_use
-            ).first()[0]
-        )
-    return default_dict
+        default_value = execute_statement(
+            engine,
+            select(text(sql_text)),
```
silentninja, line 104: Is it possible to set a `sql command` as the default char value with the intention of causing a SQL injection? Since a cast accompanies it, I am assuming it gets converted into a value, but not sure if this can be bypassed

silentninja:
> Also, it might make sense to return an error if the client attempts to a changed value for is_dynamic, since silently ignoring it could be confusing.

I think, writes for `read_only` fields should be ignored as they are treated as a hidden field in an `html form` rather than throwing up an error as this would mean we would need to have the same behaviour for any field that is not specified but passed in a payload. This should be dealt with on the client-side  either as documentation or an sdk

db/columns/operations/select.py
mathemancer: In this specific instance, I think we're safe since the default-setting mechanism seems to sanitize input (wrapping in quotes and casting).  However, I'm not 100% sure.  I _definitely_ think we need to figure out some automated audit for SQL injection problems with our app.

mathemancer: On a side note:  I consider this a beta-or-later problem, since our initial users should be considered "friendly", i.e., they own the DB they're running queries against.

kgodey: Can we create an issue to track potential SQL injection issues?

kgodey:
> Also, @kgodey what do you think about altering dynamic defaults to be static for non-default columns? On the one hand, there are good reasons to do that sometimes. On the other hand, if they do it mistakenly, they won't be able to recover through the Mathesar UI, since we can't create dynamic defaults.

I commented on this in #956, I don't think it should have anything do with default columns.

> I did some reading, and this seems to be the intended behavior for `read_only` fields.  Wouldn't this imply changing such behavior for every `read_only` field in the API?  For example, the same thing happens if you try to `PATCH` "valid_target_types" to add/remove some (I could actually envision a user attempting this).  I think it should be consistent for our API, and if we want to change that for this field, we should probably create a separate (meta) issue to track all the spots we'll need to change.  From a quick look, we'll need to deal with both the columns endpoint and the data files endpoint.  There might be other spots with the same issue.  @kgodey let me know what you think.

If this is the intended behavior and our API is already consistent with that, then I don't think we need to make any changes.

kgodey (approved):
Approved, my concerns were addressed and #956 can be handled separately.

kgodey:
I'm going to go ahead and merge since @dmos62 approved and @silentninja didn't request changes.

db/columns/operations/select.py
mathemancer: > Can we create an issue to track potential SQL injection issues?

#958

Issue #928, opened 2022-01-03
Wait until user is done editing a cell before saving


seancolsen:
## Current behavior

1. Double click a cell to enter edit mode.
1. Modify the value by typing something new.
1. Wait 1 second.
1. Expect nothing.
1. Observe the value of the cell is saved.
1. Press `Esc`.
1. Mode changes from edit to select.
1. Expect the cell value to be its original value before edits began, because I pressed `Esc`.
1. Observe the cell value to be changed.

## Proposed behavior

- Follow the same UX that spreadsheets use. LibreOffice Calc and Google Sheets both behave as follows... The cell value does not get saved until you leave edit mode with one of the following options:

    - `Enter`
    - `Shift+Enter`
    - `Tab`
    - `Shift+Tab`
    - a click outside the cell

## Notes

- I think this is relevant to our recent [discussion](https://github.com/centerofci/mathesar-wiki/pull/13#discussion_r770537811) about data integrity during editing. If our users expect the same behavior that I expect, then they might be annoyed that pressing `Esc` did not "undo" (so-to-speak) the edits they made to a cell.

kgodey:
This sounds good to me.

seancolsen:
Thanks @kgodey. I'm marking this as ready now that we have support from one other person.

@ghislaineguerin @pavish you two might still like to weigh in if you have any concerns here.

seancolsen:
#1080 seems to have fixed this, so I'm closing it.

@pavish I noticed you marked that PR as "related to" this issue instead of "fixes" this issue. Is that just because of the difference in behavior with regard to the Esc key? (If so, I'm planning to open a separate issue for that where we can discuss it.)

PR #961, opened 2022-01-08
Display null while editing a null cell

### LINKED ISSUE #939: Display NULL while editing a NULL cell
## Current behavior

1. Find a NULL cell.
1. Double click to enter edit mode.
1. Observe that `NULL` no longer displays when in edit mode, leaving the user with no way to distinguish NULL from empty string while editing.

## Desired behavior

1. Double click to enter edit mode on a NULL cell.
1. `NULL` appears after the cursor, styled the same as when viewing the cell in select mode.
1. Begin entering text.
1. Observe that `NULL` disappears, being replaced by the text entered.
1. Delete all text entered.
1. Observe the cell to be blank (no `NULL` displayed) indicating the value is an empty string.

## Notes

- In edit mode, the user won't have a way to set the cell to NULL. That's okay for now. #930 and #936 will give users the capability to set a cell to NULL (but only from select mode).
- We should also style the NULL value the same in edit mode as it is styled in select mode. See discussion within #778 to learn more about our thinking on style.

## See also

- See #929 for a bug that you may hit while working on this ticket. (Note that you can still preserve your NULL value by clicking outside the cell instead of pressing Esc.)

## Additional context

See [NULL Value Specs, scenario 2b](https://wiki.mathesar.org/design/specs/specs-null-values#steps-for-2b)

gagan2005:
Fixes #939 

The cells which have null values will now show null when we edit the value in the cell.Previously the cell will show up as empty when we start editing it

**Technical details**
I added a flag to check whether NULL needs to be displayed or not. When we need to display null , we shift the Input behind the null element. If  user enters any input the flag is set to false and the input element shifts back to in front of null element  

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
Thanks for this PR, @gagan2005. I'll review it sometime within the next few days.

seancolsen:
@kgodey can you help troubleshoot the failing test in this PR?

kgodey:
@seancolsen It looks like the automatic add-to-project failed because this is based on a fork. I updated the GitHub action event to compensate in `master` https://github.com/centerofci/mathesar/commit/e90460b918011dd20471d3e5b8377a75e3d4438d and merged it in.

seancolsen:
@gagan2005 Thanks for this PR!

Using the `placeholder` attribute is a clever quick-fix, but this approach has some shortcomings particular to our needs here.

- In step 6 of the desired behavior in #939, we need to show an empty cell instead of the placeholder. Your changes _effectively_ satisfy that goal, but only as a consequence of #928 still being unresolved. You can simulate the behavior we're aiming for with that fix by commenting out the contents of `debounceAndSet` within `RowCell.svelte`. When the cell doesn't auto-save itself, then the placeholder continues to display after typing text into a NULL cell and then subsequently deleting that text.
- On a slightly lower-priority stylistic note, I'd like to style the NULL value the same during edit mode as it is styled during select mode. Apologies that I did not include a note about that in the original ticket. I've edited it now to mention this additional goal, and I've linked to a #778 where we discussed some of the reasons to display NULL as stylized wherever possible.

With the above two considerations in-mind. I'd recommend that instead of using the `placeholder` attribute, we conditionally render a `Null` component (already exists) and absolutely position it within the cell. Would you be up for taking another stab at this issue using that approach?

seancolsen (changes requested):
@gagan2005 Thanks for this PR!

Using the `placeholder` attribute is a clever quick-fix, but this approach has some shortcomings particular to our needs here.

- In step 6 of the desired behavior in #939, we need to show an empty cell instead of the placeholder. Your changes _effectively_ satisfy that goal, but only as a consequence of #928 still being unresolved. You can simulate the behavior we're aiming for with that fix by commenting out the contents of `debounceAndSet` within `RowCell.svelte`. When the cell doesn't auto-save itself, then the placeholder continues to display after typing text into a NULL cell and then subsequently deleting that text.
- On a slightly lower-priority stylistic note, I'd like to style the NULL value the same during edit mode as it is styled during select mode. Apologies that I did not include a note about that in the original ticket. I've edited it now to mention this additional goal, and I've linked to a #778 where we discussed some of the reasons to display NULL as stylized wherever possible.

With the above two considerations in-mind. I'd recommend that instead of using the `placeholder` attribute, we conditionally render a `Null` component (already exists) and absolutely position it within the cell. Would you be up for taking another stab at this issue using that approach?

gagan2005:
ok , I will make changes and Update this PR

gagan2005:
Sorry, could not work on it as I have been busy last few days. I will resume the work today and inform you if I need any help

gagan2005:
@seancolsen Earlier everything was fine but Now I can't get docker-compose up to work. It says arrow module not found. I checked the dockerfile, it is using  requirements.txt and the file has arrow module listed. I am unable to figure out what might be causing the issue

seancolsen:
@gagan2005 You might need to run `docker-compose up --build` 

Sometimes (though rarely) you'll need to re-build the container if after pulling in other changes.

If you're still having trouble, post a message in the "Code - General" [Matrix](https://wiki.mathesar.org/en/community/matrix) channel with the specific error message you're seeing.

gagan2005:
@seancolsen Please take a look and inform me if any other changes are required.

seancolsen (changes requested):
@gagan2005 Thanks for continuing to work on this.

I like that I can see `NULL` stylized now during edit mode. However, we still need to address a few more things before merging this.

Even though we're not using the `placeholder` functionality (like you originally did), we're essentially trying to mimic it with a stylized placeholder that permanently disappears once the user has modified the value of the input. In other respects though, we are aiming for a UX that mirrors the placeholder behavior because users are familiar with it. As such, we need to make the following changes to this PR:

- Keep the `NULL` placeholder when the user presses a key that does not modify the value of the input.

    For example, if the user presses a key like `Up`, `Home`,  `PgDown`, `Delete`, `Shift` when the placeholder is shown, the placeholder will disappear. There are a lot of keys like that, and we don't want to have to list them all. Instead of the `keydown` event, we should probably be listening to the `input` event to determine the appropriate point at which we can remove the placeholder.

- Display the user's cursor before the placeholder, not after.

    Instead of shifting the input with `left: 54px;`, I'd recommend that we keep its current position and conditionally render an additional `<Null />` component absolutely positioned on top of the input.

seancolsen:
Thanks @gagan2005 

This is working well now.

I added 0a60b175296bfe49e329bbbf18c10de7ba453f21 to simplify the `hideNullElement` function for better readability, and I've set this PR to merge when all the tests pass.

seancolsen (approved):

Issue #292, opened 2021-06-28
Frontend component: Pagination improvements


pavish:
The followed needs to be handled:

- [x] Styling
- [x] ARIA, keyboard accessibility
- [ ] Tests
- [x] Documentation 

PRs that are related to this issue:
- https://github.com/centerofci/mathesar/pull/886
- https://github.com/centerofci/mathesar/pull/931
- https://github.com/centerofci/mathesar/pull/925
- https://github.com/centerofci/mathesar/pull/963

kgodey:
@pavish I think you could repurpose this issue to cover virtual scrolling. Please update the issue.

kgodey:
@pavish Since the infinite scrolling PR is already in review, I don't think there's a need to create an issue for it. If there's additional accessibility, documentation, or testing concerns, I think you're a better person to make issues for those than me.

agrawalanshul053:
I am trying to setup this repository in my local. I am using Windows OS. I have installed docker. I am facing difficulty in the following step.
Clone the repository and then copy the .env.example file to .env like so:
cp .env.example .env
Please help me.

pavish:
@agrawalanshul053 These commands are to copy files in a posix-like environment.

In windows, you can copy the `.env.example` to a new file and name it `.env`.

Alternatively, you can use your WSL (Windows subsystem for Linux) shell, and use posix-like commands.

silentninja:
@agrawalanshul053 This issue is for the frontend component, I have moved this to a new issue here #830. Please use it for further discussion

mr-gabe49:
@pavish I'll take the documentation

pavish:
Sure @mr-gabe49. Thanks!

mr-gabe49:
@pavish @seancolsen I would like some guidance on the behavior we are looking for on keyboard accessibility. I have seen different behaviors through the web: 

- Left/up and Right/down arrow change the current page
- Left/up and Right/down arrow change the focused page, then clicking enter would change the current page
- Other ideas: 
   -  double click left arrow sets current page to the first page
   -  eliminate left arrow when first page is active (same for right arrow and last page active). 

Let me know your thoughts. Thanks!

pavish:
@mr-gabe49 I prefer Left and Right arrow changing focused page and clicking enter would change current page. I would not use up and down since browsers by default use them for scrolling the webpage.

> double click left arrow sets current page to the first page

I think this would be more confusing and we can avoid this for now.

[...]

PR #963, opened 2022-01-10
Keyboard Accessibility Pagination Component

mr-gabe49:
Related to #292 

**Technical details**

- This PR adds keyboard accessibility to pagination component
- Adds an aria-label prop to the component
- Adds a small hover effect
- Disables options that can't be clicked

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
I'm going to let @pavish take this one since he already expressed some [opinions](https://github.com/centerofci/mathesar/issues/292#issuecomment-1005547793) and I have not formed any opinions about it.

pavish:
@mr-gabe49 I think you could remove the hovered state implementation (hoverIndex and hover class) and the left, right arrow handling.

I've added a comment on the issue: https://github.com/centerofci/mathesar/issues/292#issuecomment-1012462872

This gives a general set of guidelines to follow to add accessibility for the pagination component. I believe points 2, 3 and 4 need to be handled.

Let me know if the comment answers your questions.

mr-gabe49:
@Pavish Regarding point 4 from your [292 Comment](https://github.com/centerofci/mathesar/issues/292#issuecomment-1012462872),  point 3 of [Accessibility Matters](https://www.a11ymatters.com/pattern/pagination/) seems to suggest that `aria-selected` and `aria-current` could be used interchangeably depending on the presence of the `role`  attribute.

mathesar_ui/src/component-library/pagination/Pagination.svelte
```diff
@@ -112,39 +116,41 @@
...
             on:click={(e) => setPage(e, pageInfo.nextPageWindow)}
           >
-            <Icon class="ellipsis" data={faEllipsisH} />
+            <Icon class="ellipsis" tabindex="-1" data={faEllipsisH} />
             <Icon class="arrow" data={faAngleDoubleRight} />
-          </span>
+          </button>
         </li>
       {/if}
       <li class:active={currentPage === pageInfo.end}>
```
mr-gabe49, line 156: @pavish Is `class:active={currentPage === pageInfo.end}` necessary? Wouldn't this never be true since when `currentPage === pageInfo.end`, `pageInfo.end` will equal `pageCount`, and thus this block of code will not run?

pavish: Yes, that's right. This can be removed.

mathesar_ui/src/component-library/pagination/Pagination.svelte
```diff
@@ -74,28 +78,28 @@
...
         <li>
-          <span
+          <button
             tabindex="0"
             role="link"
             aria-label="Goto Page {pageInfo.prevPageWindow}"
             on:click={(e) => setPage(e, pageInfo.prevPageWindow)}
           >
-            <Icon class="ellipsis" data={faEllipsisH} />
+            <Icon class="ellipsis" tabindex="-1" data={faEllipsisH} />
```
pavish, line 100: Is there a reason why `tabindex` is set for icons?

mathesar_ui/src/component-library/pagination/Pagination.svelte
```diff
@@ -112,39 +116,41 @@
               : `Goto Page ${_page}`}
             aria-selected={currentPage === _page}
             on:click={(e) => setPage(e, _page)}
+            on:mousedown={(e) => e.preventDefault()}
```
pavish, line 119: What is the purpose of preventing default of mousedown event here?

pavish (changes requested):
@mr-gabe49 The PR looks pretty good.

I have a couple questions. Once resolved, we can merge this PR.

mathesar_ui/src/component-library/pagination/Pagination.svelte
mr-gabe49: @pavish Yes, when an ellipsis is shown and one uses the tab key to navigate the pagination component, this prevents the focus from disappearing for a single click. I suspect it disappears because it focuses the arrow, which is not shown on the page.  
The only thing that confuses me a little is that the only way it works is if `tabindex="-1"` is set on the ellipsis and not on the arrow which should be the one that is not focusable since it is not showing on the page. I tried doing it with CSS but didn't work.

mathesar_ui/src/component-library/pagination/Pagination.svelte
```diff
@@ -45,23 +48,24 @@
...
         >
-          <Icon data={faAngleLeft} />
-        </span>
+          <Icon data={faAngleLeft} tabindex="-1" />
+        </button>
       </li>
     {/if}
 
     {#if pageInfo.start > 1}
-      <li class:active={currentPage === pageInfo.start}>
```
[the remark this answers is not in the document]
pavish: Yes, this can be removed.

mathesar_ui/src/component-library/pagination/Pagination.svelte
pavish: I think you're worried that the user might get confused on which one is the active one. It would be better to change the css of focus state instead, so that there is clear visual distinction on which one is focused and which one is active.

For the focus state, it would be better to add an outline instead of changing border color.

mathesar_ui/src/component-library/pagination/Pagination.svelte
pavish: Let me try to understand this better. I think the following is the case:
1. Ellipsis is shown and user clicks on it (i.e the arrow).
2. The page number changes but ellipsis/arrow is now not present because it is not required. Eg., Ellipsis is present when active page is 7. Clicking on it changes active page to 2, and ellipsis is no longer present.
3. Since the last focused element is no longer present on the dom, the focus state is reset by the browser.

If you're referring to this problem, we should be resolving it by manually setting focus on the current active page when ellipsis/arrow disappears. In the above example, when we click on the ellipsis and page number is changed to 2, we should be doing `element.focus()` to focus the button with page number 2. You can add a `data-page` attribute to help query the element.

mr-gabe49: @pavish I'm not referring to this problem. I'll make a loom video showing you what it does, it will be easier. 
For the problem that you have mentioned, isn't that the standard behavior for [Ant Design](https://ant.design/components/pagination/)?

pavish: Thanks, I'll take a look at the loom video once you post it.

Regarding the problem I've described, Ant Design may not have handled it, but it'll be better for us to handle it. We don't have to do that in this PR. Please feel free to open an issue linking to the above comment and we can address it later.

mr-gabe49: @pavish [Video Explaining the problem I described](https://www.loom.com/share/49c31d28668543f9865a4da6c33d1c03)

mr-gabe49: @pavish Sounds good!

pavish: @mr-gabe49 Interesting! I didn't notice this behavior on Firefox earlier so I wasn't sure what the issue was. Seems like this issue is present in both Chrome and Safari.

This is because both Blink and Webkit find any svg with a focus event handler focusable. So this issue is present wherever we use icons on our application. Nice catch.

I don't think our `Icon` component needs to handle focus events or any event for that matter. Please remove all `on:<event>` handlers within our `Icon` component. That should fix this issue.

mr-gabe49: @pavish Sounds good, I'll make the changes.

mathesar_ui/src/component-library/pagination/Pagination.svelte
mr-gabe49: @pavish I'll take them out. [This video](https://www.loom.com/share/84ce82efe06b4d1690f8cfbdaccf7cc4) explains the issue I was  seeing.

mathesar_ui/src/component-library/pagination/Pagination.svelte
mr-gabe49: Created issue https://github.com/centerofci/mathesar/issues/1040#issue-1122334518

mathesar_ui/src/component-library/pagination/Pagination.svelte
pavish: @mr-gabe49 I think this should be fixed by separating out the css for active and focus states, as mentioned above.

pavish (approved):

PR #976, opened 2022-01-14
Implement renaming a column

### LINKED ISSUE #497: Implement renaming a column

seancolsen:
Fixes #497

## Notes

- This PR was originally #870 which contains significant discussion. I opened a new PR because the first one contained too many merge conflicts.
- This is marked as a draft because @pavish and I still need to discuss some implementation strategies around re-fetching store data.

## Demos and explanation

- [Basic column rename](https://www.loom.com/share/8b813dc217184b0899db4e08cc42b551)
- [Error scenarios when renaming a column](https://www.loom.com/share/05d1f2c76a1c47bcb19c03d62a5250bf)
- [Usability of renaming a column](https://www.loom.com/share/5e9ae39de5564b1c94bf8d3950c7035a)

## Issues

- The records within the column appear blank for a short time during the process of renaming. That's while we're awaiting `this.recordsData.fetch()` within `TabularData.ts`. I'm re-fetching the records because that's the easiest way to associate them with the new column name. There are some alternate approaches we could consider to improve this behavior. We could mutate the records store with the new column name. That seems somewhat messy though. We could modify the records API to associate cell values with column _id_ values instead of column names. For now though I think this behavior is acceptable.

- `ColumnsDataStore.rename` is inconsistent with other methods like `add` because it does not re-fetch the columns data after it renames. I don't like this inconsistency, but I'm not sure of the best way to resolve it.

    - After we rename a column, we _do_ need to re-fetch the columns data, but we also need to re-fetch the records data and the constraints data. We need to re-fetch the whole TabularData structure, and we don't (and shouldn't) have a way to do that from within the `Columns` class. So instead of re-fetching within `ColumnsDataStore.rename`, I'm re-fetching within `HeaderCell.svelte`.

    - Alternate approaches:

        - Re-fetch within `ColumnsDataStore.rename` and also call `TabularData.refresh` within `HeaderCell.svelte`. That's duplicate fetches. Not great for performance.
        - Re-fetch within `ColumnsDataStore.rename` and call `RecordsData.fetch` and `ConstraintsDataStore.fetch` from within `HeaderCell.svelte`. Messy because if we change the implementation of `TabularData.refresh` later on, it won't be apparent that we might also need to also change `HeaderCell.svelte`.

    - Several weeks ago @pavish and I chatted about the best re-fetch pattern to establish. I initially argued that store functions should _never_ re-fetch and that we should always do re-fetching at the view layer (as I've done in this PR) to give ourselves maximum granularity and control. We discussed several other potential patterns to employ and ultimately agreed to consistently re-fetch within store functions that update data and avoid re-fetching at the view layer. I think we might need to re-evaluate that pattern in light of this new example. As our codebase grows, I predict that we'll continue to need more granular control over the re-fetching behavior than our current pattern will allow.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
This isn't related to this specific issue, but are all the patterns that you and @pavish decide on for frontend code being documented somewhere for new contributors and/or future hires?

If not, I would suggest documenting them on the [Architecture](https://wiki.mathesar.org/engineering/architecture) section on the wiki.

seancolsen:
Thanks for the reminder @kgodey 

I think Pavish and I could both be doing a better job here. I'll try to keep this in-mind more as we have sync discussions that are not captured anywhere. We're planning one for early next week. 

In the mean time, I gave some thought to question of what documentation we might already lack. I came up with two ideas and asked Pavish for help within each.

- [Document front end coding standards for imports](https://github.com/centerofci/mathesar-wiki/issues/23)
- [Document front end coding standards for `.ts` vs `.d.ts`](https://github.com/centerofci/mathesar-wiki/issues/24)

I'll continue thinking on this and add some more documentation if I can. There's still quite a lot of flux in the front end, especially within our component library, component docs, and css, so at this stage I'm hesitant to start codifying rules for some of that code because it's continuing to shift regularly.

kgodey:
Thanks @seancolsen!

> There's still quite a lot of flux in the front end, especially within our component library, component docs, and css, so at this stage I'm hesitant to start codifying rules for some of that code because it's continuing to shift regularly.

At this stage, I think it would be most useful for you and @pavish to establish a pattern of documentation into your day-to-day workflows; it's okay if the documentation changes frequently. I think it's harder to remember to document decisions if you only do when it seems final in some way.

seancolsen:
@pavish This is ready for another look now.

In ec34b3c09fc7a87a4f73683b88826b1b09d8c5f4 I'm trying to implement the re-fetching approach we discussed yesterday. I know you said it'd be fine to create a function within TabularData to rename the column, but I figured I ought to try out the final pattern that we're seeking.

One issue with this approach is that `dispatch` in synchronous, so I can't await it. That means that the UI loading spinner completes too soon. Previously the loading spinner would continue to spin until all stores were re-fetched. After ec34b3c09fc7a87a4f73683b88826b1b09d8c5f4, the spinner completes as soon as the column is renamed but before the stores are re-fetched. I'd say that's a minor but noticeable regression over the column-renaming behavior before I added that commit.

So I also added 23f9392727ff5c44b9bcee99163842f7fc67cd66 which changes EventHandler to use async callbacks. This fixes the UI regression described above, but perhaps you might spot other concerns with this approach.

Additionally, in 23f9392727ff5c44b9bcee99163842f7fc67cd66 I:

- Changed `EventHandler.dispatch` so that it throws an error instead of silently returning if it can't find the event. I'm preferring an error in this case because it's a scenario I'd never expect to happen at runtime. Right? If so, I'd rather throw an error so that if we have a bug it doesn't go unnoticed.
- On a similar note, I changed `EventHandler.dispatch` so that it doesn't catch errors from within the callback function (directing them to the console). This way, if there's an error during re-fetch, that error will bubble up to the UI layer where the user will see it in a toast message.

mathesar_ui/src/stores/table-data/tabularData.ts
```diff
@@ -61,8 +61,9 @@ export class TabularData extends EventHandler {
     );
 
     this.metaParametersUnsubscriber = this.meta.metaParameters.subscribe(() => {
-      this.dispatch('paramsUpdated', this.parameterize());
+      void this.dispatch('paramsUpdated', this.parameterize());
     });
+    this.columnsDataStore.on('columnRenamed', () => this.refresh());
```
pavish, line 66: We also need to update the meta object. Perhaps have a method within meta to handle renames and call it from here.

This would not be essential after column name changes to id in our APIs, but until then I'd like the existing features not to break.

mathesar_ui/src/stores/table-data/columns.ts
```diff
@@ -172,6 +172,11 @@ export class ColumnsDataStore
     return column;
   }
 
+  async rename(id: Column['id'], newName: string): Promise<void> {
+    await this.api.update(id, { name: newName });
+    await this.dispatch('columnRenamed', id);
```
pavish, line 177: Currently when a column is renamed and the rename request succeeds, the values displayed on table for that particular column goes blank until the records request completes.

This would be fixed when the records endpoint uses column id instead of name. I just wanted to make a note of it here.

pavish (changes requested):
@seancolsen Nice work! I like the inline edit.

I have a few comments that can be handled before we merge this in.

mathesar_ui/src/component-library/common/utils/EventHandler.ts
```diff
@@ -15,14 +22,14 @@ export default class EventHandler {
...
-        console.error(`Failed to call a listener for ${eventName}`, err);
-      }
-    });
+  protected async dispatch(eventName: string, value: unknown): Promise<void> {
+    const callbacks = this.listeners.get(eventName);
+    if (!callbacks) {
+      throw new Error(`Unable to dispatch unknown event '${eventName}'.`);
+    }
+    await Promise.all(
+      [...callbacks.values()].map((callback) => callback(value)),
```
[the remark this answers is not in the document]
seancolsen: Right. Ok, my bad. I wasn't thinking about this correctly. Fixed in 9d30e3b0a4a0f8680aa1766e1eb6b2353c14a4a1

mathesar_ui/src/stores/table-data/tabularData.ts
seancolsen: > I'd like the existing features not to break.

Given that renaming a column is not an existing feature, what existing features will break by merging this PR?

Even if something small _does_ break, I'd rather not do the extra work now, given that we know we'll need to undo it soon.

seancolsen:

mathesar_ui/src/stores/table-data/tabularData.ts
pavish: I'm sorry, there aren't breaking features, I meant to resolve the delete bug that is already present. I see an issue for it here: https://github.com/centerofci/mathesar/issues/987. 

I'll resolve this and we can handle the issue there.

pavish (approved):

PR #980, opened 2022-01-17
Boolean DB type casting extension

### LINKED ISSUE #387: Handle `BOOLEAN` data type in the backend
This issue is to ensure that Mathesar can handle the Postgres `BOOLEAN` data type.

- Users can use the API to change a column to `BOOLEAN` if it's possible to do so.
- Automatic type inference during file import suggests `BOOLEAN` when it makes sense to do so. 

Most of this is already implemented, this issue is to review the current implementation, ensure it works, and see if we can make improvements to it.

**Additional Context** 

mathemancer:
Fixes #387 

This adds some strings that can be cast to boolean:
- yes -> true
- no -> false
- y -> true
- n -> false
- on -> true
- off -> false

**Technical details**

Note that this does _not_ restrict columns to have only two values for inferring boolean.  So, a column with '0', '1', 'true', 'TRUE', 'yes', 'off', ... would be inferred to be of boolean DB type.  This seems like the best compromise to make at the moment.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/tests/types/operations/test_cast.py
+++ b/db/tests/types/operations/test_cast.py
@@ -531,7 +531,9 @@ MASTER_DB_TYPE_MAP_SPEC = {
             },
             BOOLEAN: {
                 VALID: [
-                    ("true", True), ("false", False), ("t", True), ("f", False)
+                    ("true", True), ("false", False), ("t", True), ("f", False),
+                    ("yes", True), ("y", True), ("no", False), ("n", False),
+                    ("on", True), ("off", False),
                 ],
                 INVALID: ["cat"],
             },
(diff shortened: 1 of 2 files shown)


silentninja (commented):
Looks good to me!

silentninja (approved):

kgodey (approved):

silentninja:
@mathemancer I am able to cast string values('on', 'y') using the default cast, `SELECT 'on'::bool'` works, any reason why you didn't go for that?. I approved thinking that was not possible

mathemancer:
@silentninja It is possible for all strings we _do_ want to handle.  To control things a bit, I wanted to set aside separate "approved" strings (to avoid, e.g., `'of' -> false`).  For numbers, we need the custom behavior since the ways and types of numbers cast to booleans are a bit confusing in their default state.  I'm open to changing the behavior if we want later, but I think the current setup is the best compromise with the current constraints.

silentninja:
Makes sense, seems extensible too.

PR #988, opened 2022-01-19
Use composition not inheritance for EventHandler

seancolsen:
I wanted to start using some of the EventHandler code in continuing on #976 but I found the mental model of `ColumnsDataStore` _being_ an `EventHandler` to be quite unintuitive. This is a quick stab at converting that code to use composition instead of inheritance.

@pavish what do you think of this? We can scrap it if you're attached to using inheritance. I know we already discussed this topic previously (not able to find the link quickly). But I figured it would be easiest to continue the discussion by looking at something concrete. I recall you saying you were concerned that we'd need to "duplicate methods on different classes" or something to that effect. This PR demonstrates that we don't need to do that.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/common/utils/EventHandler.ts
+++ b/mathesar_ui/src/component-library/common/utils/EventHandler.ts
@@ -1,5 +1,5 @@
 export default class EventHandler {
-  protected listeners: Map<string, Set<(value?: unknown) => unknown>>;
+  private listeners: Map<string, Set<(value?: unknown) => unknown>>;
 
   constructor() {
     this.listeners = new Map();
@@ -15,7 +15,7 @@ export default class EventHandler {
     };
   }
 
-  protected dispatch(eventName: string, value: unknown): void {
+  dispatch(eventName: string, value: unknown): void {
     this.listeners?.get(eventName)?.forEach((entry) => {
       try {
         entry?.(value);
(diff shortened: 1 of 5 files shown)


pavish:
@seancolsen This would open up all of eventHandler's methods to be accessible outside of the scope it's intended to be used in.

For eg., This would allow `tab.tabularData.eventHandler.dispatch` and `tab.tabularData.eventHandler.destroy` to be called from anywhere, while we do not want them to be exposed outside of TabularData.

> but I found the mental model of ColumnsDataStore being an EventHandler to be quite unintuitive. 
> use composition instead of inheritance

I also generally prefer composition over inheritance, but this is a scenario where inheritance works much better. I would argue using inheritance here is more intuitive than composition based on how we'd expect events to work.

seancolsen:
Ok no worries.

> This would open up all of eventHandler's methods to be accessible outside of the scope it's intended to be used in.

I gave thought to that shortcoming too, but figured it wouldn't be problematic as the intended usage of `eventHandler` would be clear enough to other devs. But I can understand wanting more safety here.

To gain that safety without using inheritance, I would consider using a decorator. I feel somewhat ambivalent on decorators -- but I guess I just really hate inheritance! Not worth spending more time on this though. I don't want to distract us from larger priorities. I'll use the code that's already there.

> This is the duplication I was talking about.

Thanks for clarifying. Now I understand your opinion on this topic better.

PR #999, opened 2022-01-20
Improve line-height CSS

seancolsen:
@pavish why have we had `line-height: 1em;` set globally?

That CSS has some weird effects, including making underscores nearly invisible in tab names at certain zoom settings.

Adjusting the line-height is the best way to fix that issue I think, and to me it makes the most sense to unset it globally (as this PR does).

| Before | After |
| -- | -- |
| image | image |

I imagine we'll want to tweak a bunch of the typography stuff once we do our overall styling. If we do want to customizes the line height, MDN recommends [always using unitless values for line-height](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/templates/mathesar/base.html
+++ b/mathesar/templates/mathesar/base.html
@@ -23,7 +23,6 @@
         margin: 0;
         padding: 0;
         font-size: 14px;
-        line-height: 1em;
         color: #1E1E1E;
       }
       * {
--- a/mathesar/templates/mathesar/base.html
+++ b/mathesar/templates/mathesar/base.html
@@ -8,69 +8,84 @@
     <title>Mathesar - {% block title %}{% endblock %}</title>
 
     <!-- TODO: Specify Mathesar favicon. Using a dummy data URI until then. -->
-    <link rel="icon" href="data:;base64,iVBORw0KGgo=">
+    <link rel="icon" href="data:;base64,iVBORw0KGgo=" />
 
     <style type="text/css">
-      html, body {
+      html,
+      body {
         position: absolute;
         width: 100%;
         height: 100%;
         overflow: hidden;
-        font-family: -apple-system,system-ui,BlinkMacSystemFont,'Segoe UI',
-                      Roboto,Oxygen-Sans,Ubuntu,Cantarell,'Helvetica Neue',
-                      Helvetica, Arial, sans-serif,'Apple Color Emoji','Segoe UI Emoji','Segoe UI Symbol';
+        font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI",
+          Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", Helvetica,
+          Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
+          "Segoe UI Symbol";
         font-weight: 400;
         margin: 0;
         padding: 0;
         font-size: 14px;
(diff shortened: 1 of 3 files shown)


kgodey:
Perhaps we should add `mathesar/templates/` to the frontend list in our CODEOWNERS file?

pavish:
@kgodey 

> Perhaps we should add mathesar/templates/ to the frontend list in our CODEOWNERS file?

Yes, we should. We could also include `mathesar/urls.py` and `mathesar/views.py` since they also deal with the frontend urls and preloading.

pavish:
@seancolsen 
> Why have we had line-height: 1em; set globally?

Line height with no default varies oddly across browsers, leading to slight misalignment of elements in browsers we do not test on.

Setting a proper line height also depends on the font we use and we only use system fonts in the moment, which also makes it differ across operating systems.

Setting a line-height globally avoids this issue. We should be setting individual line heights to parts of code where we encounter issues as you've posted above. 

As of now, removing line-height has effects in other parts of the page, leading to misalignment of icons and extra leading heights for inputs.

We could change `1em` to `1` instead. 

We can come back to this when we do our overall styling.

pavish:
@kgodey 

> Those should be @centerofci/mathesar-maintainers since they're also Python code, backend should look at them too. Can you go ahead and make the update?

Raised https://github.com/centerofci/mathesar/pull/1000 for this.

seancolsen:
@pavish 

> Line height with no default varies oddly across browsers

Cool, let's set a default then.

But a default of `1` seems too narrow. Can we use `1.2` instead (as in the most recent commit I pushed to this PR)? That looks better, is more consistent with browser defaults, and fixes the UX issue with overflow of the underscore character in tabs.

pavish (approved):
The changes look fine.

The issue with finding the correct line-height is that fonts differ on the baseline to descender distance, as well as the cap-height distance.

Currently we only use system fonts, so our font will differ based on the OS and this will also lead to subtle visual differences in positions. Usually, that wouldn't be of concern, but we have a few places where this is clearly visible.

Screenshot 2022-01-29 at 1 35 32 AM

In the screenshot, you can find the additional leading space that is introduced with line-height `1.2`. This makes the cell content appear slightly aligned to the bottom instead of center.

The clear visual difference appears when we enter and exit edit mode, causing a slight jerk in the text, as shown in the below video:

https://user-images.githubusercontent.com/11850603/151618072-c55a86a8-96fb-41f7-b082-56864732edb1.mov

 The other place where it appears is the header images (which are temporary at this point).

I've added a commit https://github.com/centerofci/mathesar/pull/999/commits/40e9b2aec4e31600a19c5d051dbe1140f8af0353 which fixes these immediately visible places. We could improve upon the css during our styling update tasks. Please take a look at the commit and feel free to merge.

seancolsen:
@pavish I think the slight jerk in text is better than table names appearing to display underscore characters as space characters. So I'm going to merge this. 

If we're going to get particular about spacing, I think we ought to ship a web font to the browser. I can certainly see the appeal of using system fonts. But I'd lean towards being explicit about the font -- for the same reason we want to be explicit about the line height -- because we want predictable spacing across different platforms.

PR #926, opened 2022-01-02
Type configuration for Text, Number, Boolean, and all mathesar types with single associated db type

pavish:
Related to: 
* #254 
* #255 
* #256 
* #257 
* #258 

The PR handles the `database` options in the data type selection dropdown, for Text, Number, Boolean and all mathesar types with a single db type association.

Screenshot 2022-02-01 at 3 01 51 PM

The discussion [Dynamic Type selection & customization on the frontend #1035](https://github.com/centerofci/mathesar/discussions/1035) provides the declarative format and additional concerns. I advise the reviewers to review this PR after reading the discussion.

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1005, opened 2022-01-21
UI for adding multi-column unique constraints

### LINKED ISSUE #531: Frontend work for adding multi-column unique constraints

seancolsen:
Fixes #531

## Demos

- [Demo 1 -- Second modal for confirming delete](https://www.loom.com/share/3db085e8fc6b42bca85c2d373c1ffb47)
- [Demo 2 -- Second modal for adding constraint](https://www.loom.com/share/864f877a0a804b72992eaf8293f55363)
- [Demo 3 -- Form for adding new unique constraints](https://www.loom.com/share/9067d4d94412402f8caee3e7916f152d)

## Changes

- Modify table constraints dialog to allow users to add new unique constraints.

- Allow multiple modals to display at once -- so that we can display the form for adding a new unique constraint on top of the modal which displays all constraints.

- Change the UI for confirming the deletion of a constraint to use our modal confirmation system instead of the inline sliding confirmation. This reduces code complexity and leads to a more consistent UI.

- Split `Modal` component into

    - `IndependentModal` -- a simple base component for use-cases where we don't have a controller. (In Mathesar, I don't think we'll ever want to use this component, but people using the component library might.)
    - `ControlledModal` -- a higher-level component that composes `IndependentModal` and requires a `ModalController`

- Refactor `AddEditSchema` and `RenameTableModal` into simpler components that both compose `ModalTextInputForm` (a new component) to reduce code duplication and ensure consistent UI. 

## Review tips

- Play with the process of adding new unique constraints and deleting constraints.
- Play with the modal system in storybook
- Play with other modals in the app, since I did some refactoring which touches all modals.
    - Confirmations used when deleting anything, e.g. delete table
    - Create/edit schema
    - Rename table

## Issues

- The `Form` and `FormField` components (which exist only for styling) might overlap a bit with the form-builder system in terms of intended purpose. I haven't yet looked into this more deeply.

- Modal z index stacking is very primitive at the moment. All modals have the same z index, which means they are layered in the order in which they are _first_ opened. Technically, this could lead to bugs, but it's a little hard to imagine a real-world scenario that would trigger the bug. You'd need to begin with no modals, then open modal A, then open modal B, then re-open modal A again (even though it's already open). At that point, you'd expect modal A to be re-layered on top of modal B, but due to its z-index being identical to modal B it would remain on bottom. The data structure already supports this re-layering correctly. It would just be a little extra work to set the z-index based on the stacking order we're storing. That's totally feasible. I just haven't yet taken the time to do it. We could take that up later if this whole multi-modal approach actually seems viable.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish this is ready for review now.

@ghislaineguerin I'd also like you to look at this, since the UX is different than what we initially discussed. I explain a bit about this divergence in the "Demo 2" video. The direction I'm going here is to display multiple modals on top of each other. Pavish and I chatted a bit about this approach on a call. He expressed some reservations about it but figured it would be worth exploring.

Pavish, we initially talked about hiding the lower modal in some cases when the upper modal is displaying. After giving this some more thought I decided to always show the lower modal. For one thing it's simpler to implement. But also I prefer it from a UX perspective too. I think that placing the lower modal under the overlay sufficiently communicates the UI state to the user, making it clear that they can only interact with the top-most modal. It also adds additional clarity that the steps they took within the lower modal have persisted.

ghislaineguerin:
@seancolsen, your video provides a reasonable explanation for taking this approach. I don't feel that the modal on top of the modal is bad. If they are different enough, as you describe, it makes sense to overlap them.

ghislaineguerin (approved):

kgodey:
I'm not the biggest fan of the double modal UX, but I think it's fine for now. I think in the long run, perhaps we should show table constraints in a separate pane or page rather than a modal.

seancolsen:
@kgodey 

> I think in the long run, perhaps we should show table constraints in a separate pane or page rather than a modal.

Yeah I think that's definitely worth considering. We could also consider separating the DDL and DML operations for each table into separate tabs. Many DB GUIs I've worked with behave this way. You get one tab that lists all the columns, constraints, indexes, allowing you to add/update/delete those things. Then separately there's a different tab which lists all the rows in the table. Right now that's not super easy to do without some refactoring to the tabs system.

I think it's definitely worth trying to avoid multiple modals and we should be keeping this in mind going forward. If we're inclined to put something in a modal we should always be asking _"is there anything the user might want to do inside this modal that would require another modal?"_.

kgodey:
> We could also consider separating the DDL and DML operations for each table into separate tabs. Many DB GUIs I've worked with behave this way. You get one tab that lists all the columns, constraints, indexes, allowing you to add/update/delete those things. Then separately there's a different tab which lists all the rows in the table. Right now that's not super easy to do without some refactoring to the tabs system.

I think in Mathesar's use case, we could tuck DML operations into a separate page. I think users will be using Mathesar for DDL far more often than DML. We can figure out a good pattern for this when we start working on the Data Modeling milestone cc @ghislaineguerin.

> I think it's definitely worth trying to avoid multiple modals and we should be keeping this in mind going forward. If we're inclined to put something in a modal we should always be asking "is there anything the user might want to do inside this modal that would require another modal?".

Agreed, also cc @ghislaineguerin.

pavish (changes requested):
@seancolsen The PR looks good to me.

I do have my reservations of allowing multiple modals but I think we can come back to this after our alpha release. The discussion on this PR regarding a separate page for DML operations makes sense to me.

There are some action items:
1.  Update this PR to resolve merge conflicts.
2. Stack this PR on top of https://github.com/centerofci/mathesar/pull/976. I'd like to merge https://github.com/centerofci/mathesar/pull/976 first.
3. The `Form` and `FormField` components focus on composing forms using slots, whereas the `form-builder` component focuses on generating a form using a json config. Even though there is a clear different in usecase, they share the same component name and css classes. I'd like to avoid that. I think it makes more sense to keep the names of these components the same. I'll change the names of form-builder components as part of https://github.com/centerofci/mathesar/pull/926

seancolsen:
@pavish I've resolved merge conflicts and stacked this on top of #976. Back to you now.

pavish (approved):

PR #1022, opened 2022-01-25
Introduce function API

### LINKED ISSUE #385: Implement filtering options for Number types.

### LINKED ISSUE #846: Duplicates filter should work with other filters

dmos62:
This is a continuation of the Backend Filtering Refactor PR (#921).

This PR does not affect the current APIs. It only adds the endpoints and data structures related to the new function API, but does not replace the previous APIs. There will be another PR that will replace the current filtering API with a new filtering API that will use this function API.

**Technical details**

Should be considered a minor PR, since the functions API doesn't affect the rest of the code and its main "user" in the foreseeable future will be the to-be-implemented filtering API. In other words, this adds an internal API that will not yet be used on merge.

There are some minimal coverage tests for checking if `DbFunction` subclasses can be used for filtering. I expect to add tests as implementation breadth improves and as the upcoming filtering API is built.

There are two significant improvements planned for the functions API (though not for this PR):
- named parameters (currently it's using indexed parameters);
- generics (for hinting at the relationship between various inputs and outputs).

Adds the endpoints `/api/v0/functions` and `/api/v0/db_types`.

Here are their sample responses. Note that the hints currently applied are not accurate and are not meant to be at this stage. The uses of the hints system are somewhat unclear at this stage, since the frontend will not be using the functions API directly to build up the UI. The near-term future of the hints system depends on what the new filtering API will be like, which, as I said, is yet to be seen. I'd expect the principle of the hints system to be under review and not the details of their current use.

<details>
  <summary>`/api/v0/functions` endpoint sample response</summary>
  
  ```

GET /api/v0/functions/
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

[
    {
        "id": "and",
        "name": "And",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            }
        ]
    },
    {
        "id": "column_reference",
        "name": "Column Reference",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "parameter",
                "index": 1,
                "hints": [
                    {
                        "id": "column"
                    }
                ]
            }
        ]
    },
    {
        "id": "empty",
        "name": "Empty",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 1
            }
        ]
    },
    {
        "id": "equal",
        "name": "Equal",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            }
        ]
    },
    {
        "id": "extract_uri_authority",
        "name": "Extract URI Authority",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "parameter",
                "index": 1,
                "hints": [
                    {
                        "id": "uri"
                    }
                ]
            }
        ]
    },
    {
        "id": "greater",
        "name": "Greater",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "comparable"
                    }
                ]
            }
        ]
    },
    {
        "id": "in",
        "name": "In",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "parameter",
                "index": 2,
                "hints": [
                    {
                        "id": "array"
                    }
                ]
            }
        ]
    },
    {
        "id": "lesser",
        "name": "Lesser",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "comparable"
                    }
                ]
            }
        ]
    },
    {
        "id": "list",
        "name": "List",
        "hints": null
    },
    {
        "id": "literal",
        "name": "Literal",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "parameter",
                "index": 1,
                "hints": [
                    {
                        "id": "literal"
                    }
                ]
            }
        ]
    },
    {
        "id": "not",
        "name": "Not",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 1
            }
        ]
    },
    {
        "id": "or",
        "name": "Or",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            }
        ]
    },
    {
        "id": "starts_with",
        "name": "Starts With",
        "hints": [
            {
                "id": "returns",
                "hints": [
                    {
                        "id": "boolean"
                    }
                ]
            },
            {
                "id": "parameter_count",
                "count": 2
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "string_like"
                    }
                ]
            }
        ]
    },
    {
        "id": "to_lowercase",
        "name": "To Lowercase",
        "hints": [
            {
                "id": "parameter_count",
                "count": 1
            },
            {
                "id": "all_parameters",
                "hints": [
                    {
                        "id": "string_like"
                    }
                ]
            }
        ]
    }
]
```
  
</details>
<details>
  <summary>`/api/v0/db_types` endpoint sample response</summary>
  
```
GET /api/v0/db_types/
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

[
    {
        "id": "_array",
        "hints": null
    },
    {
        "id": "bigint",
        "hints": null
    },
    {
        "id": "bit varying",
        "hints": null
    },
    {
        "id": "bit",
        "hints": null
    },
    {
        "id": "boolean",
        "hints": [
            {
                "id": "boolean"
            }
        ]
    },
    {
        "id": "bytea",
        "hints": null
    },
    {
        "id": "\"char\"",
        "hints": null
    },
    {
        "id": "character varying",
        "hints": [
            {
                "id": "string_like"
            }
        ]
    },
    {
        "id": "character",
        "hints": [
            {
                "id": "string_like"
            }
        ]
    },
    {
        "id": "cidr",
        "hints": null
    },
    {
        "id": "date",
        "hints": null
    },
    {
        "id": "daterange",
        "hints": null
    },
    {
        "id": "decimal",
        "hints": null
    },
    {
        "id": "double precision",
        "hints": null
    },
    {
        "id": "float",
        "hints": null
    },
    {
        "id": "hstore",
        "hints": null
    },
    {
        "id": "inet",
        "hints": null
    },
    {
        "id": "int4range",
        "hints": null
    },
    {
        "id": "int8range",
        "hints": null
    },
    {
        "id": "integer",
        "hints": null
    },
    {
        "id": "interval",
        "hints": null
    },
    {
        "id": "json",
        "hints": null
    },
    {
        "id": "jsonb",
        "hints": null
    },
    {
        "id": "macaddr",
        "hints": null
    },
    {
        "id": "money",
        "hints": null
    },
    {
        "id": "name",
        "hints": null
    },
    {
        "id": "numeric",
        "hints": [
            {
                "id": "comparable"
            }
        ]
    },
    {
        "id": "numrange",
        "hints": null
    },
    {
        "id": "oid",
        "hints": null
    },
    {
        "id": "real",
        "hints": null
    },
    {
        "id": "regclass",
        "hints": null
    },
    {
        "id": "smallint",
        "hints": null
    },
    {
        "id": "text",
        "hints": [
            {
                "id": "string_like"
            }
        ]
    },
    {
        "id": "time",
        "hints": null
    },
    {
        "id": "time with time zone",
        "hints": null
    },
    {
        "id": "time without time zone",
        "hints": null
    },
    {
        "id": "timestamp",
        "hints": null
    },
    {
        "id": "timestamp with time zone",
        "hints": null
    },
    {
        "id": "timestamp without time zone",
        "hints": null
    },
    {
        "id": "tsrange",
        "hints": null
    },
    {
        "id": "tstzrange",
        "hints": null
    },
    {
        "id": "tsvector",
        "hints": null
    },
    {
        "id": "uuid",
        "hints": null
    },
    {
        "id": "email",
        "hints": null
    },
    {
        "id": "uri",
        "hints": [
            {
                "id": "uri"
            }
        ]
    },
    {
        "id": "money",
        "hints": null
    }
]
```
</details>

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### CODE REVIEW

File: db/functions/base.py [hunk 1]
```diff
@@ -0,0 +1,261 @@
...
+class Literal(DbFunction):
+    id = 'literal'
+    name = 'Literal'
+    hints = tuple([
+        hints.parameter_count(1),
+        hints.parameter(1, hints.literal),
+    ])
+
+    @staticmethod
+    def to_sa_expression(p):
```
kgodey, line 65: Could you use a more verbose variable than `p`?

kgodey: Also applies to other instances.

db/functions/base.py
```diff
@@ -0,0 +1,261 @@
...
+        hints.parameter_count(1),
+        hints.all_parameters(hints.string_like),
+    ])
+
+    @staticmethod
+    def to_sa_expression(p1):
+        return func.lower(p1)
+
+
+class ExtractURIAuthority(DbFunction):
```
kgodey, line 227: Since this depends on our custom URI type, I would not put this in the general functions namespace. I'd store it with the URI type instead.

Keep in mind that our custom types may not be installed on all Mathesar deployments.

kgodey: This should apply to any functions that depend on our custom types.

db/functions/base.py
```diff
@@ -0,0 +1,261 @@
...
+        hints.parameter(1, hints.uri),
+    ])
+
+    @staticmethod
+    def to_sa_expression(p1):
+        return func.getattr(URIFunction.AUTHORITY)(p1)
+
+
+def _get_defining_module_members_that_satisfy(predicate):
+    # NOTE: the value returned by globals() (when it's called within a function) is set when the
```
kgodey, line 241: Could you explain in a docstring what this function does?

db/functions/operations/apply.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def _get_columns_that_exist(relation):
+    columns = relation.selected_columns
+    return set(column.name for column in columns)
+
+
+def _db_function_to_sa_expression(db_function):
+    """
+    Everything is considered to be either a DbFunction subclass or a literal.
```
kgodey, line 37: What is "everything" here?

db/tests/functions/operations/test_filter.py
```diff
@@ -0,0 +1,132 @@
...
+    ("numeric", "is_not_null", None, 100),
+    ("date", "is_not_null", None, 100),
+    ("array", "is_not_null", None, 100),
+    # eq
+    ("varchar", "eq", "string42", 1),
+    ("numeric", "eq", 1, 1),
+    ("date", "eq", "2000-01-01", 1),
+    ("array", "eq", "{0,0}", 1),
+    # ne
+    # ("varchar", "ne", "string42", 99),
```
kgodey, line 66: Please remove commented out code.

kgodey: Applies below as well.

mathesar/api/viewsets/db_types.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import viewsets
+from rest_framework.response import Response
+from db.types.base import db_types_hinted, PostgresType, MathesarCustomType
+from itertools import chain
+
+
+class DbTypeViewSet(viewsets.ViewSet):
```
kgodey, line 7: Should be `DBType`, see above.

mathesar/api/viewsets/functions.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import viewsets
+from rest_framework.response import Response
+from db.functions.base import supported_db_functions
+from mathesar.api.serializers.functions import DbFunctionSerializer
+
+
+class DbFunctionViewSet(viewsets.ViewSet):
```
kgodey, line 7: Please write a test for this API to ensure that it works and has the expected format.

mathesar/api/viewsets/db_types.py
```diff
@@ -0,0 +1,18 @@
...
+    def list(self, _):
+        known_vanilla_types = (postgres_type for postgres_type in PostgresType)
+        known_custom_types = (mathesar_custom_type for mathesar_custom_type in MathesarCustomType)
+        known_db_types = chain(known_vanilla_types, known_custom_types)
+        data = [
```
kgodey, line 12: Please use a serializer to generate this, this logic does not belong in the viewset.

kgodey: Also, you will need to specify a list of databases for every type. Mathesar might be connected to multiple databases and not all of them will have every type installed. We need to check what types are installed where.

mathesar/api/viewsets/db_types.py
```diff
@@ -0,0 +1,18 @@
...
+            {
+                "id": db_type.value,
+                "hints": db_types_hinted.get(db_type, None)
```
kgodey, line 15: What is this hint expected to be used for?

mathesar/api/viewsets/db_types.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class DbTypeViewSet(viewsets.ViewSet):
+    def list(self, _):
+        known_vanilla_types = (postgres_type for postgres_type in PostgresType)
+        known_custom_types = (mathesar_custom_type for mathesar_custom_type in MathesarCustomType)
+        known_db_types = chain(known_vanilla_types, known_custom_types)
+        data = [
+            {
+                "id": db_type.value,
```
kgodey, line 14: This is not how we are representing DB types anywhere else in the API. How we represent types should be consistent with columns in `/api/v0/tables/` (and `/api/v0/databases/<id>/types`, although I assume that API will be deprecated by this one).

mathesar/api/viewsets/db_types.py
kgodey: It would be ideal to support filtering the types endpoint by database ID so that if a client is working with a single database, they can filter to types available in that database. I think we have this implemented for schemas already so you can reference that.

mathesar/urls.py
```diff
@@ -9,6 +9,8 @@
 router.register(r'schemas', viewsets.SchemaViewSet, basename='schema')
 router.register(r'databases', viewsets.DatabaseViewSet, basename='database')
 router.register(r'data_files', viewsets.DataFileViewSet, basename='data-file')
+router.register(r'functions', viewsets.DbFunctionViewSet, basename='functions')
+router.register(r'db_types', viewsets.DbTypeViewSet, basename='db_types')
```
kgodey, line 13: I would just call this `types`. We'll have a separate abstractions namespace for Mathesar types.

mathesar/api/serializers/functions.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework import serializers
+
+
+class DbFunctionSerializer(serializers.Serializer):
+    id = serializers.CharField()
```
kgodey, line 5: You'll need to also include a list of database IDs that the function is available on. Not all functions will be installed or available on all databases.

kgodey (changes requested):
@dmos62 Nice work, this code flows well overall and seems like it will be easy to work with. I did have some feedback, please see individual comments.

mathesar/api/viewsets/db_types.py
dmos62: When a `DBFunction` has a hint with id `boolean` or `string_like`, the API client can check against the `db_types` endpoint to tell which database types are meant by that (which db types have the hints `boolean` or `string_like`).

mathesar/urls.py
dmos62: It's ambiguous until the `abstractions` and `db` API namespaces are introduced. I'd change this then.

mathesar/api/viewsets/db_types.py
dmos62: Hm, this is giving me some pause.

I presume you meant changing `id` to `type` in the below response:

```
GET /api/v0/db_types/
    {
        "id": "_array",
        "hints": null
    },
```

But, looking at the `tables` endpoint, it does use `id` fields for identifying an instansce of some taxonomy. For example, a table is identified by its `id` and so is a column. The column described below says that it has the `type` `INTEGER`, but in the `db_types` response the `id` field signifies a DB type's identifier, not its `type` attribute.
```
GET /api/v0/tables/

          {
            "id": 3,
            "columns": [
                {
                    "id": 29,
                    "type": "INTEGER",
                },
```

Does this make sense?

dmos62:
Changes requested in Kriti's review
  - [ ] different capitalization Db -> DB
    - DbFunction
    - DbTypeViewSet
    - DbFunctionViewSet
  - [ ] parametrize some endpoints on the database
    - functions
      - installed functions might be different depending on database
    - db_types
      - db types might be different depending on database
  - [ ] more argument variables in `DBFunction.to_sa_expression`
  - [ ] move ExtractURIAuthority to the URI type's namespace
    - since it might not be installed on a given database
  - [ ] improve doc string in `_db_function_to_sa_expression`
    - don't use the ambiguous "everything"
  - [ ] write test for `DBFunctionViewSet`
  - [ ] move `db_types` endpoint response generation logic from the viewset to a serializer
  - [ ] (possibly?) remove commented tests in `db/tests/functions/operations/test_filter.py`
  - [ ] (possibly?) resolve database type reference conflict between the `id: type_id`  
    - https://github.com/centerofci/mathesar/pull/1022#discussion_r793590671

mathesar/api/viewsets/db_types.py
kgodey: I actually mean the the types are represented in all caps, the `id` nomenclature is fine. We should also make sure usage of spaces and underscores are consistent.

mathesar/urls.py
kgodey: Okay, sounds good.

mathesar/api/viewsets/db_types.py
dmos62: Ok, I'll look into this.

mathesar/urls.py
kgodey: On second thought, why don't we start the `db-api` namespace for new APIs introduced in this PR instead of needing to rename this later?

db/functions/operations/deserialize.py
```diff
@@ -0,0 +1,51 @@
...
+        # Everything except for a dict is considered a literal parameter.
+        # And, only the Literal and ColumnReference DbFunctions can have a literal parameter.
+        return parameter
+    else:
+        raise BadDbFunctionFormat(
+            "A literal must be specified as such by wrapping it in the literal function."
+        )
+
+
+def _get_db_function_subclass_by_id(subclass_id):
```
mathemancer, line 43: This is the sort of thing that could be tidied up by my wacky `type` function idea above.

mathemancer (changes requested):
Overall, very cool.  It seems like things are coming together.  I won't really be able to comment too much on usability until I'm getting my hands dirty trying to put some grouping functions together using this stuff.

I had a suggestion for how to reduce some repetition and make the many `DBFunction` subclasses a bit easier to use and define.  See the specific comment.  Feel free to disregard the idea if it doesn't make sense to you.

mathesar/urls.py
dmos62: @kgodey maybe the changes related to introducing a new API namespace will be easier to review in a dedicated PR? These endpoints won't be used immediately. I can go either way, I'm just wondering what's faster.

db/functions/base.py
```diff
@@ -0,0 +1,261 @@
...
+        if predicate(member)
+    )
+
+
+def _is_concrete_db_function_subclass(member):
+    return inspect.isclass(member) and member != DbFunction and issubclass(member, DbFunction)
+
+
+# Enumeration of supported DbFunction subclasses; needed when parsing.
+supported_db_functions = _get_defining_module_members_that_satisfy(_is_concrete_db_function_subclass)
```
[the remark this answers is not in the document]
dmos62: I like this. If generating a dict or list of `DBFunction` subclasses, that would mean they're not top-level type objects and thus can't be imported like `from db.functions.base import ToLowercase`. Might not be a problem.

This makes me wonder how I'll get `DBFunction`s defined in different namespaces collected into a single list. The functions endpoint has to be parametrized on the database because not all types and not all database functions might be installed on all databases.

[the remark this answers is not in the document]
mathemancer: You could generate them in a dict, then (in the same loop) add them to `globals` (I've used this in other projects).  I suggest _not_ doing that, though, since most of the places you want them, you need to look them up anyway, so you may as well grab them from a dict or list or whatever.  You could import the various dicts into the `base` module (this also nicely avoids circular imports) .  You should be able to restrict which dictionaries get passed through the `DBFunction` factory based on what's installed, I think.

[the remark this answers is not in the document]
dmos62: Thanks for the tips. Useful.

dmos62:
Requesting re-review @kgodey @mathemancer. Changes implemented. See [this comment](https://github.com/centerofci/mathesar/pull/1022#issuecomment-1023204247) for a list.

@mathemancer I did not end up using your suggested setup. Current setup works ok and seems simple enough. I had fun exploring your idea though.


Most changes are minimal and to the point.

A noteworthy change is that [I lowercased database type names in the table API](https://github.com/centerofci/mathesar/pull/1022/commits/06cea56dd185ee623af0c74dcb0d2153dc717b25). It seems that the SQLAlchemy method used to get those type names was uppercasing them. The rest of our code and at least some of the other SA methods return/expect database type names in lower case, so I forced it on the output of that method too.

dmos62:
I may have jumped the gun. There are failing tests due to switching to lowercase casing for database type names on the table API.

kgodey:
@dmos62 Why update the existing API, why not use the same SQLAlchemy method in the new API?

dmos62:
@kgodey I'm reverting my changes to the casing of database types. After investigating further, it seems that there's a fair amount of logic about the casing of those types, which I don't fully understand yet, and I'm not comfortable making any changes to it at the moment.

I think that resolving this is outside the scope of this PR. I could uppercase the ids in the db_types endpoint as a *partial* workaround.

> Why update the existing API, why not use the same SQLAlchemy method in the new API?

Because I'm working with database type Enums found under `db.types`. The code I referred to that's returning uppercased names is working with SQLAlchemy classes, so I can't pass it the data I have. Wrapping in SQLAlchemy classes and then unwrapping could be a workaround, but doesn't feel right: we should be able to compare database type ids without SA.

kgodey:
> I think that resolving this is outside the scope of this PR. I could uppercase the ids in the db_types endpoint as a partial workaround.

@dmos62 Can you open a new issue to track this? I agree it makes sense to not block this PR on it, but I don't think the `db_types` endpoint is usable by the frontend until the ID is consistent with other usage of types.

dmos62:
> I don't think the db_types endpoint is usable by the frontend until the ID is consistent with other usage of types

> Ideally, we would have a single source of truth for type IDs and use those everywhere.

@kgodey agreed. I'll open a ticket for this tomorrow.

dmos62:
Opened the ticket for inconsistent type name casing https://github.com/centerofci/mathesar/issues/1036

mathesar/api/viewsets/db_types.py
```diff
@@ -0,0 +1,26 @@
...
+
+from mathesar.api.serializers.db_types import DBTypeSerializer
+
+from mathesar.models import Database
+
+
+class DBTypeViewSet(viewsets.ViewSet):
+    def list(self, request):
+        try:
+            db_name = request.query_params['db_name']
```
kgodey, line 16: We should be using IDs to refer to API resources, not names. Also, this shouldn't be a required parameter, we should be able to loop through all installed DBs if this is not provided. Filtering by DB ID should be handled similarly to other APIs, see `mathesar/api/filters.py`.

If there's some reason to _require_ a database, then it's better to put this under `/api/v0/databases/<id>/types/`. We should not have multiple patterns to provide a database identifier in the URL.

All of this feedback also applies to the functions API.

mathesar/api/serializers/functions.py
kgodey: It looks like you've instead chosen to only make these available on a single database. In that case, they should not be a top level resource, please see my comment below. If we do choose to make them available on multiple databases, then you'll need to specify the database ID in the serializer.

kgodey (changes requested):
@dmos62 Nice work, I had some more comments, mainly about the APIs. One of my previous comments is also unaddressed and I added a comment about that.

db/tests/functions/operations/test_filter.py
dmos62: Weird, I remember you making this comment, then I wrote a reply, which is absent and then I didn't see this "conversation" anymore, so I presumed that you marked it as resolved.

Anyway, I originally replied that the commented tests serve to remind me of test cases that will likely be later implemented, but I'll just remove them to make review faster.

dmos62:
@kgodey @mathemancer re-requesting review. Made requested changes. Namely, removed commented tests and made new endpoints' routing coherent with the rest.

kgodey (approved):
Looks good to me, I'll defer to @mathemancer to complete review and merge.

db/types/base.py
```diff
@@ -109,9 +109,36 @@ def get_qualified_name(name):
...
+
+# Known database types are those that are defined on our PostgresType and MathesarCustomType Enums.
+known_db_types = _known_vanilla_db_types + _known_custom_db_types
+
+
+def get_available_known_db_types(engine):
+    """
+    Returns database types that are both available on the database and known through our Enums
+    above.
+    """
```
mathemancer, line 133: This doesn't check the DB, but rather gets the types which are both available in the engine mapping, as well as the Enums.  Does it need to ensure the type is installed on the DB in order to fulfill its purpose?   If not, the doc string should change.  Otherwise, we should add a way to check that.

mathemancer (approved):
I noticed a problem with the function that checks whether a type is installed.  To be honest, that logic is a bit all over the place both in the `db/types/base.py` file as well as the `db/types/operations/cast.py` file, and should probably be tidied up.  I'm okay with deferring fixing the problem I noted for a future PR that also cleans up those functions generally.  I'll approve, and leave it to you to decide whether to make the change now or later (or perhaps it's just not needed).

db/types/base.py
dmos62: Available in this context means that it's installed on the DB. I made a mistake thinking that the engine mapping is a proxy for that.

dmos62: I was going by the docstring on `db/types/base.py::get_available_types`:

```
def get_available_types(engine):
    """
    Returns a dict where the keys are database type names defined on the database associated with
    provided Engine, and the values are their SQLAlchemy classes.
    """
    return engine.dialect.ischema_names
```

dmos62:
I'll merge this PR in now, and I'll open a ticket to track the problem with checking whether a type is installed.

db/types/base.py
mathemancer: The original docstring was also a mistake it seems.

mathemancer:
> @mathemancer I'll merge this PR in now, and I'll open a ticket to track the problem with checking whether a type is installed.

If you don't mind, tag me (or ping me in matrix) when you do.  We should make sure the ticket covers the whole problem, which has sprawled across multiple files at this point.

dmos62:
@mathemancer sure, here's the ticket: #1050.

Issue #971, opened 2022-01-11
Improve focus indicator of `InputGroup` component


pavish:
The [form-builder PR](https://github.com/centerofci/mathesar/pull/915) introduced the `InputGroup` component, which helped make the input components more cleaner but lead to a stylistic regression.

Comments from @seancolsen on the PR:

> I really like the changes that simplify `TextInput` to a single DOM element, and I want to make sure to keep those changes. There's also clearly value in keeping the "grouped input" functionality. But the changes in this PR lead to a minor stylistic regressions in my opinion.
> 
> Examples:
> 
> - Filter tables
> 
>     | Before | After |
>     | -- | -- |
>     ![image](https://user-images.githubusercontent.com/42411/148478215-7999cfa4-f5ae-444b-8135-9af315affa5a.png) | ![image](https://user-images.githubusercontent.com/42411/148478304-b98116a8-9b5e-4b8c-9b97-21283fa1da89.png) |
> 
> - Add column
> 
>     | Before | After |
>     | -- | -- |
>     | ![image](https://user-images.githubusercontent.com/42411/148479870-a1ee2b9c-9aa9-478f-b350-9c6d1e4d045e.png) | ![image](https://user-images.githubusercontent.com/42411/148479532-95a1eedd-b15c-45ab-af43-40c7e0cf4cc3.png) |
> 
> I'd prefer that the focus indicator encompass the entire group, like it did before.
> 
> I'd like to keep the changes that add a grey background to the group items as it helps to distinguish the label from the input.
> 
> In Storybook, I see that `InputGroup` is also intended to handle the use case of multiple inputs grouped together, in which case the focus indicators would _need_ to be separate. But this use-case seems quite esoteric to me. I can't recall ever having seen a UI like that. I'd suggest we abandon any attempts to support it.

## Proposed solution

1. We need to analyze if it's worth it to bind focus indicator between the InputGroup component and the slotted input components.
2. [Bootstrap's InputGroup](https://getbootstrap.com/docs/5.0/forms/input-group/#multiple-inputs) was the main inspiration for this component. We can further analyze the need to group multiple inputs together instead of just text with input.

Based on the analysis, we'd like to perform necessary improvements on the component.

pavish:
Blocked by https://github.com/centerofci/mathesar/pull/915

seancolsen:
I splintered off some of the content that was originally in the body of this ticket and put it into #1042

pavish:
[...]

PR #1024, opened 2022-01-25
In-app help content

### LINKED ISSUE #914: Create components for building in-app documentation and help content
We need a Tooltip or Help component, akin to [Carbon's Tooltip component](https://carbon-components-svelte.onrender.com/components/Tooltip) that will allow us to add documentation and help messages within the UI.

seancolsen:
Fixes #914

## Goals

- Add a `Help` component that shows help content in a popover.

    Use is as follows:

    ```svelte
    This is text on the page. <Help>This is help content.</Help>
    ```

- Establish a convention of placing help content inside dedicated components which live in `__help__` directories close to the code where they get used. This pattern helps to separate concerns of content from concerns of functionality, keeping the higher-level components less-cluttered.

- Move some existing help content from the Schemas page into a `Help` popover.

- Add some new help content to the Table Constraints modal.

## Demo

The following gif shows all the in-app help content I've added in this PR.

Peek 2022-01-25 11-59

## Issues

- In `Help.scss`, I had to use a couple awkward CSS selectors, `button.dropdown.trigger.no-arrow.help-trigger` and `.dropdown.help-content` because I was fighting against the specificity of the CSS within the lower-level components. @pavish I'm curious to hear your recommendation for a better way to handle this. If it were all up to me, I would be setting up our CSS very differently, so I'd like some guidance here. And I want to make sure not to use `!important` anywhere in our component library. With my `padding: 0;` CSS as an example, what's the correct way to do this? I tried setting `triggerClass="padding-zero"` on `Dropdown` in `Help.scss`, but that didn't work because `button.btn.padding-zero` has lower specificity than `button.dropdown.trigger.no-arrow`. I feel like we need to make some modifications to `Button` and/or `Dropdown` to support my use-case of a dropdown without padding, but I'm not sure how to approach it and stay consistent with the our CSS patterns.

- Line height within help content is noticeably narrow. That should be addressed in #999

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey From a product perspective, would you like to give an opinion on this new pattern of sprinkling help content popovers throughout the UI?

@ghislaineguerin From a design perspective, would you like to give an opinion on the help messages?

kgodey:
@seancolsen I think it's a great pattern!

ghislaineguerin:
@seancolsen I'm wondering if we need to have some rules for using the pop-over instead of inline text. This pattern looks perfect for definitions, but it seemed a bit lengthy to explain more in-depth concepts when adding the unique constraint. Should we separate hints from definitions maybe? For now, I think we can go ahead and have both under this pattern, and later we might decide to treat hints differently or link to the relevant documentation for longer texts.

seancolsen:
@ghislaineguerin 

> link to the relevant documentation for longer texts

Linking to external docs is exactly what I'd like to do once we have them set up. For now, this in-app pattern will help users and give us a head start on the content for the content we want to place externally.

silentninja:
I like this, was aiming for a similar approach with error handling, where we could have links to the doc that would explain in detail the cause of error(Null violation exception for example)

pavish:
@seancolsen Please feel free to update the css and dom structuring of our `Dropdown` and `Button` components. Our CSS patterns have diverged from when we initially started and these components were some of the earliest ones. I'll leave it to you to play around and update the components.

Here are a few ideas:
* It's better to keep padding-zero as default for the Dropdown trigger i.e in the composed Button component.
* We could set the padding to the slot content instead, for both label and arrow.
* Arrow is currently absolutely positioned. Since the entire Button component is inline-flexed, we could remove absolute positioning for the arrow.
* The initial plan was to avoid utility classes as much as possible, but we may still need to use them in cases like these. I would suggest to move the padding utility classes to our common css file, and remove any and all padding specifications in both Button and Dropdown scss files.

pavish:
@ghislaineguerin @kgodey @seancolsen 

Regarding the UX for the help content, it is currently displayed only upon clicking the help icon. There are two widely used patterns for inline-help:
1. Show the help content on hovering upon the icon, like a tooltip. This is generally preferred when the help content does not have links or any form of user interaction.
2. Show the help content upon clicking the icon, like a popup. This is preferred when the help content has links or any form of user interaction.

We have gone with option 2 in this PR, which I also think is best suited for us.

I'd like to confirm this from a UX perspective and how we envision inline-helps to function in the future.

kgodey:
I think option 2 is fine.

seancolsen:
@pavish regarding the CSS.... It sounds like we're in agreement that the code in this PR _works_, but accentuates the need for some lower-level CSS refactoring.

This situation reminds me of #915 in which I [critiqued](https://github.com/centerofci/mathesar/pull/915#pullrequestreview-848217134) your use of `!important` and you responded by opening a [new ticket](https://github.com/centerofci/mathesar/issues/971) to address it later.

With our desire to ship features faster, I suggest taking a similar approach here. If you're okay with merging as-is, I'll create a new ticket for the CSS refactoring.

pavish:
@seancolsen Sounds good.

Please create a new issue for streamlining css, and remove the css update part of #971 and place it in the new issue.

pavish (approved):
Looks good to me!

seancolsen:
@pavish 

> Please create a new issue for streamlining css, and remove the css update part of #971 and place it in the new issue.

I added two new issues:

- #1043 for the concerns we both discussed here
- #1042 for the concerns I originally raised in #915

With our [agreed intent to de-prioritize the component library](https://github.com/centerofci/mathesar/discussions/1033#discussioncomment-2088021) I've placed both issues in the "Future Consideration" milestone so as to avoid concerning ourselves with them before the Alpha release.

Issue #775, opened 2021-10-29
Improve UX when adding new records to a table with a NOT NULL column


seancolsen:
## Reproduce

1. Set up a table with a NOT NULL column that doesn't have a default value.
2. Click "New Record".
3. Observe new row with yellow background, spinner that spins several seconds, finally a "Couldn't save changes" error.

![image](https://user-images.githubusercontent.com/42411/139354253-40fc87e5-9a4b-4c58-9154-ca01824a9f43.png)

I'm not sure what the expected behavior would be here. I think we need some more clarity around that from a UX perspective. @ghislaineguerin @pavish would you like to weight in?

seancolsen:
I'm marking this as blocked by #859 because we'll need more granular error info from the API before we're able to display a warning inside the cell

seancolsen:
This should be ready to implement now that #859 is closed

seancolsen:
In a separate [discussion](https://github.com/centerofci/mathesar/pull/1053#issuecomment-1032841315), we came to the conclusion that we should shift the "NOT NULL" validation responsibility to the front end. This means that before sending the API request to create an new empty record, we'll check the columns store to see which columns have a NOT NULL constraint. Then we'll provide some UX for the user to fill in the values. We'll need to think this through a bit more to hammer out the details.

kgodey:
@seancolsen Is this unblocked since we decided on client side validation (#1057)?

seancolsen:
@kgodey Indeed! Thanks for the follow-up. I've marked this ticket as unblocked now. Although I'll also mention that before resuming work on this ticket, I'd like to complete our (hopefully quick) planned synchronous discussion on "UX for failure to save a cell" at this Wednesday's [meeting](https://hackmd.io/OEXWY519TceOUi53q0Ay1A).

PR #1053, opened 2022-02-08
Fix `Not Null Violation` to return detailed error response

### LINKED ISSUE #1051: Provide detailed error info for POST to records API when in violation of a not-null constraint
## Current behavior

1. Set up a table which contains a `NOT NULL` column.
1. POST an empty request (`{}`) to the `records` API (e.g. by clicking the `+` icon to add a new row to the table).
1. Observe the following response:

    ```json
    [
      {
        "code": 4999,
        "message": "(psycopg2.errors.NotNullViolation) null value in column \"common_name\" of relation \"trees\" violates not-null constraint\nDETAIL:  Failing row contains (11, null, null, null, null, null, null).\n\n[SQL: INSERT INTO public.trees DEFAULT VALUES RETURNING public.trees.id]\n(Background on this error at: http://sqlalche.me/e/14/gkpj)",
        "field": null,
        "detail": null
      }
    ]
    ```

1. Within the response, observe no easy way to identify _which column_ (or columns) was responsible for the error.

## Expected behavior

- For #775, we need enough detail for the front end to be able to display error messages specific to certain cells. For example, if a POST request has two columns with invalid data (perhaps even invalid in _different ways_), we need to be able to match specific error messages to each of those columns.

- @pavish and I are flexible on the exact response schema from the API here, as long as it satisfies the above goal.

- We should also keep #896 in-mind with this change.

## Status

- blocked by #896

silentninja:
Fixes #1051 

Extracts details from Postgres diagnostics and returns a detailed error response that consists of the the column name causing the exception along with other details
```js
// Sample error response
[
  {
    'code': 4204,
    'message': 'null value in column "Case Number" of relation "NASA Record Create" violates not-null constraint',
    'field': None,
    'detail': {
      'row_parameters': {
        'Center': 'NASA Example Space Center',
        'Status': 'Application',
        'Case Number': None,
        'Patent Number': '01234',
        'Application SN': '01/000,001',
        'Title': 'Example Patent Name',
        'Patent Expiration Date': ''
      },
      'row_detail': 'Failing row contains (1394, NASA Example Space Center, Application, null, 01234, 01/000,001, Example Patent Name, ).',
      'column': 'Case Number'
    }
  }
]

```

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
I'll review this.

seancolsen (changes requested):
- Can we return the column id instead of the column name?
- If I try to add a new record to table that has _two_ NOT NULL columns, I only see the error for the first column. I'd like the response to give me enough info to display error messages within _both_ cells. I think it would be best to handle this case by returning two separate error objects within the top-level array of the JSON response body. But I'm open to other approaches too.

kgodey (changes requested):
Some general comments about the format, I haven't looked at the code.

- Why do we need `row_parameters` in the response? Also, why is it called `row`?, We call it `record` everywhere else.
- We need to be able to return errors for multiple columns, a single `column` key isn't going to work.
- We should be using IDs to identify columns to be consistent with the rest of the API. We should also be using PKs to identify records.

seancolsen:
@kgodey 

> We need to be able to return errors for multiple columns, a single column key isn't going to work.

What do you think about my suggestion above?

> I think it would be best to handle this case by returning two separate error objects within the top-level array of the JSON response body

seancolsen:
### Reactions

- @pavish I'd like to loop you into this convo too. The goal of this PR is to help the front end get enough info to display per-cell error messages. @silentninja said that it's hard for us to get info from Postgres about multiple constraint violations at once.

    > This is the actual behaviour of the database, it returns just the first column responsible for the not null violation.

### Thoughts about the larger picture

Here are some situations where the API should respond with error info if an attempt to POST to `records` results in an error from Postgres:

- (a) NULL value for a NOT NULL column (as #1051 describes)
- (b) value in violation of a unique constraint
- (c) value that does not fit within the column's specified length
- (d) FK constraint violation
- ...yada yada

For some of these situations (e.g. (a) and (c)), the front end already has sufficient info to perform client-side validation. Others (e.g. (b), (d)), it's not feasible for the front end to perform the validation.

@pavish and I briefly discussed using client-side validation to satisfy #775 but I argued that we should opt for server-side validation to reduce complexity.

### Moving forward

- From the perspective of designing an intuitive API, the challenge of getting multiple errors from Postgres seems problematic.

- However, from the more pragmatic perspective of our existing front end, we might be abe to cope with this limitation somewhat easily if we use client-side validation. Let's see if others agree with my train of thought (especially @pavish)...

- As far as I can tell, the front end currently has two situations where it makes POST requests to the records API:

    - When updating the value of an existing cell
    
        Here, we only send _one_ field at a time, because it's only one cell. So the API will never need to respond with multiple errors. Even in situations like (b) and (d) above, we'd still be okay because I don't imagine we'd ever need to display two errors for the same cell.

        Further (and only relevant to situation (a) above), if I recall, our specs for setting a single cell to NULL (not yet implemented) require client-side validation anyway.

    - When adding a new record

        Here we always send an empty object as the request (`{}`).

        I _think_ we should only need to worry about errors from situation (a) above. Right @kgodey @silentninja? In theory, a table could have a default value set for a column which also contains a unique constraint. That would generate a (b) type error. But that seems really unlikely. Maybe there's a situation I'm not considering though.

        If I'm not seeing some special failure scenario where the attempt to add a new empty row violates a constraint which only the server can validate, then we'll need to put some more thought into this!


If my train of thought is correct here, then we _could_ rely on client-side validation to solve #775 (although at the cost of increasing complexity on the front end). That would change the priority of #1051 from "blocking important work" to "nice to have someday".

kgodey:
> I will rename `row` to `record`. As for `row_paramters`, I thought it would be helpful information when debugging as it makes sure the information received by the database is correct. I can remove it if it seems unnecessary.  

I think we should remove it. Let's keep errors minimal and only have the information that's needed.

> I think it would be best to handle this case by returning two separate error objects within the top-level array of the JSON response body.

Since the database can't return multiple errors, I think the current format is fine. @silentninja I do think we should rename the `column` field to `column_id` for clarity.

--

Generally, I agree with the approach of the frontend valdiating what it can with the information that it is. For one thing, it's the fastest feedback loop for the user.

> I _think_ we should only need to worry about errors from situation (a) above.

I do think (b) errors are possible, but it should be unlikely since those columns would have to have been created outside of Mathesar. I think in that case, it's fine for the server validation to be incomplete, we can just show a frontend error for the first cell with an error.

seancolsen:
I feel good about merging this PR once @kgodey's concerns are addressed.

I have some more thoughts regarding implementation of #775 that I will raise somewhere else.

pavish:
@seancolsen I'll add some quick thoughts here.

* I would prefer performing frontend validations in scenarios wherever possible, for a better UX for the user. We can validate (a) on the frontend before sending the request.
* For (b), I do think it's possible to show targeted error messages. When there's an error for the unique constraint due to the default value, if the server could respond with the name of the constraint that failed, we can use that to highlight all associated columns.

seancolsen:
> if the server could respond with the name of the constraint that failed, we can use that to highlight all associated columns.

This is a great idea, and addresses a concern that's been rattling around in my head about how we'll deal with the case of a row that violates a multi-column unique constraint.

For this PR, I think it makes sense to stick with `column_id` as a way to identify the column (since, strictly speaking, there's no actual "constraint" for the column). But we'll need to consider other approaches once we take up the validation of unique constraints. I've opened #1059 for that.

kgodey (approved):
Looks good to me.

kgodey:
I'll leave it to @seancolsen to approve and merge.

seancolsen (approved):

PR #1062, opened 2022-02-11
Finish adding E2E tests

### LINKED ISSUE #640: Write integration tests for some important user scenarios
## Problem

We need integration tests written for each user scenario so far so that we don't break things as we develop new features.

## Tests to add

- [ ] Add new empty table
- [ ] Import via clipboard
- [ ] Import via file
- [ ] Add schema
- [ ] Add column
- [ ] Add row
- [ ] Sort table by column
- [ ] Edit existing record
- [ ] Open tab
- [ ] Close tab
- [ ] Delete two records
- [ ] Increment table pagination
- [ ] Remove multi-column unique constraint
- [ ] Try to disallow NULL on a column with NULL values
- [ ] Add multi-column unique constraint

[discussion](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$d_enIAaHE1sIa-N6czmXs3QBX8KsOrU89D_E4imjbBw?via=matrix.mathesar.org&via=matrix.org)

## Additional context

- Work on this is blocked by #639 
- #89

seancolsen:
Fixes #640

## Notes

- #1032 lays some of the groundwork for closing #640 by making it easier to write E2E tests (and demonstrating that ease by providing two tests). This PR will go the _rest_ of the way by adding all the remaining tests specified in the ticket.

## Tests to add _(as spec'ed in ticket)_

- Already added in #1032
    - [x] Add new empty table
    - [x] Import via clipboard
- Added in this PR
    - [x] Import via file
    - [x] Add schema
    - [x] Add column
    - [x] Add row
    - [x] Sort table by column
    - [x] Edit existing record
    - [x] Open tab
    - [x] Close tab
    - [x] Delete two records
    - [x] Increment table pagination
    - [x] Remove multi-column unique constraint
    - [x] Add multi-column unique constraint
    - [x] Try to disallow NULL on a column with NULL values

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
This PR is ready for review except for the fact that it's based on top of #1032 which is still in draft state.

silentninja:
@seancolsen it looks like some of the e2e tests are failing and there is also linting issues you would need to take care of

seancolsen:
@silentninja From what I can tell, the work you did in #1032 to ensure that the tests are isolated from one another does not work for the `test_create_schema` test which is why we're seeing the CI pipeline fail here. The schema gets created when running that test on Chromium. Then when we run it on Firefox and Webkit it fails because of my assertion that the schema should not already exist. 

What's your recommended fix for this problem?

Even if it's quick and dirty for now, I think that's fine, just so we can get this merged and move on. If you can't identify easy fix, I wouldn't be opposed to adding some code within that test which deletes the schema too (and adding some code-level docs explaining that hacky fix).

silentninja:
@seancolsen The `mathesar/tests/integration/conftest.py::base_schema_url` fixture uses `mathesar/tests/integration/conftest.py::schema` fixture to create a `schema` so that it is easier to drop all the related objects that were created during test rather than having to drop them individually. 

But if you are testing `schema` related api's, don't use the schema fixture rather just use the `mathesar/tests/integration/conftest.py::test_db_model` fixture alone and use the `live_server_url/<db_name>/` url as your base url. 

Please take a look at the [backend schema api tests](https://github.com/centerofci/mathesar/blob/f3a32c9fd4861d01b3bc36d4d07860bfe34398fe/mathesar/tests/api/test_schema_api.py#L185)

seancolsen:
@silentninja

> don't use the schema fixture

I made changes in a6a777e6373a3471d595fd5076ce9f62fabff835 that I _think_ follow your recommendation, but I was still observing the test to fail when run on two browsers in the same test session.

So in 4fce76bed1600ae5ae25968d8db6274a5ec12f27 I implemented my hacky fix which cleans up the state accumulated in the test by deleting the schema through the UI in the same test. Is this okay for now? I'm open to suggestions for fixing it, as long as they're quick. I'm also okay with deleting the test entirely just for the sake of moving forward.

seancolsen:
Because the CI is now passing, I've removed the draft status. This PR is now ready for review.

@silentninja Would you be up for reviewing/merging?

silentninja (changes requested):
Nice work @sean. I noticed few patterns across the codebase which might need some changes
- Type annotation - We don't use type annotations for the backend codebase. Though I am not sure if it applies for integration tests. I am leaving that to @kgodey .
- Page objects - I think it is a good time to discuss about it in this PR, as I found few functions which should belong to a page object.

kgodey:
> We don't use type annotations for the backend codebase. Though I am not sure if it applies to integration tests. I am tagging @kgodey to decide on it.

We shouldn't be using type annotations. I think we should aim for consistency in Python code style wherever we use Python.

kgodey:
Also, a general note, @silentninja, I see you've requested changes in the structure of the tests. I don't think that's worth blocking this PR for. I recommend getting this PR merged as soon as possible and creating new tickets for follow up work (which we can prioritize separately).

mathesar/tests/integration/utils/locators.py
```diff
@@ -0,0 +1,9 @@
+from playwright.sync_api import Page
+
+
+def get_tables_list(page: Page):
+    return page.locator("#sidebar li[aria-level='1']:has(button:has-text('Tables')) ul")
```
silentninja, line 5: These are much better suited as properties of a Page object model as they belong to a particular page(tables list page).

mathesar/tests/integration/conftest.py
```diff
@@ -67,3 +68,23 @@ def schema(create_schema, schema_name):
...
+def go_to_patents_data_table(page, create_table, schema_name, base_schema_url):
+    """
+    Imports the `patents.csv` data into a table named "patents" and navigates to
+    the view of that table before starting the test.
+    """
+    table_name = "patents"
+    table = create_table(table_name, schema_name)
+    table.import_verified = True
+    table.save()
+    page.goto(base_schema_url)
```
silentninja, line 88: It would be better to have the page navigation as a different fixture or as a Page object. I would prefer to keep navigation logic as part of a Page object. I have written down an incomplete example to give a rough idea.
```
class TablePage:
    def __init__(self, page, base_schema_url):
        self.page = page

    def navigate(self):
        self.get_table_entry(page, table_name).click()
```

silentninja, line 88: It would be better to have the page navigation as a different fixture or as a Page object. I would prefer to keep navigation logic as part of a Page object. I have written down an incomplete example to give a rough idea.
```
class TablePage:
    def __init__(self, page, base_schema_url):
        self.page = page

    def navigate(self):
        self.get_table_entry(page, table_name).click()
```

silentninja (commented):
- Page objects - I think it is a good time to discuss it in this PR, as I found a few functions which should belong to a page object. Page objects used to encapsulate locators, navigation logic, so that we don't have to repeat it for each test cases. It makes the codebase maintainable and less fragile by 
  - They keep all page element selectors in one place
  - Contains navigational logic, so we can use it for combinatorial test flows.
  
  I have made more specific comments on how it could be used, to make the codebase better.

silentninja:
> Also, a general note, @silentninja, I see you've requested changes in the structure of the tests. I don't think that's worth blocking this PR for. I recommend getting this PR merged as soon as possible and creating new tickets for follow up work (which we can prioritize separately).

I marked the Page object-related information as a single comment but it got merged with the requested changes, looks like I cannot do it in a single operation

mathesar/tests/api/conftest.py
```diff
@@ -23,31 +21,6 @@ def client():
     return APIClient()
 
 
-@pytest.fixture
-def create_schema(engine, test_db_model):
```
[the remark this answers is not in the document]
seancolsen: In 733f4bfc927c4f609a23f84038ef7fa0fdafb972 I moved `create_schema` and `create_table` up so that I could use them in my own `go_to_patents_data_table` fixture that I added in 2e06b25d7565b8e6cda1e9c06812b748ced3568b. My understanding is that I can only access fixtures defined within ancestor directories -- not within sibling/cousin directories. Correct?

mathesar/tests/integration/test_schemas.py
```diff
@@ -0,0 +1,18 @@
...
+    schema_entry = page.locator(f".schema-list .schema-row:has-text('{schema_name}')")
+    expect(schema_entry).not_to_be_visible()
+    page.click("text=New Schema")
+    page.fill("[aria-label='name']", schema_name)
+    page.click("button:has-text('Save')")
+    expect(schema_entry).to_be_visible()
+    # We're also deleting the schema in the same test as a way of cleaning up
+    # the state created in this test so as not to interfer with other tests.
+    # This is a hack for now.
+    schema_entry.locator("button[aria-label='Delete Schema']").click()
```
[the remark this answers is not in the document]
seancolsen: I agree that this is hacky. Can we clean it up in a subsequent PR? Being that I've already [spent some time](https://github.com/centerofci/mathesar/pull/1062#issuecomment-1043357864) trying to find a better solution to this hack, how would you feel about writing the fixture for it yourself since I think you'd be able to get it working quicker than I could?

mathesar/tests/integration/test_ui.py
```diff
@@ -0,0 +1,72 @@
...
+    table_1_tab = get_tab(page, "Table 1")
+    close_tab(table_1_tab)
+
+    # No tabs should be open
+    expect(table_0_tab).not_to_be_visible()
+    expect(table_1_tab).not_to_be_visible()
+
+    # Open Table 0
+    table_0_entry.click()
+    expect(table_0_entry.locator(".item")).to_have_class(re.compile("active"))
```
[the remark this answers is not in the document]
seancolsen: > Do we need to use a regex

Unfortunately, yes -- as far as I can tell. I initially tried a plain string and it did not work. The [`to_have_class` locator assertion](https://playwright.dev/python/docs/1.18/api/class-locatorassertions#locator-assertions-to-have-class) compares the expected value to the _entire_ value of the `class` attribute. So if you have `<div class="foo bar" />` a call of `.to_have_class("foo")` will fail because `"foo" != "foo bar"`. I found this behavior surprising. It also made me wonder whether

mathesar/tests/integration/utils/locators.py
seancolsen: How would you feel about making a PR with these changes after merging this PR?

mathesar/tests/integration/conftest.py
seancolsen: How would you feel about making a PR with these changes after merging this PR?

seancolsen:
> We shouldn't be using type annotations.

I thought I could sneak them in there without anyone noticing! haha! 🤣

Actually I meant to raise the topic within #1032. As food-for-thought in case this topic comes up in the future, I'll explain my rationale for including these type annotations here even though I'm aware the back end team has chosen to deliberately avoid typing. I expect E2E test authoring to be done primarily by front end devs who will be accustomed to type annotations but likely not familiar with Playwright's API (since it's so new). For me, test authoring is faster and easier with the aid of documentation and autocomplete within my editor -- and those features only work when the `page` argument has a type annotation. Was it `page.selector`? Nope. It's `page.locator`! The intellisense helps me reach that conclusion _much_ faster than I can when browsing the Playwright docs. If the only harm is a little inconsistency, then I figured the annotations do more good than harm in this case. I think productivity trumps consistency.

In 172b52e91629a59753e3af87bad04761c8599634 I've removed the type annotations.

I won't add any more going forward. I might put them in while I writing the test and then delete them before committing.

seancolsen:
@silentninja this is ready for re-review now.

silentninja (approved):

silentninja:
@kgodey Can you merge this PR, I cannot seem to merge it as it needs an additional approving review

kgodey:
@silentninja It need a review from a frontend codeowner so @pavish needs to merge this.

seancolsen:
@pavish To speed up your review, you should only need to focus on the front-end files. The only thing I've done there is make some adjustments to `aria-label` attributes (because I wanted to rely on them within my tests).

pavish (approved):

PR #1067, opened 2022-02-14
Interval type improvements

### LINKED ISSUE #430: Handle `INTERVAL` type in the backend
This issue is to ensure that Mathesar can handle the Postgres `INTERVAL` data type.

- Users can use the API to change a column to `INTERVAL`if it's possible to do so (this should already be complete).
- Users can set and change the `precision` and `fields` of a given `INTERVAL` column via a `type_options` field in the API.
- Automatic type inference during file import suggests `INTERVAL` when it makes sense to do so. 
  - We could use something like https://github.com/oleiade/durations or https://github.com/wroberts/pytimeparse
  - Some work on this is complete, but there are probably improvements to be made in detecting durations.

**Additional Context** 
- #251  (parent issue)
- Inference PRs:
  - #127
  - #135
- See comments on #132 for conversation about inferring durations.

mathemancer:
Fixes #430 

This adds a custom interval type _at the SQLAlchemy level_ that maps to the default PostgreSQL type.  Further, we can now accept `precision` and `fields` arguments when creating or altering columns involving the `INTERVAL` type. 

**Technical details**

The `precision` type option takes an integer (1-6) as input.  The `fields` type option takes a string, and defines which fields the interval stores.  Acceptable strings are:
```
YEAR
MONTH
DAY
HOUR
MINUTE
SECOND
YEAR TO MONTH
DAY TO HOUR
DAY TO MINUTE
DAY TO SECOND
HOUR TO MINUTE
HOUR TO SECOND
MINUTE TO SECOND
```
If both `precision` and `fields` are specified, then `fields` must include seconds, since `precision` applies to the seconds field.

The reason for a custom type was initially to ensure that we didn't convert PostgreSQL `INTERVAL`s into Python `timedelta`s, since that conversion is lossy.  It also standardizes the output and some aspects of input for intervals.  Future PRs will similarly standardize output of other time-related types.

For reference, we'll be using the ISO 8601 spec as reduced by [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) for standardized output, and always-acceptable (at the DB layer) input.  We will also attempt to handle _any_ string as input using the default PostgreSQL date / time / duration parsing.  So, in the case of intervals, we have strings like
```
f"P{years}Y{months}M{days}DT{hours}H{minutes}M{seconds}S"
```
Each variable is an integer with the exception of seconds for output (seconds can be a float).  For input, decimals are allowed, and will be converted appropriately.  Also, inputs will "carry over" when possible.  Seconds and minutes will aggregate into hours, but hours won't aggregate into days since some days are different numbers of hours around DST changes.  Days won't aggregate into months, but months will aggregate into years.  For output, any missing units (e.g. zero minutes) will have actual zeroes so the client can count on each part being in the returned string.  For input, this is not necessary (but you _do_ need to include the `T` separator between the date and time sections if you include time values).

As a bonus, this PR also fixes some bugs in the constraints API tests that were preventing the pipeline from passing.

Finally, there 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/filters/operations/apply.py
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env python3
```
kgodey, line 1: What is this empty file for?

kgodey (approved):
Looks good to me overall, I added a couple of comments. @mathemancer, please resolve before merge.

I also think @silentninja should review this since he more recently worked on date and time types.

db/filters/operations/apply.py
mathemancer: Oops.  That appears to be cruft from a branch change.  I must have had a file open in a different branch.

db/tests/types/test_interval.py
```diff
@@ -0,0 +1,179 @@
...
+    three_days_interval = cast('3 days 4 hours 30 minutes', interval.Interval)
+    the_date = cast(dt(2020, 1, 1), datetime.TIMESTAMP_WITHOUT_TIME_ZONE)
+    with engine_with_types.begin() as conn:
+        res = conn.execute(
+            select(the_date + three_days_interval)
+        ).scalar()
+    assert res == dt(2020, 1, 4, 4, 30)
+
+
+def test_interval_interval_addition(engine_with_types):
```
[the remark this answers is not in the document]
mathemancer: Good idea.

db/types/interval.py
```diff
@@ -0,0 +1,75 @@
...
+                )
+        elif self.impl.fields is not None:
+            try:
+                assert self.impl.fields.upper() in all_fields
+            except AssertionError:
+                raise InvalidTypeParameters(
+                    f'fields "{self.impl.fields}" is not in {all_fields}'
+                )
+
+    def column_expression(self, col):
```
silentninja, line 58: I don't have any issues with using `column_expression`,  I am just curious to know if you have thought about setting a local `intervalstyle` instead.

silentninja (changes requested):
Looks good to me. Nice work @mathemancer, I learned a few sqlalchemy tricks, thanks. Can you add the interval standard as `ISO 8601` to the  [Engineering  decision](https://wiki.mathesar.org/en/engineering/decisions)

db/types/interval.py
mathemancer: The problem with that is we need to avoid `psycopg2` interpreting the result as an actual `INTERVAL`.  The column expression casts to `TEXT` at the DB layer, which then gets picked up by `psycopg2` as a python `str`.  The problem with letting `psycopg2` pick up `INTERVAL`s otherwise is that it uses a python `timedelta` to represent them, but `timedelta` makes some different choices (e.g., assuming 30 days can be accumulated into 1 month) and the conversion is lossy.  This would result in erroneous info being passed to the UI, and in fact make viewing some values (e.g., 37 days) impossible.

`intervalstyle` doesn't change the actual return type, and so doesn't really solve the problem by itself.  We'd still need to cast to text using a `column_expression`.  Given that, I opted to handle the formatting ourselves, since the PostgreSQL `iso_8601` style is slightly out-of-spec, and made a few choices that don't work that well for our use case (IMO).  It also makes the formatting that would happen completely visible, rather than hidden in the PostgreSQL docs.

mathesar/api/db/viewsets/columns.py
```diff
@@ -69,19 +70,13 @@ def create(self, request, table_pk=None):
                     )
                 else:
                     raise database_base_api_exceptions.ProgrammingAPIException(e)
-            except TypeError as e:
```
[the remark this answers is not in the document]
mathemancer: Ah, good catch.  I changed that so I could get more debugging output at some point and forgot to revert the change.

mathemancer:
> Looks good to me. Nice work @mathemancer, I learned a few sqlalchemy tricks, thanks. Can you add the interval standard as `ISO 8601` to the [Engineering decision](https://wiki.mathesar.org/en/engineering/decisions)

I'm currently writing up a wiki page that will cover the spec.  It's a little more involved.

db/types/interval.py
silentninja: That makes sense, thanks for the explanation!

silentninja (approved):

Issue #1028, opened 2022-01-26
Improve support for finding duplicate rows


kgodey:
## Problem
We have implemented a filter to identify duplicate rows. However, after implementing it, we've realized that it doesn't make sense as a filter like our other filters.

Our other filters work the same regardless of the order in which they are applied. This is not the case for "has duplicates". Depending on the other filters applied, we may end up returning different results and confusing the user. We don't face this problem with any other filter because other filters only rely on the data in a specific row. "Has duplicates" relies on all of the rows visible.

As an example, here's a table:

| ID | Title | Release Year | Favorite |
|-|-|-|-|
| 1 | Dante's Peak | 1997 | FALSE |
| 2 | Forrest Gump | 1994 | TRUE |
| 3 | The Karate Kid | 1984 | TRUE |
| 4 | Dante's Peak | 1997 | TRUE |
| 5 | The Karate Kid | 1984 | TRUE |

### Order 1
Imagine the user applies filters in this order: `"Year" > 1993`, `"Favorite" is TRUE`, `"Title" has duplicates`

#### Filter 1: `"Year" > 1993`
| ID | Title | Release Year | Favorite |
|-|-|-|-|
| 1 | Dante's Peak | 1997 | FALSE |
| 2 | Forrest Gump | 1994 | TRUE |
| 4 | Dante's Peak | 1997 | TRUE |

####  Filter 2: `"Favorite" is TRUE`
| ID | Title | Release Year | Favorite |
|-|-|-|-|
| 2 | Forrest Gump | 1994 | TRUE |
| 4 | Dante's Peak | 1997 | TRUE |

#### Filter 3:`"Title" has duplicates`
_0 results_.

### Order 2
But instead if the user applies filter in this order `"Year" > 1993`,`"Title" has duplicates` and `"Favorite" is TRUE`

#### Filter 1: `"Year" > 1993`
| ID | Title | Release Year | Favorite |
|-|-|-|-|
| 1 | Dante's Peak | 1997 | FALSE |
| 2 | Forrest Gump | 1994 | TRUE |
| 4 | Dante's Peak | 1997 | TRUE |

#### Filter 2:`"Title" has duplicates`
| ID | Title | Release Year | Favorite |
|-|-|-|-|
| 1 | Dante's Peak | 1997 | FALSE |
| 4 | Dante's Peak | 1997 | TRUE |

####  Filter 3: `"Favorite" is TRUE`
| ID | Title | Release Year | Favorite |
[...]

Issue #898, opened 2021-12-16
Modify constraints API to use column id instead of column name


seancolsen:
The constraints API refers to columns by name in its response and its expected request. I'd like to use the column id instead.

## GET Response

Example of current response:

```json
{
  "count": 2,
  "results": [
    {
      "id": 100,
      "name": "birds_pkey",
      "type": "primary",
      "columns": [ "id" ]
    },
    {
      "id": 101,
      "name": "birds_name_key",
      "type": "unique",
      "columns": [ "name" ]
    }
  ]
}
```

Example of desired response:

```json
{
  "count": 2,
  "results": [
    {
      "id": 100,
      "name": "birds_pkey",
      "type": "primary",
      "columns": [ "1" ]
    },
    {
      "id": 101,
      "name": "birds_name_key",
      "type": "unique",
      "columns": [ "2" ]
    }
  ]
}
```

## Other requests and responses

The same logic should apply to all the requests/responses for the other HTTP methods on this endpoint. For example, if I add a constraint, I'd like to specify the columns by id instead of name.

jjram20:
Hi, I would like to work on this issue.

Issue #896, opened 2021-12-16
Modify records API to use column id instead of column name


seancolsen:
I'd like to be able to use the records API when all I know and care about is the column id for my columns.

## GET request params

The `order_by`, `group_count_by`, and `filters` params all expect references to columns to be given by name. For example:

- `order_by=[{"field":"name","direction":"asc"}]`
- `group_count_by=["name"]`
- `filters={"and":[{"field":"name","op":"eq","value":"Sean"}]}`

I'd like to the API to accept column references by `id`.

### As a breaking change

A simple way to address this need would be to break the API, no longer accepting column references by name

- `order_by=[{"field":"1","direction":"asc"}]`
- `group_count_by=["1"]`
- `filters={"and":[{"field":"1","op":"eq","value":"Sean"}]}`

### As a backwards compatible change

Another approach would be to accept _both_ references by name and by id. If I want to use the column id, I would do:

- `order_by=[{"field_id":"1","direction":"asc"}]`
- `group_count_by_ids=["1"]`
- `filters={"and":[{"field_id":"1","op":"eq","value":"Sean"}]}`

## GET Response

Currently the records API response look like:

```json
{
  "count": 2,
  "results": [
    {
      "id": 1,
      "first_name": "Bart",
      "last_name": "Simpson",
    },
    {
      "id": 2,
      "first_name": "Lisa",
      "last_name": "Simpson",
    }
  ]
}
```

To avoid relying on the column name, I'd like the response to give the column id as keys instead.

```json
{
  "count": 2,
  "results": [
    {
      "1": 1,
      "2": "Bart",
      "3": "Simpson",
    },
    {
      "1": 2,
      "2": "Lisa",
      "3": "Simpson",
    }
  ]
}
```

This would need to be a breaking change.

## POST and PATCH request and response 

Request and response bodies look like:

```json
{
  "name": "Sean"
}
```

I'd like to be able to update a record or add a new record by sending something like:

```json
{
  "1": "Sean"
}
```

## Additional context

- Related:
  - #839

mathemancer:
I think the breaking change option is superior (though a bit of a pain).

kgodey:
Agreed re: breaking change.

silentninja:
I have a few concerns with this change

1. I assume we would use the same format for views too and views might be having computed columns that would not go well with `id`.
2. This would be useful if `filters` are long-lived, but I think we would be creating views rather than storing the filter queries
[...]

Issue #839, opened 2021-11-23
Replace column index usage with attnums and Column model `id`


silentninja:
## Problem
Column index that is being used now is not robust and has few bugs related to its usage. For example `db.constraints.operations.select.get_column_constraints` will break if columns have been deleted.

## Proposed solution
PR #658 introduces the Column model which uses attnum, a robust alternate to column index. Hence column model `id` or `attnum` should be used in the place of column index

mathemancer:
I think the `db` layer identifier for the column should be the `attnum`, and the `mathesar` service layer should map between the column id used in the API and the `attnum` used by the database.

kgodey:
Marking as ready since the prerequisite work has been merged.

silentninja:
@kgodey Should these be part of the `Initial data type` milestone or improvements as the broken down issues seem to be part of improvements?

kgodey:
@silentninja nice catch, I moved the broken down issues to `Initial Data Types` as well. I think it makes sense there since we're expanding the usage of columns through display options in that milestone.

PR #1069, opened 2022-02-14
Replace filtering API; includes support for filtering numbers

### LINKED ISSUE #385: Implement filtering options for Number types.

dmos62:
Fixes #385

Related to #1028

This is a continuation of the #1061 PR. Contains 2 major changes:

- replaces current filtering API with the new filtering API (introduced in #1061);
  - new syntax; frontend will use the filters declared in `/api/ui/v0/database/0/filters/`;
  - the `records` query param is now `filter`, not `filters` (singular, not plural);
- it also converts the `duplicate_only` operation from a filter into a query param (see #1028 for some context); to be clear, this had to be done, there wasn't a coherent way to consider it a Mathesar filter.

Also, includes number filters necessary to fulfil requirements in #385. Actually, it's the previous PR, #1061, that included these filters, but now they will be usable.

**Technical details**

Filters are specified by frontend in the function API format:

```python
{"and": [
    {"empty": [
        {"column_reference": ["some_column"]},
    ]},
    {"equal": [
        {"column_reference": ["some_column"]},
        {"literal": ["some_literal_value"]},
    ]},
]}
```

The expected data structure is made up of dicts containing one (and only one) key-value pair each. The key is the filter id, and the value is always a list of parameters. Every parameter is either a nested filter, or a string/number when it's the sole parameter to `literal` or `column_reference`.

There are 5 "special filters" that are not declared with the rest (in `/api/ui/v0/database/0/filters/`) and have to be hardcoded:
- the operators `and`, `or`, `not`; their syntax is equivalent to that of regular filters;
- the `column_reference`, whose parameter is a plain string denoting the column's name;
- the `literal`, whose parameter is some non-collection JSON primitive whose value will be used verbatim (see parameters of the `equal` sub-filter above).

`column_reference` and `literal` "special filters" must always be used when you want to pass a column reference or a literal. The API should throw an error otherwise.

New filter API uses the functions API under-the-hood. Filters are actually `DBFunction`s presented in a simplified way. The syntax for submitting filters to backend is actually just the function API syntax. The filter API could be seen as a thin layer on top of the function. API.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
> the `column_reference`, whose parameter is a plain string denoting the column's name;

We should be using column ID as the primary way to refer to a column, we're in the process of switching the entire API to do this, see:
- #839 
- #896 
- #898

pavish:
@dmos62 Please let me know when this PR is ready for frontend work. I'll take it up.

dmos62:
@kgodey I'm aware of the move to using column IDs. It's not yet in master though, or is it? Should I refactor this PR to use column ids?

I'll rename `column_reference` to `column_id`.

kgodey:
> It's not yet in master though, or is it?

@silentninja is working on these, hopefully they should be in `master` by the end of the week.

dmos62:
Added support for specifying columns via their Django IDs.

Now the JSON expected by the `records` endpoint's `filter` query param looks like this:

```python
{"and": [
    {"empty": [
        {"column_id": [1]},
    ]},
    {"equal": [
        {"column_id": [3]},
        {"literal": ["some_literal_value"]},
    ]},
]}
```

The difference from the sample posted in the first post is that `column_reference`s are replaced by `column_id`s, which take Django column IDs.

### Details

Made DB layer unaware of Django IDs, by preprocessing the passed DBFunction JSON specification (for example to the `records` endpoint `filter` query param) to rewrite column id references to column name references. Before preprocessing/rewritting, the JSON looks like `dict(empty=[dict(column_id=1)])`, and after it looks like `dict(empty=[dict(column_name="foo")])`. The DB layer only sees the latter (`column_name`) version.

kgodey:
>  {"column_id": [1]},

Why is `column_id` a list? Presumably it's just one `id`.

dmos62:
@kgodey `column_id` is a `DBFunction` and all `DBFunction`s take a sequence of parameters. It's optimized for simple automatic compiling and parsing, not for compilation by-hand. I've considered handling these single-parameter cases differently, but opted not to in order to keep the parsing simple.

kgodey:
@dmos62 To make the API easier to read, I recommend dropping the list from `column_id` and other single parameter options, you can always check if the input is only a single parameter and turn it into a list before passing it on to the `DBFunction`. It doesn't seem like it would be much effort or complicate parsing much.

dmos62:
@kgodey I think this makes the syntax more complicated and thus less robust, but I'll make the change to keep discussions to a minimum.

dmos62:
@kgodey I replied that I'd make the change, but then thought about it some more and came to the conclusion that it's not a good idea.

Having two forms for passing parameters grows the code complexity more than you'd expect, since we're parsing the function JSON spec twice. Once to rewrite IDs to names and again to instantiate a `DBFunction` instance. A change in one parsing routine, has to be kept in sync with the other routine.

I don't think that the readability improvement is significant enough to justify that. I'd like to keep the syntax as simple as possible.

I have considered switching syntax format from `{id: [param1,param2]}` to `{"id": id, "parameters": [param1,param2]}`. That way it's clear that `"parameters"` is always a sequence. I like this for being more orthodox (predictable) JSON, though what's more readable is debatable, since once you understand the current syntax, it's very efficient in terms of information density.

db/functions/operations/deserialize.py
```diff
@@ -45,7 +62,20 @@ def _get_db_function_subclass_by_id(subclass_id):
     for db_function_subclass in known_db_functions:
         if db_function_subclass.id == subclass_id:
             return db_function_subclass
-    raise UnknownDBFunctionId
+    raise UnknownDBFunctionId(
```
kgodey, line 65: `Id` --> `ID`

mathesar/functions/operations/convert.py
```diff
@@ -0,0 +1,44 @@
...
+    IDs and can only accept columns referenced by name.
+    """
+    # Using a private method to do the heavy lifting since it uses different parameters.
+    return _rewrite(
+        column_ids_to_names=column_ids_to_names,
+        spec_or_literal=spec
+    )
+
+
+def _rewrite(column_ids_to_names, spec_or_literal):
```
kgodey, line 18: I only see the JSON spec being parsed here - couldn't you check for single parameters here and convert them to a list during this rewrite?

kgodey (changes requested):
@dmos62 This looks good overall, added some minor comments. I'm fine with the current API spec for now if it's going to result in repeated logic, but I did have a question about why we couldn't put the logic in a single place.

Could you open a separate GitHub issue to reevaluate the filtering API spec with a link to this discussion? It should go in the API v1 milestone (which is post-alpha).

mathesar/functions/operations/convert.py
dmos62: The other place is `db.functions.operations.deserialize`.

dmos62:
I'll open a ticket to track this filter/functions API spec issue. I'll give it a technical debt label.

dmos62:
@kgodey Renamed `UnknownDBFunctionId` to `UnknownDBFunctionID` (Id -> ID) and removed the obsolete comment. Should be ready for merge.

kgodey (approved):
Looks good to me. I assume @pavish needs to make frontend changes before we merge?

pavish:
I'll get started on the frontend changes. I'll not be using the `/api/ui/v0/database/${DB}/filters/` endpoint, there's another ticket open to handle that: https://github.com/centerofci/mathesar/issues/1072. In this PR, I'll only fix parts that are breaking.

I have a couple things to confirm:

* @dmos62 I assume [this is the finalized format](https://github.com/centerofci/mathesar/pull/1069#issuecomment-1041558716).
* @kgodey I assume we will have to remove the `has_duplicates` filter option from the UI entirely.
 
Please let me know if these assumptions are correct.

kgodey:
> @kgodey I assume we will have to remove the has_duplicates filter option from the UI entirely.

Yes, we're designing a different way to handle that in #1028. We can create an implementation issue for duplicates once that is done.

kgodey:
@pavish I'm fine with breaking things in `master` temporarily as long as we can make sure that filtering functionality works as expected by the time #1072 is closed. I'll leave the schedule of which things get handled in which PR to you and @seancolsen.

kgodey:
@pavish Does that mean this PR can be merged as-is?

dmos62:
> I'll get started on the frontend changes. I'll not be using the `/api/ui/v0/database/${DB}/filters/` endpoint, there's another ticket open to handle that: #1072. In this PR, I'll only fix parts that are breaking.
> 
> I have a couple things to confirm before I begin:
> 
>     * @dmos62 I assume [this is the finalized format](https://github.com/centerofci/mathesar/pull/1069#issuecomment-1041558716).
> 
>     * @kgodey I assume we will have to remove the `has_duplicates` filter option from the UI entirely.
> 
> 
> Please let me know if these assumptions are correct.

@pavish Yes, your assumptions are correct. I've forgotten to post a summary of that for you, sorry! [This is the final format](https://github.com/centerofci/mathesar/pull/1069#issuecomment-1041558716) and the `duplicate_only` filter is not a filter anymore, but a query param on the `records` endpoint.

This PR shouldn't block any other developers for a while, so I'd say feel free to merge in whichever order you prefer.

@kgodey @pavish @seancolsen Would it make sense to merge into a dedicated branch (that is not master) until it's not broken anymore and then merge that into master, thus not breaking master?

kgodey:
> @kgodey @pavish @seancolsen Would it make sense to merge into a dedicated branch (that is not master) until it's not broken anymore and then merge that into master, thus not breaking master?

I don't think it's worth the fuss at this point since no one is using `master` for anything important. After our alpha release, we should not be breaking `master` anymore.

pavish:
> @pavish Does that mean this PR can be merged as-is?

@kgodey Yes, I think we can merge this PR as-is.

PR #1076, opened 2022-02-15
Menu components

### LINKED ISSUE #742: Create Menu component
From what I can tell, we're going to end up with multiple places throughout the app where we want dropdown menus which behave similar to the typical "File" menus atop a desktop application.

From [a discussion about styling dropdown menus](https://github.com/centerofci/mathesar/discussions/741), @ghislaineguerin proposed this mockup:

mockup

Additional thoughts:

- I'd also like to see the text for all menu items left-aligned with each other regardless of whether an icon is present.

- We might also want nested menus at some point, so it would be worth considering that when setting up this UI component.

Here's an example of a dropdown menu with other features that we may someday want to support

image

## Initially proposed designs

- For the most common use-case with a trigger that looks like a button with a downward pointing chevron

    ```svelte
    <DropdownMenu icon={} label="" >
      <DropdownMenuItem icon={} label="" onClick={}/>
      <DropdownMenuItem icon={} label="" onClick={} isDisabled/>
      <DropdownMenuDivider />
      <DropdownSubMenu icon={} label="">
        <DropdownMenuItem icon={} label="" onClick={}/>
      </DropdownSubMenu>
    </DropdownMenu>
    ```

- With a custom trigger

    ```svelte
    <DropdownMenu>
      <svelte:fragment slot="trigger">
        put anything here
      </svelte:fragment>
      <DropdownMenuItem icon={} label="" onClick={}/>
      <DropdownMenuItem icon={} label="" onClick={}/>
    </DropdownMenu>
    ```

# Finalized component design

```svelte
<Menu>
  <MenuItem icon={} label="" onClick={}/>
  <MenuItem icon={} label="" onClick={}/>
  <MenuItem>
     Custom menu item
  </MenuItem>
  <MenuDivider/>
  <NestedMenu icon={} label="">
    <MenuItem icon={} label="" onClick={}/>
    <MenuItem icon={} label="" onClick={}/>
  </NestedMenu>
</Menu>
```

seancolsen:
Fixes #742

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.d.ts
+++ b/mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.d.ts
@@ -1,17 +1,6 @@
-import type { IconDefinition } from '@fortawesome/fontawesome-common-types';
-import type {
-  IconFlip,
-  IconRotate,
-} from '@mathesar-component-library-dir/types';
-
-export interface IconDetails {
-  data: IconDefinition;
-  spin?: boolean;
-  flip?: IconFlip;
-  rotate?: IconRotate;
-}
+import type { IconProps } from '@mathesar-component-library-dir/icon/Icon.d';
 
 export interface ButtonDetails {
   label: string;
-  icon: IconDetails;
+  icon: IconProps;
 }
--- a/mathesar_ui/src/component-library/confirmation/ConfirmationController.ts
+++ b/mathesar_ui/src/component-library/confirmation/ConfirmationController.ts
@@ -1,24 +1,13 @@
-import type { IconDefinition } from '@fortawesome/fontawesome-common-types';
+import type { IconProps } from '@mathesar-component-library-dir/icon/Icon.d';
 import { faArrowLeft, faCheck } from '@fortawesome/free-solid-svg-icons';
 import type { Writable } from 'svelte/store';
 import { writable } from 'svelte/store';
 import type { ModalController } from '@mathesar-component-library-dir/modal';
-import type {
-  IconFlip,
-  IconRotate,
-  ComponentAndProps,
-} from '@mathesar-component-library/types';
-
-interface IconDetails {
-  data: IconDefinition;
-  spin?: boolean;
-  flip?: IconFlip;
-  rotate?: IconRotate;
-}
+import type { ComponentAndProps } from '@mathesar-component-library/types';
 
 interface ButtonDetails {
   label: string;
-  icon?: IconDetails;
+  icon?: IconProps;
 }
 
 export interface ConfirmationProps {
--- a/mathesar_ui/src/component-library/icon/Icon.d.ts
+++ b/mathesar_ui/src/component-library/icon/Icon.d.ts
@@ -1,3 +1,5 @@
+import type { IconDefinition } from '@fortawesome/fontawesome-common-types';
+
 export enum IconFlip {
   VERTICAL = 'vertical',
   HORIZONTAL = 'horizontal',
@@ -8,4 +10,21 @@ export enum IconRotate {
   NINETY = '90',
   ONE_EIGHTY = '180',
   TWO_SEVENTY = '270',
+}
+
+/**
+ * NOTE:
+ * This interface represents all the props listed in `Icon.svelte`.
+ *
+ * After https://github.com/sveltejs/language-tools/issues/442 is fixed, we
+ * should hopefully be able to clean up this code duplication a bit.
+ */
+export interface IconProps {
+  data: IconDefinition;
+  size?: string;
+  spin?: boolean;
+  pulse?: boolean;
+  flip?: IconFlip;
+  rotate?: IconRotate;
+  label?: string;
 }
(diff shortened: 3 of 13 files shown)


pavish (changes requested):
@seancolsen I have a few comments:

1. We decided that styles in component-library would always remain global. This component seems to go against that decision and I'd like us to stay consistent. Please move the styles to dedicated files and export them through styles.scss
2. The css for menu items and hovering needs to be improved. I am okay with going with grid but it introduces styling issues as you've mentioned. I'd suggest not to use grid. The gap issue you've specified in your comments with respect to hover also occurs with padding in Safari.
   Screenshot 2022-03-08 at 10 43 08 AM
   I'm using version 14.1.2 which is quite recent. Please make sure your code works as expected in most recent versions of all browsers.
3. Menu component needs to support keyboard interactions (similar to dropdown). We don't have to do it in this PR and we can take it up later. Please create an issue for it.

seancolsen:
@pavish ready for re-review.

1. Global styles

    I was hoping we could relax that rule now that we're aiming to focus our development resources strictly on Mathesar -- not on the component library. It's more work to maintain the global CSS pattern. I need to create more files. I need to switch between files more as I'm developing. It's harder to ensure I'm catching everything while refactoring. It's harder to avoid naming collisions. It's harder to maintain component encapsulation. It slows me down. With the component library goal off the table for now, the only benefit that I see to maintaining the global CSS pattern is adhering to "consistency". But at this point, that "consistency" would only be within one subdirectory of the project -- `component_library`. We're already inconsistent within the Mathesar project as a whole, and that project is our primary focus. To me, maintaining consistency within one subdirectory of the project provides a negligible benefit, whereas the _cost_ of maintaining that consistency is a noticeable loss of productivity for me. Does consistency within that one subdirectory offer some meaningful benefit to Mathesar that I'm not seeing?

    All that said, I don't want to push too hard on this. I can make the additional files and link them up if you really feel strongly about this. I just wanted to take a pause here to give us a chance to look at this pattern with fresh eyes and ask _"is this pattern actually helping Mathesar?"_

1. Safari style bug

    Thanks for catching this! It should now be fixed in a89dea23ace60541c4eac96585e200e808f0181d. Can you verify that it's working for you?

    > I'd suggest not to use grid.

    I had some reservations about grid, given the challenges here. But what else would you suggest? The goal of aligning the icons and controls with respect to each other seems fairly crucial to me, stylistically. Without grid, the only other approach I could come up with was to build a bidirectional communication system between `MenuItem` and `Menu` through context wherein each `MenuItem` would report the width of its control and the width of its icon. Then the `Menu` would tell each `MenuItem` how to position its constituent pieces. That's potentially even uglier than the hacks required to get grid to work. Can you think of something simpler? If so, I'm open to it. Or if you don't have substantial concerns with grid, I think it would be good to stick with it, since (as far as I can tell) it seems to be working well. The Safari bug was just due to the browser default stylesheet having margin on `<button>`. It seems to me that it was nothing to do with grid.

    > Please make sure your code works as expected in most recent versions of all browsers.

    I certainly should have done some cross-browser testing for this PR. This is a good reminder for me. Thanks!
    
    But you're not suggesting that I test _every change_ I make in all browsers, are you? Do _you_ do that? I don't think that would be a good use of time. And I don't even have access to test on actual Safari since I'm on Linux. I can sort of fake it by using Gnome Web but it's not perfect. It uses a older version of WebKit. It's hard to even figure out which version of Safari it's equivalent to, but it's definitely an older one. I actually notice that the checkboxes don't render correctly on Gnome Web. They're still usable, but they don't look as good. I notice they look fine in your screenshot though. Discrepancies like this checkbox thing make it a little hard for me to rely on Gnome Web as a workflow for regular cross-browser testing. My typical dev workflow is to pause every so often and do some cross-browser (and cross-device) testing on real Apple products. I might borrow a friend's iPad for that sort of thing, for example.
    
1. Keyboard navigation

    Good thinking. Since none of the current menus have keyboard navigation, I see that concern as being separate from this PR though. So I don't think we should hold up this PR, even for the creation of the ticket.

mathemancer:
@seancolsen Does this actually need back end review?

seancolsen:
@pavish This is ready for re-review

Pavish and I chatted on a call about the choice of CSS grid and the visual alignment goals. We agreed that we'd continue using grid here for now.

I'd rather not take the time to continue debating about global CSS right now. I've pushed 8167fb5c5a90f8b76d13d3fa7eb088e6a30414b8 which converts the CSS to global.

pavish (approved):

pavish:
@mathemancer This PR does not actually require backend review, there is [one change with regards to the locator in an e2e test file](https://github.com/centerofci/mathesar/pull/1076/files#diff-9bc31676a2c9531376b9002f3eda04942e23431cf24052d05ceb7eed4057c2fe) which tags backend reviewers. Could you please take a look and approve this PR?

kgodey (approved):
Since @mathemancer is out today, approving from the backend to unblock merge.

@pavish or @seancolsen, can you submit a PR updating the CODEOWNERS for the E2E files to frontend?

pavish:
@kgodey I think it might be best to keep the code owner as backend reviewers for the E2E files for a short while. We're still at the early stages of it and we'll be updating/creating fixtures, common utilities etc., We can change it once we have a range of tests and enough common code in place. Thoughts?

kgodey:
@pavish That's fine.

PR #1079, opened 2022-02-17
Add URI and text filters

### LINKED ISSUE #406: Implement filtering options for Text types.
## Problem
We need to ensure that records that include columns of all Text types support the following filters via API: 
- contains {x}
- is {x}
- is not {x}
- starts with {x}
- ends with {x}
- does not contain {x}
- does not start with {x}
- does not end with {x}
- is empty
- is not empty

This involves:
  - Filter information should include the number of parameters needing to be passed in (e.g. `contains` needs 1 parameter, `is empty` needs 0)

## Additional context
- This issue is blocked by implementation of Text types: 
  - #404 
  - #405 
  - #407

### LINKED ISSUE #413: Implement filtering options for the URI type
## Problem
We need to ensure that records that include columns of the custom URI type support the following filters via API: 
- contains {x}
- does not contain {x}
- is {x}
- is not {x}
- is empty
- is not empty
- **host contains {x}**
- **host does not contain {x}**
- **scheme is [select from list of choices of valid schemes]**
- **scheme is not [select from list of choices of valid schemes]**

The ones in bold are unique to the URI type.

This involves:
  - Filter information should include the number of parameters needing to be passed in (e.g. `is` needs 1 parameter, `is empty` needs 0)
  - Filter information should also include the list of valid choices for the filter (e.g. scheme filters)

## Additional context
- Blocked by #412 

dmos62:
Fixes #413, Fixes #406

Implements following filters:
- x contains y,
- x contains (case insensitive) y,
- x starts with (case insensitive) y,
- x URI authority contains y,
- x URI scheme equals y.

Other filters required in #413 and #406 have been implemented in previous PRs.

**Technical details**

The new filters look like this on the `filters` endpoint:

```json
    {
        "id": "contains",
        "name": "contains",
        "parameters": [
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            },
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            }
        ]
    },
    {
        "id": "starts_with_case_insensitive",
        "name": "starts with (case insensitive)",
        "parameters": [
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            },
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            }
        ]
    },
    {
        "id": "uri_authority_contains",
        "name": "URI authority contains",
        "parameters": [
            {
                "ui_types": [
                    "uri"
                ]
            },
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            }
        ]
    },
    {
        "id": "uri_scheme_equals",
        "name": "URI scheme equals",
        "parameters": [
            {
                "ui_types": [
                    "uri"
                ]
            },
            {
                "ui_types": [
                    "text",
                    "uri"
                ]
            }
        ]
    }
``` 

Notice that the Mathesar type `uri` is string-like, just like `text`, so it's allowed wherever `text` is allowed (see `contains` parameters' types for an example). In contrast, `uri_authority_contains` first parameter can only be `uri`. Same for `uri_scheme_equals`.

Tests for new filters have been added.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
Notice that in the issue #413, the `x URI scheme equals y` filter is specified to allow the user to choose from a list of URI schemes. This PR *does not* implement that.

It could be done somewhat easily. Just add to the relevant parameter a hint like `hint.suggested_values(("http","https","ftp","ftps",...))`. Or, give the parameter a hint specifying that it must be either just a string-like or a uri-scheme: then the frontend would know what kinds of suggestions would be appropriate there. This would require putting some `and`-`or` logic into the hint system though, which will be necessary down the road, but maybe not so much now.

At the same time, I think that scheme suggestions is not an important feature right now, so I propose we create an issue for it and put it off.

dmos62:
Note that in this PR the filters endpoint response includes some filters multiple times. This is fixed in https://github.com/centerofci/mathesar/pull/1090.

db/functions/redundant.py
```diff
@@ -4,18 +4,33 @@
...
+from db.functions.base import DBFunction, Or, Lesser, Equal, Greater, StartsWith, Contains, ToLowercase
 
 from db.functions import hints
 
 
-# TODO (tech debt) define in terms of other DBFunctions
-# would involve creating an alternative to to_sa_expression: something like to_db_function
-# execution engine would see that to_sa_expression is not implemented, and it would look for
-# to_db_function.
+class RedundantDBFunction(DBFunction):
```
kgodey, line 14: I find this name confusing. Perhaps something like `DBFunctionCombination` would be clearer?

db/functions/redundant.py
```diff
@@ -40,6 +59,48 @@ class GreaterOrEqual(DBFunction):
...
+        param1 = self.parameters[1]
+        return Or([
+            Greater([param0, param1]),
+            Equal([param0, param1]),
+        ])
+
+
+class StartsWithCaseInsensitive(RedundantDBFunction):
+    id = 'starts_with_case_insensitive'
+    name = 'starts with (case insensitive)'
```
kgodey, line 73: If this name is meant to be used in the frontend, please remove ` (case insensitive)` from the `name` parameter.

kgodey (changes requested):
A couple of changes requested in individual comments below. I'd also like @mathemancer to take a look at this before merge since it touches the URI type.

> Notice that in the issue #413, the `x URI scheme equals y` filter is specified to allow the user to choose from a list of URI schemes. This PR *does not* implement that.
> 
> At the same time, I think that scheme suggestions is not an important feature right now, so I propose we create an issue for it and put it off.

I think we should do this, it is important for making the product friendly to non-technical users, who may not know what a URI scheme is.

> It could be done somewhat easily. Just add to the relevant parameter a hint like `hint.suggested_values(("http","https","ftp","ftps",...))`.
>
> This would require putting some `and`-`or` logic into the hint system though, which will be necessary down the road, but maybe not so much now.

I'm not sure why this would need "`and`-`or` logic",  can you elaborate?

db/functions/redundant.py
```diff
@@ -40,6 +59,48 @@ class GreaterOrEqual(DBFunction):
...
+        param1 = self.parameters[1]
+        return StartsWith([
+            ToLowercase([param0]),
+            ToLowercase([param1]),
+        ])
+
+
+class ContainsCaseInsensitive(RedundantDBFunction):
+    id = 'contains_case_insensitive'
+    name = 'contains (case insensitive)'
```
kgodey, line 92: Case insensitive comment from above applies here as well.

db/functions/redundant.py
dmos62: It is. If the suffix is removed, it will have the same display name as a regular (case-sensitive) `starts with` filter: a user won't be able to distinguish the two.

db/functions/redundant.py
dmos62: I think DBFunctionCombination can be even more confusing, since a regular DBFunction instance can contain other DBFunction instances as parameters, and thus could be called a combination or supporting combination. I like redundant, because it highlights that it's made of DBFunctions that could be used without this particular DBFunction. I'm open to other names. Maybe `SecondaryDBFunction`?

dmos62:
>> It could be done somewhat easily. Just add to the relevant parameter a hint like `hint.suggested_values(("http","https","ftp","ftps",...))`.
>>
>> This would require putting some `and`-`or` logic into the hint system though, which will be necessary down the road, but maybe not so much now.

>I'm not sure why this would need "`and`-`or` logic",  can you elaborate?

I wasn't clear enough. These are the options I mentioned:

1. > It could be done somewhat easily. Just add to the relevant parameter a hint like hint.suggested_values(("http","https","ftp","ftps",...)).
2. > Or, give the parameter a hint specifying that it must be either just a string-like or a uri-scheme: then the frontend would know what kinds of suggestions would be appropriate there. This would require putting some and-or logic into the hint system though, which will be necessary down the road, but maybe not so much now.

If this is wanted for alpha, I'll hack something together into a dedicated PR.

dmos62:
@kgodey @mathemancer note that the changes in the `db/types/uri.py` are just me adding URI-specific DBFunctions to that namespace. I don't make changes to the exiting URI type code.

db/functions/redundant.py
kgodey: I like `DBFunctionPackage`.

db/functions/redundant.py
kgodey: There shouldn't be a case-sensitive "starts with" in the frontend, only the case insensitive version. 

We don't want to overwhelm users with a lot of similar filtering options, it will make finding the filter they want harder. And non-technical users may not know what "case sensitive" refers to.

kgodey:
> If this is wanted for alpha, I'll hack something together into a dedicated PR.

Okay, thanks, please do. Option 1 seems easiest.

dmos62:
Ready for review.

I've removed case-sensitive string filters from the filters endpoint.

Other requested changes will be submitted as dedicated PR/s. Those changes are:
  - rename RedundantDBFunction to something better;
  - have appropriate filters suggest popular/standard URI schemes
    - https://github.com/centerofci/mathesar/issues/1097

kgodey (approved):
Looks good to me, I'll leave it to @mathemancer to review/merge.

db/functions/redundant.py
```diff
@@ -4,18 +4,33 @@
...
+    A DBFunction that is meant to be unpacked into another DBFunction. A way to define a DBFunction
+    as a combination of DBFunctions. Its to_sa_expression method is not to used. Its concrete
+    implementations are expected to implement the unpack method.
+    """
+    @staticmethod
+    def to_sa_expression(*_):
+        raise Exception("UnpackabelDBFunction.to_sa_expression should never be used.")
+
+    @abstractmethod
+    def unpack(self):
```
mathemancer, line 25: For my own education:  Should implementations of this `unpack` method recurse "in place"?  I.e., if I happen to have a redundant (or whatever we decide for a term) function that's composed of other redundant functions, should the `unpack` method call the unpack methods of those functions as well, or should that be left to the caller?  I'm not sure since you'd want to avoid having the recursion in multiple places, and `_db_function_to_sa_expression` is already recursive.

db/types/base.py
```diff
@@ -131,10 +146,22 @@ def _add_to_db_type_hintsets(db_types, hints):
...
     all_db_types = known_db_types
     hints_for_all_db_types = (hints.any,)
     _add_to_db_type_hintsets(all_db_types, hints_for_all_db_types)
 
+    # string-like types get the "string_like" hint
+    string_like_db_types = (
+        PostgresType.CHARACTER_VARYING,
+        PostgresType.CHARACTER,
+        PostgresType.TEXT,
+        MathesarCustomType.URI,
```
mathemancer, line 159: This sort of info is also in the `cast.py` file.  Did you consider trying to unify these with those definitions?  I ask because it makes me nervous to have this defined in more than one place.

mathemancer (changes requested):
I mostly just had questions; the only real request is to try to unify the definitions of things like "string-like" "number-like", etc. between the definitions in the `db.types.operations.cast` module and the hint building logic.

db/functions/base.py
```diff
@@ -225,6 +230,36 @@ def to_sa_expression(*values):
 class StartsWith(DBFunction):
     id = 'starts_with'
     name = 'starts with'
+    hints = tuple([
```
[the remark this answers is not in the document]
dmos62: I find the shorthand tuple syntax (`(x,)`) to be errorprone. Just yesterday I ran into an annoying bug where I had forgotten the trailing comma and what was supposed to be a single-element tuple ended up being just the element. So I try to do the more awkward, but safer `tuple([x])`. This is not a very strong preference. After getting bit by this a few times you learn to append a comma to everything, I guess.

db/functions/base.py
```diff
@@ -73,7 +77,7 @@ class Literal(DBFunction):
     name = 'as literal'
     hints = tuple([
         hints.parameter_count(1),
-        hints.parameter(1, hints.literal),
+        hints.parameter(0, hints.literal),
```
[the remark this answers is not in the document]
dmos62: I wanted to use 1-based indexing, but changed my mind. Everything in Python is 0-based, so I decided to go with the flow.

db/functions/redundant.py
dmos62: I think it's desirable not to have to do manual recursion inside `unpack`. I don't immediately recall if I implemented full recursion yet. I think I did, since I was concerned about infinite loops.

db/types/base.py
dmos62: I didn't notice the duplication. I'll investigate.

dmos62:
@mathemancer could we get this merged? It's the base for a few other PRs. I'll open an issue for the duplicated logic across `db.types.operations.cast` and `db.types.base` modules and see about fixing that in a new PR.

mathemancer (approved):
Approving, since I agree we should probably get this merged to unblock downstream PRs.  Please don't forget to create the issue about the duplicated definitions of type tags (string-like, number-like, etc.)

dmos62:
@mathemancer the issue for that is here: https://github.com/centerofci/mathesar/issues/1100

Issue #257, opened 2021-06-14
Frontend implementation for Boolean Mathesar type


kgodey:
## Requirements

Users should be able to:
- view data in columns that have the Boolean Mathesar type
- change data type to and from the `BOOLEAN` DB type
- view, set, and change data type options related to the Boolean data type
- ~~filter data by Boolean-specific filtering options~~
- use display options to view data in Boolean type columns
  - The checkbox display option should display data as checkboxes
  - The dropdown display option should display data as dropdowns and only allow boolean options to be input
  - If custom labels are selected, the labels for the dropdowns should be adjusted accordingly.
 
Update:
- Filtering will be handled in #1072 

## DB Type Mapping
There is only one DB type mapping to this UI type, so no special mapping rules are necessary.

### Design
- The [Global Data Types](https://wiki.mathesar.org/design/specs/global-data-type-components) spec has the latest version of the UX.
- The [Boolean Data Type](https://wiki.mathesar.org/design/specs/boolean-data-type) spec has an older version of the UX but more information on user scenarios.
- The [NULL values](https://wiki.mathesar.org/design/specs/specs-null-values) spec has requirements for displaying and editing NULL values within checkbox columns. (See "Scenario 1c" and "Scenario 2c".)

### Backend
- `/api/v0/tables/` lists the name and DB type of each columns 
- `/api/v0/tables/<id/columns/` lists column information in more detail
- `/api/v0/databases/<id>/types/` maps Mathesar types to DB types and provides a list of filtering and grouping options per-type.
  - These should be used to populate the filter and grouping options dynamically. 
- Updates to column types and type options should happen via `PATCH` requests to `/api/v0/tables/<id>/columns/<id>/`. These updates include
  - Changes to the data type (`type`)
  - Changes to the display options (`display_options`)

## Additional Context
- #237 (design)
- #269 (concept of Mathesar types)
- #498 (conversation about data types)
- Blocked by backend work:
  - #387
  - #388 
  - #390

PR #1080, opened 2022-02-17
Text type & basic Boolean type frontend implementation

### LINKED ISSUE #254: Frontend implementation for text data types
Blocked by #234 and #244. Details will be added once those issues are closer to completion.

pavish:
Fixes #254 
Related to #257 
Related to #928 

**Features**
* Adds text type frontend support
  - Choosing db type based on field size limit
    - length: null, field size restricted -> VARCHAR
    - length < 255, field size restricted -> CHAR
    - field size un-restricted -> TEXT
  - Cell input & size changes based on dbType and length property
    - Input is textarea for TEXT
    - Input changes to textarea for VARCHAR when length > 255 and textbox otherwise
    - Input is textbox for CHAR
    - The issue description mentions:
      - > columns with large amounts of text should expand when selected
      - This is currently done based on dbtype and type_options.length
      - If content exceeds a line, it expands to two lines on selection
      - When content is edited, it opens up the textarea which is resizable
      - Note:
        - Implementation slightly varies from the Figma design
        - Expanding based on exact content length instead of type_options.length is tricky and we can consider that later
* Adds basic boolean type frontend support
  - Only supports checkbox for now

**Regressions**
* Null placeholder is no longer shown when in edit mode. This is a regression we need to handle in upcoming PRs, after we implement other types.

**Improvements/Changes**
* Cell value only gets saved when user is done editing. However, escape does not reset to old value like spreadsheets.
* Some naming improvements based on conventions.
* TextInput component is made leaner, events such as enter, esc etc., are removed and parent components check them using keydown.

**UX discussions required**
* New row now contains empty checkbox if one of the columns is a boolean column. We need a UX reconsideration for the new row placeholder.
* Interactions differ widely for each cell type, we need to analyze and hash out the interaction specs.

**Backend issues**
* Responses for CHAR columns with fixed length contains empty space to fill for the entire length. While this is expected behaviour of the DB, this results in bad UX.
* Changing types to boolean from text fails with a 500 error. This is because the request contains `type_options: null` while boolean type does not accept any type_options. However considering that type_options is a property of Column, null value should be accepted, and any other value should throw an error.
* I will create issues for these later.

**E2E tests**
* This PR does not contain any e2e tests. I will add them in upcoming PRs when the essential types are implemented. I expect more refactoring to take place as more types are implemented and it would be better to write solid tests once the component structure suits all requirements.

**Video**
<a href="https://www.loom.com/share/ec2d03b57aa043c4a226774322dbf877">
    <p>Text type and basic boolean type frontend implementation - Watch Video</p>
    
  </a>

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
I watched the video, it looks great @pavish. Nice work, it's nice to see this functionality coming together.

> New row now contains empty checkbox if one of the columns is a boolean column. We need a UX reconsideration for the new row placeholder.

We're supposed to be using indeterminate state checkboxes for `NULL` value booleans. See the [NULL value spec](https://wiki.mathesar.org/en/design/specs/specs-null-values). We should only be showing empty checkboxes for `FALSE` values.

kgodey (approved):
Approved from the backend side. Perhaps we need to change the codeowners for `mathesar/tests/integration/` to be frontend?

pavish:
@kgodey 

> We're supposed to be using indeterminate state checkboxes for NULL value booleans. See the [NULL value spec](https://wiki.mathesar.org/en/design/specs/specs-null-values). We should only be showing empty checkboxes for FALSE values.

We currently do this for existing and newly created records. It can be seen in the video. The style of indeterminate checkboxes is pretty similar to checked state, we can update that in the upcoming boolean type PR.

The new record placeholder however is very different, as it is not an actual record. The false state checkbox and empty cells in the placeholder are because the value is `undefined` instead of `null` If we set NULL values for all columns in the placeholder, the client will send `null` for the columns while creating the row and override the default value which needs to be set on the db.

This requires some UX rethought and it's the reason I've mentioned it on the UX discussions required section.

> Approved from the backend side. Perhaps we need to change the codeowners for mathesar/tests/integration/ to be frontend?

Yes, I think we should do that.

seancolsen (approved):

seancolsen:
I've looked over this and tested it out. Mostly everything is working great! Exciting to see these new features!

I have jotted down some additional comments, critique, and questions. Tomorrow I'll flesh them out into tickets for further discussion and action.

seancolsen:
Here are the tickets/PRs I created in lieu of a review:

- [x] #1111
- [ ] #1112
- [ ] #1113
- [ ] #1114
- [ ] #1115
- [ ] #1116

PR #1088, opened 2022-02-21
Datetimes to RFC 3339

### LINKED ISSUE #1086: Return RFC 3339 (ish) style date / time values from API
## Problem

Currently, the style of date/time values returned by the API is whatever the Django default for handling `python` date / time types is.  Further, since the dates and times are made into `python` types during the process, precision is lost, as well as functionality around dates before the common era, and dates with years containing more than 4 digits.

## Proposed solution

We should parse the dates and times into RFC 3339 compliant strings in the database, and then pass those strings up through the API.  We will need to make the following slight extensions and refinements
- We will append `AD` or `BC` to each date or datetime
- We will allow years with more than 4 digits
- We will always take the 'verbose' version when there's a choice (e.g., a timezone will always be `+08:00`, not `+08`).

## Additional context

There is a spec being written in the wiki in parallel with the implementation of this issue.  The PR is here:  https://github.com/centerofci/mathesar-wiki/pull/35 .

mathemancer:
Fixes #1086 

This changes the `TIME_WITH_TIME_ZONE`, `TIME_WITHOUT_TIME_ZONE`, `TIMESTAMP_WITH_TIME_ZONE`, and `TIMESTAMP_WITHOUT_TIME_ZONE` types to comply with the spec in the issue.  It adds a custom `DATE` type to comply with the spec in the issue.  It adds numerous tests to make sure the types are properly parsing and returning datetime values.

**Technical details**

It also sets up a framework in the `db/tests/types/test_datetime.py` file so we can test more date time string parsing easily in the future.  The framework set up is much faster than the column alteration tests.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@dmos62 There's a slight flaw in the filtering tests (see my changes to `db/tests/functions/operations/test_filter.py`.  Basically, we're filtering on the DB (which works), then checking whether the output matches the input.  This conflates two things.  We don't care if the output matches the input in Python.  In fact, a proper test would allow testing that we can input _anything_ that gets coerced in the DB to the proper type, then filter on that.  For example, I should be able to filter on dates which are input in non-RFC3339-compliant ways (this is fine), and it should be possible to test that those filters work (not possible to test with the current framework).

Options:
- I fix those tests in general in this PR (I assume this is going to conflict with your current work)
- You fix those tests in general in one of your current PRs (my preference)
- We make an issue and write those tests later.

What do you think?

dmos62:
@mathemancer let's open an issue for that and I'll do a dedicated PR addressing that. I'm still fuzzy on what you mean though. Could you write the issue and provide an example?

db/tests/conftest.py
```diff
@@ -37,6 +37,7 @@ def engine_with_roster(engine_with_schema):
 @pytest.fixture
 def engine_with_filter_sort(engine_with_schema):
     engine, schema = engine_with_schema
+    engine.dialect.ischema_names.update(types.CUSTOM_TYPE_DICT)
```
dmos62, line 40: Could you comment on why this is needed? I've been writing a lot of these fixture these past few days and some of the mechanics are still somewhat opaque to me.

kgodey (approved):
Looks good to me.

kgodey:
Since @dmos62 didn't request changes in his review, I'm going to merge this.

db/tests/conftest.py
mathemancer: @dmos62 You need to either use a fixture that creates (or uses) an engine that includes our custom types, or you need to manually add the custom types to the engine for the custom types to be usable in filtering.  While `DATE` isn't a custom DB type, it _is_ a custom python type.

PR #1091, opened 2022-02-21
Implement display options for Duration type

### LINKED ISSUE #433: Implement display options for Duration type
## Problem
The Duration Mathesar type supports the following display options:
- Min unit
- Max unit
- Whether to show unit labels

We will combine this into a single `format` string.

## Solution
(1) We should store these display options in the following format in the `display_options` field of the corresponding column.
```
{
  "format": "HH:mm:ss.SSS"
}
```

(2) We should also validate these so that:
- Only columns of this type can have these display options. They should not be able to be set if the column is of a different type.
- `format` should be a valid JavaScript duration format.

(3) If the column type is changed, the display options should be deleted.

(4) We should add supported display options to the `types` endpoint.

## Additional Context
- [Design of Duration type options on Figma](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=4260%3A37440&node-id=4270%3A41231&viewport=324%2C48%2C0.29&scaling=contain&starting-point-node-id=4270%3A41231&show-proto-sidebar=1)
- Blocked by #658
- #392

silentninja:
Fixes #433 

Adds display options for the `duration type`.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/api/serializers/shared_serializers.py
```diff
@@ -180,6 +180,14 @@ def validate(self, datetime_obj, display_format, serializer_field):
         return super().validate(datetime_obj, display_format, serializer_field)
 
 
+class DurationFormatValidator(TimeWithoutTimeZoneFormatValidator):
+    """
+    The validation is same as Time Without TimeZone as of now,
```
kgodey, line 185: This does not make sense. Durations can span weeks, months, days, and years. Times cannot.

kgodey (changes requested):
@silentninja It seems like this point from #433 is not addressed here:

> (4) We should add supported display options to the types endpoint.

See individual comments also.

kgodey:
I also wanted to verify that this point was addressed, I don't see it anywhere.

> (3) If the column type is changed, the display options should be deleted.

mathesar/api/serializers/shared_serializers.py
silentninja: Fixed it, it was meant to be timestamp. I have changed the validation method anyway since we need a different error message, though the logic is similar.

silentninja: Also, I haven't restricted using a `weekdate` format string(2019-W17), though the duration value returned from the database follows RFC 3339 and does not return a weekdate, as the user might have a different requirement

silentninja:
> I also wanted to verify that this point was addressed, I don't see it anywhere.
> 
> > (3) If the column type is changed, the display options should be deleted.

This was handled as part of the initial display options PR, it is taken care of [when a column is reflected](https://github.com/centerofci/mathesar/blob/94c540545d876969117b53f959b3f7b9657e7c6b/mathesar/reflection.py#L79)

silentninja:
> (4) We should add supported display options to the types endpoint.

[Available Display options are part](https://github.com/centerofci/mathesar/blob/94c540545d876969117b53f959b3f7b9657e7c6b/mathesar/api/serializers/databases.py#L37) [of the types serializer](https://github.com/centerofci/mathesar/blob/94c540545d876969117b53f959b3f7b9657e7c6b/mathesar/api/ui/viewsets/databases.py#L29)

kgodey:
> [Available Display options are part](https://github.com/centerofci/mathesar/blob/94c540545d876969117b53f959b3f7b9657e7c6b/mathesar/api/serializers/databases.py#L37) of the [types serializer](https://github.com/centerofci/mathesar/blob/94c540545d876969117b53f959b3f7b9657e7c6b/mathesar/api/ui/viewsets/databases.py#L29). I have [added it to the list](https://github.com/centerofci/mathesar/blob/9b7acc8e1373bc53b2a7e8dccd436b1cb53d729b/mathesar/api/display_options.py#L22)

Sorry I missed this, I think I reviewed too many PRs in a row. :/

kgodey (approved):

PR #1090, opened 2022-02-21
Add filters for money and boolean

dmos62:
Fixes #418, fixes #388

This PR makes sure that the relevant filters work with boolean types ([tests](https://github.com/centerofci/mathesar/pull/1090/files#diff-74daef53a2da6ea185b3bb3b02a57c533e9ded6cc11cc17686fee419eb3d272cR215) are added).

Also it adds money type support to the the filters endpoint. See below excerpt from filters endpoint response:

```
  {                                                                                                                                                            
    "id": "greater",                                                                                                                                           
    "name": "is greater than",                                                                                                                                 
    "parameters": [                                                                                                                                            
      {                                                                                                                                                        
        "ui_types": [                                                                                                                                          
          "datetime",                                                                                                                                          
          "duration",                                                                                                                                          
          "money",                                                                                                                                             
          "number"                                                                                                                                             
        ]                                                                                                                                                      
      },                                                                                                                                                       
      {                                                                                                                                                        
        "ui_types": [                                                                                                                                          
          "datetime",                                                                                                                                          
          "duration",                                                                                                                                          
          "money",                                                                                                                                             
          "number"                                                                                                                                             
        ]                                                                                                                                                      
      }                                                                                                                                                        
    ]                                                                                                                                                          
  }
```

Also included is a fix for a bug where some filters appeared in filters endpoint response multiple times.

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

[...]

PR #1098, opened 2022-02-24
Hint system improvements

### LINKED ISSUE #1097: Suggest popular/standard URI schemes as values for appropriate filter parameters
## Problem
We have a filter (`UriSchemeEquals`) that takes a URI scheme as its second parameter (first too, but that's irrelevant). We want to provide suggestions for when a user is entering/choosing a scheme to filter for.

## Proposed solution
Alter the filters endpoint so that it includes suggested values for that specific parameter.

Use [this Wikipedia article](https://en.wikipedia.org/wiki/List_of_URI_schemes) as source of known schemes.

## Additional context
https://github.com/centerofci/mathesar/issues/413
https://github.com/centerofci/mathesar/pull/1079

dmos62:
Makes some under-the-hood improvements, namely ability to add suggested values through hints, and also a way to add hints that are not for describing type attributes.

No API changes.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@dmos62 Although this PR looks good, I'd like to consider an alternate or additional implementation.

I don't think showing the entire list of possible schemes will provide a good UX to users. Most URI data will not use a majority of the schemes listed here and having to find `http` or `https` in a long list will be an annoying experience.

I think it would be ideal to return a list of schemes used in the data of a particular column in the columns endpoint. e.g. imagine the output of `/api/ui/v0/tables/1/columns/2/` could include a key that lists the schemes used in that particular column. The frontend could use these to display as choices. 

Perhaps it could look something like:
```
{
  // other column keys go here
  "filtering_choices": [
    "<filter_id_here>": ["choice1", "choice2"] 
  }
}
```

It would be extendable to choices for other filters.

Please note that this involves setting up a UI API for tables and columns, but we'll need those eventually. We should also probably move column display options to this API, but that would be out of scope here.

kgodey:
Also tagging @pavish and @mathemancer in case they have thoughts.

dmos62:
@kgodey that's something I had in mind at some point, but put it off because wasn't sure if this is the right time to implement that.

My idea is let the frontend query for unique values in a column and use that for suggesting parameters for the equality filter. That doesn't actually require anything new from the filter API. We'd just have to give frontend a way to make this single-column unique-only query.

pavish:
@kgodey @dmos62 

> I think it would be ideal to return a list of schemes used in the data of a particular column in the columns endpoint
> imagine the output of /api/ui/v0/tables/1/columns/2/ could include a key that lists the schemes used in that particular column.

I think this would help provide a better UX.

> I don't think showing the entire list of possible schemes will provide a good UX to users.

We can show an auto-complete box instead of a select menu. There could be a max of 10 values displayed at a time and it will change as the user types.

pavish:
@dmos62 

> My idea is let the frontend query for unique values in a column and use that for suggesting parameters for the equality filter. That doesn't actually require anything new from the filter API. We'd just have to give frontend a way to make this single-column unique-only query.

I had this in mind too and we can provide this for a small set of filters (equals, not equals etc.,) i.e filters that use the values stored in the columns directly.

If we need to provide suggestions from stored rows for other filters like `uri_scheme_equals`, we will require additional information like @kgodey suggested.

dmos62:
@pavish 
> If we need to provide suggestions from stored rows for other filters like uri_scheme_equals, we will require additional information like @kgodey suggested.

True. For `uri_scheme_equals` you'd need a single-column unique-only query that would also extract the scheme.

We could do something like `/api/db/v0/tables/1/columns/2/?unique_only=true&db_function={extract_uri_scheme: [{column_reference: [2]}]}` (i.e. use a DBFunction to process the column values). How to construct the `db_function` part could be suggested on the `uri_scheme_equals` filter, probably just adding a `first_parameter_preprocessor=extract_uri_scheme` attribute.

Does this sound reasonable?

At the same time, maybe there's a simpler approach to solving this problem? Maybe what @pavish suggested:

>>  I don't think showing the entire list of possible schemes will provide a good UX to users.
>
>We can show an auto-complete box instead of a select menu. There could be a max of 10 values displayed at a time and it will change as the user types.

This doesn't scale from backend's perspective, but seems easier.

dmos62:
Note, that this PR includes some under-the-hood work that subsequent PRs depend on, so I'll want to merge this either way. Question is whether to revert the changes to the filters endpoint.

kgodey:
> We could do something like `/api/db/v0/tables/1/columns/2/?unique_only=true&db_function={extract_uri_scheme: [{column_reference: [2]}]}` (i.e. use a DBFunction to process the column values). How to construct the `db_function` part could be suggested on the `uri_scheme_equals` filter, probably just adding a `first_parameter_preprocessor=extract_uri_scheme` attribute.

@dmos62 This seems more complicated from an API user perspective than my original proposal, why do this instead?

> We can show an auto-complete box instead of a select menu. There could be a max of 10 values displayed at a time and it will change as the user types.

I think the select menu will be a better experience; we could narrow down the options as the user is typing. I think users will be confused if the first 10 values are things they've never heard of and aren't in their data.

dmos62:
@kgodey 

> > We could do something like `/api/db/v0/tables/1/columns/2/?unique_only=true&db_function={extract_uri_scheme: [{column_reference: [2]}]}` (i.e. use a DBFunction to process the column values). How to construct the `db_function` part could be suggested on the `uri_scheme_equals` filter, probably just adding a `first_parameter_preprocessor=extract_uri_scheme` attribute.
> 
> @dmos62 This seems more complicated from an API user perspective than my original proposal, why do this instead?

Some points:

1. Exposing the frontend to a more expressive mechanism _feels_ better in this case (to me);
2. Whichever interface we provide, under-the-hood it should look similarly to what I described; I see your proposal as a layer on top of that: a specialized interface;
3. Your proposal would save dev-time for the frontend, but just a little; it would cost some dev time for the backend; we don't get a net gain in dev time as far as I can tell;
4. Providing unique column values preprocessed with various functions at the same time is a lot of eager calculation: will that scale?
5. I feel like your proposal is optimizing third-party API clients' dev experience: I say let's do that later;
6. I'd debate the benefit to API clients: personally I'd prefer general, expressive interfaces to highly tailored/specialized ones.

kgodey:
@dmos62 You make good points, I'm convinced (especially by points 3-5). 

I don't think there's any need to revert your existing work, I see no harm in providing a list of suggested values as well, even if the frontend doesn't use it.

dmos62:
@kgodey I'd prefer removing the unused logic so as to save on potential maintenance. If we ever need this, we can revert the removal (perks of git).

kgodey:
@dmos62 Sounds good.

dmos62:
Ready to be merged. Updated first post to say that there are no changes relevant to frontend.

kgodey (approved):
Looks fine to me, please merge after #1090.

Issue #1099, opened 2022-02-25
Refactor Select and SimpleSelect components, merging them into one


seancolsen:
## Current

- We have `Select` and `SimpleSelect` components which appear identical to the user but have different props structure.

    Why? Because `Select` came first, and then we decided on a different approach to structuring the props in order to specify the options.

## Desired

- We have only one component named `Select`.

- The props of `Select` behave exactly as the props of `SimpleSelect`.

- All code locations that have been using `Select` have been refactored to use its new props structure.

- `SimpleSelect` is deleted.

- All code locations that have been using `SimpleSelect` have been changed to use `Select` instead (and refactoring should not be necessary here, beyond a simple name change).

- `Select.d.ts` is renamed to `SelectUtils.ts`

- Inside `SelectUtils.ts`, we no longer have `SelectOption` or `SelectOptions` because they're not necessary anymore.

pavish:
I'm picking this up because this refactoring makes things easier for #257

PR #1047, opened 2022-02-03
Mathesar 896 Change records api parameters to use column id instead of name

silentninja:
Fixes #896 

Change `/records/` api query parameters(`order_by`, `group_count_by`, and `filters`) to use column id instead of the column name.

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1109, opened 2022-03-01
Refactor front end `Meta` class and a lot of surrounding code.

seancolsen:
## Context

This is laying groundwork for moving the front end to use column ids instead of column names, as required by #1047

## Goals

- Loosen the coupling between: (A): the functionality which saves tab data to the URL; and (B): the underlying data structures used within the tabs system and the tabular data. Previously, we were passing around `MetaParams` arrays like this 

    ```ts
    export type MetaParams = [
      number[], // [pageSize, page]
      string[], // [sortcolumn, sortorder:'a'|'d', sc, so ...]
      string[], // [groupcolumn, gc, gc]
      string[], // [filtercombination:'a'|'o', filtercolumn, condition, value ...]
    ];
    ```

    which led to code like [this](https://github.com/centerofci/mathesar/blob/8591a147f8af19004e1ab3431f84ae7044f06c6f/mathesar_ui/src/stores/tabs/tabList.ts#L72-L72), which is very hard to read

    ```ts
    function getTabsFromConfig(
      schemaId: SchemaEntry['id'],
      tabListConfig?: TabListConfig,
    ): MathesarTab[] {
      const tabs: MathesarTab[] = [];
      const tableStoreData = get(getTablesStoreForSchema(schemaId));
      tabListConfig?.tabularDataParamList?.forEach((entry) => {
        const table = tableStoreData.data.get(entry[1]);
        if (table) {
          tabs.push({
            id: calculateTabularTabId(entry[0], entry[1]),
            label: table.name,
            isNew: false,
            tabularData: getTabularContent(entry[0], entry[1], entry),
          });
        }
      });
      return tabs;
    }
    ```

    (That `getTabsFromConfig` function doesn't really have an analog after my changes -- it's been exploded into many different functions.)

- Splinter off some of the logic within `Meta` into separate classes for `Pagination`, `Filtering`, `Sorting`, and `Grouping`.
- Move `TabularType` out of `d.ts` file. Jest would not run my `tabDataSaver.test.ts` file the `TabularType` value being in `d.ts` file.
- Provide strong typing for the records API get request params.
- Add `SimpleSelect` component which provides cleaner DX than the `Select` component. In #1099 we'll merge the two components into one component called `Select`.
- Remove unused `type` and `parentId` properties from `Display` and `Meta` classes.
- Add base64 encoding to the serialized JSON tab data so that it doesn't require any URL encoding.

## Caveats

- Filtering UI is commented out because that will all be changing soon and is already broken in `master`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/App.svelte
```diff
@@ -6,6 +6,15 @@
...
   import { confirmationController } from '@mathesar/stores/confirmation';
+  import { currentSchemaId } from '@mathesar/stores/schemas';
+  import { beginUpdatingUrlWhenSchemaChanges } from './utils/routing';
+
+  // This is a bit of a hack to deal with our routing still being a patchwork of
+  // declarative and imperative logic. Without this call, the URL will not
+  // reliably set the query params when the schema changes. It actually _will_
+  // set the query params _sometimes_, but we weren't able to figure out why the
+  // behavior is inconsistent.
+  beginUpdatingUrlWhenSchemaChanges(currentSchemaId);
```
seancolsen, line 17: @pavish This is how I'm handling that finicky routing issue we spent some time discussing yesterday.

I played with it for a while after our call. There's definitely something fishing going on which _inconsistently_ causes the stores within `TabList` to update after tinro has performed client-side navigation. I tried to trace it down using the in-browser debugging tools but just ended up in a rat's nest of compiled Svelte code.

So this call to `beginUpdatingUrlWhenSchemaChanges` is my hacky work-around. I had to put it within a pretty top-level location in the app in order to avoid cyclical dependency linting errors.

I'm interested in eventually moving Mathesar to SvelteKit (once it's stable), in which case we'll need do re-think all the routing anyway. So I'm okay with some hack in the mean time.

pavish: > I'm interested in eventually moving Mathesar to SvelteKit (once it's stable), in which case we'll need do re-think all the routing anyway. So I'm okay with some hack in the mean time.

I estimate this mean time to be _atleast_ a couple years. Even then I wouldn't want to switch to SvelteKit until it has an official python adapter. Currently, if we need to switch, we'll need a node js middle layer, which is just pointless for us. There is a SvelteKit static adapter but that'll essentially the same (if not less perfomant) than directly using Svelte with pre-rendered data as we do now. So, I doubt we'd ever move to SvelteKit.

I'll take a look at this issue. We need to find the cause and fix it.

pavish: It seems to be an issue with Svelte, or maybe it was intended but I don't see the behaviour mentioned in the docs. I opened https://github.com/sveltejs/svelte/issues/7330 for this.

mathesar_ui/src/component-library/common/utils/typeUtils.ts
```diff
@@ -0,0 +1,7 @@
+function hasLabelProperty(v: unknown): v is { label: unknown } {
+  return typeof v === 'object' && v !== null && 'label' in v;
```
pavish, line 2: We could probably go another step and make the 'label' key configurable. This will retain existing Select functionality as well as retain type safety. We can think about it in the upcoming Select improvements PR.

mathesar_ui/src/component-library/pagination/Pagination.svelte
```diff
@@ -29,6 +29,12 @@
     undefined;
 
   // Total number of pages.
+  //
+  // TODO: @seancolsen says:
+  // > Refactor `pageCount` to no longer be an exported prop. We're exporting it
+  // > just so the parent component can access the calculation done within this
+  // > component. That's an unconventional flow of data. I'd rather do the
```
pavish, line 36: It's unconventional for other frameworks but the whole point of the `bind` keyword is to do exactly this. There are even dedicated readonly `bind` properties that Svelte provides. I'd say this is one of the intended ways of doing things in Svelte.

mathesar_ui/src/utils/storeUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+        results = arrayWithValueSetAtIndex(results, index, value);
+        set(results);
+      }),
+    );
+    // This is memory safe because when the last subscriber unsubscribes from
+    // the `unite` store, the function below will ensure that we're
+    // unsubscribing from all the inner stores.
+    return () => unsubscribers.forEach((unsubscriber) => unsubscriber());
+  });
+}
```
pavish, line 41: I love these utility functions!

pavish (approved):
@seancolsen Nice work! I didn't test it thoroughly but I didn't see any issues.

mathesar_ui/src/component-library/pagination/Pagination.svelte
seancolsen: > the whole point of the bind keyword is to do exactly this

I disagree. I would say the point of `bind` is to allow data to flow from child to parent. `bind:value` on an `input` element allows data that _originates_ in the `input` element (from user input) to flow up to the parent. That's great.

This `Pagination` component is different. The data for `pageCount` originates in the parent component as `total` and `pageSize`. The `pageCount` value is simply a derivation of those two values, which `Pagination` never modifies. Thus the source-of-truth data here is being passed _down_. Then a _derivation_ of that data is being passed back _up_. That's weird an unconventional. We have lots of `bind` usage in our codebase which is fine because the source of truth for the data actually originates in the child. But this is the only one I've seen which, I would say, abuses the `bind` pattern. Grappling with this weirdness costed me time during this refactor, which is why I suggested we clean it up.

mathesar_ui/src/App.svelte
seancolsen: I'll add this to my list of things to discuss during our next sync call. I totally understand the need to manipulate the query params imperatively. I still think it's worth wrapping up this topic a bit with some more discussion. For example, I'd like to understand from you (if you're able to explain it) why the routing behavior is _inconsistent_. I understand the upstream issue with Svelte. But I'd still like to better understand the cause of the _inconsistency_.

mathesar_ui/src/component-library/pagination/Pagination.svelte
pavish: I see your point regarding `bind` keyword.

For the `pageCount` property however, I strongly feel that it should be the child component's property and not be passed down from the parent.

Even though `pageSize` and `total` are passed down, `pageCount` is still a property which is calculated within the child component inorder for it to work as expected. The source of truth for `pageCount` is where the calculation happens and not where it's dependencies are passed from.

We can remove the `bind` keyword here and dispatch an event when `pageCount` changes and use that to update any property on the parent.

mathesar_ui/src/App.svelte
seancolsen: > There is a SvelteKit static adapter but that'll essentially the same (if not less perfomant) than directly using Svelte with pre-rendered data as we do now.

During our sync call, I'm also curious to explore this point above more, because that definitely does not match not my understanding.

pavish: > I'd like to understand from you (if you're able to explain it) why the routing behavior is inconsistent

The routing behaviour is not inconsistent, it works exactly as it is intended. The router is not at fault here. The upstream issue from svelte caused our url update method to get called which modified the query params.

Can you explain what exactly you find inconsistent?

mathesar_ui/src/component-library/pagination/Pagination.svelte
seancolsen: > We can remove the bind keyword here and dispatch an event when pageCount changes and use that to update any property on the parent.

That would be problematic in the same way.

If I were designing this from scratch, I'd make a `PaginationStats` object like this:

```ts
class PaginationStats {
  readonly recordCount: number;
  readonly pageSize: number;
  readonly pageCount: number;

  constructor({
    recordCount,
    pageSize
  }: {
    recordCount: number,
    pageSize: number,
  }) {
    this.recordCount = recordCount;
    this.pageSize = pageSize;
    this.pageCount = Math.ceil(this.recordCount / this.pageSize);
  }
}
```

Then I'd make a `stats` prop on `Pagination` which accepts an instance of `PaginationStats`.

In our code currently, the `StatusPane` needs access to the `pageCount`. The _only_ way it can get the `pageCount` is to render a `Pagination` component. That's deeply wierd to me. What if we want to take `Pagination` out of `StatusPane` but keep `pageCount` in `StatusPane`? Since `StatusPane` needs access to `pageCount`, `pageCount` should not be coupled to `Pagination`. The design above decouples `pageCount` from `Pagination`.

mathesar_ui/src/App.svelte
pavish: It has to do with `TabContainer` being created and destroyed, since it is wrapped within an `#if` block.

Assume you opened a table in `Schema 1`, which creates the `TabContainer` and `TabList` instance, which inturn updates the url. Then you open `Schema 2`, which contains no tables, so `TabContainer` gets destroyed. Note, we create a new `TabList` instance when a schema is opened for the first time, so the url update will happen when `Schema 2` is opened.

When we switch back to `Schema 1`, the `TabContainer` gets created again, which updates the url. If you then open a table in `Schema 2`, switching between schemas do not update the url.

I assume in your local environment, you mostly tested by opening tables in both schemas, so you noticed the url update happening ~5% of the time.

seancolsen: Interesting. Ok it sounds like there's a good chance I was not being observant enough of these small details while testing, and likely the behavior was consistent but my test scenarios were inconsistent. Thanks for explaining!

mathesar_ui/src/component-library/pagination/Pagination.svelte
pavish: > That would be problematic in the same way.

Not necessarily. `pageCount` would be decoupled since the parent's pageCount and the child's pageCount are different variables. 

> What if we want to take Pagination out of StatusPane but keep pageCount in StatusPane

We'd pass down `pageCount` to` StatusPane`.

> The only way it can get the pageCount is to render a Pagination component.

It's the only `current` way to do it. If we absolutely need to calculate pageCount on the parent, we'd best expose a util method in the module part of the Pagination component. That way any file can do:
`import { calculatePageCount } from 'Pagination.svelte'`

> If I were designing this from scratch, I'd make a PaginationStats object like this
> Then I'd make a stats prop on Pagination which accepts an instance of PaginationStats.

This is a design pattern I'd go with only when it is absolutely essential.

It makes the Pagination component reliant on the PaginationStats object, which means that anyone wanting to use the Pagination component needs to create an instance of the PaginationStats class. Reactive updates of `page` and `pageSize` would always need creating a new object, leading to boilerplate code wherever we use this. It's one of the annoying patterns/aspects of React that made me like Svelte.

We can solve this simply by exporting some util methods, as I mentioned above.

seancolsen: The utils pattern you suggest above is great too. It's certainly less code than my example. The calculation function would get called more times than in my proposal, but that's fine since it's not a perf concern here. I'll update my code comment to point to this conversation, and recommend the utils function approach.

PR #1141, opened 2022-03-04
UI changes for column ids in records API

seancolsen:
These are the front end changes for #1047.

I put them in a separate PR to aid review. I figure we can merge this first (into `#1047`), then merge `#1047`.

## Issues

- Grouping is broken because the response from the records API is still using column names within the `grouping` properties.

    Here's the GET response with grouping.

    ```json5
    {
      count: 10,
      grouping: {
        columns: ["country"], // <-- column name here
        mode: "distinct",
        num_groups: null,
        ranged: false,
        groups: [
          {
            count: 1,
            first_value: { country: "Bhutan" }, // <-- column name here
            last_value: { country: "Bhutan" }, // <-- column name here
            result_indices: [0],
          },
          // ...
        ],
      },
      results: [
        //...
      ],
    }
    ```

    I'm not sure how we want to handle this. I _thought_ I tested this when I reviewed `#1047` initially, but it may have slipped through the cracks. Or it may be a regression from changes pushed to that PR more recently. @silentninja any thoughts?

    I'd lean towards getting all of this work merged into `master` (thereby breaking the grouping feature temporarily) and fixing it with subsequent issues/PRs. My understanding is that some of the grouping API stuff is about to change anyway.

    We don't have any E2E tests for grouping yet which is why you don't see test failures.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/display.ts
+++ b/mathesar_ui/src/stores/table-data/display.ts
@@ -32,12 +32,21 @@ const movementKeys = new Set([
 ]);
 
 /**
- * I don't understand what this is but I had to create it during a refactor.
- * This value is used as a key in a ColumnPositionMap where column ids are also
- * used, so should have the same type as a column id but needs to have a value
- * that no column id will ever have. That's why we're using -1. If you
- * understand this design and can suggest a better name for this variable, then
- * please do so.
+ * This value is used as a key in a `ColumnPositionMap` where each entry
+ * corresponds to the position of the column, as indexed by the column id.
+ * However, the entry indexed by `ROW_POSITION_INDEX` consists of the total
+ * width & left values (i.e the position) of the row. It's placed within
+ * `ColumnPositionMap` in order to avoid calculation within the component, which
+ * will run for each row. Thus, `ROW_POSITION_INDEX` should have the same type
+ * as a column id but needs to have a value that no column id will ever have.
+ * That's why we're using -1.
+ *
+ * We could use a dedicated store for it or even a new class containing both
+ * columnPosition and row width.
+ *
+ * Pavish put it within ColumnPositionMap because we were passing around a lot
+ * of props to the child components and he wanted to reduce the number of props.
+ * Now, it is all passed down using context, so that's no longer an issue.
  */
 export const ROW_POSITION_INDEX = -1;
 
--- a/mathesar_ui/src/stores/table-data/display.ts
+++ b/mathesar_ui/src/stores/table-data/display.ts
@@ -9,7 +9,8 @@ export interface ColumnPosition {
   width: number;
   left: number;
 }
-export type ColumnPositionMap = Map<string, ColumnPosition>;
+/** keys are column ids */
+export type ColumnPositionMap = Map<number, ColumnPosition>;
 
 // TODO: Select active cell using primary key instead of index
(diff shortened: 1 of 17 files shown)


mathesar_ui/src/stores/table-data/display.ts
```diff
@@ -30,23 +31,33 @@ const movementKeys = new Set([
...
+/**
+ * I don't understand what this is but I had to create it during a refactor.
+ * This value is used as a key in a ColumnPositionMap where column ids are also
+ * used, so should have the same type as a column id but needs to have a value
+ * that no column id will ever have. That's why we're using -1. If you
+ * understand this design and can suggest a better name for this variable, then
+ * please do so.
+ */
+export const ROW_POSITION_INDEX = -1;
+
```
seancolsen, line 43: @pavish Would you like to lend a hand at clarifying this design? Maybe you could push a commit that cleans this up a bit by setting a better name for this constant and re-writing the comments above it? I also considered allowing strings as keys in the `ColumnPositionMap` so that we could keep the previous value of `__row`, but `-1` seemed a bit cleaner to me.

pavish: Each entry within the columnPositionMap corresponds to the position of the column. However, the `_row` value consists of the total width & left values (i.e the position) of the row. It's placed within columnPositionMap inorder to avoid calculation within the component, which will run for each row.

We could use a dedicated store for it or even a new class containing both columnPosition and row width.

It was placed within `ColumnPositionMap` because we were passing around a lot of props to the child components and I wanted to reduce the number of props. Now, it is all passed down using context, so that's no longer an issue.

mathemancer:
> * I'd lean towards getting all of this work merged into `master` (thereby breaking the grouping feature temporarily) and fixing it with subsequent issues/PRs. My understanding is that some of the grouping API stuff is about to change anyway.

I would support that.  I'm currently blocked on grouping until the column ID stuff is merged, and it might be more efficient for me to change it as part of the next grouping PR instead.

kgodey:
I agree re: getting it merged sooner and breaking grouping temporarily.

mathesar_ui/src/stores/table-data/display.ts
seancolsen: Thanks for the explanation, @pavish. Are my changes in this PR with respect to `ColumnPositionMap` okay for now? Or would you like further adjustment? If you'd like to see more clean up here, can you either add a commit or let me know how you'd like me to address it?

pavish: @seancolsen I think the change is fine. I wanted to specify it since you weren't sure about the context. You could just update your comment in this PR.

silentninja:
@seancolsen I agree on merging along with breaking changes. I created a discussion to decide on [Which column identifier should we use for DML operation](https://github.com/centerofci/mathesar/discussions/1147). There are a few conflicting thoughts and there could be a possibility of using `column name`

pavish (approved):
Looks good to me, I didn't test it thoroughly and I didn't see any breaking functionality so far (apart from grouping, which is mentioned in the PR description).

PR #1148, opened 2022-03-07
API for getting all valid parameters for an equality filter

### LINKED ISSUE #1097: Suggest popular/standard URI schemes as values for appropriate filter parameters

dmos62:
Fixes #1097 

The added functionality is not easy to summarize. In essence, this allows the frontend to look up and suggest parameter values to the user when constructing an equality-based comparison.

### Technical details

For example, given a column that has URIs like `x://a/b/c` and `y://e/f/g`, when constructing a filter with `URISchemeEquals`, the new functionality will tell you that `URISchemeEquals`'s first parameter is preprocessed with `ExtractURIScheme`, and the new functionality will also let you apply a function of your choice, `ExtractURIScheme` in this case, to that column, which will return `x` and `y`, and those will be the only values for the second parameter that will test positive. A user can thus receive exhaustive suggestions on what second parameter to use with that filter.

The `/api/ui/v0/databases/1/filters/` endpoint is extended to specify which filters are `equality_based`, and when an equality-based filter's parameter is processed with (`processed_with`) a DBFunction, that's also included:

```
  {
    "id": "email_domain_equals",
    "name": "email domain is",
    "aliases": [],
    "parameters": [
      {
        "ui_types": [
          "email"
        ],
        "processed_with": "extract_email_domain"
      },
      {
        "ui_types": [
          "email",
          "text",
          "uri"
        ]
      }
    ],
    "equality_based": true
  },

```

The `/api/db/v0/table/1/records/` endpoint is extended to accept `db_function` and `deduplicate` query parameters, which allow you to apply a DBFunction to a table and deduplicate its result.

### Example scenario

There's a column that contains emails with id `X`. After querying `/api/ui/v0/databases/1/filters/`, frontend knows to let the user apply email column `X` to the first parameter of the above copy-pasted `email_domain_equals` filter.

The frontend sees that `email_domain_equals` attribute `equality-based` is `true`, and thus knows that it can make suggestions about what values to use for the second parameter, because only values that are in the column referenced by the first parameter (or are derived from it) make sense as the second parameter.

The frontend sees that the first parameter has the attribute `processed_with` set to `extract_email_domain`, which means that to get the values in the column `X` processed in the same way that `email_domain_equals` processes it (for its first parameter) it has to apply `extract_email_domain` to it. It does that by querying `/api/db/v0/table/1/records/?db_function={extract_email_domain: [{column_id: X}]}&deduplicate=true` and gets something like:

```
{
  "count": 4,
  "grouping": null,
  "results": [
    {
      "mathesar_types.email_domain_name_1": "orange.fr"
    },
    {
      "mathesar_types.email_domain_name_1": "bluewin.ch"
    },
    {
      "mathesar_types.email_domain_name_1": "windstream.net"
    },
    {
      "mathesar_types.email_domain_name_1": "frontiernet.net"
    }
  ]
}
```

You'll notice that `results` sub-dict keys are not Django IDs, but a name dynamically given to the temporary column by Postgres. That's because the results here are a column computed by running the table through the specified `db_function`, which is temporary, and thus does not (cannot) have a Django ID.

The values in the `results` sub-dict key-value pairs are the deduplicated email domains present in the column `X` and can be forwarded to the user as choices for the second parameter for the filter in question.

### Some noteworthy points for the frontend team

- A filter, as returned by the `filters` endpoint, now has the `equality_based` boolean attribute: it tells you that the filter is based on an equality check between its parameters;
- in the case where one of the parameters is additionally preprocessed, that parameter has the `processed_with` attribute, whose value is a DB function id:
  - it may be used to process a column with the `db_function` query param on the `records` endpoint;
  - the `processed_with` DB function will always take only one parameter;
  - a DBFunction is formulated in JSON in the same way as a filter (actually, a filter is secretely a DB function);
  - see *Example scenario* section above for why you'd use this;
- the `records` endpoint has two new query parameters:
  - `db_function` specifies a DB function to apply to the table: see the copy-paste above for what the output looks like;
    - its format is identical to that of `filter`;
  - `deduplicate` takes a boolean and specifies whether to deduplicate the rows returned;
    - for example, if the above example didn't use `deduplicate` it would have returned more rows that would have been duplicates of some of the rows that were returned;
- the suggested algorithm for suggesting exact values for the filter's second parameter is:
  - when filter's `equality_based` attribute is `true`,
    - if the other parameter has `processed_with` attribute
      - do `records/?db_function={extract_email_domain: [{column_id: X}]}&deduplicate=true`;
        - replacing `extract_email_domain` with the value of the `processed_with` attribute;
    - else
      - do `records/?db_function={identity: [{column_id: X}]}&deduplicate=true` (i.e. use the `identity` db function, which just returns the column referenced);
    - in either case, replacing `X` with the column referenced by the filter's first parameter;
    - disregard the column name in the returned result list;
    - the returned values are the values that may be suggested for the filter's second parameter.

### Notes about architectural decisions

There were a few discussions about how to implement the application of a DB function to a table. I intimately explored more than one option and finally went with piggy-backing the `records` endpoint, because it already has most of the infrastructure needed for this query, or in other words this query (application of a function) can be formulated as an additional step in the more general "get records" query. This meant less duplicated logic.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
There's currently a bug where application of a function returns a superfluous row with the value `None`. I'm working on it. It shouldn't affect the rest of the PR.

dmos62:
@mathemancer I've a weird bug that's leaving me stumped. I do `SELECT DISTINCT` on a column and the `DISTINCT` part seems to work, but it also adds an additional row that is just `(None,)`. My debug queries look like this:

with `DISTINCT`:
```
ipdb> pp str(query)
('SELECT DISTINCT "Application SN" \n'
 'FROM (SELECT "Patents"."NASA Record List Filter Duplicates".id AS id, '
 '"Patents"."NASA Record List Filter Duplicates"."Center" AS "Center", '
 '"Patents"."NASA Record List Filter Duplicates"."Status" AS "Status", '
 '"Patents"."NASA Record List Filter Duplicates"."Case Number" AS "Case '
 'Number", "Patents"."NASA Record List Filter Duplicates"."Patent Number" AS '
 '"Patent Number", "Patents"."NASA Record List Filter Duplicates"."Application '
 'SN" AS "Application SN", "Patents"."NASA Record List Filter '
 'Duplicates"."Title" AS "Title", "Patents"."NASA Record List Filter '
 'Duplicates"."Patent Expiration Date" AS "Patent Expiration Date" \n'
 'FROM "Patents"."NASA Record List Filter Duplicates" ORDER BY "Patents"."NASA '
 'Record List Filter Duplicates".id ASC) AS anon_1 \n'
 ' LIMIT %(param_1)s')
```

notice the `(None,)` at the top:
```
ipdb> pp results
[(None,),
 ('13/053,633',),
 ('12/564084',),
 ('11/689,431',),
 ('12/900644',),
 ('14/021812',),
 ('12/39,506',),
 ('11/924,766',),
 ('11/537,280',),
 ('09/906,014',),
 ('10/943,649',),
 ('13/534,951',),
 ('12/560,535',),
 ('12/875254',),
 ('13/291,372',),
 ('08/414,661',),
 ('12/839171',),
 ('12/536,969',),
 ('10/302,323',),
 ('13/213,022',),
 ('13/191,882',),
 ('14/095,442',),
 ('08/917,581',),
 ('12/893,627',),
 ('13/600992',),
 ('13/600826',),
 ('13/615,850',),
 ('12/857,004',),
 ('13/673,360',),
 ('11/76,824',),
 ('10/73,957',),
 ('11/421,196',),
 ('09/667,426',),
 ('10/693,850',),
 ('13/463,780',),
 ('11/066649',),
 ('10/320,698',),
 ('12/849,906',),
 ('12/248,339',),
 ('12/533,520',),
 ('10/845,418',),
 ('13/326,779',),
 ('09/607,211',),
 ('13/951671',),
 ('09/560,532',),
 ('13/771,920',),
 ('12/390,606',),
 ('11/148,778',),
 ('12/870,475',),
 ('12/721,833',)]
```

without distinct:
```
ipdb> pp str(query)
('SELECT "Application SN" \n'
 'FROM (SELECT "Patents"."NASA Record List Filter Duplicates".id AS id, '
 '"Patents"."NASA Record List Filter Duplicates"."Center" AS "Center", '
 '"Patents"."NASA Record List Filter Duplicates"."Status" AS "Status", '
 '"Patents"."NASA Record List Filter Duplicates"."Case Number" AS "Case '
 'Number", "Patents"."NASA Record List Filter Duplicates"."Patent Number" AS '
 '"Patent Number", "Patents"."NASA Record List Filter Duplicates"."Application '
 'SN" AS "Application SN", "Patents"."NASA Record List Filter '
 'Duplicates"."Title" AS "Title", "Patents"."NASA Record List Filter '
 'Duplicates"."Patent Expiration Date" AS "Patent Expiration Date" \n'
 'FROM "Patents"."NASA Record List Filter Duplicates" ORDER BY "Patents"."NASA '
 'Record List Filter Duplicates".id ASC) AS anon_1 \n'
 ' LIMIT %(param_1)s')
```
```
ipdb> pp results
[('13/033,085',),
 ('08/543,093',),
 ('09/017,519',),
 ('10/874,003',),
 ('09/652,299',),
 ('09/226,673',),
 ('09/374,491',),
 ('10/637,087',),
 ('09/498,123',),
 ('09/606,107',),
 ('09/800,309',),
 ('09/800,311',),
 ('09/800,313',),
 ('09/800,310',),
 ('09/822470',),
 ('11/045,041',),
 ('12/807,375',),
 ('11/274,744',),
 ('10/099,247',),
 ('10/956,517',),
 ('10/914,783',),
 ('12/694,966',),
 ('10/320,698',),
 ('10/828,524',),
 ('11/387,503',),
 ('10/232,975',),
 ('10/885,533',),
 ('10/706,478',),
 ('10/135,013',),
 ('10/758611',),
 ('12/45,970',),
 ('10/494,853',),
 ('11/194,991',),
 ('10/238,515',),
 ('10/198,672',),
 ('10/928,874',),
 ('12/211,439',),
 ('10/789,049',),
 ('10/457,696',),
 ('10/923,156',),
 ('10/816,576',),
 ('10/789,029',),
 ('10/703,039',),
 ('10/608,884',),
 ('10/808,704',),
 ('10/923,160',),
 ('11/239,456',),
 ('10/825,795',),
 ('11/900,131',),
 ('10/779,504',)]
```

Why is the use of `DISTINCT` seemingly causing the appearance of a seemingly nonsensical `(None,)` row?

Also, note that this doesn't happen in all settings. I've a test that's failing because of this and another test that works fine. The failing test is `mathesar/tests/api/test_record_api.py::test_record_db_function_and_deduplicate`.

pavish:
@dmos62 The structure sounds good to me. I have a couple comments/questions:

* > You'll notice that results sub-dict keys are not Django IDs, but a name dynamically given to the temporary column by Postgres

   Can we have another step in the backend where the dynamic names are replaced by a static key? If this was a ui endpoint, it's fine for the frontend to ignore the response, but this is a db endpoint which is not specific to the frontend. Having dynamic keys here makes it harder to consume the API. It might even be okay if the response specifies the dynamic name as a separate property.

* This is not related to this particular PR, it's more of a general filter question which I got when reading this PR's description.
  ```
  "parameters": [
    {
      "ui_types": [
        "email"
      ],
      "processed_with": "extract_email_domain"
    },
    {
      "ui_types": [
        "email",
        "text",
        "uri"
      ]
    }
  ]
  ```
  Here, the second parameter accepts email, text and uri. This is a field entered by the user and we would only be showing one input box here. We need to be able to identify what kind of input we should show to the user. Also, can we get conflicting types of input, like:
  ```
  "parameters": [
    {
      "ui_types": [
        "email"
      ],
      "processed_with": "some_fn"
      ...
    },
    {
      "ui_types": [
        "boolean",
        "number",
        "
      ]
    }
  ]
  ```

  What is the best match logic we should be following to identify the input to show to the user? For the previous config, it has to be 'text' which is the superset of the three. For the second, it could be one among 'boolean' or 'number'.

dmos62:
>> You'll notice that results sub-dict keys are not Django IDs, but a name dynamically given to the temporary column by Postgres
>
>Can we have another step in the backend where the dynamic names are replaced by a static key? If this was a ui endpoint, it's fine for the frontend to ignore the response, but this is a db endpoint which is not specific to the frontend. Having dynamic keys here makes it harder to consume the API. It might even be okay if the response specifies the dynamic name as a separate property.

We still don't have a policy for how to deal with dynamic identifiers. I think that using string names for dynamic identifiers and integers for persisted objects might be robust enough.

I can check if the column is not a known persisted column and replace its name with a static string (like `DYNAMIC`), but I don't see a workflow where this might improve things.

Maybe let's see where the current discussions about dynamic ids go and refactor based on that in a separate PR?

> What is the best match logic we should be following to identify the input to show to the user? For the previous config, it has to be 'text' which is the superset of the three. For the second, it could be one among 'boolean' or 'number'.

That's a good question. For some context, currently, there's no `OR` logic in the hint system, so if you have `"ui_types": ["email","text","uri"]`, that's because all these types have the attribute/s deemed necessary for this parameter (they are all string-like, in this case).

What a frontend could do is make some sort of ordering of UI types based on how "general" they are, such that email and uri are lower than text, and then choose the type that is highest (text in this case).

For seemingly conflicting ui_types like boolean and number, follow that same ordering. Another way to phrase the ordering I'm talking about is to say that it should capture how "expressive" a type is. For example, a boolean can only express things like 0, 1, true, false, but a number can do 0, 1, 2.99, etc. (number is more general and hence would have a higher ordering).

This might not be full-proof. If we come upon cases where ambiguity is possible, we can provide new hints for that. Also, it would be great if in the UI you could switch between whether you're inputing just text, an email or a URI (override the default). That would solve any edge cases.

pavish:
@dmos62 

> Maybe let's see where the current discussions about dynamic ids go and refactor based on that in a separate PR?

Sounds good.

> What a frontend could do is make some sort of ordering of UI types based on how "general" they are, such that email and uri are lower than text, and then choose the type that is highest (text in this case).
> For seemingly conflicting ui_types like boolean and number, follow that same ordering.

I was thinking the same, but I felt it might be better if the backend provided that information since this is a ui specific endpoint. I could place it in the mocked type layer on the frontend for now, and we could get back to it when we try to implement the type-agnostic architecture in the future.

> Also, it would be great if in the UI you could switch between whether you're inputing just text, an email or a URI (override the default)

We would have to think of a UX which would allow that, this could probably be taken up after the initial release. @kgodey Would we need such a functionality for the query builder at the moment?

dmos62:
@mathemancer I found the reason for the [previously mentioned](https://github.com/centerofci/mathesar/pull/1148#issuecomment-1063217090) bug. The column had empty cells (they had the `NULL` value), which caused `SELECT DISTINCT` to return an empty cell too. In other words, this isn't a bug.

I expected when applying a boolean-returning SQL function to a table, and doing DISTINCT on that, that I'd get at most 2 distinct values (true and false), but now I see that it can also return `NULL`.

mathemancer:
> Why is the use of `DISTINCT` seemingly causing the appearance of a seemingly nonsensical `(None,)` row?

I'd need to see the data you're selecting from to be sure, but you're ordering by a different column that isn't displayed. I suspect the row with `None` just has a high enough `id` that it isn't shown in the first page of rows in the non-distinct select.

dmos62:
> > Why is the use of `DISTINCT` seemingly causing the appearance of a seemingly nonsensical `(None,)` row?
> 
> I'd need to see the data you're selecting from to be sure, but you're ordering by a different column that isn't displayed. I suspect the row with `None` just has a high enough `id` that it isn't shown in the first page of rows in the non-distinct select.

Apparently, I just didn't consider what happens when a column has a `NULL` value.

kgodey:
> I was thinking the same, but I felt it might be better if the backend provided that information since this is a ui specific endpoint. I could place it in the mocked type layer on the frontend for now, and we could get back to it when we try to implement the type-agnostic architecture in the future.

@dmos62 I agree with @pavish here, we should only be returning the UI types that are specced for that filter (e.g. email domain equals should only be returning a UI type for email). The UI API is meant to support our specific UI and reduce the need for frontend logic.

> Also, it would be great if in the UI you could switch between whether you're inputing just text, an email or a URI (override the default)

We do not currently have plans to support this (even in the query builder interface). If we do, we can then modify the API.

dmos62:
>> I was thinking the same, but I felt it might be better if the backend provided that information since this is a ui specific endpoint. I could place it in the mocked type layer on the frontend for now, and we could get back to it when we try to implement the type-agnostic architecture in the future.

> @dmos62 I agree with @pavish here, we should only be returning the UI types that are specced for that filter (e.g. email domain equals should only be returning a UI type for email). The UI API is meant to support our specific UI and reduce the need for frontend logic.

Here we're actually talking about the second parameter. The first parameter is declared by the `filters` endpoint to only support `email` UI type. The second parameter supports any type that is string-like, like text, but also email and URI. Effectively, we're saying that anything that is string-like is potentially an email domain.

What @pavish was saying is, given a choice of `email`, `URI` and `text`, and given that all of them require different input forms, how to choose one? As I see things, that's the right question to ask right now, and I'll try to explain.

I think what you're saying is, why would we compare an email domain to another email or a URI? Those are suggested in this case, because the filter's second parameter is only constrained to be string-like, which emails and URIs are.

One solution is to consider this a problem of precise definition and be thorough and use precise hints to specify what kinds of strings email domains are. That's probably not a good time investment at this point. It also wouldn't scale.

What might seem like a solution is to remove the less "general" suggested UI types like `email` and `uri` from the `[email, uri, text]` set above. But, that wouldn't work, since then the frontend couldn't use a specialized column (like `uri`) where a more general column (like `text`) is currently also accepted. For example, the second parameter of `starts_with` filter takes this same UI type set (`[email, uri, text]`), and you can check if the first parameter starts with an email or a uri or a general string, but if the second parameter's UI type set was just `[text]`, it could only take input that is strictly `text`, not `uri` and not `email`.

I think what @pavish suggested about keeping a generality or expressiveness ordering of UI types on an endpoint, that he will temporarily mock, is a good way to solve this.

kgodey (changes requested):
@dmos62 As-built, this feature breaks [our API standards](https://wiki.mathesar.org/en/engineering/architecture/api-standards), specifically the [Responses section](https://wiki.mathesar.org/en/engineering/architecture/api-standards#responses)

We have a REST API, which takes an object-oriented approach. That means that each endpoint represents a resource (which is a single type of object). The records list endpoint should always return list of records, regardless of what query parameters have been applied to it. In this PR, applying certain query parameters completely transform the response.

Although we are focusing on a feature-first approach, I don't want to go so far as to reverse some of the architectural consistencies that we've managed to maintain so far.

Please find an alternate way to solve this that is either RESTful or introduce an explicit RPC based API `functions/run/` as we discussed on Matrix.

kgodey:
> Here we're actually talking about the second parameter.

Sorry, I didn't catch that while reading the previous conversation. You can ignore my previous comment.

> I think what @pavish suggested about keeping a generality or expressiveness ordering of UI types on an endpoint, that he will temporarily mock, is a good way to solve this.

Yeah, that's fine with me.

dmos62:
> The records list endpoint should always return list of records, regardless of what query parameters have been applied to it.

@kgodey This has been a pain point that I wrestled with. Thing is that everything on the backend is set up to think of this endpoint (and its foundational infrastructure) not in terms of records (I'd like to not rehash the discussion on what that is precisely), but in terms of persistence-agnostic rows. On the backend we're essentially building SQL pipelines and SQL is not aware of the difference between rows and records.

I advocate, with determination, that we embrace that and rename the endpoint from `/db/tables/1/records` to the more general `/db/tables/1/rows`. Its definition is then broadened to whatever tabular output that can be arrived at by any sequence of transformations made available by this endpoint, the starting point being that table, and the ending point being anything tabular. Or, in other terms, the `rows` endpoint response would be the result of a pipeline, whose starting point is the table, intermediate steps are the transformations like `order_by` or `filter` or `db_function`, and the output is tabular.

For illustration of what I mean when I say that the backend is building pipelines, this is the workhorse behind the `records` endpoint:

```python
def get_query(
    table,
    limit=None,
    offset=None,
    order_by=None,
    filter=None,
    group_by=None,
    duplicate_only=None,
    db_function=None,
    deduplicate=False,
):
    if duplicate_only:
        select_target = _get_duplicate_only_cte(table, duplicate_only)
    else:
        select_target = table

    if isinstance(group_by, group.GroupBy):
        selectable = group.get_group_augmented_records_query(select_target, group_by)
    else:
        selectable = select(select_target)

    if order_by:
        selectable = apply_sort(selectable, order_by)

    if filter:
        db_function_instance = get_db_function_from_ma_function_spec(filter)
        selectable = apply_db_function_as_filter(selectable, db_function_instance)

    if db_function:
        db_function_instance = get_db_function_from_ma_function_spec(db_function)
        selectable = apply_db_function_as_function(selectable, db_function_instance)

    if deduplicate:
        selectable = selectable.distinct()

    if limit:
        selectable = selectable.limit(limit)

    if offset:
        selectable = selectable.offset(offset)

    return selectable
```

Notice its pipish nature. Here, `db_function` is just another step, same as `offset`, `deduplicate`, `group_by`, etc.

You mentioned that our API is based on objects: I don't think that my suggestion is in conflict with that. I'm saying let's take another step in that direction and say that our API is based on SQL objects: schemas, tables, rows and columns. But, let's not constrain the definition of those objects to only things that are currently persisted on the disk. I mentioned this in another thread today: I don't think that data transience should play a role in the definitions of these basic objects.

I also don't think that there's a resource/RPC conflict here. Let's consider a resource to be whatever a given pipeline outputs when applied to a given resource's path. Path, as in an address like `database -> schema -> table`. Let's consider all transformations as fundamentally equivalent, whatever concrete transformations they apply, wheter it's the redefinition of the order of rows (`order_by`) or the passing of the entire table through a function (e.g. `db_function`).

I think that this approach would scale great. For example, summarizing transformations, just like `db_function`, would not require a separate endpoint, or deviate into RPC territory.

Also, it's very flexible, since it's basically using a single pipeline builder, the `get_query` function above. Currently it's pretty primitive (e.g. fixed order of transformations), but who knows what it will grow into. This also mirrors the fact that any SQL query/view is a single pipeline.

To address the probable future comment about me prioritizing architecture over product, this is not it! My intent here is to manipulate architecture to reduce the workload on developers, make hard problems easier, make annoying problems less common, and to get the product out better and faster. I'll try to summarize the benefits briefly, since this has been a long post:

- a single endpoint for all table transformations is easier to maintain and implement than multiple endpoints: less duplicated logic and bloat on the service-layer;
- the semantics would closer mirror SQL and thus reduce cognitive load, and make leveraging SQL generally easier;
  - I think this would have noteworthy, positive knock on effects, especially when it comes to implementing the data explorer.

dmos62:
Closing. See https://github.com/centerofci/mathesar/discussions/1154.

PR #1155, opened 2022-03-10
Number component

### LINKED ISSUE #970: NumberInput component improvements
## Problem
The NumberInput component implemented as part of https://github.com/centerofci/mathesar/pull/915 is a low level component with few practical uses. The basic version was implemented inorder to support the form-builder component. Inorder to make it usable, it requires a number of improvements.

## Proposed solution
The following contents are taken from @seancolsen's comments on the [form-builder PR](https://github.com/centerofci/mathesar/pull/915#pullrequestreview-848217134)

### Fix `NumberInput` setting `value` to string

The two-way binding for `value` should reliably produce a type of `number | undefined`.

Case 1

1. Instantiate a `NumberInput` component with a bound `value` equal to `7`.
1. In the browser focus on the input, placing the cursor at the end.
1. Enter `8`.
1. Expect `value` to be `78` (number).
1. Observe `value` to be `"78"` (string).

Case 2

1. Instantiate a `NumberInput` component with a bound `value` equal to `7`.
1. In the browser focus on the input, placing the cursor at the end.
1. Press Backspace once, clearing the contents of the input.
1. Expect `value` to be `undefined`.
1. Observe `value` to be `""` (empty string).

Furthermore (and on a related note), I'm a bit confused about what's happening with `interface HTMLNumberInputElement`, and my hunch is that we should remove it.

My understanding is that the `value` property of an `<input>` will always be a string, even when setting `type='number'`. That means that writing `e.target as HTMLNumberInputElement` (in `NumberInput.svelte`) is incorrectly providing type assurance that `value` will be a number.

I would expect `NumberInput` to assume that the underlying DOM element will have a string value, and to be converting that string to a number (i.e. with `Number(inputElement.value)`) to produce a binding to the `value` prop that satisfies `number | undefined`.

### More gracefully handle some edge cases in `NumberInput`

Independent sub-issues, with steps to reproduce.

- Support pasting multi-character values

    1. Begin with a value of `1`
    1. Into your clipboard, copy `22`
    1. Focus on the input and select all its contents
    1. Paste
    1. Expect `22`
    1. Observe `1`

    Adding a `+` after the group in `validKeyRegex` appear to fix this.

- Support entering `-0.1`

    1. Begin with an empty input
    1. Enter `-0` and observe `-0` which is good
    1. Type `.`
    1. Expect `-0.`
    1. Observe `0`

- Allow empty integer

    1. Begin with `1` and `isInteger`
    1. Place cursor at end
    1. Press `Backspace` once
    1. Expect empty input
    1. Observe `0`

- Don't wipe contents when entering `7..`

    1. Begin with an empty input
    1. Enter `7.` and observe `7.` which is good
    1. Type `.`
    1. Expect `7.`
    1. Observe an empty input

- Retain cursor position

    1. Begin with `78`
    1. Move cursor between `7` and `8`
    1. Enter `a`
    1. Expect `78` with cursor between `7` and `8`
    1. Observe `78` with cursor after `8`

- Don't auto-delete decimal

    1. Begin with `1.2`
    1. Delete the `2`
    1. Expect `1.`
    1. Observe `1`

- Don't auto-delete minus sign

    1. Begin with `-8`
    1. Delete the `8`
    1. Expect `-`
    1. Observe an empty input

- Don't truncate integer value when entering decimal midway through number

    1. Begin with `23` and `isInteger`
    1. Place cursor between `2` and `3`
    1. Enter `.`
    1. Expect `23`
    1. Observe `2`

- Remove support for numbers like `8e9`, or allow users to enter them. I can enter `89`, then place my cursor between the two digits and type `e`. But I can't enter `8e9` directly. I don't think we need support entering these kinds of numbers into the input though.
 
Additional thoughts

- I'd also like to stress test this a bit more after the string type issue is fixed.
- It might be worth using the [`beforeinput`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event) event instead of the `input` event.

### Add tests for `NumberInput`

This is the kind of component that really benefits from having some tests in my opinion.

We could pretty easily write tests for the scenarios I've laid out above, as well as other ones that already work correctly.

### Remove `type='number'` on `NumberInput`

I hate the browser native number input, and I think most people do too.

- I don't like the tiny up/down buttons. They look bad and don't add much value.
- The value of the input changes in response to scroll events, which can be really annoying sometimes.

I'd prefer that we remove

Considerations:

- We'll need to set [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) to either `decimal` or `numeric`, potentially with some logic based on the props.
- We'll need to adjust some of the logic within the `onInput` function to account for the different behavior of `inputElement.value` when the user enters a non-numeric string like `x` into the input.

seancolsen:
## Summary

- This PR alters the behavior or our `NumberInput` component, making it much more robust.

- Example changes in behavior to `NumberInput`

    | | Before | After |
    | -- | -- | -- |
    | Value produced when user enters `e009..e` | `"e009..e"` (string) | `9` (number) |
    | Formatted display when user enters `e009..e` | `e009..e` | `9.` |
    | Formatted display of number `1234` | `1234` | `1,234` |

- Within Mathesar, `NumberInput` is currently used _only_ for the form that configures the DB settings on a text field. But we eventually want to it for editing cells, and potentially other places too.

## Additional context

- Many of the fundamental design decisions were solidified in https://github.com/centerofci/mathesar-wiki/pull/39 where we added [Number Display and Entry](https://wiki.mathesar.org/en/engineering/specs/numbers) specs to the wiki.

## Demo

Peek 2022-03-21 11-52

## Review tips

- Play with the `NumberInput` story. In particular, try fiddling with `allowFloat`, `allowNegative` and `locale` to see how the behavior of the input changes.
- Play with the input that sets the field size limit of a text column. (From what I can tell, that's the only place within Mathesar that's affected by this PR.)

## Requirements from #970

- [x] Fix NumberInput setting value to string
- [x] Support pasting multi-character values
- [x] Support entering `-0.1`
- [x] Allow empty integer
- [x] Don't wipe contents when entering `7..`
- [ ] Retain cursor position -- _to be addressed via new ticket_
- [x] Don't auto-delete decimal
- [x] Don't auto-delete minus sign
- [x] Don't truncate integer value when entering decimal midway through number
- [x] Remove support for numbers like `8e9`
- [x] More stress testing (I monkeyed around with it quite a bit)
- [x] use the beforeinput event
- [x] Add tests
- [x] Remove type='number' on NumberInput -- _(already done in #1144)_
- [x] Set inputmode to either decimal or numeric

## Additional work completed

- Make a generalized `FormattedInput` component that can be composed to build other kinds of specialized inputs.

- Allow input of numbers in other locales.

- Use `NumberInput` for `number` numbers and `StringifiedNumberInput` for high precision numbers (stored as `string`).

- Disallow user input into `NumberInput` that will result in loss of precision.

    You cay try this out by pressing `9` repeatedly on the input. You'll notice it stops accepting input after fifteen digits. In any situation where we're using `NumberInput`, hitting the precision limit should be a very unlikely scenario. That said, an alternate behavior (that might be nice to add at some point) would be to _round_ the input value to the nearest storable value (instead of preventing input). This would mean you could keep typing `9` and when you get to sixteen digits, the value of the input would change to `10000000000000000`. I _could_ open a ticket for this, but I don't really think it's worth it. If someone needs this later we could address it. It wouldn't be too hard to add a prop on `NumberInput` to configure this behavior.

- Set `allowFloat` and `allowNegative` to false by default. (I figure better to be strict by default. We can discuss if necessary.)

## To handle later via tickets

Here are the templates to create the tickets

<details>
    <summary><h3>Support scientific notation</h3></summary>

`cat | gh issue create -F - -t 'Add support for scientific notation in number input components' -m '[Beta] Better Editing Experience' -l "status: ready" -l "type: enhancement" -l "work: frontend"`

```
## Current behavior

- Entering numbers like `8E9` into a `NumberInput` (or `StringifiedNumberInput`) is not possible.

## Desired behavior

- `NumberInput` and `StringifiedNumberInput` accept a prop `allowScientificNotation` which defaults to `false`. When `true` the user can enter numbers in scientific notation and the component will display numbers in scientific notation when appropriate.

## Additional context

- See [Number display and entry](https://wiki.mathesar.org/en/engineering/specs/numbers) for more information about how scientific notation should work.

## Implementation

- Begin by adding `allowScientificNotation` to `Options` within `src/component-library/number-input/number-formatter/options.ts`. This change will allow the components to accept that prop.
- Many of the functions in `cleaners.ts` will need to be modified to work with scientific notation. Make sure to add tests in `cleaners.test.ts`.
- `formatter.ts` will also need modification to display numbers in scientific notation.
```
</details>

<details>
    <summary><h3>Retain cursor position</h3></summary>

`cat | gh issue create -F - -t 'Retain cursor position when changing text within FormattedInput' -m '[Beta] Better Editing Experience' -l "status: ready" -l "type: bug" -l "work: frontend"`

```
## Current behavior

1. In Storybook, open the NumberInput story.
1. Set the input to contain text `134`.
1. place the cursor between `1` and `3`.
1. Type `2`.
1. Expect the cursor to be between `2` and `3`.
1. Observe the cursor to be after `4`.

    This happens because the value of the input gets reformatted to add the comma.

## Desired behavior

- After modifying the input (via inserting text or deleting text), the cursor position generally meets users' expectations.

- There may be some edge cases where users expectations are not crystal clear. That's okay. We just want to offer some improvement that meets the expectations under most circumstances. For example, let's say we begin with `12,345` and the cursor between `1` and `2`. Then we press forwards-delete. We end up with the text `1,345`. Where should the cursor go? Between `1` and `,` or between `,` and `3`? Either would be fine and would be an improvement over the current behavior (where the cursor ends up at the very end).

## Implementation

1. In `FormattedInput.svelte`, in the function `handleChildValueChange`, you have access to:

    - the value before re-format (as `userInput`)
    - the cursor position before re-format (as the `cursorPosition` property of the return value from `getOutcomeOfBeforeInputEvent`)
    - the value after reformat (as `parseResult.intermediateDisplay`)

1. Write a pure function that takes input of the three values above and returns the resulting cursor position after the text has changed.

1. I suggest using https://www.npmjs.com/package/fast-diff to diff the text values. Then use the old cursor position to figure out where _in the diff_ the cursor should go, and use the diff to figure out where the cursor should go in the resulting text.
```
</details>

<details>
    <summary><h3>Research possibility of setting <tt>inputmode</tt> when negative</h3></summary>

`cat | gh issue create -F - -t 'For numeric inputs, research the possibility of setting the inputmode attribute when allowing negative numbers' -m '[Beta] Better Editing Experience' -l "status: ready" -l "type: enhancement" -l "work: frontend"`

```
- In In `src/component-library/number-input/numberInputUtils.ts` we have a function, `getInputMode` which derives the value of the [`inputmode` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) to set on the `<input>` element within our `NumberInput` and `StringifiedNumberInput` components. This helps the user's device display the correct software keyboard when necessary (e.g. on a phone).

- For positive numbers, we set `inputmode` to `decimal` or `numeric` and users get a nice keyboard with big number buttons.

    But MDN says:

    > Devices may or may not show a minus key (-).

    Because of this caveat in the MDN docs, we leave `inputmode` as `text` when the input allows negative numbers. Are we being overly cautious here? Are there _actually_ devices that don't show a minus key? Which ones? Can we set `inputmode` to `decimal` or `numeric` anyway? Or will there be users that can't enter negative numbers? 
```
</details>

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/component-library/number-input/number-formatter/NumberFormatter.ts
```diff
@@ -0,0 +1,62 @@
...
+  // normalizing the input for an expected integer, and all numbers are floats
+  // in javascript anyway.
+  const value = parseFloat(normalizedInput);
+  if (Number.isNaN(value)) {
+    return undefined;
+  }
+  return value;
+}
+
+export default class NumberFormatter implements InputFormatter<number> {
```
pavish, line 22: @seancolsen I think the number formatter should always return a JSON number compliant string instead of a number, when parsed.

In this scenario, assume if the user enters the following:
```
parse("123456.232423124353445"): 123456.23242312435
parse("123456232423124353445): 123456232423124350000
```

Notice that we lose precision in double precision and big integers. Also, when performing `parseFloat` or `parseInt` there are several other cases where we would get lossy values.

I think it's best if parsing only produces a JSON number compliant string by cleaning the string up, stripping down separators and only using `.` as the decimal separator, and we use that wherever user input is required.

seancolsen: Thanks for pointing this out, @pavish. This limitation has been in the back of my mind, but I haven't put enough thought yet into dealing with high-precision numbers. I'll think on this more and follow up, either by adopting your suggestion or by proposing something else.

seancolsen: @pavish Although this PR still needs more work, the changes I just pushed should address your concern. Now there's a `StringifiedNumberFormatter` and a `StringifiedNumberInput`.

per your agreement here https://github.com/centerofci/mathesar-wiki/pull/39#discussion_r828113600 I'm guessing this new approach will be cool with you. If so, can you resolve this thread?

pavish (approved):
@seancolsen Looks good to me! Nice work!

pavish:
@seancolsen I assume you'll create tickets mentioned in the description.

I'm going to mark https://github.com/centerofci/mathesar/issues/970 as closed.

seancolsen:
@pavish 

> create tickets mentioned in the description.

Thanks for the reminder. Here are the tickets:

- #1247
- #1248
- #1249

PR #1174, opened 2022-03-15
Minor refactor recycled from closed PR

dmos62:
This is a small collection of refactors recycled from a rejected PR (https://github.com/centerofci/mathesar/pull/1148).

It's also a minor extension of the filters endpoint to show some information relevant for equality-based filters. This will become useful again when we come back to the issue addressed by #1148. No breaking changes.

Some changes are only formatting, because, for example, I added some parameters to a function in the PR this is based on, making the declaration multi-line, but then removed the new parameters for this PR, after which only formatting changes were left. I don't think any of the formatting changes are detrimental on their own, so I didn't bother undoing them. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/api/db/viewsets/records.py
```diff
@@ -36,27 +36,28 @@ def list(self, request, table_pk=None):
...
         if grouping:
             group_by_columns_names = [column_ids_to_names[column_id] for column_id in grouping['columns']]
             name_converted_group_by = {**grouping, 'columns': group_by_columns_names}
         name_converted_order_by = [{**column, 'field': column_ids_to_names[column['field']]} for column in order_by]
-
         try:
-
+            filter_processed = _rewrite_db_function_spec(
+                filter_unprocessed, column_ids_to_names,
+            )
```
mathemancer, line 52: I think this logic would fit much better in the models utility module, `mathesar.utils.models`.

mathemancer (changes requested):
Looks pretty good overall. I think we should move the column id to name wrangling out of the viewset, and down into a model utility, since it's more related to the models (and so that would be easier to find/understand later).

dmos62:
@mathemancer in reply to:

[> I think this logic would fit much better in the models utility module, `mathesar.utils.models`.](https://github.com/centerofci/mathesar/pull/1174#discussion_r827653822):

The actual logic that does the rewriting is actually in the `mathesar.functions.operations.convert` namespace. This method in the records viewset file is just a wrapped call to that, which also adds handling for input being None, and includes a (hopefully helpful) doc string and some comments, so I didn't move it to another namespace.

[I removed this method, unwrapping the call to the rewriting method.](https://github.com/centerofci/mathesar/pull/1174/commits/4b82f5a70ce1bb5d587fa0f4894a4086d0e2b51f) I think it was better before. If you agree, we can revert back.

----------
@mathemancer in reply to:

[> Since you've set things up with default `None`s in the function sig, this conditional isn't needed: `limit(self, limit)` is a no-op when `limit = None`.](https://github.com/centerofci/mathesar/pull/1174#discussion_r827645339)

I'm aware that some of the `None` checks are redundant. But, the idea here is to treat each transformation in this pipeline the same. I tried to underline their conceptual equality by wrapping each of them in an identical `if` block, even if it's redundant in a few cases. I wanted to make it obvious that this is ripe for some refactoring: maybe something that would make the pipish nature of this routine more explicit.

Can we keep the redundant `if` checks?

dmos62:
@kgodey in this particular case, it's more confusing to have something not `None`-checked. I think a reader would take a second to realise that the SA call probably does a no-op if it's passed a `None`, but he can't be sure, because the enclosing method might never get called with that parameter set to `None`. These "redundant" if blocks tell the reader "there's nothing to think about: sometimes this is `None`, and in that case this block doesn't happen".

Some comments might be a good idea, but some code structure can be worth a lot of comments.

mathemancer:
> @mathemancer in reply to:
> 
> [> I think this logic would fit much better in the models utility module, `mathesar.utils.models`.](https://github.com/centerofci/mathesar/pull/1174#discussion_r827653822)
> 
> The actual logic that does the rewriting is actually in the `mathesar.functions.operations.convert` namespace. This method in the records viewset file is just a wrapped call to that, which also adds handling for input being None, and includes a (hopefully helpful) doc string and some comments, so I didn't move it to another namespace.
>
> [I removed this method, unwrapping the call to the rewriting method.](https://github.com/centerofci/mathesar/pull/1174/commits/4b82f5a70ce1bb5d587fa0f4894a4086d0e2b51f) I think it was better before. If you agree, we can revert back.

I strongly prefer the separate function version to inline. My problem is having non-exception branching in a viewset at all. That said, as I look at our other viewset code, maybe I'm in the minority there. I'm okay with reverting to the wrapped version, getting that in, and reassessing where we want what logic later.

> @mathemancer in reply to:
> 
> [> Since you've set things up with default `None`s in the function sig, this conditional isn't needed: `limit(self, limit)` is a no-op when `limit = None`.](https://github.com/centerofci/mathesar/pull/1174#discussion_r827645339)
> 
> I'm aware that some of the `None` checks are redundant. But, the idea here is to treat each transformation in this pipeline the same. I tried to underline their conceptual equality by wrapping each of them in an identical `if` block, even if it's redundant in a few cases. I wanted to make it obvious that this is ripe for some refactoring: maybe something that would make the pipish nature of this routine more explicit.
> 
> Can we keep the redundant `if` checks?

I think this will be a problem, since different pipeline bits might treat `None` differently. There might be cases where some explicit `None` is used. Further, even in the case we have, this added complexity adds the potential for bugs, and these are more complicated to test for in this setting than they would be function-by-function. `limit` and `offset` treat the falsy `0` value differently in comparison to the falsy `None`. A query with a limit of `0` can be used to return just column names (here, that possibility is short-circuited), whereas the offset of `0` is equal to an offset of `None` (with our current logic). While this specific example can be sorted by changing all checks to `if foo is None:` rather than `if foo:`, my broader point is that what to do about a `None` input should belong to the bit of the pipeline rather than to the caller, and tested for each part separately. In the cases of functions in the pipeline that _don't_ do the right thing when handed `None`, I think we should instead get those sorted out bit by bit.

dmos62:
This PR will likely be closed, since it's being superseded by https://github.com/centerofci/mathesar/pull/1222, but I'm keeping it open until after I merge the other one, since they might not completely overlap.

PR #1188, opened 2022-03-17
Avoid inferring type varchar

### LINKED ISSUE #1121: The type inference algorithm should use `TEXT` rather than `VARCHAR`
## Reproduce

1. "New Table" > "Import Data" > "Copy and Paste Text"

1. Paste the following data and proceed to create and view the table.

    ```txt
    first_name
    Marge
    Homer
    Lisa
    Bart
    Maggie
    ```

1. From the `columns` API, expect the response for the `first_name` column to have `"type": "TEXT"`

1. Observe instead that the column is `VARCHAR` without a length set.

## Rationale

- I spoke with @kgodey about the Mathesar Text type today and she say that Mathesar should only be configuring either: `TEXT` columns or `VARCHAR` columns with a length specified. She may be able to elaborate on the thinking that went into this decision.

## Additional context

- In #1118, we are doing some work to bring the front end into alignment with the above expectations when the user manually configures the DB settings for the Text type.

vrutik2809:
Fixes #1121 

While importing the data from .csv file, previously default datatype for the string was `VARCHAR` but now it is changed to `TEXT` (check the Screenshot)

**Screenshots**

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/operations/infer_types.py
+++ b/db/columns/operations/infer_types.py
@@ -15,6 +15,7 @@ logger = logging.getLogger(__name__)
 MAX_INFERENCE_DAG_DEPTH = 100
 
 TYPE_INFERENCE_DAG = {
+    base.PostgresType.TEXT.value:[],
     base.PostgresType.BOOLEAN.value: [],
     base.MathesarCustomType.EMAIL.value: [],
     base.PostgresType.INTERVAL.value: [],
--- a/db/columns/operations/infer_types.py
+++ b/db/columns/operations/infer_types.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 MAX_INFERENCE_DAG_DEPTH = 100
 
 TYPE_INFERENCE_DAG = {
-    base.PostgresType.TEXT.value:[],
+    base.PostgresType.TEXT.value: [],
     base.PostgresType.BOOLEAN.value: [],
     base.MathesarCustomType.EMAIL.value: [],
     base.PostgresType.INTERVAL.value: [],
--- a/db/columns/operations/infer_types.py
+++ b/db/columns/operations/infer_types.py
@@ -15,14 +15,13 @@ logger = logging.getLogger(__name__)
 MAX_INFERENCE_DAG_DEPTH = 100
 
 TYPE_INFERENCE_DAG = {
-    base.PostgresType.TEXT.value: [],
     base.PostgresType.BOOLEAN.value: [],
     base.MathesarCustomType.EMAIL.value: [],
     base.PostgresType.INTERVAL.value: [],
     base.PostgresType.NUMERIC.value: [
         base.PostgresType.BOOLEAN.value,
     ],
-    base.STRING: [
+    base.PostgresType.TEXT.value: [
         base.PostgresType.BOOLEAN.value,
         base.PostgresType.DATE.value,
         base.PostgresType.NUMERIC.value,
--- a/db/columns/operations/infer_types.py
+++ b/db/columns/operations/infer_types.py
@@ -44,9 +44,9 @@ def _get_reverse_type_map(engine):
     reverse_type_map = {v: k for k, v in supported_types.items()}
     reverse_type_map.update(
         {
-            Text: base.STRING,
-            TEXT: base.STRING,
-            VARCHAR: base.STRING,
+            Text: base.PostgresType.TEXT.value,
+            TEXT: base.PostgresType.TEXT.value,
+            VARCHAR: base.PostgresType.TEXT.value,
         }
     )
     return reverse_type_map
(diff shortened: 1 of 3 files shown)


db/columns/operations/infer_types.py
```diff
@@ -15,6 +15,7 @@
 MAX_INFERENCE_DAG_DEPTH = 100
 
 TYPE_INFERENCE_DAG = {
+    base.PostgresType.TEXT.value: [],
```
mathemancer, line 18: The way this is set up, you'll never infer a column to be any type except `TEXT`, if it starts as `TEXT`. You need to put `TEXT` upstream of all other types (i.e., where `String` is now).

You need to change the reverse type map function to avoid an infinite loop.

mathemancer (changes requested):
See specific comment regarding the DAG.

db/columns/operations/infer_types.py
dmos62: > The way this is set up, you'll never infer a column to be any type except TEXT, if it starts as TEXT.

@mathemancer I'm surprised that this didn't trigger a test failure.

vrutik2809: Okay. So i need to add `base.PostgresType.TEXT.value` as first value in `base.STRING` and then change the `reverse_type_function()` to avoid an infinite loop.
Please confirm it

vrutik2809: Can you please elaborate what `_get_reverse_type_map()` function is doing?

mathemancer: The way the type inference algorithm works is:
 1. reflect a column.
 2. get its type.
 3. map that type to a string representation through `_get_reverse_type_map`.
 4. Look in the `TYPE_INFERENCE_DAG` dictionary for that string as a key
 5. Get the list value associated with that key.
 6. Attempt to cast the column to the type represented by each list value in turn.
 7. If all fail, you're done (leave the column as-is) if you succeed, go to (1).

@dmos62 Actually, there is a test that checks this, and it's failing, but for the wrong reasons. :hankey: I interpreted this failure incorrectly in previous comments. It's actually failing just because it wasn't updated to the new regime where the default column type is `TEXT`.

We treat all string-like types as the same, except for the extra-restrictive `CHAR`. This is what the little dictionary update in `_get_reverse_type_map` is for. It maps all those types to the same string representation.

@vrutik2809 You should try to accomplish the following:

If a column is of type `TEXT`, in the database, it should have every type as a downstream option. Columns of type `TEXT` in the database will have an SQLAlchemy type of either `sqlalchemy.TEXT` or `sqlalchemy.Text`. (The former when reflected, and the latter when freshly created in the Python layer). To represent this, I prefer replacing the `base.STRING` key with `base.PostgresType.TEXT.value` in the `TYPE_INFERENCE_DAG`, and then modifying the `_get_reverse_type_map` function to point all of the python types `Text`, `TEXT`, and `VARCHAR` (as imported in that file) to `base.PostgresType.TEXT.value` in the reverse type map.

vrutik2809: Okay. I will look into that.
Thank you

vrutik2809:
These tests are failing:
```
1. mathesar/tests/api/test_table_api.py::test_table_type_suggestion - Ass...
2. mathesar/tests/api/test_table_api.py::test_table_patch_columns_invalid_type_with_name
3. mathesar/tests/api/test_table_api.py::test_table_patch_columns_invalid_type_with_type
4. mathesar/tests/api/test_table_api.py::test_table_patch_columns_invalid_type_with_drop
5. mathesar/tests/api/test_table_api.py::test_table_patch_columns_invalid_type_with_multiple_changes
```
The 1st one I got it that it is failing because the default is set to `varchar`. So that I need to change that to `TEXT`. But I can't figure out the rest. I'm assuming that they are failing because they all are using the `_get_data_types_column_data()` function in which `VARCHAR` is there.

kgodey:
@vrutik2809 For many of the tests you listed, the data is being tested against `_get_data_types_column_data()` on line 1116, which sets types as `VARCHAR`, not `TEXT`. If you change those, I think your tests will pass. 

```python
def test_table_patch_columns_invalid_type_with_name(create_data_types_table, client, engine_email_type):
    table_name = 'PATCH columns 15'
    table = create_data_types_table(table_name)
    column_data = _get_data_types_column_data()
    column_data[1]['name'] = 'hello'
    column_data[3]['type'] = 'BOOLEAN'

    body = {
        'columns': column_data
    }
    response = client.patch(f'/api/db/v0/tables/{table.id}/', body)
    response_json = response.json()
    assert response.status_code == 400
    assert 'Pizza is not a boolean' in response_json[0]['message']

    current_table_response = client.get(f'/api/db/v0/tables/{table.id}/')
    # The table should not have changed
    original_column_data = _get_data_types_column_data()
    _check_columns(current_table_response.json()['columns'], original_column_data)
```

For example, see the penultimate line of this test, which is getting `_get_data_types_column_data()` and comparing that to the API response.

mathemancer (approved):
I think this looks good. Thank you for your effort on this, @vrutik2809 !

PR #1190, opened 2022-03-17
Add db type alias information to db types endpoint

### LINKED ISSUE #1037: Add database type alias information to database type endpoint
## Problem
We have some db types that are equivalent. In such cases, we want to tell the frontend which of those types is the default or canonical one, so as to prevent it from having to hardcode it. E.g. `DECIMAL` and `NUMERIC` are [different in the SQL standard, but equivalent in Postgres](https://stackoverflow.com/questions/33730538/difference-between-decimal-and-numeric-datatype-in-psql) and thus equivalent in Mathesar.

## Proposed solution

Off the top of my head I see two approaches:

1. Define a type alias as being the non-canonical equivalent and append the following attribute in returned JSON to aliases `'alias-of': 'SOME CANONICAL TYPE'`.

2. Define all equivalent types, canonical and not, as aliases and append an `aliases` attribute containing a list of the rest of the equivalent types to all aliases (`'aliases': ['SOME OTHER ALIAS', 'SOME YET ANOTHER ALIAS']`), then on each alias append whether it's canonical or not: `'is_canonical_alias': true | false`.

## Additional context

Proposed here: https://github.com/centerofci/mathesar/discussions/1035#discussioncomment-2087541

dmos62:
Fixes #1037 

This adds database type alias information to the `/api/db/v0/databases/1/types/` endpoint.

**Technical details**

Alias information copied from table in https://www.postgresql.org/docs/11/datatype.html.

Adds two new fields to objects in the `types/` endpoint output: `aliases` (contains aliases for this type, if this type is canonical and has known aliases) and `alias_of` (contains the canonical alias for this type, if this type is known to be a non-canonical alias).

So, for example, if `alias_for` is `null`, according to the information in PG's documentation, it's a canonical.

Mentioned endpoint's output after changes:

```json
[
  {
    "id": "_ARRAY",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "BIGINT",
    "aliases": [
      "INT8"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "BIT VARYING",
    "aliases": [
      "VARBIT"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "BIT",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "BOOLEAN",
    "aliases": [
      "BOOL"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "boolean"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "BYTEA",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "\"CHAR\"",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "string_like"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "CHARACTER VARYING",
    "aliases": [
      "VARCHAR"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "string_like"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "CHARACTER",
    "aliases": [
      "CHAR"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "string_like"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "CIDR",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "DATE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "point_in_time"
      },
      {
        "id": "date"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "DATERANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "DECIMAL",
    "aliases": [],
    "alias_of": "NUMERIC",
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "DOUBLE PRECISION",
    "aliases": [
      "FLOAT8"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "FLOAT",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "HSTORE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "INET",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "INT4RANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "INT8RANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "INTEGER",
    "aliases": [
      "INT",
      "INT4"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "INTERVAL",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "duration"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "JSON",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "JSONB",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "MACADDR",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "MONEY",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "NAME",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "string_like"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "NUMERIC",
    "aliases": [
      "DECIMAL"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "NUMRANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "OID",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "REAL",
    "aliases": [
      "FLOAT4"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "REGCLASS",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "SMALLINT",
    "aliases": [
      "INT2"
    ],
    "alias_of": null,
    "hints": [
      {
        "id": "comparable"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TEXT",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "string_like"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TIME",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      },
      {
        "id": "comparable"
      },
      {
        "id": "time"
      },
      {
        "id": "point_in_time"
      }
    ]
  },
  {
    "id": "TIME WITH TIME ZONE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      },
      {
        "id": "comparable"
      },
      {
        "id": "time"
      },
      {
        "id": "point_in_time"
      }
    ]
  },
  {
    "id": "TIME WITHOUT TIME ZONE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      },
      {
        "id": "comparable"
      },
      {
        "id": "time"
      },
      {
        "id": "point_in_time"
      }
    ]
  },
  {
    "id": "TIMESTAMP",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "time"
      },
      {
        "id": "date"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TIMESTAMP WITH TIME ZONE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "time"
      },
      {
        "id": "date"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TIMESTAMP WITHOUT TIME ZONE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "time"
      },
      {
        "id": "date"
      },
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TSRANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TSTZRANGE",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "TSVECTOR",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  },
  {
    "id": "UUID",
    "aliases": [],
    "alias_of": null,
    "hints": [
      {
        "id": "any"
      }
    ]
  }
]

```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/types/aliases.py
+++ b/db/types/aliases.py
@@ -1,0 +1,34 @@
+from frozendict import frozendict
+
+
+# Copied (and reordered) from table in
+# https://www.postgresql.org/docs/11/datatype.html
+canonical_to_aliases = frozendict(
+    {
+        'boolean': ['bool'],
+        'bit varying': ['varbit'],
+        'smallint': ['int2'],
+        'integer': ['int', 'int4'],
+        'bigint': ['int8'],
+        'smallserial': ['serial2'],
+        'serial': ['serial4'],
+        'bigserial': ['serial8'],
+        'real': ['float4'],
+        'double precision': ['float8'],
+        'numeric': ['decimal'],
+        'character': ['char'],
+        'character varying': ['varchar'],
+        'time with timezone': ['timetz'],
+        'timestamp with timezone': ['timestamptz'],
+    }
+)
+
+alias_to_canonical = frozendict(
+    {
+        alias: canonical
+        for canonical, aliases
+        in canonical_to_aliases.items()
+        for alias
+        in aliases
+    }
+)
--- a/db/types/base.py
+++ b/db/types/base.py
@@ -3,7 +3,7 @@ from enum import Enum
 from sqlalchemy import create_engine
 
 from db import constants
-
+from db.types.aliases import canonical_to_aliases, alias_to_canonical
 from db.functions import hints
 
 from frozendict import frozendict
@@ -14,7 +14,21 @@ STRING = 'string'
 VARCHAR = 'varchar'
 
 
-class PostgresType(Enum):
+class DatabaseType:
+    @property
+    def is_canonical(self):
+        return self.value not in aliases
+
+    @property
+    def aliases(self):
+        return canonical_to_aliases.get(self.value, [])
+
+    @property
+    def alias_of(self):
+        return alias_to_canonical.get(self.value)
+
+
+class PostgresType(DatabaseType, Enum):
     """
     This only includes built-in Postgres types that SQLAlchemy supports.
     SQLAlchemy doesn't support XML. See zzzeek's comment on:
@@ -66,7 +80,7 @@ class PostgresType(Enum):
     UUID = 'uuid'
 
 
-class MathesarCustomType(Enum):
+class MathesarCustomType(DatabaseType, Enum):
     """
     This is a list of custom Mathesar DB types.
     Keys returned by get_available_types are of the format 'mathesar_types.VALUE'
--- a/db/types/base.py
+++ b/db/types/base.py
@@ -17,7 +17,7 @@ VARCHAR = 'varchar'
 class DatabaseType:
     @property
     def is_canonical(self):
-        return self.value not in aliases
+        return self.value in canonical_to_aliases
 
     @property
     def aliases(self):
--- a/mathesar/api/serializers/db_types.py
+++ b/mathesar/api/serializers/db_types.py
@@ -8,7 +8,14 @@ class DBTypeSerializer(serializers.Serializer):
     hints = serializers.ListField(child=serializers.DictField())
 
     def to_representation(self, db_type):
+        # TODO solve db type casing holistically
+        # https://github.com/centerofci/mathesar/issues/1036
+        uppercase_id = db_type.value.upper()
+        uppercased_aliases = [db_type_id.upper() for db_type_id in db_type.aliases]
+        uppercased_alias_of = db_type.alias_of.upper() if db_type.alias_of else None
         return {
-            "id": db_type.value,
+            "id": uppercase_id,
+            "aliases": uppercased_aliases,
+            "alias_of": uppercased_alias_of,
             "hints": db_types_hinted.get(db_type, None),
         }


dmos62:
@pavish can you take a look and see if this suits frontend needs? I considered creating a separate endpoint, something like `aliases`, with a mapping from aliases to canonicals, but instead added the `aliases` and `alias_of` fields to the `types/` endpoint. If a separate endpoint makes more sense to you, it's not a lot of work on my end.

pavish:
@dmos62 This looks good and makes it a lot easier for the frontend than having a separate endpoint.

kgodey:
The structure looks good to me.

@dmos62 FYI, I edited your issue description to mark the snippet as JSON, the syntax highlighting makes it easier to read.

mathemancer (changes requested):
If we take as a given that this feature is desirable, then the canonical type name shouldn't be related to the DB docs, but rather to the string names for types in the `engine.dialect.ischema_names` dictionary. The reasons are:
  - This is the dictionary that defines the type for SQLAlchemy, and 
  - these strings are _guaranteed_ to be the name of the type upon reflection (otherwise reflection wouldn't work).

However, as I noted in [my comment](https://github.com/centerofci/mathesar/issues/1037#issuecomment-1073532790) on the linked issue, I don't think we should add this feature.
- If the front end is being handed two options as names for some type, that's a bug.
- If the type name is different at different endpoints for some type, that's a bug.

Taking the example of both `DECIMAL` and `NUMERIC` being given as options, we should only give `NUMERIC`. Even if you create a column of type `DECIMAL`, it will be reflected as `NUMERIC`. This is in stark contrast to the situation with `VARCHAR` and `TEXT`. These are distinct types, since `VARCHAR` takes a `length` type option, whereas `TEXT` does not.

@pavish I'd be interested in your thoughts here. Do you actually see a need for alias information in the front end, or would it be better to simply avoid multiple names for a given type?

dmos62:
@mathemancer good catch. I agree.

We had a call about this with Brent after he made this comment.

I summarized our thoughts in this discussion: https://github.com/centerofci/mathesar/discussions/1184#discussioncomment-2409037

The short of it is that, if we're right that the only reason to expose type aliases to frontend was because the frontend is exposed to database aliases by the db types endpoint, then a better solution is to fix that at the db types endpoint. Or, from the backend's perspective, to alter `ischema_names` or wrap it in an abstraction.

@pavish if getting rid of database aliases would solve your use case, feel free to close this PR.

pavish:
@dmos62 It's not only on the db types endpoint but also the `column.type` property on the columns endpoint. From the previous comments and the linked thread, I assume that is covered as well.

@mathemancer 
> Do you actually see a need for alias information in the front end, or would it be better to simply avoid multiple names for a given type?

I'm on board with not having any alias information and just have a single name for a single type.

However, we do show the DB type to the user on the frontend, so I'm not sure if it will be an issue for the user if we do not show the exact name of the type that the db shows. Eg., The DB cli tools show it as `DECIMAL` and we show it as `NUMERIC`. 

@kgodey I'd like your thoughts on this.

kgodey:
For types with aliases, I think it's preferable to have a canonical type that the API always shows rather than have the frontend deal with aliases. I don't think it matters, for example, if we show `DECIMAL` or `NUMERIC` for the DB type since they're the same thing.

I'm in support of closing this PR and repurposing #1037 to do the work of ensuring that the API has a single name for a single type. I do think the API should be able to _accept_ aliases (e.g. you could create a type as either `DECIMAL` or `NUMERIC`, and it would always reflect as `NUMERIC`), but we can handle that post-alpha if needed.

pavish:
@dmos62 Since we are all in agreement, I'm going to close this PR.

Issue #259, opened 2021-06-14
Frontend implementation for Money Mathesar type


kgodey:
## Requirements

Users should be able to:
- view data in columns that have the Money Mathesar type
- change data type to and from the associated DB types based on options selected 
  - `MATHESAR_TYPES.MONEY_[XXX]`
  - `MONEY`
- view, set, and change data type options related to the Money data types
- ~~filter data by Money-specific filtering options~~
- ~~group data by Money-specific grouping options~~
- use display options to view data in Money type columns
  - Data should be rounded according to the precision specified in the display options
- see Money data shown in a monospace font or using tabular figures (see design spec)
- see Money data formatted according to the locale of the currency 

Update:
- Grouping will be handled in #1073 
- Filtering will be handled in #1072 

## DB Type Mapping
- If the custom `MATHESAR_TYPES.MONEY` type is available:
  - all new columns of UI type Money should use it.
  - converting existing columns to the UI type Money should use it.
- If the custom `MATHESAR_TYPES.MONEY` type is not available:
  - all new columns of UI type Money or converting existing columns to UI type Money should use default `MONEY` type 

Existing columns of type `MONEY` can only be converted to `MATHESAR_TYPES.MONEY` if `MATHESAR_TYPES.MONEY` is available by converting first to Number and then converting back to Money.

Please note that the design for `MONEY` vary on whether `MATHESAR_TYPES.MONEY` is installed.

### Design
- The [Global Data Types](https://wiki.mathesar.org/design/specs/global-data-type-components) spec has the latest version of the UX.
- The [Money Data Type](https://wiki.mathesar.org/design/specs/money-data-type) spec has an older version of the UX but more information on user scenarios.

### Backend
- `/api/v0/tables/` lists the name and DB type of each columns 
- `/api/v0/tables/<id/columns/` lists column information in more detail
- `/api/v0/databases/<id>/types/` maps Mathesar types to DB types and provides a list of filtering and grouping options per-type.
  - These should be used to populate the filter and grouping options dynamically. 
- Updates to column types and type options should happen via `PATCH` requests to `/api/v0/tables/<id>/columns/<id>/`. These updates include
  - Changes to the data type (`type`)
  - Changes to the display options (`display_options`)

## Additional Context
- Design work:
  - #239
  - #655
- #269 (concept of Mathesar types)
- #498 (conversation about data types)
- Blocked by backend work:
  - #417 
  - #420 
  - #1094 
  - #1105

PR #1205, opened 2022-03-21
Accept display options for MathesarMoney type

### LINKED ISSUE #1105: Implement display options for the custom Money type
## Problem
We need to implement support for display options for the custom Money type.

## Solution
The custom money type supports the following display options. We need to represent them in the API, validate them, and also set them automatically on inference of Money types.

### Format Type
This stores the currency that the Money type should be displayed as. We will store a pre-set number of currencies that the frontend can pick from, or we will allow users to use "Custom" formatting that doesn't obey the rules of a specific currency.
#### Options
- Either a currency code or "Custom"
- If the user manually updates any of the other display options, we should automatically set this to "Custom"

### Symbol
Users can set a custom currency symbol here. 

#### Options
- Any text.
- Whenever the "Format Type"  setting is changed to a currency code, this should be set to the appropriate currency symbol for that code.

### Symbol Location
Location of the currency symbol.

#### Options
- "Beginning" or "End"
- Whenever the "Format Type"  setting is changed to a currency code, this should be set to the appropriate symbol location for that code.

### Decimal Symbol
Location of the decimal symbol.

#### Options
- "." or ","
- Whenever the "Format Type"  setting is changed to a currency code, this should be set to the appropriate decimal symbol for that code.

### Digit Grouping
How the digits are grouped.

#### Options 
- ‘123456789" (represented as `[0]`)
- "123,456,789" (represented as `[3]`)
- "123456,789" (represented as `[3, 0]`)
- ‘12,34,56,789" (represented as `[3, 2]`)

Whenever the "Format Type"  setting is changed to a currency code, this should be set to the appropriate grouping for that code.

Note: These representations are based on [.NET docs](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.numbergroupsizes?view=net-6.0).

### Digit Grouping Symbol
What symbol is used for digit grouping

#### Options
- None
- Thin Space
- ","
- "."

Whenever the "Format Type"  setting is changed to a currency code, this should be set to the appropriate grouping for that code.

## Expected Outcomes
- The frontend should be able to set any of these display options.
- The frontend should know what the choices are for any display option that supports a set of choices.
- The display options should be validated.
- If any of the display options end up changing other display options, the backend should handle that.
    - This means the backend should know this data for a set of known currencies,  we should use a library or dataset for that.
- When we infer a column as a money type, we should set the display options for that column based on the currency symbols in the original CSV/other file.

## Additional context
- Blocked by #1094
- [Money Data Type design spec](https://wiki.mathesar.org/en/design/specs/money-data-type)
    - [Design prototype](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=7552%3A83433&node-id=7590%3A84021&viewport=241%2C48%2C0.46&scaling=contain)

silentninja:
Fixes #1105 by adding display options support for `MathesarMoney` type.

## Features List
- [X] The frontend should be able to set any of these display options.
 - mathesar.tests.api.test_column_api.test_column_create_display_options

- [x] The frontend should know what the choices are for any display option that supports a set of choices.
- mathesar.tests.api.test_ui_types_api.test_type_list
- mathesar.api.display_options.money_display_options_schema
- [x] The display options should be validated.
 - mathesar.tests.api.test_column_api.test_column_create_wrong_display_options
 
- [X] If any of the display options end up changing other display options, the backend should handle that.
   - mathesar.tests.api.test_column_api.test_column_update_mathesar_money_display_options

- [X] When we infer a column as a money type, we should set the display options for that column based on the currency symbols in the original CSV/other file.
 - mathesar.tests.display_options_inference.test_money.test_display_options_inference

## Additional fix
- Fix DRF error mixin to allow custom error exceptions inside Serializer validations

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@silentninja I happened to see your new `currency_info.json` in the context of a related conversation and I just wanted to make a small comment here that the property names in that file are quite cryptic. At a glance, it's hard for me to understand what properties like `p_cs_precedes` and `n_sign_posn` represent. I would suggest always using complete words. No abbreviations. And nesting objects might help too, for example, 

```json5
{
  // ...
  "negative_sign": {
    "text": "-",
    "position": 1,
  },
  // ...
}
```

silentninja:
> At a glance, it's hard for me to understand what properties like p_cs_precedes and n_sign_posn represent

@seancolsen I just extracted data from the [locale files](https://docs.python.org/3/library/locale.html#locale.localeconv), it is used internally by the backend. There are an issue #1218 to clean the dataset, to make the currency codes readable. As for the property names, we could make it readable, but I feel it would better to stick with the standard locale naming convention rather than introducing our own for compatibility reasons. The frontend won't be receiving these property names, rather it would be the display option names listed in the spec

kgodey:
I'd like @mathemancer to take a look at this.

mathesar/api/display_options.py
```diff
@@ -1,34 +1,70 @@
...
+            {
+                "name": "show_as_percentage",
+                "type": "string",
+                "enum": ['dropdown', 'checkbox']
+            },
+            {
+                "name": "locale",
+                "type": "string"
+            }
+        ]
```
mathemancer, line 59: I realize this is carried over, but I think we should change this to be more like the money display options, as I mentioned in a comment elsewhere. There are plenty of reasons to display numbers using configurations not represented by a locale. It has to do with preferences, not just your location.

db/types/money.py
```diff
@@ -25,3 +29,18 @@ def install(engine):
         conn.execute(text(drop_domain_query))
         conn.execute(text(create_domain_query))
         conn.commit()
+
+
+def get_money_array(table_oid, engine, column_attnum):
```
mathemancer, line 34: I think you should rename this or add some clear comments that this only gets an array representing the pieces of the string in the first row. It's a bit confusing, and we want to make sure it's clear that the currency display option decisions are based on a single entry.

db/types/money.py
```diff
@@ -25,3 +29,18 @@ def install(engine):
...
+    table = reflect_table_from_oid(table_oid, engine)
+    column_name = get_column_name_from_attnum(table_oid, column_attnum, engine)
+    with engine.begin() as conn:
+        package_func = getattr(func, get_qualifier_prefix())
+        money_func = getattr(package_func, MONEY_ARR_FUNC_NAME)
+        money_select = money_func((table.c[column_name]), type_=ARRAY(String))
+        sel = select(money_select).where(money_select[4] is not None)
+        data = conn.execute(sel).fetchone()
```
mathemancer, line 42: Did you double-check that this doesn't run the function for all rows before returning one in `fetchone`? The function in question is pretty intense, and I think it might make sense to limit the query (unless it's not actually running for all rows).

mathemancer (changes requested):
Overall, I think this makes the right compromises given the goals. I do think you should add comments about that make it clear for future readers that the inference isn't over the whole column, even though the results are applied column-wide. I noted a couple specific spots.

Also, please double-check or confirm that the function that gets the money array doesn't actually run the underlying function on the whole column unnecessarily.

kgodey:
I think we should mark this PR as a draft until #1243 is resolved since we might end up changing the structure of display options.

db/types/money.py
silentninja: Nice catch! Got accustomed to Django's behaviour and forgot we are dealing with cursors here.

silentninja:
Blocked by #1335

seancolsen (approved):
I haven't looked at the diff at all here. (It's kind of baffling to me that so many changes would be required for what seems like a relatively limited addition to the API, so I haven't even bothered to try understanding what's going on under the hood.)

I did verify that the API is working as I need it to work though.

The following display options, seem to work well according to these [recent and somewhat ad-hoc specs](https://github.com/centerofci/mathesar/discussions/1243#discussioncomment-2519195):

- `number_format`
- `currency_symbol`
- `currency_symbol_location`

And that's all I need in order to continue working on #259, so I'd like to get this merged.

But I'll hold off on merging until we get a proper backend review.

kgodey:
@seancolsen Most of the complexity here is because money has required display options that we need to set from the backend. If we infer a column in a CSV as `MONEY`, we need to also figure out what the currency is and set the display option accordingly. No other type has anything like that.

I've assigned @mathemancer to review this since it touches type casting logic.

mathemancer (approved):
I think this looks good. Let's get it merged!

PR #1210, opened 2022-03-22
Fixes #1197 - raised the playwright page default from 5sec to 10 sec

### LINKED ISSUE #1197: Raise the default Playwright page timeout value
## Problem

- Some E2E tests are failing intermittently. For example, [here](https://github.com/centerofci/mathesar/runs/5605572101?check_suite_focus=true) is a CI job that failed but passed when re-ran.
- I'm theorizing that we're hitting Playwright's default 5-second timeout on some operations that take longer.

## Solution

- I think we should raise the timeout value globally.
- Playwright has a [set_default_timeout](https://playwright.dev/python/docs/api/class-page#page-set-default-timeout) method that we can run on the `page` object.
- We need to figure out where in our E2E testing stack is appropriate to call this function so that it applies everywhere.
- I think we should set it to 10 seconds.

## Tips

- If you're working on this, I'd suggest the following. In order to test that your fix works, first set the timeout to something very small, like 50ms, and verify that you see some tests which fail (giving you an indication that your setting has been correctly applied). Then tweak the setting to 10 seconds.

## Additional context

- [Discussion on Matrix](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$xZcuhHRDrvqO7cBYXODdqs1tmRxuWcPlHmxobSR_beo?via=matrix.mathesar.org&via=matrix.org)

Cronus1007:
Fixes #1197 

**Technical details**
raised the playwright page default from 5sec to 10 sec
**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/run-flake8.yml
+++ b/.github/workflows/run-flake8.yml
@@ -16,5 +16,6 @@ jobs:
         with:
           checkName: "flake8"
           path: "."
+          plugins: flake8-no-types
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/db/columns/base.py
+++ b/db/columns/base.py
@@ -2,7 +2,7 @@ from sqlalchemy import Column, ForeignKey, inspect
 
 from db.columns.defaults import TYPE, PRIMARY_KEY, NULLABLE, DEFAULT_COLUMNS
 from db.columns.operations.select import (
-    get_column_attnum_from_name, get_column_default, get_column_default_dict, get_column_index_from_name,
+    get_column_attnum_from_name, get_column_default, get_column_default_dict,
 )
 from db.tables.operations.select import get_oid_from_table
 from db.types.operations.cast import get_full_cast_map
@@ -16,6 +16,7 @@ class MathesarColumn(Column):
     column definition that we care about, and this class defines that
     subset.
     """
+
     def __init__(
             self,
             name,
@@ -120,23 +121,6 @@ class MathesarColumn(Column):
                 )
             )
             return valid_target_types if valid_target_types else None
-
-    @property
-    def column_index(self):
-        """
-        Get the ordinal index of this column in its table, if it is
-        attached to a table that is associated with the column's engine.
-        """
-        if (
-                self.engine is not None
-                and self.table_ is not None
-                and inspect(self.engine).has_table(self.table_.name, schema=self.table_.schema)
-        ):
-            return get_column_index_from_name(
-                self.table_oid,
-                self.name,
-                self.engine
-            )
 
     @property
     def column_attnum(self):
--- a/db/columns/operations/create.py
+++ b/db/columns/operations/create.py
@@ -19,8 +19,14 @@ from db.constraints import utils as constraint_utils
 from db.tables.operations.select import reflect_table_from_oid
 from db.types.operations.cast import get_supported_alter_column_types
 
+COLUMN_NAME_TEMPLATE = 'Column'
+
 
 def create_column(engine, table_oid, column_data):
+    table = reflect_table_from_oid(table_oid, engine)
+    column_name = column_data.get(NAME, '').strip()
+    if column_name == '':
+        column_data[NAME] = gen_col_name(table)
     column_type = column_data.get(TYPE, column_data.get("type"))
     column_type_options = column_data.get("type_options", {})
     column_nullable = column_data.get(NULLABLE, True)
@@ -65,6 +71,13 @@ def create_column(engine, table_oid, column_data):
         reflect_table_from_oid(table_oid, engine).columns[column_data[NAME]],
         engine
     )
+
+
+def gen_col_name(table):
+    base_name = COLUMN_NAME_TEMPLATE
+    col_num = len(table.c)
+    name = f'{base_name} {col_num}'
+    return name
 
 
 def _gen_col_name(table, column_name):
--- a/db/columns/operations/infer_types.py
+++ b/db/columns/operations/infer_types.py
@@ -21,7 +21,7 @@ TYPE_INFERENCE_DAG = {
     base.PostgresType.NUMERIC.value: [
         base.PostgresType.BOOLEAN.value,
     ],
-    base.STRING: [
+    base.PostgresType.TEXT.value: [
         base.PostgresType.BOOLEAN.value,
         base.PostgresType.DATE.value,
         base.PostgresType.NUMERIC.value,
@@ -44,9 +44,9 @@ def _get_reverse_type_map(engine):
     reverse_type_map = {v: k for k, v in supported_types.items()}
     reverse_type_map.update(
         {
-            Text: base.STRING,
-            TEXT: base.STRING,
-            VARCHAR: base.STRING,
+            Text: base.PostgresType.TEXT.value,
+            TEXT: base.PostgresType.TEXT.value,
+            VARCHAR: base.PostgresType.TEXT.value,
         }
     )
     return reverse_type_map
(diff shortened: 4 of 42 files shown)


aviwad:
why did you increase timeout in test_import_from_clipboard? because aside from that our PR is identical haha

seancolsen (changes requested):
@Cronus1007 Thanks for this PR. It's not quite what I had in-mind though when I wrote #1197. From the ticket:

> we should raise the timeout value **globally**

and 

> figure out where in our E2E testing stack is appropriate to call this function so that it applies **everywhere**

(emphasis added)

As I said in the ticket, I don't know how to raise the timeout value globally, because I'm new to Playwright (as is everyone else on our team). But I'm confident there is a way. The PR should specify the 10 second value in only _one place_, and that value should apply to _all_ E2E tests.

Because the steps to resolve this issue are somewhat unknown (at least to me), I was initially hesitant to apply the "Good first issue" label, but someone else on the team suggested it. If you want to move on to another ticket that's fine. (Just let us know so that we can un-assign you). Or if you want to keep working on this one you'll probably need to do a little research into configuring Playwright.

Cronus1007:
@seancolsen Sure this is just an initial stage where I wanted to reflect those tests which would need a timeout.
Plus I would like to make an addition to the problem statement that this where [this workflow](https://github.com/centerofci/mathesar/runs/5605572101?check_suite_focus=true) failed at the expect statement which has a default timeout of 5s so we need to set the default timeout in  config in this manner 
`config = { expect: { timeout: 10000 } }`

This is my solution for this problem. Shall I proceed it in this manner.

seancolsen:
@Cronus1007 I don't understand what you're proposing in your last message. Sorry.

> Can you please tell me where the config file has been placed in the codebase?

No, because I don't know. Figuring it out is part of what's required to close this issue. If you want to move to another issue that's fine. Just let us know.

Cronus1007:
@seancolsen 

> @Cronus1007 I don't understand what you're proposing in your last message. Sorry.
> 
> > Can you please tell me where the config file has been placed in the codebase?
> 
> No, because I don't know. Figuring it out is part of what's required to close this issue. If you want to move to another issue that's fine. Just let us know.

No I would work upon this issue. Since I am little bit familiar with playwright with javascript I guess I would figure it out. What's your opinion about the 

> Plus I would like to make an addition to the problem statement that [this workflow]
> (https://github.com/centerofci/mathesar/runs/5605572101?check_suite_focus=true) failed at the expect statement which has > a default timeout of 5s

Cronus1007:
@seancolsen Have a look at this [documentation](https://playwright.dev/docs/test-timeouts). This one is related to the playwright javascript since I am unable to find out the such sort of documentation for the playwright python. As per the documentation we need not to worry about the global timeout since it has no default timeout. For solving this issue we only need to increase the expect timeout from 5s to 10s. For a particular expect statement it can be done in this manner 
`
{expect(locator).toBeVisible({ timeout: 10000 })}
`
I am raising an issue at the microsoft/playwright-python since in the [documentation](https://playwright.dev/python/docs/intro) of playwright python they had no such guide which deals with the configuration and timeouts in playwright python.

pavish:
@Cronus1007 @seancolsen 

I'd like to jump in to provide a suggestion here. The python configuration of playwright differs significantly from it's Node.js equivalent, as in there is no config file. We use the `pytest-playwright` plugin, which provides us with a `page` fixture.

This also gives us more freedom. Inorder to use global page configurations, all we have to do is create a new fixture, maybe `global_page` (I'm terrible at names) and use that fixture instead of `page` everywhere. This `global_page` fixture can return a page object with the timeout set to 30s.

Cronus1007:
@pavish @seancolsen 
We can use it in this manner
def new_global_page_fixture(playwright):
    browser = playwright.chromium.launch(channel="chrome")
    context = browser.new_context()
    global_page = context.new_page()
    global_page.set_default_navigation_timeout(30000)
    browser.close()

with sync_playwright() as playwright:
    new_global_page_fixture(playwright)

silentninja:
@Cronus1007 Take a look at this issue https://github.com/microsoft/playwright-pytest/issues/23 and see if it works

Cronus1007:
@silentninja @seancolsen I am increasing the default timeout value to 20 sec. Since while setting up the context's default timeout at 10 sec few tests were failing so the most optimum timeout value to which I reached was 20sec.

@silentninja They have not implemented their fixtures feature until now.

pavish:
@Cronus1007 @silentninja

This context timeout is not the timeout of locators but rather the timeout for launching the browser. This is different from the page locator timeout.

Please override the page fixture instead.

Cronus1007:
@pavish @silentninja @seancolsen This approach of mine is working since here I have created a new instance of page object from Page class and that function gave an additional property regarding the page's default navigation timeout value to be 20 sec. 
When I reduced the timeout to 2sec the tests failed. Also if we wanted to override the new timeout settings to a test we need to pass the new_default_timeout function to the test.

Cronus1007:
@pavish I guess this is what you intent to tell because locally by running the command 
`docker exec mathesar_service pytest --browser webkit  --no-cov mathesar/tests/integration`
I am able to reslve this issue. Can you please approve [this workflow](https://github.com/centerofci/mathesar/actions/runs/2036456199)  for a final check?

Cronus1007:
> @Cronus1007 @seancolsen
> 
> I'd like to jump in to provide a suggestion here. The python configuration of playwright differs significantly from it's Node.js equivalent, as in there is no config file. We use the `pytest-playwright` plugin, which provides us with a `page` fixture.
> 
> This also gives us more freedom. Inorder to use global page configurations, all we have to do is create a new fixture, maybe `global_page` (I'm terrible at names) and use that fixture instead of `page` everywhere. This `global_page` fixture can return a page object with the timeout set to 30s.

Shall I increase it from 10s to 30s. Since locally I was able to complete the entire job in 10s.

pavish:
@Cronus1007 We could actually only add this fixture:

```
@pytest.fixture
def page(page):
    page.set_default_navigation_timeout(30000)
    yeild page
```

This way we don't have to modify usages of the page fixture everywhere.

Cronus1007:
@pavish I have done the required changes.

seancolsen (approved):
This looks good. 

@pavish thank you for suggesting a clean solution here that only requires minimal changes.

I tested this locally and reduced the timeout setting to 100ms to verify that it is indeed taking effect. The default of 30s seems fine. The large timeout does have a downside of slowing down the process of confirming that a test _fails_ during development, but I think that's okay.

I added a commit which fixed a misspelling of `yeild` to `yield`. This misspelling seems to have likely originated in the code that Pavish suggested. A misspelling like this seems perfectly reasonable in the context of a code suggestion comment. However @Cronus1007 the misspelling in your committed code does make me wonder if you tested the changes locally before committing and pushing them. I can't imagine you would have been able to run those changes locally with success. That's fine if the PR is still in draft state and not yet ready for re-review. But your latest comment communicated to me that you consider this code ready for re-review. For your future PRs, the review process would be easier for us if you could put in the effort ahead of time to make sure the code is working locally for you.

Cronus1007:
> But your latest comment communicated to me that you consider this code ready for re-review. For your future PRs, the review process would be easier for us if you could put in the effort ahead of time to make sure the code is working locally for you.

@seancolsen Sure I would keep it in mind. Sorry for the inconvenince caused.

Issue #930, opened 2022-01-03
Add context menu to set cell value to NULL


seancolsen:
## Current behavior

The user has no way to set a cell value to NULL.

## Desired behavior

1. Right click within a cell (or click-hold on a touch device).
1. The browser's native context menu is suppressed.
1. The cell is selected.
1. A Mathesar-specific context menu appears.
1. An item exists within the context menu labeled "Set to NULL".
1. The menu entry is disabled if any of the following conditions are met:
    - The value is already NULL.
    - The column does not allow NULL.
1. After the user clicks on the menu entry, the context menu disappears and Mathesar exhibits the same loading, success, and error behavior as when updating a value within a cell via typing new text and pressing Enter.

## Notes

- `faBackspace`, or `faTimes` would probably be acceptable icons to use in the menu.

## Additional context

- See [NULL Value Specs, scenario 3a](https://wiki.mathesar.org/design/specs/specs-null-values#steps-for-3a)

## See also

- #936

## Status

- Blocked by #742, especially because we should rely on `DropdownMenu` functionality to implement the disabled feature.

seancolsen:
@ghislaineguerin the specs require the context menu entry to be "Set as NULL". That wording seems a bit awkward. Can we use "Set to NULL" instead?

kgodey:
@seancolsen I think it's fine to use "Set to `NULL`".

seancolsen:
Thanks @kgodey I've updated the issue description to reflect that adjustment.

PR #1213, opened 2022-03-22
Add ContextMenu component

seancolsen:
Related to #930

## Summary

- This PR adds a `ContextMenu` component to our component library.
- No changes to Mathesar.

## Rationale

- I need this component to implement #930.

## Notes

- In `clickOffBounds.ts`, I changed the event from `click` to `pointerdown` so that the event is triggered for context menu interactions as well. I did this to ensure that only one context menu would be open at a time, but it also changes the behavior of things like Dropdown and Select so that they close when the user right-clicks somewhere else. To me that seems slightly better. The `pointerdown` event is part of the [Pointer Events API](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) which has [decent support, but not perfect](https://caniuse.com/pointer). Some older versions of Safari don't support it. Personally, I think we're probably fine here, but if you have concerns about this approach, I can add multiple event listeners instead.

## Demo

Peek 2022-03-22 10-09

## Usage

1. Add a context menu to a `div` like this

    ```svelte
    <div>
      <p>Other stuff in this div</p>
      <ContextMenu>
        <MenuItem>Lorem Ipsum</MenuItem>
        <MenuItem>Dolor Emit</MenuItem>
      </ContextMenu>
    </div>
    ```

1. Clicking anywhere within the `div` will open the menu.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/common/actions/clickOffBounds.ts
+++ b/mathesar_ui/src/component-library/common/actions/clickOffBounds.ts
@@ -25,7 +25,20 @@ export default function clickOffBounds(
     }
   }
 
-  document.body.addEventListener('pointerdown', outOfBoundsListener, true);
+  /**
+   * When the browser supports pointer events, we use the pointerdown event
+   * which is fired for all mouse buttons and touches. However, older Safari
+   * versions don't have pointer events, so we fallback to mouse events. Touches
+   * should fire a mousedown event too.
+   */
+  const events =
+    'onpointerdown' in document.body
+      ? ['pointerdown']
+      : ['mousedown', 'contextmenu'];
+
+  events.forEach((event) => {
+    document.body.addEventListener(event, outOfBoundsListener, true);
+  });
 
   function update(opts: Options) {
     callback = opts.callback;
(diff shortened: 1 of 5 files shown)


pavish (changes requested):
@seancolsen Nice work! The component looks good to me.

I have a couple comments:
* Older versions of Safari do not support pointer events as you've mentioned on the PR description (Safari is the new Internet Explorer.. sigh). These older versions are only 2 years old (2019) and are still in considerable use. I'd suggest writing a mini polyfill, where we check if `pointerdown` is present and use it, and if not, we use `mousedown` (which is also supported by all touch devices, although certain devices may behave inconsistently).
* The styles in `ContextMenu` component should be marked global.

seancolsen:
@pavish ready for re-review

> writing a mini polyfill, where we check if pointerdown is present and use it, and if not, we use mousedown

Done in c7c83a2dea8f33474db0aa60628262840953c33f

> The styles in ContextMenu component should be marked global.

Addressed in 449a2440d7c0679b036537117ac2409915328004 by removing that CSS since it was actually redundant. I meant to take that out earlier. Thanks for catching that.

pavish (approved):
@seancolsen Looks good to me!

PR #1240, opened 2022-03-26
Merge Select and SimpleSelect

### LINKED ISSUE #1099: Refactor Select and SimpleSelect components, merging them into one

pavish:
Fixes #1099 

**This PR:**
* Merges Select and SimpleSelect

**Additional decisions that differ/not-included in issue description:**
* Name of `options` prop retained instead of `values` since options is more consistent with dom convention.
* `labelKey` is retained. The function `getLabel` checks using `labelKey` instead of checking with a static label property.
* `onChange` callback is not included. It is better to follow svelte's convention of dispatching events on lower order components. `onChange` is retained on the top level component `SelectSortDirection`.
* `change` event now dispatches only the value instead of an object comprising value.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/component-library/select/Select.svelte
```diff
@@ -190,7 +181,11 @@
   on:open={setSelectedItem}
 >
   <svelte:fragment slot="trigger">
-    {value?.[labelKey]}
+    {#if typeof value !== 'undefined'}
+      {getLabel(value, labelKey)}
+    {:else}
+      No option selected
```
seancolsen, line 187: Displaying "No option selected" here is a little weird to me. But I don't think it's worth addressing now. We can fix this if/when it becomes a problem. I'd prefer to show it as blank.

seancolsen (approved):
> Name of options prop retained instead of values since options is more consistent with dom convention.

- Sounds good

> labelKey is retained. The function getLabel checks using labelKey instead of checking with a static label property.

- Good idea!

> onChange callback is not included. It is better to follow svelte's convention of dispatching events on lower order components. onChange is retained on the top level component SelectSortDirection.

- Yeah. I hear this. That's fine. I actually prefer using props over events, especially when the callback has an argument -- though I realize this opinion goes against the conventions of Svelte (and Vue). I don't like events because: (A) you can't `await` an async event handler (e.g. see our SpinnerButton component for a use-case), (B) making the event handler arguments type-safe requires using `$$Events` which is more boilerplate and still experimental, and (C) using props makes it easier for other devs to see the parent-child data flow up front (instead of requiring them to search the component for `dispatch`). The events functionality in Vue and Svelte is nearly identical, but I prefer the way React does it. However this is all just food for thought. We can stick with using events because that's certainly what other Svelte devs will expect. The only thing I like about events is the succinct syntax for forwarding them from child to parent. I'd be curious to hear if there are other things you actually like about events though.

> change event now dispatches only the value instead of an object comprising value.

- Good.

pavish:
>  I actually prefer using props over events, especially when the callback has an argument
> The events functionality in Vue and Svelte is nearly identical, but I prefer the way React does it.

I too prefer the React way, but I think the balance I previously established and am still following is:
* Use events for anything the child component does not have to worry about. Here the child component does not have to worry about what the parent does for on:change.
* Use props for anything the child has to worry about. For SpinnerButton, the child needs to know whether the parent has finished processing the callback.

I'd like to apply these especially on lower order components, since it is essential we don't deviate much from the conventions in the Svelte world. I don't worry about the higher order components, it's the reason why I didn't modify SelectSortDirection.

PR #1257, opened 2022-03-31
Cursor position in FormattedInput

### LINKED ISSUE #1248: Retain cursor position when changing text within FormattedInput
## Current behavior

1. In Storybook, open the NumberInput story.
1. Set the input to contain text `134`.
1. place the cursor between `1` and `3`.
1. Type `2`.
1. Expect the cursor to be between `2` and `3`.
1. Observe the cursor to be after `4`.

    This happens because the value of the input gets reformatted to add the comma.

## Desired behavior

- After modifying the input (via inserting text or deleting text), the cursor position generally meets users' expectations.

- There may be some edge cases where users expectations are not crystal clear. That's okay. We just want to offer some improvement that meets the expectations under most circumstances. For example, let's say we begin with `12,345` and the cursor between `1` and `2`. Then we press forwards-delete. We end up with the text `1,345`. Where should the cursor go? Between `1` and `,` or between `,` and `3`? Either would be fine and would be an improvement over the current behavior (where the cursor ends up at the very end).

## Implementation

1. In `FormattedInput.svelte`, in the function `handleChildValueChange`, you have access to:

    - the value before re-format (as `userInput`)
    - the cursor position before re-format (as the `cursorPosition` property of the return value from `getOutcomeOfBeforeInputEvent`)
    - the value after reformat (as `parseResult.intermediateDisplay`)

1. Write a pure function that takes input of the three values above and returns the resulting cursor position after the text has changed.

1. I suggest using https://www.npmjs.com/package/fast-diff to diff the text values. Then use the old cursor position to figure out where _in the diff_ the cursor should go, and use the diff to figure out where the cursor should go in the resulting text.

priyang12:
Fixes #1248

**Technical details**
Corrected Cursor Position using Diff package by creating a function.

**Screenshots**

https://user-images.githubusercontent.com/72823974/161086514-23820d1a-fd37-4bf6-a431-d44dc405ca5c.mov

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
@priyang12 Good job on this PR! 

It does exactly what the issue was specced for. There is a race condition I noticed, when you enter a lengthy number with a repeated number in it, eg.,100000000, the cursor position is not placed accurately.

However, before handling this, I am re-thinking the whole input. Here's the relevant discussion: https://github.com/centerofci/mathesar/discussions/1268.

I'm placing the review and merge on hold until that discussion is resolved.

priyang12:
@pavish ok thanks for the update.

seancolsen (approved):
@priyang12 Thanks again for this PR.

For the most part, your work has improved the behavior of the input, so I'm merging this PR. However, as @pavish mentioned, there are some additional edge cases that we'll still need to handle. To track the work of handling those additional edge cases, I've opened #1284.

I pushed a commit to this PR which restructures your code as follows:

- I renamed the function `setCursorToPostion` to `getCursorPositionAfterReformat`. That fixes the spelling error in "position" and makes the name more specific.
- I made the function into a pure function and moved it to a utils file so that it's easier to test with unit tests.
- I changed the argument structure to use named arguments with names that pertain to the purpose of the function instead of the purpose of the component. This increases readability when calling the function and reduces the risk of calling the function with the arguments mixed up.
- I removed the TypeScript type `any` from within the function. We don't need to use it here. We should only use `any` as a last resort.
- I made some other small naming changes within the function to increase readability.
- I wrote a unit test with several test cases that pass. And I added more test cases that fail (with those test cases commented out.)
- I removed the logic which checks to see if the last character is a number. We shouldn't need to do that. The process of setting the cursor position should work regardless of the characters in the text. One important reason is that the `FormattedInput` component is a general purpose component which serves as a building block for other components. So far, we only have `NumberInput`. But we may someday have others, like `PhoneNumberInput` or `DateInput` or `CreditCardNumberInput`. Further, within `NumberInput` specifically, sometimes the last character of the number _isn't_ a digit. For example, we need to gracefully deal with an input of `12.`.

PR #1265, opened 2022-04-05
Rename .d.ts files to .ts

### LINKED ISSUE #1135: Change .d.ts files to .ts files
## Context

- We have an undocumented front end coding standard of placing TypeScript types in `*.d.ts` in certain specific scenarios.
- Shortly after I joined the team I raised a concern with this practice when talking to @pavish. At the time, my concern was only mild. I mentioned that the practice seemed unconventional, but I didn't have any specific problems with it. Pavish wanted to continue with it, so I let it go.
- Recently, this practice has been bugging me more, and it was one source of slowness during my refactoring in #1109. So I'm raising the concern again, this time with stronger arguments.

## Request

- I'd like to stop using `*.d.ts` files, moving all their contents to `*.ts` files instead.

- For situations where we want to export types which are relevant to a specific component, we currently have:

    - `Icon.svelte`
    - `Icon.d.ts`

    I'd like to change the pattern to:

    - `Icon.svelte`
    - `iconUtils.ts`

- For situations where we're re-exporting stuff, we currently have:

    - `index.ts`
    - `types.d.ts`

    I'd like to move everything to `index.ts`.

### More type safety and easier refactoring

It seems to me that TypeScript does not check `*.d.ts` files with the same rigor that it checks `*.ts` files.

Follow me on an example...

1. In `src/stores/table-data/types.d.ts`, change:

    ```ts
    export type { Display } from './display';
    ```

    to

    ```ts
    export type { Display } from './columns';
    ```

    Note that `columns.ts` has no exported `Display` member. TypeScript sees no problem in with this change in `types.d.ts` though! Not good. Code like that in a `*.ts` file would immediately be complaining at me.

1. Running `svelte-check` produces no errors! Really? That can't be. Let's look at `RowCell.svelte` which imports `Display` from `'@mathesar/stores/table-data/types'`.

    All properties of `Display` (e.g. `handleKeyEventsOnActiveCell`) are now typed as `any`. Ugh! And we don't even get linting errors about the `any` types!

    This leads to some bugs during refactoring that can be trickier to track down.

In the above example, `Display` is already exported from `src/stores/table-data/index.ts`. So within `RowCell.svelte`, we can import it with:

```ts
import type { Display } from '@mathesar/stores/table-data';
```

I don't see the use of also exporting it from `src/stores/table-data/types.d.ts`.

### Simplicity

It's simpler to have one TypeScript file type. Fewer decisions to make when adding new files. It's especially simpler when re-exporting stuff because all the re-exports can go in one file instead of two.

### Easier auto-imports

Let's say I want to use `AbstractTypeResponse` (defined in `App.d.ts`). When I let VS Code auto-import that type, it gives:

```ts
import type { AbstractTypeResponse } from '@mathesar/App';
```

But this is problematic because some module loaders will see `'@mathesar/App'` and have difficulty resolving the ambiguity between `'@mathesar/App.svelte'` and `'@mathesar/App.d.ts'`. I've run into this within our codebase before. I can't remember the exact steps I took to reproduce the problem, but I do remember that I had to add `.d` on the end of my import it to get it to work. It may have been a problem loading imports while running Jest or Storybook. That's additional friction that would not be present when importing from `appUtils.ts`.

### Issues with enums

Some `*.d.ts` files contain enums. That's problematic because enums are not strictly types -- but also objects used at runtime. In #1109 I hit this issue when running tests from a `test.ts` files. Jest was unable to find the runtime value for the enum. I had to move `TabularType` out of a `*d.ts` file just to get my test to run. All enums should be moved to `.ts` files. The fact that we even need to think about this demonstrates the additional overhead that comes with our pattern of using `*.d.ts` files.

### Consistency with TypeScript community norms

For example...

- https://www.reddit.com/r/typescript/comments/t23p7o/confused_on_use_of_dts_file_in_reacttypescript_app/

    - > If the types are for code you control that's in your application's source directories, they should be defined in regular TypeScript files and consumed via the module system.

    - > Declaration files... are for telling TypeScript about code that will exist at runtime that it cannot discover through your own code.

- https://www.reddit.com/r/typescript/comments/r69jmi/whats_the_difference_between_ts_and_dts_file/

    - > In your case, you probably just want to write everything in .ts files

    - > *.d.ts is typically used to add types to untyped JavaScript

- https://www.reddit.com/r/typescript/comments/rirjiw/when_do_you_use_dts_files_and_when_do_you_use_ts/

    - > The .d.ts is a file that let's typescript know the types for a javascript file so you don't have to annotate someone else's code to use the library.

- https://www.reddit.com/r/typescript/comments/rz62fa/dts_files_vs_ts_can_someone_clear_this_up_for_me/?sort=top

    - > > ...When would I actually write a .d.ts file?...
        >
        > ...when you are typing something that's written in Javascript

I couldn't find a single case of someone arguing to use `*.d.ts` file like we're using them.

seancolsen:
Fixes #1135

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish (approved):
@seancolsen Looks good to me.

I can see the naming convention for the type definition files are:
* Name of `component/systems/class/utils` + `Types`.
* Name is either PascalCased or camelCased based on the Component or util's naming convention.

I'd prefer a single naming convention for type definition files (eg., always use camelCased file names for them). But I don't have a strong opinion on this, I'm just laying out the thought.

PR #1266, opened 2022-04-05
One SQLAlchemy operation per function

### LINKED ISSUE #1096: Reduce db functions to one SQLAlchemy operation per function
## Problem
We have decided to use `attnums` when passing a column reference as arguments but most of the SQLAlchemy operations use column name, so we need to fetch the column name before an SQLAlchemy operation, this can lead to referencing a wrong column in case there are multiple SQLAlchemny operations within a single function and one of those happens to change a column name.

## Proposed solution
Each function should use only one SQLAlchemy operation so a column name altering operation won't have any effect on other operations.

Anish9901:
Related To #1096

This PR reduces each function to have only one SQLAlchemy operation.

**Technical details**

Currently, in the function `_batch_alter_table_columns` in `db/columns/operations/alter.py`, there are two SQLAlchemy operations. This might cause referencing a wrong column. So this PR  splits the function `_batch_alter_table_columns` to separate functions with only one SQLAlchemy operation.
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
Great work on this. 

This could be merged in its current state, but I'm not comfortable claiming that it "Fixes #1096 " without a test to verify that. Writing such a test is likely to be pretty technical. If you feel up to it, I encourage you to write such a test. Otherwise, please change the description to say "Related to #1096 " so that the issue stays open until we write the appropriate test.

Anish9901:
@mathemancer I'll try to write some test cases.

silentninja:
@mathemancer Can you give a rough idea of how the test should work or be implemented. I can only think of a test case that is fairly complex which involves mocking and inferring mocked calls.

silentninja:
I agree with @mathemancer comment. Other than that the PR looks good to me

mathemancer:
> @mathemancer Can you give a rough idea of how the test should work or be implemented. I can only think of a test case that is fairly complex which involves mocking and inferring mocked calls.

I'd agree on the single test case. I suppose it could be method by method, but I'd want those individual "cases" to be generated to ensure we don't forget to make a new one when we make a new method. The way I imagine it would be to use the `ast` module to parse each source file (or maybe you could go for top-level imports; not sure if `ast` will go into the files imported or not) for function definitions and imports. Catch any import from `sqlalchemy.*` or `alembic.*`, and then look for `Call` objects with one of the imported names within function definitions, making sure there's at most one per `def`. 

As I said, this will be quite technical, and even the version I described doesn't take branching within functions into account; it's fine if some function has more than one SQLAlchemy call, as long as they're in mutually-exclusive logical branches within the function.

The goal would be for the test to fail if any function calls more than one unwrapped SQLAlchemy method, and as a bonus, would list the offending functions.

Anish9901:
@mathemancer I have updated the PR description as "Related To" we can work on the tests later.

mathemancer (approved):
This looks good with the modified description.

PR #1275, opened 2022-04-08
Referenced column deleted-issue 987

### LINKED ISSUE #987: Filter/Sort/Group are broken when referencing a deleted column
## Reproduce

1. Sort a table by one column.
1. Delete that column.
1. Expect the column to be removed from the sort criteria.
1. Observe the sort criteria to still contain the column.
1. Refresh the table.
1. Expect the records to load.
1. Observe an error 400 from the records API when attempting to fetch the records because the sort criteria is now invalid.

## Implementation

- The `Columns` store should dispatch an `EventListener` event whenever a column is deleted. `TabularData` should listen to that event and react by altering `Meta` to remove all references to the deleted column from within the `filtering`, `sorting`, and `grouping` properties on the `Meta` class.
- We should provide some E2E test coverage along with this fix for this bug.

Jinxiao0302:
Since the filtering funtion is still in progress, I modified the deleteColumn() function so that it has no more reference gregaring sorting and grouping relationships with the deleted columns.

Fixes #987 

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@Jinxiao0302 It looks like you've added a `zip` file to the repository, I don't see any changes to the code.

Jinxiao0302:
> @Jinxiao0302 It looks like you've added a `zip` file to the repository, I don't see any changes to the code.

Yeah, I have now removed the zip file and directly made change to the modified part. It went wrong when uploading the whole file systems. Next time I will try the git command instead.

seancolsen (changes requested):
Thanks for this PR @Jinxiao0302. 

You have the right idea with your changes, but to follow some of our patterns, we'll need to move the code to a different place before merging.

To see an example of what we need, look at the `Columns` class, in the `add` method. There, we have:

```ts
await this.dispatch('columnAdded', column);
```

Then in the constructor for `TabularData` we have:

```ts
this.columnsDataStore.on('columnAdded', () => this.recordsData.fetch());
```

This code makes it so that whenever the `add` method is called, the required side-effects are executed. That's nice because it shifts responsibility away from the view layer and into the store layer. And it means we don't accidentally miss some of the side-effects if we have multiple UI code points which perform the same store mutations.

We should follow this pattern here too. That means picking a name for the event. I'd suggest `'columnDeleted'`. Then dispatching that event from within `Columns`. Then responding to that event from within `TabularData`, and within that event listener, _that's_ where we'll put the code to perform the side effects which mutate other stores.

Do you want to give that a shot? You can keep this PR and add another commit which makes those changes.

Jinxiao0302:
> Thanks for this PR @Jinxiao0302.
> 
> You have the right idea with your changes, but to follow some of our patterns, we'll need to move the code to a different place before merging.
> 
> To see an example of what we need, look at the `Columns` class, in the `add` method. There, we have:
> 
> ```ts
> await this.dispatch('columnAdded', column);
> ```
> 
> Then in the constructor for `TabularData` we have:
> 
> ```ts
> this.columnsDataStore.on('columnAdded', () => this.recordsData.fetch());
> ```
> 
> This code makes it so that whenever the `add` method is called, the required side-effects are executed. That's nice because it shifts responsibility away from the view layer and into the store layer. And it means we don't accidentally miss some of the side-effects if we have multiple UI code points which perform the same store mutations.
> 
> We should follow this pattern here too. That means picking a name for the event. I'd suggest `'columnDeleted'`. Then dispatching that event from within `Columns`. Then responding to that event from within `TabularData`, and within that event listener, _that's_ where we'll put the code to perform the side effects which mutate other stores.
> 
> Do you want to give that a shot? You can keep this PR and add another commit which makes those changes.

Thanks so much for your review, @seancolsen ! I have followed your guidance and moved my code into the constructor of `TabularData` instead. Also I have tested the functionality and it is still working. I'd also glad to see if there is other enhance suggestions regarding the design pattern.

seancolsen (approved):
Nice! Thanks for the adjustment. I added some further tweaks that you can see in my last commit. Plus I reverted changes to `package-lock.json`. You'll see changes to that file sometimes after restarting docker. It's sort of annoying. Always wipe away those changes if you see them. 

There were some type errors too. I suppressed them by using the TypeScript `as` keyword, which we generally want to avoid because we lose type safety when doing so. I opened #1302 with a solution proposed that will allow us to avoid those `as` keywords here once we add some more complex generics to the `EventHandler` class.

PR #1283, opened 2022-04-13
Configure scroll state reset and preserving

### LINKED ISSUE #632: Preserve table scroll across live table updates
## Problem
At the moment, while browsing a table in the UI, when a table update is performed (for example a column type is changed or a sort is added) the scroll state of the table view is reset to top left. In terms of UX that might not be desirable or expected.

After some table updates the vertical scroll might lose meaning (for example after changing row sort) and therefore it could be argued that resetting vertical scroll automatically might be desirable at least in some cases. However, I think that resetting of horizontal sort is always undesirable.

## Proposed solution
Persist scroll state across table updates; don't perform automatic scroll resets. I've not looked into what this amounts to technically.

Jyuart:
Fixes #632 

The vertical scroll state preserves if a user:
- Switches between tabs (either on the actions panel or left pane)
- Changes table's or columns' properties

It's being reset when a user:
- Applies sorting, grouping, filtering (not at the moment, cause filtering is commented out in the UI)
- Changes table's page or its page size

## Checklist
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
Thanks for this PR! I'll probably get to this review next week. In the mean time I wanted to offer one very quick comment in reaction to a quick glance at the diff (without yet having gotten deeper into the inner-workings of your code)...

I see a lot of `subscribe` calls in the Svelte code. Any time we're subscribing, we also need to handle the unsubscribe event too (by calling the function that gets returned when subscribing). This is a pain. It's hard to know when to unsubscribe. It's verbose. As such, I try to avoid calling `subscribe` unless there's no alternative. You won't see many `subscribe` calls in our code. You might be able to refactor those `subscribe` calls into reactive statements. Without getting deeper into the code I'm not sure. But while you're waiting for me to get the time to offer a comprehensive review, it might be worth trying to eliminate all `subscribe` calls from your diff or explain why they're necessary (and add code to handle the unsubscriptions).

Jyuart:
> Thanks for this PR! I'll probably get to this review next week. In the mean time I wanted to offer one very quick comment in reaction to a quick glance at the diff (without yet having gotten deeper into the inner-workings of your code)...
> 
> I see a lot of `subscribe` calls in the Svelte code. Any time we're subscribing, we also need to handle the unsubscribe event too (by calling the function that gets returned when subscribing). This is a pain. It's hard to know when to unsubscribe. It's verbose. As such, I try to avoid calling `subscribe` unless there's no alternative. You won't see many `subscribe` calls in our code. You might be able to refactor those `subscribe` calls into reactive statements. Without getting deeper into the code I'm not sure. But while you're waiting for me to get the time to offer a comprehensive review, it might be worth trying to eliminate all `subscribe` calls from your diff or explain why they're necessary (and add code to handle the unsubscriptions).

Thanks for the quick response. I have some ideas about how to do it differently. I'll start testing them on Monday and update the code once I find something worthwhile

seancolsen (approved):
Looks good. Thanks for this @Jyuart and thanks for adding the tests too!

It's great you figured out how to eliminate the `subscribe` calls. I added a commit which further simplifies the reactive logic, making the code a bit cleaner. If you have any concerns with my changes, let me know and we can address it with a follow-up PR.

seancolsen:
Well shoot. The CI picked up a linting issue with my changes. In troubleshooting that, I identified an issue with my changes. And in troubleshooting _that_ I identified a separate issue with the changes before my commit. This PR needs more work before merge. I'm actively working on figuring out what it needs. I wish GitHub offered a way for me to invalidate my approving review. Oh well. Will report back soon.

kgodey:
@seancolsen You can invalidate ("dismiss") your review, here's a screenshot:
Screen Shot 2022-04-18 at 9 37 25 AM

seancolsen (changes requested):
Ok apologies for jumping the gun with my code cleanup commit and approving review. I've reverted my commit and rescinded my review.

As @Jyuart [outlined](https://github.com/centerofci/mathesar/issues/632#issuecomment-1092975173), the ticket requires the following independent changes to the persistence of vertical scroll position:

> 1. When switching between tabs scroll should be persisted (currently it's not)
> 2. _(omitted for brevity)_
> 3. When adding filtering, sorting, or grouping it should not be persisted (currently it is)

From what I can tell, this PR satisfies requirement `3` successfully. However, I've identified the following problem with satisfying requirement `1`. When I open two tables, the _second_ one has its vertical scroll position preserved when switching tabs, but the first does not. Here's a demo:

https://user-images.githubusercontent.com/42411/163831005-740a8929-b776-46c8-8a00-07c130a0faa4.mp4

Also, if the second table has a sorting condition applied, then its vertical scroll position will not be preserved.

If I remove all the sorting/filtering/grouping/pagination code from `Body.svelte`, then requirement `1` seems to be satisfied in those two problematic scenarios above. (But requirement `3` is no longer satisfied.)

The code you've added to address requirement `1` is straightforward and works well. The code you've added to address requirement `3` is less straightforward due of some of the underlying architecture of the tabs system and virtual list system. Consequently, that code for `3` is breaking `1` in certain scenarios. Without spending more time on this problem, I'm not able to identify a simple approach that will satisfy both `1` and `3`. It's possible that a clean solution to both of these requirements will require some deeper re-architecting.

@Jyuart if you want to remove that code for `3`, I would merge this PR and open a new ticket to track `3`. Or if you prefer to keep working on `3` within this PR I'm happy to continuing reviewing further changes you make towards that end. I'll leave it to you.

@pavish Would you care to take a quick look at this and see if you have any ideas. We can chat on a sync call if needed to speed things up. And if you think it'll take too long to get your head into the problem, I think it's fine to ignore this. I figured since the architecture of the tabs system and virtual list system are more familiar to you, you might be able to identify an approach more quickly than me or @Jyuart.

As I see it, one of the fundamental problems is that, within `Body.svelte` it's difficult to distinguish between: (A) a change in `sorting` that originates from the user switching tabs (imagine two tabs with different sorting params), and (B) a change in `sorting` that originates from the user manipulating the sorting params. I can imagine several different pathways towards dealing with this problem, but none of them seem very simple.

@Jyuart It's also worth mentioning that the core team is currently at the start of adopting a new workflow wherein we prioritize issues within the current feature milestone for a highly ambitions three-week sprint, at the expense of other work. We are experimenting with news way so accelerate our pace towards releasing the alpha version of Mathesar. This PR, while useful, is unfortunately not part of our current sprint, so it's possible we'll need to sideline it for a bit while we focus on finishing data types and foreign keys. That's why I can't spend more time today to give this the attention it deserves.

Jyuart:
> 

Wow, thanks for such a detailed reply.
And sorry for the last-minute change that broke everything. Initially, I was using `beforeUpdate` instead of `onMount` and it worked nicely for me. Here's a demo:

https://user-images.githubusercontent.com/20342130/163838292-7b6fca12-7fbd-480d-bfde-dc4e95b978ec.mp4

Currently, all the scenarios work as expected, including changing the page and the page size number. I've also added a `console.log()` statement to the demo to show how many calls to the reactive statement are done on each action.

> As I see it, one of the fundamental problems is that, within Body.svelte it's difficult to distinguish between: (A) a change in sorting that originates from the user switching tabs (imagine two tabs with different sorting params), and (B) a change in sorting that originates from the user manipulating the sorting params

For me, the most difficult part was differentiating between the initial set for sorting/paging/filtering values and the subsequent updates initiated by the user. I've dealt with it saving the initial state each time before updating data (probably, that is the reason it stopped working with `onMount`).

Please take a look at the updated code when you have some time and tell me what you think. If there are still other issues in there that I didn't spot, I'd be glad to fix them or separate the PR.

> so it's possible we'll need to sideline it for a bit while we focus on finishing data types and foreign keys

Absolutely ok. I was thinking about what issue to take next, maybe I'll be able to pick something that will be both useful and prioritized.

seancolsen (approved):
@Jyuart Nice! This works now! I've played with your lasted cod a bit locally and everything seems to be meeting the expectations laid out in the ticket. Good thinking with `beforeUpdate`. And thanks for the quick response. Very well done on this PR!

Issue #1228, opened 2022-03-25
Internal server error after column type is converted from text to boolean with display options 


pavish:
## To Reproduce
* Create a new text column
* Patch the column from text to boolean with display options.
  ```javascript
  {
    type: 'boolean',
    type_options: {},
    display_options: {
      input: 'dropdown',
      custom_labels: {
        TRUE: 'TRUE'
        FALSE: 'FALSE'
      }
    }
  }
  ```
* Notice that the request succeeds.
* Send a subsequent request to get all columns of the table. Notice the 500 error.
* Try loading the frontend, a debug page with the error message appears.
* **This does not happen always, happens intermittently and I end up having to drop the table directly in the db inorder to get things working again.**

## Error message:
```
[{"code":4999,"message":"\"Got KeyError when attempting to get a value for field `input` on serializer `BooleanDisplayOptionSerializer`.\\nThe serializer field might be named incorrectly and not match any attribute or key on the `dict` instance.\\nOriginal exception text was: 'input'.\"","field":null,"detail":null}]
```

## Error log:
```
Environment:


Request Method: GET
Request URL: http://localhost:8000/mathesar_tables/1/?t=W1tdLG51bGxd

Django Version: 3.1.14
Python Version: 3.9.8
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'django_filters',
 'django_property_filter',
 'mathesar']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/rest_framework/fields.py", line 457, in get_attribute
    return get_attribute(instance, self.source_attrs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/fields.py", line 95, in get_attribute
    instance = instance[attr]

During handling of the above exception ('input'), another exception occurred:
[...]

Issue #1276, opened 2022-04-11
500s when patching column to a different type


pavish:
## Description
When patching a column to a different type, having `display_options` or `type_options` in the request results in a 500 if the new type does not allow them.

### Scenario 1:
* Create a text column
* Patch the column to URI type with `null` values for display and type options:
   ```
  {
    "type": "MATHESAR_TYPES.URI",
    "display_options": null,
    "type_options": null
  }
   ```
* Notice the following response with status code 500:
  ```
  [
    {
        "code": 4999,
        "message": "'NoneType' object is not iterable",
        "field": null,
        "detail": null
    }
  ]
  ```

### Scenario 2:
* Create a text column
* Patch the column to URI type with an empty object: `{}` for display and type:
   ```
  {
    "type": "MATHESAR_TYPES.URI",
    "display_options": {},
    "type_options": {}
  }
   ```
* Notice the following response with status code 500:
  ```
  [
    {
        "code": 4999,
        "message": "Cannot find a matching serializer for the specified type uri",
        "field": null,
        "detail": null
    }
  ]
  ```

## Expected behavior
* The response should not result in 500 for both scenarios.
* For Scenario 1, the request should succeed.
* For Scenario 2, the it would be easier for the client if the request succeeds. If that's against backend convention, the response should result in a 400.

kgodey:
> For Scenario 2, the it would be easier for the client if the request succeeds. If that's against backend convention, the response should result in a 400.

If `display_options` is set to an empty object, we should remove all display options.

Anish9901:
Can I work on this?

silentninja:
@Anish9901 I have assigned the issue to you. This issue blocks a lot of frontend work, so a core contributor might take it if it takes too long to be completed(more than 2-3 days)

silentninja:
Thanks! @Anish9901

kgodey:
@silentninja I think if the frontend sends a request to remove all display options, the backend should treat it as a request to reset the display options to their default state (i.e. the user is removing all customizations). We should not let the display options end up in a unusable state.

Alternatively, you can validate required display options, but the first option seems more flexible.

pavish:
@silentninja  Here are my thoughts:

1. I think the default state should be `None`. Like you mentioned, an empty object is set by the client, for which currently we do not have any validation. But we might possibly have it in the future.
[...]

PR #1297, opened 2022-04-19
Fix DisplayOptions Serializer to validate display options when updating a column

silentninja:
Related to #1228

When a column is partially updated with an invalid display option, it does not get validated and the request succeeds due to weird behaviour with the DRF serializer which treats nested data as a partial update too, but we completely replace the display options object when saving it, so we end up with a scenario where a required field could be missing. So when the same object is requested by the API, the serializer throws up an error when looking up for the required field in the display options JSON object. This PR fixes it by validating display options during a partial update

## Technical Details
During a partial update of a parent serializer, drf calls the nested serializer with `partial` flag even if we haven't explicitly specified it. We consider update to display options as a full update but since DRF takes a partial flag from the parent serializer, the nested display options serializer gets treated as a partial update,  so the serializer becomes valid even if a required field is missing. A mixin was used to fix the behaviour for the individual display options serializer but it wasn't applied to the mapping serializer. This PR adds the mixin that fixes the behaviour of the DRF serializer to the mapping serializer. 

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
@silentninja 

>  I am assuming you would be sending empty object even if the column type has the required display options

I do not understand this. While patching a column, I'd send an empty object if the user has not set any display options even if the column has display options associated with it. See [Kriti's comment](https://github.com/centerofci/mathesar/issues/1276#issuecomment-1094989622) here.

silentninja:
> Also the sample requests mentioned in https://github.com/centerofci/mathesar/issues/1228 contain values for all the required fields.

Sorry got confused with issue #1276 scenario 2, please ignore that message. But this PR should solve the invalid state, so now it has to be `None` or a valid object at the time of validation. I have added a comment to the issue on how to deal with empty object.

mathemancer (changes requested):
While I do think this change is needed, and the test looks good, I don't understand how this PR fixes #1228 . Can you add some explanation about that to the description for posterity?

silentninja:
@mathemancer #1228 happens due to a bugs that can lead to invalid data being saved, this PR fixes one of the bugs that can cause the error. I updated the description to mention it is related to #1228 rather being a complete fix. I have also updated the description to mention the cause of the issue and how this PR aims to solve it.

mathemancer (approved):
Okay, thanks for the extra explanation.

PR #1307, opened 2022-04-23
Batch update columns using id instead of column_index

### LINKED ISSUE #1167: Batch updating column using `/tables/` api should reference column using `id` instead of `column_index`
## Problem
Batch updating Columns using the `columns` field in the `patch` method of the `/tables/{id}/` does not use `id` of the column rather the column to update is identified based on the position of the column objects. This is not ideal due to the following reasons:
- Similar to the `column_index` which is [deprecated in favour](https://github.com/centerofci/mathesar/issues/839) of `id`.
- Column position could have changed resulting in updating a wrong column

## Proposed solution
The following changes have to be made:

-`PATCH` method of the `/tables/{id}` api which has the following structure 

```
{
    "columns": [
        {
            "name": "id",
            "type": "INTEGER"
        },
        {
            "name": "Center",
            "type": "TEXT"
        }
    ]
}
```

should be replaced with 

```
{
    "columns": [
        {
            "id": 1, // Use column id to reference a column, rather than assuming the column based on its index.
            "name": "id",
            "type": "INTEGER"
        },
        {
            "id": 2,
            "name": "Center",
            "type": "TEXT"
        }
    ]
}
```

- The column id's have to be converted into `attnums` which can be fetched from the column object before sending the column data from the service layer to the database layer(before calling any function from the `db` module)

Anish9901:
Fixes #1167 

Currently, batch updating columns was done based on the position of the column object which could be unreliable, to fix this, we make use of the `id` field instead of the `column_index` for referencing a column.

**Technical details**

### Updated request for renaming or updating the type of columns:
```json
{
    "columns": [
        {
            "id": 1,
            "name": "id",
            "type": "INTEGER"
        },
        {
            "id": 2,
            "name": "Center",
            "type": "TEXT"
        }
    ]
}
```
### Updated request for dropping columns:
```json
{
    "columns": [
        {
            "id": 1,
            "name": "id",
            "type": "INTEGER"
        },
        {
            "id": 2
        }
    ]
}
```
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
@silentninja Can you please review this PR? I will work on refactoring failing tests once you approve the changes in the PR.

mathesar/utils/models.py
```diff
@@ -43,3 +44,15 @@ def update_sa_schema(schema, validated_data):
...
     alter_schema(schema.name, schema._sa_engine, validated_data)
+
+
+def update_id_to_attnum(table, validated_data):
+    if 'columns' in validated_data:
+        data = validated_data.get('columns')
+        for column_data in data:
+            col_id = column_data.get('id', None)
+            if col_id is not None:
+                attnum = table.columns.get(id=col_id).attnum
```
silentninja, line 55: It would be nice if column id's can be fetched from the database in a single query instead of having to fetch from the db for each column.

silentninja (changes requested):
Nice work @Anish9901. I have few requests though. Also update the documentation to mention that the api has added an additional field `id`.

silentninja:
@pavish or @seancolsen This PR brings in a breaking change and might need additional work on the frontend

seancolsen:
I don't think we need any front end changes here. I'm reasonably certain that the front end never sends a `columns` param in a PATCH request to the tables endpoint. After some brief searching of our codebase, the closest thing I found was [here](https://github.com/centerofci/mathesar/blob/e2fedadfc1a7fe2c319e1cbf4f2cc62637779b88/mathesar_ui/src/stores/tables.ts#L142) where we update the name of a table. The `tables.ts` file seems to be the only place where we use that endpoint. And that code location seems to be the only PATCH request in that file.

pavish:
@seancolsen @silentninja 

It is actually used in one place on the frontend: while saving the preview table. We will need frontend changes in this PR.

@seancolsen You can find the usage here: https://github.com/centerofci/mathesar/blob/e2fedadfc1a7fe2c319e1cbf4f2cc62637779b88/mathesar_ui/src/sections/import-data/importUtils.ts#L360

pavish:
Considering that the frontend changes are supposed to be part of this PR, @Anish9901 would you be upto handling it yourself?

The core team might take a while to get to this PR.

Anish9901:
@pavish I am not currently unfamiliar with JavaScript and Svelte so I won't be able to contribute to the frontend part. Maybe we can create a separate issue for the frontend work required for this PR.

pavish:
@Anish9901 No worries. We'll handle the frontend changes in a day or two, but it has to be in this PR since this breaks file import at the moment. This PR can be placed on hold till then, since I also see some requested changes and failing backend tests, which might take you some time to fix.

Anish9901:
Sure @pavish, I'll make the required changes.

Anish9901:
@silentninja Can you please point me to the docs that need to be updated?

Anish9901:
@kgodey @silentninja can you please change the label of this PR from waiting to review?

Anish9901:
@silentninja If I am not wrong the import preview API spec needs change right?

silentninja:
> If I am not wrong the import preview API spec needs change right?

Yes, right now the preview data contains `column name`, we would need to extend it to include the column `id`

Anish9901:
@silentninja I have updated the docs can you please take a look?

mathesar/tests/api/test_table_api.py
```diff
@@ -1025,6 +1025,14 @@ def _get_patents_column_data():
     }]
 
 
+def get_column_data_with_id(table, column_data):
```
silentninja, line 1028: This fixture can also be converted to accept `table_oid` and return column data  with id

silentninja (changes requested):
Overall looks good, I have few minor changes that needs to be addressed.

silentninja (approved):

seancolsen:
UI fixes are in #1390

PR #1333, opened 2022-04-29
Set `allowFloat` prop for number inputs

seancolsen:
## Before

- All Number cells allow float, regardless of type options.
- All default value inputs for Number columns prohibit float, regardless of type options.

## After

- Cells and default value inputs both utilize the column type options to allow or prohibit entry of floats accordingly.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/components/cell/data-types/number.ts
```diff
@@ -21,28 +14,50 @@ const localeMap = new Map<NumberFormat, string>([
...
+    case 'DOUBLE PRECISION':
+    case 'REAL':
+      return false;
+    case 'DECIMAL':
+    case 'NUMERIC':
+      return (column.type_options?.scale ?? Infinity) === 0;
+    default:
+      return false;
+  }
+}
```
pavish, line 35: This hardcodes DB types on the frontend specific layer. A pattern I've been establishing is to avoid this kind of logic which makes it hard to separate frontend & backend concerns in the long run.

Instead of specifying this here, please refer `multiLine` within `cell.conditionalConfig` in `type-configs/text.ts`. Here, the db specific options and logic are present only in the stubbing layer.

This may seem like unnecessary additional work considering our sprint goals, but the large part of the work is already done and is in place, and this is a minor change. Moving it to the stubbing layer also reduces the total number of lines in code.

You could probably have a `conditionalConfig` called `allowFloat: true | false | 'auto'` in the stub. In 'auto' mode, it checks and sets value based on the scale, in the frontend specific layer.

pavish (changes requested):
@seancolsen Looks good. I have a single comment, which I hope should be a minor change.

mathesar_ui/src/components/cell/data-types/number.ts
seancolsen: @pavish do the changes in ef3a5112e0a2ca56694a85f09c640a78bedc2d09 align with your intended goals?

seancolsen: For readability's sake, I chose the name `floatAllowanceStrategy` instead of `allowFloat` because their values do not map 1-to-1 to each other.

seancolsen: Oh, and I missed a small thing that I fixed in 735eb1e148603e5034f64cc058371153c342e24b too

pavish (approved):
@seancolsen Looks good to me!

PR #1350, opened 2022-05-05
Duration frontend implementation

### LINKED ISSUE #261: Frontend implementation for duration data type
Blocked by #241 and #251. Details will be added once those issues are closer to completion.

### LINKED ISSUE #1108: Decide on date/time formatting string syntax
## Context

When displaying dates and times, the front end will recieve the following two pieces of info:

- a string representation of the value in [canonical form](https://wiki.mathesar.org/en/engineering/architecture/dateandtime).
- a "formatting string", containing information about how to display the value to the user.

The front end needs to take these two strings and combine them into a formatted value.

The front end has total control over setting the formatting string and reading it. The back end only validates that it's 255 characters or less ([discussion](https://wiki.mathesar.org/en/meeting-notes/2022-03#display-options-discussion)).

## To-do

We need to decide how our formatting strings will work.

## Prior art

- [LibreOffice](https://help.libreoffice.org/6.4/en-US/text/shared/01/05020301.html?&DbPAR=CALC&System=UNIX#hd_id3157309)

- [Moment](https://momentjs.com/docs/#/displaying/format/)

- [Moment](https://moment.github.io/luxon/#/formatting?id=table-of-tokens)

## Considerations

- We'll likely want to use a 3rd-party library to perform the formatting. As such, the work of evaluating and choosing that library may, in fact, be the only thing we need to do to close this ticket.

- Once users have set their formatting strings, it will not be easy for us to make backwards-incompatible changes to the formatting algorithm. This means that it's important that we choose a library which is stable and likey to have long-term support.

### LINKED ISSUE #1337: After saving, the cell should display the saved value from Postgres instead of the user input
## Reproduce

1. Set up a number column with "Decimal Places = 2 and "Max digits" = 1000.

1. Edit a cell in this column. Enter `1.23` and save. Observe `1.23` to be displayed. Refresh the table. Observe `1.23` to be displayed. This is all good.

1. Now edit the cell and change the value to `1.239` and save.

1. Expect `1.24` to be displayed after the value is saved, because the saved value is rounded by Postgres.

1. Observe `1.239` to be displayed, as the user entered it.

1. Refresh the table, and observe `1.24` to be displayed, as stored in Postgres.

pavish:
Fixes #261 , Fixes #1337 , Fixes #1108 

Screenshot 2022-05-05 at 6 24 56 AM

https://user-images.githubusercontent.com/11850603/166850095-53e73bc1-16d6-46e4-b14f-cf9585947cdd.mov

## Note:
* Fractional values for each unit is allowed.
   * User may save duration in `mm:ss` and then change format to `hh`. In this case, the hour value will be fractional.
   * Fractional values are rounded off to 3 decimal places.
* `show_units` display option is removed for the sake of faster completion.
   * Several UX considerations need to be taken into account to allow user to edit a duration field with unit strings.
   * It is very different from number where the user need not enter the grouping separators.
* `dayjs` is installed inorder to be used with duration and for date & time types. The main consideration was the library size and `dayjs` is a lot smaller than the other options. Since we not require heavy date time operations on the frontend, this seems like the best choice.   

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/api/tables/columns.ts
+++ b/mathesar_ui/src/api/tables/columns.ts
@@ -1,5 +1,5 @@
 import type { DbType } from '@mathesar/AppTypes';
-import type { DurationUnit } from '@mathesar/utils/Duration';
+import type { DurationUnit } from '@mathesar/utils/duration/types';
 
 /**
  * | value     | example locale | example format |
@@ -110,6 +110,7 @@ export interface Column extends BaseColumn {
   display_options: Record<string, unknown> | null;
 }
 
+// TODO: Remove specification DB types here
 export interface NumberColumn extends Column {
   type:
     | 'BIGINT'
--- a/mathesar_ui/src/api/tables/columns.ts
+++ b/mathesar_ui/src/api/tables/columns.ts
@@ -1,4 +1,5 @@
 import type { DbType } from '@mathesar/AppTypes';
+import type { DurationUnit } from '@mathesar/utils/Duration';
 
 /**
  * | value     | example locale | example format |
@@ -76,8 +77,6 @@ export interface BooleanDisplayOptions extends Record<string, unknown> {
     FALSE: string;
   } | null;
 }
-
-export type DurationUnit = 'd' | 'h' | 'm' | 's' | 'ms';
 
 export interface DurationDisplayOptions extends Record<string, unknown> {
   min: DurationUnit | null;
--- a/mathesar_ui/src/api/tables/columns.ts
+++ b/mathesar_ui/src/api/tables/columns.ts
@@ -77,6 +77,14 @@ export interface BooleanDisplayOptions extends Record<string, unknown> {
   } | null;
 }
 
+export type DurationUnit = 'd' | 'h' | 'm' | 's' | 'ms';
+
+export interface DurationDisplayOptions extends Record<string, unknown> {
+  min: DurationUnit | null;
+  max: DurationUnit | null;
+  show_units: boolean | null;
+}
+
 export interface BaseColumn {
   id: number;
   name: string;
--- a/mathesar_ui/package-lock.json
+++ b/mathesar_ui/package-lock.json
@@ -15513,6 +15513,11 @@
         "whatwg-url": "^8.0.0"
       }
     },
+    "dayjs": {
+      "version": "1.11.1",
+      "resolved": "https://registry.npmjs.org/dayjs/-/dayjs-1.11.1.tgz",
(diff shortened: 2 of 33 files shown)


mathesar_ui/src/api/tables/columns.ts
```diff
@@ -1,4 +1,5 @@
 import type { DbType } from '@mathesar/AppTypes';
+import type { DurationUnit } from '@mathesar/utils/duration/types';
```
seancolsen, line 2: I'd prefer to keep our application types dependent on the API types, and not the other way around. Can you move `DurationUnit` into the `src/api` somewhere? I recognize that it's not exactly relevant to `columns`, so we could potentially create a new directory like `src/api/_common` or something similar. I don't really care _where_ it goes within `src/api`, but I would like to put it in there somewhere.

It bears mentioning that I'd ideally like for `DbType` to live inside `src/api`, too -- but that type already existed when I created `src/api` and I wanted to minimize my refactoring at the time. (I'm not requesting you refactor `DbType` too -- only `DurationUnit` because that's new code.)

The fact that `DurationUnit` is a display option type does muddy the water somewhat, because (in theory) the front end should have control over display options. But even so, I find storing the type in the API directory to be helpful because it serves to communicate that any refactoring to the type requires extra care and thought beyond the references to that type within the rest of the font-end codebase. For example, if we refactor this type in a backwards incompatible way, we could still have runtime errors which slip through the type-checker due to user data that has not been migrated.

seancolsen (changes requested):
This is very cool!

- I played with the functionality a bit and didn't find any bugs. Very nice! I like that the UI prevents "Max Time Unit" from being less than "Min Time Unit".

- I support the choice of day.js. I looked through the [formatting options for day.js](https://day.js.org/docs/en/display/format) vs [those for date-fns](https://date-fns.org/v2.28.0/docs/format). Day.js is definitely more limited. For example, I can't format values like `May 5th, 2022` or `12:00 noon` with it, whereas I can obtain those formatted results using date-fns. But all-in-all, these limitations seem acceptable. Users will definitely be able to format their date/time data in a manner that looks good and properly conveys the underlying values -- even if they can't tweak it for display _exactly_ as they might want, were they to render the data in some other context like a document.
    
- I have noted one small change request which I hope is agreeable and easy enough to implement that we can handle it before merging.

mathesar_ui/src/api/tables/columns.ts
pavish: @seancolsen I've made the changes in this commit: https://github.com/centerofci/mathesar/pull/1350/commits/6cac69c84e45f1486274a6bfbcdc817fc39acf88

I've placed it within `columns.ts` for the time-being because it also contains types like `NumberFormat` which aren't exactly relevant to `Columns`, same as `DurationUnit`. We can think of refactoring and moving these to a separate file later.

PR #1365, opened 2022-05-06
Add a default value for precision field if only scale is present

### LINKED ISSUE #1300: Number scale not persisting
## Reproduce

1. In the front end, edit the DB options for a Number column.
1. Set "Number Type" to "Decimal". Set "Decimal Places" to "3". Click Save.
1. Observe a PATCH request to the columns API which specifies `type_options.scale` to `2`. Good.
1. In the response of this request (and also in the response of the subsequent GET request to the columns endpoint), expect the column to show `type_options` with an object containing a `scale` of `2`.
1. Instead, observe `type_options` to be `null`.

silentninja:
Fixes #1300 

Assigns a default value of `1000` for the `precision` type_option if only `scale` type_option is sent as a request.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/api/db/viewsets/columns.py
+++ b/mathesar/api/db/viewsets/columns.py
@@ -49,6 +49,7 @@ class ColumnViewSet(viewsets.ModelViewSet):
             )
         else:
             try:
+                # TODO Refactor add_column to user serializer validated date instead of request data
                 column = table.add_column(request.data)
             except ProgrammingError as e:
                 if type(e.orig) == DuplicateColumn:
--- a/mathesar/api/db/viewsets/columns.py
+++ b/mathesar/api/db/viewsets/columns.py
@@ -34,7 +34,8 @@ class ColumnViewSet(viewsets.ModelViewSet):
         # We only support adding a single column through the API.
         serializer = ColumnSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
-
+        if request.data.get('type_options', None) is not None and request.data['type_options'].get('scale', None) is not None and request.data['type_options'].get('precision', None) is None:
+            request.data['type_options']['precision'] = 1000
         if 'source_column' in serializer.validated_data:
             column = table.duplicate_column(
                 serializer.validated_data['source_column'],
--- a/mathesar/api/db/viewsets/columns.py
+++ b/mathesar/api/db/viewsets/columns.py
@@ -34,8 +34,12 @@ class ColumnViewSet(viewsets.ModelViewSet):
         # We only support adding a single column through the API.
         serializer = ColumnSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
-        if request.data.get('type_options', None) is not None and request.data['type_options'].get('scale', None) is not None and request.data['type_options'].get('precision', None) is None:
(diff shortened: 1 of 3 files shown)


silentninja:
@seancolsen Can you review it?

mathesar/api/db/viewsets/columns.py
```diff
@@ -34,7 +34,8 @@ def create(self, request, table_pk=None):
         # We only support adding a single column through the API.
         serializer = ColumnSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
-
+        if request.data.get('type_options', None) is not None and request.data['type_options'].get('scale', None) is not None and request.data['type_options'].get('precision', None) is None:
```
kgodey, line 37: This is long and hard to parse quickly, can you put each conditional into a separate variable?

seancolsen (changes requested):
@silentninja This is still not working for me using the exact repro steps from the ticket.

- ❌ When I send

    ```json5
    {
      "type_options": { "scale": 2 }
    }
    ```

    I expect:

    ```json5
    {
      // ...
      "type_options": { "scale": 2, "precision": 1000 }
      // ...
    }
    ```

    Instead, I receive:

    ```json5
    {
      // ...
      "type_options": null
      // ...
    }
    ```

- ❌ When I send

    ```json5
    {
      "type_options": { "scale": null, "precision": null }
    }
    ```

    I expect either of the following:

    ```json5
    {
      // ...
      "type_options": { "scale": null, "precision": null }
      // ...
    }
    ```

    ```json5
    {
      // ...
      "type_options": null
      // ...
    }
    ```

    Instead, I receive an error response:

    ```json5
    [
      {
        "code": 2023,
        "field": "precision",
        "message": "This field may not be null.",
        "detail": {}
      },
      {
        "code": 2023,
        "field": "scale",
        "message": "This field may not be null.",
        "detail": {}
      }
    ]
    ```

silentninja:
@seancolsen 

- The last commit should fix the first error. 
-As for the second error, the backend response won't contain the fields `precision` or `scale` if their value is `None` and this being done intentionally.
https://github.com/centerofci/mathesar/blob/e33b5eed35882645f0344959fcc3f7e6c77662b6/db/columns/base.py#L180

I think the `None` value type_options are removed in order to remove irrelevant `type_options` as the backend is unable to differentiate `type_options` applicable for a type, so if I allow `null`, `type_options` will leak between type(`boolean` type will return {'fields': None, 'length': None, precision': None, 'scale': None}

The proper solution would be, to have `type` aware `type_options` which isn't a small refactor and can take quite some time.

So we have two temporary options here
- Accept `type_options` leaking into other types
- Frontend should avoid passing null values and instead remove them

seancolsen (approved):

seancolsen:
Thanks @silentninja. This is looking good now.

> Frontend should avoid passing null values and instead remove them

That's fine for now. At some point it would be nice to do the refactoring you mention, so the API can better follow the principle of least surprise. But this approach works well in the short term

I've opened #1367 to track a similar problem I uncovered after I was able to test more.

PR #1374, opened 2022-05-10
Add link Table API

### LINKED ISSUE #972: API support for "Link Table" dialog
## Problem

- In the ["Link Table" dialog](https://wiki.mathesar.org/en/design/specs/foreign-key-constraints#scenario-1a-automatically-adding-foreign-key-constraints-from-the-link-table-dialog), the front end needs to create a new foreign key reference between two tables.

    - For the one-to-one case, we need to add a column, add a foreign key constraint, and add a unique constraint.
    - For the many-to-one and one-to-many cases, we need to add a column and add a foreign key constraint.
    - For the many-to-many case, we need to add a mapping table, add columns to the table, and add foreign key constraints.

- The front end can already perform all of those operations through the API. However, at the DB level, we'de like for those operations to occur within the same transaction, which necessitates some additional API features.

## Solution

The implementer should write a proposed API spec and get feedback from other members of the engineering team before implementing this issue.

### Ideas
_This section written by @kgodey._
- An `/api/v0/links/` endpoint that lists all links and their types and allows creation and deletion of links. Creating and delete links should encapsulate all operations involved (creation and deletion of columns, constraints, and/or tables) in a single transaction.
- Links can be one of three types:
  - one-to-one: this involves any single-column FK with a unique constraint applied.
  - many-to-one: this is any single-column FK
  - many-to-many: this is any table with two FKs
- Links should be populated entirely by reflection and shouldn't need to be created within Mathesar

Details about how to define many-to-many and many-to-one tables need to be worked out as part of the spec. Please also consider the case of tables with FKs to themselves while doing this.

## Additional context
- [FK design spec](https://wiki.mathesar.org/design/specs/foreign-key-constraints)
- #942 
- #943

silentninja:
Fixes #972 

Adds APIs to support creating relationships between columns.

**Technical details**
As per the discussion in #972, the finalised APIs are

```typescript
interface OneToOneOrOneToMany {
  link_type: "one-to-one" | "one-to-many";
  reference_table: number;
  reference_column_name: string;
  referent_table: number;
}

interface ManyToMany {
  link_type: "many-to-many";
  mapping_table_name: string;
  referents: {
    referent_table: number;
    column_name: string;
  }[];
}

type LinkTablePostRequest = OneToOneOrOneToMany | ManyToMany;
```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/links/operations/create.py
```diff
@@ -0,0 +1,67 @@
...
+    with engine.begin() as conn:
+        referent_tables_oid = [referent['referent_table'] for referent in referents]
+        referent_tables = reflect_tables_from_oids(referent_tables_oid, engine, conn)
+        metadata = MetaData(bind=engine, schema=schema, naming_convention=naming_convention)
+        opts = {
+            'target_metadata': metadata
+        }
+        ctx = MigrationContext.configure(conn, opts=opts)
+        op = Operations(ctx)
+        op.create_table(map_table_name, schema=schema)
```
mathemancer, line 49: I think you should use the standard Mathesar function for creating a table. If nothing else, we'll want the default columns so this fits into the rest of the framework, correct?

db/links/operations/create.py
```diff
@@ -0,0 +1,67 @@
...
+            op.create_unique_constraint(None, referrer_table.name, [column_name], schema=schema)
+        op.create_foreign_key(
+            None,
+            referrer_table.name,
+            referent_table.name,
+            [column.name],
+            [primary_key_column.name],
+            source_schema=schema,
+            referent_schema=schema
+        )
```
mathemancer, line 36: Unless there's a convincing reason not to, I think you should create the column with the unique constraint and foreign key included, then add that column to the table. This would make this simpler and more aligned with the foreign key creation in `tb/tables/operations/split.py`. I can think of no reason to do it in separate operations given that we're not supporting multi-column foreign key creation with this.

db/links/operations/create.py
```diff
@@ -0,0 +1,67 @@
...
+            op.add_column(map_table_name, column, schema=schema)
+            op.create_foreign_key(
+                None,
+                map_table_name,
+                referent_table.name,
+                [column.name],
+                [primary_key_column.name],
+                source_schema=schema,
+                referent_schema=schema
+            )
```
mathemancer, line 67: Why not use your `create_foreign_key_link` function? Nested transactions don't commit until the outermost transaction is complete. If there's a reason not to call the other function, the comment from above about creating the column with the needed pieces then adding it applies here as well.

db/tables/operations/select.py
```diff
@@ -35,6 +35,31 @@ def reflect_table_from_oid(oid, engine, connection_to_use=None):
     return reflect_table(table_name, schema, engine, connection_to_use=connection_to_use)
 
 
+def reflect_tables_from_oids(oids, engine, connection_to_use=None):
```
mathemancer, line 38: Why not combine this function with the previous one? I think you should either loop through the singular version in this plural version, or (my preference) modify the singular one to call this one with appropriate arguments.

mathemancer (changes requested):
I left some specific comments. The big one to deal with is about creating the foreign key columns as a single step.

seancolsen (dismissed):
@silentninja I'm not able to get any requests to succeed, as spec'ed in your schema. Below are four scenarios that I've tried. Each scenario lists the request I sent. In each scenario, I expect a success response. And in each scenario, I observed an error (with the exact response shown that I received).

To me it seems like there are probably a number of small things in this PR that need to be adjusted in order for my API requests to succeed. For example, I'd like to use the property `reference_table_id` instead of `reference_table`. And in the "one-to-one" case, I think we should be specifying `reference_column_name` not `referent_column_name`.

## One to one

Request:

```json
{
  "link_type": "one-to-one",
  "reference_table_id": 4,
  "reference_column_name": "planet_id_1",
  "referent_table_id": 2
}
```

Response:

```json
[
  {
    "code": 2002,
    "field": "referent_column_name",
    "message": "This field is required.",
    "detail": {}
  },
  {
    "code": 2007,
    "field": "reference_table",
    "message": "This field is required.",
    "detail": {}
  },
  {
    "code": 2007,
    "field": "referent_table",
    "message": "This field is required.",
    "detail": {}
  }
]
```


## Many to one

Request:

```json
{
  "link_type": "many-to-one",
  "reference_table_id": 4,
  "reference_column": { "name": "planet_id_1" },
  "referent_table_id": 2
}
```

Response:

```json
[
  {
    "code": 4999,
    "message": "'NoneType' object has no attribute 'create'",
    "field": null,
    "detail": null
  }
]
```

## One to many

Request:

```json
{
  "link_type": "one-to-many",
  "reference_table_id": 4,
  "referent_table_id": 2,
  "referent_column_name": "moon_id"
}
```

Response:

```json
[
  {
    "code": 2007,
    "field": "reference_table",
    "message": "This field is required.",
    "detail": {}
  },
  {
    "code": 2007,
    "field": "referent_table",
    "message": "This field is required.",
    "detail": {}
  }
]
```

## Many to many

Request:

```json
{
  "link_type": "many-to-many",
  "mapping_table_name": "moon_planet",
  "referents": [
    { "table_id": 4, "column_name": "moon_id" },
    { "table_id": 2, "column_name": "planet_id" }
  ]
}
```

Response:

```json
[
  {
    "code": 4999,
    "message": "unhashable type: 'dict'",
    "field": null,
    "detail": null
  }
]
```

db/links/operations/create.py
silentninja: `create_mathesar_table` doesn't respect the transaction. The below test case fails due to `create_table` not being part of the transaction and ends up creating the tables although the foreign key constraint step fails. I am not sure if I am doing something wrong 

```
def test_table_creation(engine_with_schema):
    engine, schema = engine_with_schema
    supported_types = get_supported_alter_column_types(
        engine, friendly_names=False,
    )
    sa_type = supported_types.get("INTEGER")
    metadata = MetaData(bind=engine, schema=schema, naming_convention=naming_convention)
    with engine.begin() as conn:
        metadata.reflect()
        opts = {
            'target_metadata': metadata
        }
        ctx = MigrationContext.configure(conn, opts=opts)
        referent_column = MathesarColumn(
            "reference",
            sa_type()
        )
        target_table = create_mathesar_table(
            "target_table",
            schema,
            [referent_column],
            engine,
            metadata,
        )
        fk_column = MathesarColumn(
            "fk_col",
            sa_type()
        )
        base_table = create_mathesar_table(
            "base_table",
            schema,
            [fk_column],
            engine,
            metadata
        )
        op = Operations(ctx)
        # op.create_unique_constraint(None, target_table.name, [referent_column.name], schema=schema)
        try:
            op.create_foreign_key(
                None, base_table.name,
                target_table.name,
                [fk_column.name], [referent_column.name], source_schema=schema, referent_schema=schema
            )
        except:
            pass
    schema_obj = reflect_schema(engine, schema)
    tables = get_table_oids_from_schema(schema_obj.oid, engine)
    # Table should not be created as foreign key creation step would have failed due to referent column missing the unique constraint 
    assert len(tables) == 0
```

silentninja: The transaction fails if `create_mathesar_table` uses a connection to bind instead of engine when creating the table 

https://github.com/centerofci/mathesar/blob/b2467dcf63de786c0e8063821ff317ef608ec3a6/db/tables/operations/create.py#L28
~table.create(engine)~
table.create(connection)

db/links/operations/create.py
silentninja: Isn't creating a foreign key constraint in individual steps much more flexible, as it could allow for multiple columns, although we don't use it now. Does creating a Foreign key as part of the column bring in any advantage?

db/links/operations/create.py
silentninja: It seemed redundant as both the signature are the same, but I agree it would be better to use a common API.

silentninja:
@seancolsen 

- One to one error - I made changes to the API schema, please do try out the schema listed in the PR's description
- Many to one error - There is no `Many to one` type, I added correct validation error instead of throwing a 500 error
- One to many - Please do try out the schema listed in the PR's description.
- Many to many - Fixed the error handler to  return correct validation, it should throw a correct error now

db/links/operations/create.py
mathemancer: I think the main advantage is that it's easier to see the logic. I acknowledge, however, that it's more flexible in separate steps. If you want to keep separate steps, I can go with it. Consider using your already-written `create_foreign_key_link` function to add the foreign key.

db/links/operations/create.py
mathemancer: Argh, what a pain.

I think it would still be better to use the `create_mathesar_table` function to create the table, since it's important that we have the pieces needed (`id` column at a minimum). We should then file a bug to deal with the fact that transactions aren't working properly.

I really want to avoid ad-hoc table creation calls scattered through the codebase since we already have a default column involved and we're going to add more default metadata, etc. over time.

db/links/operations/create.py
mathemancer: In that case, consider extracting the common code from the two functions and using it as a private method.

mathemancer (changes requested):
The only blocking change I have to request is to figure out how to avoid the bare table creation. I think it's already a problem since it's not created with the default column(s), and it may get further from how other tables are created over time. The other request is a suggestion.

Issue #950, opened 2022-01-05
Add visual cues to foreign key columns


seancolsen:
## Current behavior

- When viewing a table, the user has no indication which columns have foreign key constraints.

## Desired behavior

- Column header displays referenced table and referenced column.
- Column header menu displays referenced table.
- Cell values are stylized differently.

## Notes

- If two columns are both part of the same multi-column FK, then they should both receive the foreign key styling. 

## Specs

- [Foreign Key Constraints Spec](https://wiki.mathesar.org/en/design/specs/foreign-key-constraints) -- see Scenario 3a.

## See also

- #952

seancolsen:
@ghislaineguerin one small concern -- the cell styling of a background and border radius on the text looks great with simple numbers -- and I imagine styling will suffice in the vast majority of cases. But I wonder if it's worth considering the scenario where the FK value is a text that wraps within the cell. In that case the border radius won't look so good. Would it be better to apply the background color to the entire cell instead of just the text? Or perhaps the text-wrapping scenario will be rare enough to disregard?

ghislaineguerin:
@seancolsen I worry about how that might communicate the fact that the records are 'linked', meaning they exist elsewhere. My idea with separating them visually from the table was that. I believe that if we merge them into the cell we are losing that. I'll test it out and see if we can work something out, or test with wrapping as well to see how that looks.

pavish:
> FK value is a text that wraps within the cell

@seancolsen @ghislaineguerin 
For cases where the FK value is a lengthy text, I'd suggest that we rather add ellipsis to the end instead of wrapping it.
1. The cell height has to be explicitly mentioned and not controlled by the browser, to support virtual scrolling.
2. Spreadsheets do not wrap text usually.
3. The cell styling from the UX would work for all cases when we do not wrap the text.

kgodey:
@pavish Won't we need to support different cell heights for the record preview feature? Also, how would editing the value work if the user can't see the whole value?

pavish:
@kgodey 

> Won't we need to support different cell heights for the record preview feature?

* Yes, we will be supporting this. This is possible because the cell heights for the preview feature can be calculated prior to rendering, considering that each key value pair takes a single line for itself within the cell. However, if the text is wrapped and can take any arbitrary number of lines per key value pair, we cannot estimate the space it would possibly take without rendering it first, which we cannot support. 
* We can provide options for users to still manually resize both the column and the row (similar to spreadsheets).

[...]

Issue #942, opened 2022-01-05
Modify constraints API to support creating foreign key constraints


seancolsen:
## Current behavior

- I can't create FK constraints.
- I can create a _unique_ constraint with a POST to `/tables/{table_id}/constraints/` and `type: "unique"`.

## Desired behavior

- Creating FK constraints should function similarly to creating unique constraints.

## Notes
- The API for creating unique constraints is currently in flux due to #898. We should be sure to follow the same pattern for FK constraints, which will likely mean requiring the API request to specify all tables and columns by id instead of name.
- We don't plan to ever support _updating_ FK constraints. If the user wants to change the constraint, they'll need to drop it and add a new one.
- If it is fairly quick to implement, the API should support as many of the following features as it can, even though there's no UI for them yet. If you do not support these in your PR, please design the API in a way that can be easily extended to support them in the future without breaking changes. It would also be helpful to create separate placeholder issues to track any issues not handled by the PR.
    - Creating deferred FK constraints
    - Setting the `ON DELETE` attribute to any of the five possible actions.
    - Setting the `ON UPDATE` attribute to any of the five possible actions.
    - Creating multi-column FKs, including the ability to pass `MATCH FULL`.

## Additional context
- Blocked by API definition in #943
- [Foreign Key Spec](https://wiki.mathesar.org/en/design/specs/foreign-key-constraints) -- Scenario 1
- [Postgres docs for Constraints](https://www.postgresql.org/docs/13/ddl-constraints.html)
- [Postgres docs for `SET CONSTRAINTS`](https://www.postgresql.org/docs/13/sql-set-constraints.html)
- The UI only supports single-column FKs, but we might want to add support for creating multi-column FKs in the future ([discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$ZMwTJrIU7sWuqtrSndMHaRZ3hR2rAW9A8JEemqQOW6w?via=matrix.mathesar.org&via=matrix.org)).

seancolsen:
@kgodey I'm assigning this to you to help answer the above questions and loop others in as needed.

kgodey:
@seancolsen I've updated the ticket description to resolve questions. I'm going to mark the ticket as blocked until #898 is resolved and the API for constraints stabilizes.

PR #1375, opened 2022-05-10
Foreign key constraints list

### LINKED ISSUE #943: Modify constraints API to list foreign key constraints
## Current behavior

- I can see unique and primary key constraints with a GET to `/tables/{table_id}/constraints/`, but the response does not include FK constraints.

## Desired behavior

- FK constraints should be included in the constraints response.
- Additional FK-specific details about each FK constraint should be present too, including:
    - Target columns of the FK
    - `ON DELETE` behavior
    - `ON UPDATE` behavior
    - `MATCH` behavior
    - `DEFERRED` behavior 

## Considerations
- A foreign key might be associated with more than one column, and in such cases will reference more than one column too. We're not planning to add support for creating multi-column FKs any time soon, but we do need support for displaying them, so the API needs to respond by showing all the constraint columns and all the target columns.
- Note that API consumers also need a way to associate each constraint column with its corresponding target column. If we return the constraint columns separately from the target columns, that's fine as long as the ordering of those two arrays is suitable for associating the columns by array index.
- The UI eventually want to display a specific warning in the front end when the user is deleting rows that will cascade in case of constraints with `ON DELETE CASCADE`.

## Spec Prompts
The person implementing this issue should spec out their proposed API changes and run them by the Mathesar core team. Questions the spec should answer include:
- How does the response indicate the target table and target columns?
- How does the response provide info about the FK-specific details listed above?

## Additional context
- Blocked by API changes in #898
- [Foreign Key Spec](https://wiki.mathesar.org/en/design/specs/foreign-key-constraints) -- Scenarios 2 and 3
- [Postgres docs for Constraints](https://www.postgresql.org/docs/13/ddl-constraints.html)
- [Postgres docs for `SET CONSTRAINTS`](https://www.postgresql.org/docs/13/sql-set-constraints.html)

silentninja:
Fixes #943 

Adds support for listing foreign key constraints to the `table/constraints/` API

**Technical details**
API schema

```typescript
interface ForeignKeyConstraint {
  id: number;
  name: string;
  type: 'foreignkey';
  referent_columns: number[];
  referent_table: number;
  onupdate: 'RESTRICT' | 'CASCADE' | 'SET NULL' | 'NO ACTION' | 'SET DEFAULT' | None;
  ondelete: 'RESTRICT' | 'CASCADE' | 'SET NULL' | 'NO ACTION' | 'SET DEFAULT' | None;
  deferrable: boolean | None;
  match: 'SIMPLE' | 'PARTIAL' | 'FULL' | None;
}
```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
@silentninja

To specify the referent column, I see that this implementation currently returns only the id.

I understand that, in theory, the id alone is sufficient to convey the association to API consumers because ids are unique across all columns of all tables. However, in practice, that's difficult for the API consumers...

Pretend I have a `moon` table with a `planet_id` column referencing a `planet` table. Issue #950 requires that, when rendering the `moon` table, the front end displays the text "planet.id" within the column header of the `planet_id` column. To render that text, I need the name of the referent table ("moon") and the name of the referent column ("id"). If I only have the column id, then I need to iterate through all tables, loading their columns one-by-one, until I find the table which contains the referent column. That's messy and slow.

I'd like for the constraints API to give me the id of the referent table too.

I'm open to various ways of solving this problem. From an API design perspective, I'm not sure what the best solution might be.

Here's one proposal, just to make this discussion as concrete as possible...

```ts
interface ForeignKeyConstraint {
  // ...
  referents: {
    table_id: number;
    column_id: number;
  }[];
  // ...
}
```

I recognize that when adding a new FK constraint via a `POST` request (to satisfy #942), the schema I'm proposing adds an unnecessary field. The API client should not need to pass `table_id` because `column_id` is sufficient. So perhaps `table_id` could be optional in the request but always present in the response.

mathesar/api/serializers/constraints.py
```diff
@@ -1,13 +1,58 @@
...
+            'referent_columns',
+            'onupdate',
+            'ondelete',
+            'deferrable',
+            'match'
+        ]
+
+    referent_columns = serializers.PrimaryKeyRelatedField(queryset=Column.current_objects.all(), many=True)
+    onupdate = serializers.ChoiceField(choices=['CASCADE', 'DELETE', 'RESTRICT'])
+    ondelete = serializers.ChoiceField(choices=['CASCADE', 'DELETE', 'RESTRICT'])
```
mathemancer, line 33: Assuming these are supposed to represent and cover the options for `ON UPDATE` and `ON DELETE` in the DB, I think these lists should be `['CASCADE', 'RESTRICT', 'NO ACTION', 'SET NULL', 'SET DEFAULT']`. Is `'DELETE'` supposed to be in place of `'SET NULL'`? See [the docs](https://www.postgresql.org/docs/13/ddl-constraints.html#DDL-CONSTRAINTS-FK).

mathemancer (changes requested):
I mostly focused on Python and the DB terms. I assume you'll handle the API negotiation with @seancolsen .

mathesar/api/serializers/constraints.py
silentninja: Thanks for pointing it out @mathemancer, I was using the options listed in[alembic docs](https://alembic.sqlalchemy.org/en/latest/ops.html?highlight=foreign#alembic.operations.Operations.create_foreign_key), with the assumption that they were using generic terms to make it compatible across databases, but it looks like it isn't the case.

silentninja:
@seancolsen I made the change to the API schema, let me know if it works for you.

mathemancer (approved):
Okay, I think it looks good from a DB/Python perspective.

seancolsen (approved):
Looks good! Thanks @silentninja

PR #1376, opened 2022-05-11
Add support for creating Foreign key constraints

### LINKED ISSUE #942: Modify constraints API to support creating foreign key constraints

silentninja:
Fixes #942 

Adds support for adding foreign key constraints using `/constraints/` API. The schema is the same as the foreign key constraint list response

**Technical details**
There are few changes to the existing method for creating constraints

- Rather than use specific functions(`create_unique_constraint`) for creating each constraint, we have constraints classes that have abstract methods for each constraint type which take in the alembic `migration operation` object and call the equivalent constraint operation on the passed object. This makes it easier to have specific fields for different constraints and call multiple constraints in a single operation(bonus).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
Blocked by #1375

seancolsen:
I've marked this as unblocked now that #1375  is merged.

seancolsen (approved):
I haven't reviewed any of the code, but the functionality works for me, and I've opened #1378 with the UI changes on top of this work.

db/constraints/base.py
```diff
@@ -0,0 +1,72 @@
...
+            options
+    ):
+        self.name = name
+        self.table_oid = table_oid
+        self.columns_attnum = columns_attnum
+        self.referent_table_oid = referent_table_oid
+        self.referent_columns = referent_columns_attnum
+        self.options = options
+
+    def add_constraint(self, migration_operation, engine, connection_to_use):
```
mathemancer, line 37: It's strange to me that you're passing the `Operations` object (`migration_operation`) here. Why not just create it from the engine you have in hand? Passing all of these separately is a bit wonky, since if they're out-of-sync (i.e., if the engine isn't the one used for the `MigrationContext` behind the `Operations` class, weird things could happen (what if the engine is reflecting from some other DB, for example?).

mathemancer (changes requested):
Overall, looks good. Please reconsider creating the `Operations` object (and its `MigrationContext` in the function. Alternatively, get the connection info and engine from those objects rather than passing them separately.

mathemancer (approved):
I think this looks good. Nice work, @silentninja !

PR #1378, opened 2022-05-12
UI to add FK manually

### LINKED ISSUE #948: Front end support for manually creating foreign key constraints
## Goal

- Allow users to create foreign key constraints through the Table Constraints dialog.

## Limitations

- Multi-column FK constraints will _not_ be supported yet. We may add support for that at some point in the future ([discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$ZMwTJrIU7sWuqtrSndMHaRZ3hR2rAW9A8JEemqQOW6w?via=matrix.mathesar.org&via=matrix.org)) but we'll let the need arise organically. During the spec process, it was somewhat unclear when/if we'd support adding multi-column FK constraints, so there are some aspects of the specs which still communicate intent to support it. You can disregard that when implementing this ticket. For example, the video and prototype show a checkbox group to select the constraint columns. That should be a radio group instead so that the user can only select one.

## Specs

- In the [Foreign Key Constraints Spec](https://wiki.mathesar.org/en/design/specs/foreign-key-constraints), see:
    - the video for Scenario 1
    - the notes for Scenario 1b

## Notes

- The specs are somewhat out-of-date with the most current design for the table constraints dialog. We should follow the patterns set within the work to close #531.

## See also

- #947

## Status

- Blocked by #942

seancolsen:
Fixes #948

## Demo

https://www.loom.com/share/943b776f3b0747108ef9383986efe742

## Issues to handle later

- To handle the logic required for the constraint name, there's a bit of duplicated code between `NewUniqueConstraintModal.svelte` and `NewFkConstraintModal.svelte`. At some point we could add an abstraction to reduce this duplicated code.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/api/tables/constraints.ts
+++ b/mathesar_ui/src/api/tables/constraints.ts
@@ -5,7 +5,7 @@ export type ConstraintType =
   | 'check'
   | 'exclude';
 
-export interface Constraint {
+export interface BasicConstraint {
   id: number;
   name: string;
   type: ConstraintType;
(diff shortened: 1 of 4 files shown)


pavish (approved):
@seancolsen Looks good to me.

There are a couple improvements we can make:
* The 'Add' button is not disabled when the target table is switched and the columns are still loading. It would be better to disable it.
* We allow the user to select columns for FK mappings regardless of the column types, which then fails on the backend. Ideally, we should be only showing valid target columns (unique columns of the same type).

I'm approving and merging in this PR. These improvements can be taken care of in a later PR.

PR #1385, opened 2022-05-18
Improve UX for new row with NOT NULL constraint

### LINKED ISSUE #775: Improve UX when adding new records to a table with a NOT NULL column

### LINKED ISSUE #1362: Show DEFAULT in new record cells by default
## "Limbo"

When adding a new record, there is a to which I'll refer as "limbo". The new record row is in limbo after it's been added to the UI but before it's been saved to the DB. Most times, the new record row only remains in limbo transiently, for a split second, while the request is loading. Any behavior of the row for such a short time is only of negligible concern. However, if there are errors adding a new row, the row may reach stasis in limbo, as described in #775 for example. In this prolonged limbo state, the UX becomes more important and currently has some shortcomings.

## Current behavior in limbo

- When a cell is poised to set the Postgres value to `DEFAULT`, it appears blank.

- After the user edits the cell, the user has no way to revert the cell back to its `DEFAULT` state.

## Desired behavior

- All cells appear as `DEFAULT` by default.

- `DEFAULT` is styled similarly to `NULL`.

- Editing a cell, brings the value out of `DEFAULT`.

- The user can return to `DEFAULT` using the context menu, in a similar fashion to `NULL`.

seancolsen:
Fixes #775
Fixes #1362

## Demo

https://www.loom.com/share/b8da99a10eec4719b71c4ade93102c85

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
Ok, the E2E test failure here is actually a good catch. Thanks, Playwright! I need to spend some more time on this, so I'm sending it back to draft state.

seancolsen:
I pushed another commit to fix the bug that the E2E test caught. It brought to light an interesting edge case which warrants its own demo. Please also watch this video when reviewing:

https://www.loom.com/share/fd29aae20abf4877bb7f62e433cfaf8d

pavish (approved):
@seancolsen Nice work on this!

I tested on a few scenarios and the UX looks good to me. There are a few scenarios that are okay with me but I'm not sure on how the target users might perceive it.

For eg., For checkbox, when DEFAULT is displayed and the user clicks on the cell, it toggles the value and the checkbox is then displayed. This seems reasonable to me, but this is one of the areas where we should try to get the user's opinions when possible after the alpha.

PR #1398, opened 2022-05-25
Add API and UI support for toggling grouping separators within numbers

### LINKED ISSUE #1329: Allow the user to hide grouping separators for Number and Money
## Current behavior

- Look at this table of numbers. The right-most column is storing years as in integer, which I think will be common. But they are displayed with grouping separators. Ugh. That looks terrible!

    image

- There is currently no way for the user to disable the display of grouping separators.

## Desired behavior

- The user can use display options to enable/disable grouping separators.

### Schema

- Use the `use_grouping` property as [spec'ed in this comment](https://github.com/centerofci/mathesar/discussions/1243#discussioncomment-2519195).

### UX

- Add a new form field labeled "Use Grouping Separators", with a dropdown input having options "Yes", "No", and "Auto". Display this new field before the already-existing "Format" field.

    For integers, the default is "No".

    For non-integers, and Money, the default is "Yes".

- Modify the "Format" form field to behave as follows:

    - If we have an **integer with grouping separators**, then present the following options:

        | option    | value |
        | --------- | ------- |
        | Use browser locale | `null` |
        | 1,234,567 | `'english'` |
        | 1.234.567 | `'german'`  |
        | 1 234 567 | `'french'`  |
        | 12,34,567 | `'hindi'`   |
        | 1'234'567 | `'swiss'`   |

    - If we have an **integer without grouping separators**, then don't show the "format" option, because it's irrelevant. Keep value as `null`.

    - If we have a **non-integer with grouping separators**, then present the following options:

        | option    | value |
        | --------- | ------- |
        | Use browser locale | `null` |
        | 1,234,567.89 | `'english'` |
        | 1.234.567,89 | `'german'`  |
        | 1 234 567,89 | `'french'`  |
        | 12,34,567.89 | `'hindi'`   |
        | 1'234'567.89 | `'swiss'`   |

    - If we have a **non-integer without grouping separators**, then present the following options:

        | option    | value |
        | --------- | ------- |
        | Use browser locale | `null` |
        | 1234.56 | `'english'` |
        | 1234,56 | `'german'`  |

- When changing the "Use Grouping Separators" field from "No" to "Yes", we set a more specific value for the "Format" field, using the browser locale to resolve ambiguity. For example, if the user has configured `format` to `'german'` (because they just wanted a comma decimal separator), but their locale is `fr-FR`, then we change `'german'` to `'french'`, which also has a comma decimal separator but is more likely to match their desired grouping separator. We can use the browser locale to determine the format option by formatting a number and inspecting it. As a further edge case, if the user has configured `format` to `'german'`, but their locale is `en-US` (which is _not_ a more specific form of `'german'`), then we'd leave the value untouched.

seancolsen:
Fixes #1329

## Demo

Peek 2022-05-25 16-50

## Before

- All numbers display grouping separators. This is annoying sometimes, e.g. if you have an integer column to represent "year".

## After

- We have a display option for it.

## Notes

- I don't completely understand the behavior of `useGrouping: "auto"` from [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat). Under what scenarios do `true` and `"auto"` differ? I'm not sure. But I stuck with using a separate option for it because that was the approach which seemed to involve the least amount of thought and research into edge cases.
- I didn't implement the custom "Number Format" options as spec'ed in #1329 because it's tricky to do within the serializable form generator. I think we might need a custom component for that. I'll open a ticket for that work separately after we merge this PR.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/tests/api/test_column_api_display_options.py
+++ b/mathesar/tests/api/test_column_api_display_options.py
@@ -314,8 +314,6 @@ def test_column_alter_same_type_display_options(
     column_index = 2
     column = columns[column_index]
     pre_alter_display_options = column.display_options
-    print("----------------------------------------------------")
-    print(pre_alter_display_options)
     with engine.begin() as conn:
         alter_column_type(table.oid, column.name, engine, conn, PostgresType.NUMERIC)
     column_id = column.id
--- a/mathesar/tests/api/test_column_api_display_options.py
+++ b/mathesar/tests/api/test_column_api_display_options.py
@@ -28,7 +28,7 @@ def column_test_table_with_service_layer_options(patent_schema):
     column_data_list = [
         {},
         {'display_options': {'input': "dropdown", "custom_labels": {"TRUE": "yes", "FALSE": "no"}}},
-        {'display_options': {'show_as_percentage': True, 'number_format': "english"}},
+        {'display_options': {'show_as_percentage': True, 'number_format': "english", "use_grouping": 'auto'}},
         {'display_options': None},
         {},
         {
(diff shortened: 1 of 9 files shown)


seancolsen:
This is a full-stack PR, so I've assigned @pavish to review the front end code and @silentninja to review the back end code.

mathesar/tests/api/test_column_api_display_options.py
```diff
@@ -307,6 +312,8 @@ def test_column_alter_same_type_display_options(
     column_index = 2
     column = columns[column_index]
     pre_alter_display_options = column.display_options
+    print("----------------------------------------------------")
```
silentninja, line 315: Please remove `print` statements, We use `print` statements only to output messages to the user.

silentninja (changes requested):
Looks good to me! I have requested a small change, please do fix it.

mathesar/tests/api/test_column_api_display_options.py
seancolsen: Oh gosh, that's embarrassing 🙄. Thanks for catching! I've fixed it in f98ad2eabc9da427bf7fc9e02764860d9a7cb740

silentninja (approved):
Nice work @seancolsen, I have approved the PR but let's wait until we get a review from @pavish before merging it.

pavish (approved):
@seancolsen Looks good to me!

I am not sure about the 'Auto' option, and what it actually does. Maybe we could restrict it to just 'On' and 'Off', which would simplify things for the user as well. I'm merging the PR, we could come back to this later.

PR #1406, opened 2022-05-31
Fix constraints endpoint's error behaviour

### LINKED ISSUE #1030: Improve failure behavior of constraints endpoint
## Description

This ticket lists many small (somewhat independent) problems with the response of the constraints API when subjected to POST requests containing invalid data.

## Caveats

- I tested the constraints API using the changes for #1026, which I'm presuming will soon be merged.
- I'm not mentioning some of the more obvious failure cases which also have the problems with non-conformant response (e.g. case where the request is missing the "columns" field), because I imagine we'll fix that at a lower level, so as to apply to all endpoints.

## Addition context

- #883
- [API Error Standards](https://wiki.mathesar.org/en/engineering/architecture/api-standards#errors)

## Response schema problems

In the following situations, the API correctly responds with a HTTP 4xx status code, but the response schema does not conform to our error standards.

- ~~Attempting to set a unique constraint for a column which contains duplicate values~~ (now fixed)
- ~~Attempting to set a unique constraint with a supplied constraint name equal to that of an existing table~~ (now fixed)

## Unhandled exception problems

In the following situations, the API responds with a HTTP 5xx status code when it should respond with a 4xx status code and a JSON response body conforming to our error standards.

- Invalid "type", e.g.:

    ```json
    {"type":"foo","columns":[1]}
    ```

- Attempting to set a unique constraint by supplying the id of a column that exists in a different table from the table whose id is supplied in the URL.

- An empty `columns` array, e.g.:

    ```json
    {"type":"unique","columns":[]}
    ```

- Non-existent column id, e.g.:

    ```json
    {"type":"unique","columns":[9999]}
    ```

Anish9901:
Related to #1030

This PR Fixes many small unhandled exceptions and 500s caused by the constraints API endpoint.

**Technical details**

- An empty columns array:
```json
{"type":"unique","columns":[]}
```
- Response:
```json
{
    "code": 4999,
    "message": "Columns field cannot be empty.",
    "field": null,
    "detail": null
}
```
- Invalid "type"
```json
{"type":"foo","columns":[1]}
```
- Response:
```json
{
    "code": 4408,
    "message": "Unknown database type identifier foo.",
    "field": null,
    "detail": null
}
```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
@silentninja I am getting a KeyError while sending this request:
```json 
{"type":"unique","columns":[9999]}
``` 
I suspect the error is generated by this part of the code https://github.com/centerofci/mathesar/blob/4ccea635e390732823f1294811d6a623ab864043/mathesar/api/serializers/shared_serializers.py#L116-L118
How can I solve this error?

silentninja:
> @silentninja I am getting a KeyError while sending this request:
> 
> ```json
> {"type":"unique","columns":[9999]}
> ```

The actual error is from this line

https://github.com/centerofci/mathesar/blob/4716b0fb999b164a7354ad52dc4bb985e7481bfc/mathesar/api/exceptions/mixins.py#L64


The column id `999` does not exist and the queryset throws up an error correctly but when it gets passed to the error handling mixin, it gets handled incorrectly. You would need to fix the error handling mixin to convert `ManyRelatedField` type errors correctly

mathesar/api/serializers/constraints.py
```diff
@@ -129,3 +151,15 @@ def get_mapping_field(self, data):
         else:
             constraint_type = data.get('type', None)
         return constraint_type
+
+    def run_validation(self, data):
+        print(f'context === {self.context}')
```
silentninja, line 156: Please remove the `print` statement

mathesar/api/serializers/constraints.py
```diff
@@ -129,3 +151,15 @@ def get_mapping_field(self, data):
...
+        columns = data.get('columns', None)
+        if columns == []:
```
silentninja, line 158: Any reason why you are not setting the parameter `allow_empty=False` to the field for the empty value check

mathesar/api/serializers/constraints.py
```diff
@@ -121,6 +136,13 @@ class Meta:
 
     def create(self, validated_data):
         serializer = self.get_serializer_class(self.get_mapping_field(validated_data))
+        if serializer is None:
+            message = 'Unknown type passed.'
+            raise base_api_exceptions.NotFoundAPIException(
```
silentninja, line 141: Replace it with the correct error class, check `mathesar.api.exceptions.validation_exceptions.exceptions.UnknownDatabaseTypeIdentifier` for reference

mathesar/api/serializers/constraints.py
```diff
@@ -32,7 +42,12 @@ def create(self, validated_data):
         table = self.context['table']
         constraint_obj = self.construct_constraint_obj(table, validated_data)
         if constraint_obj is None:
-            raise ValueError('Only creating unique constraints is currently supported.')
+            message = 'Only creating unique constraints is currently supported.'
+            raise base_api_exceptions.ValueAPIException(
```
silentninja, line 46: Please replace this with correct error class

silentninja (changes requested):
Overall good, but I noticed that you are not using the correct error class when throwing up an error. Each error type should also have a unique error code so that the frontend can use the code to figure out the error that is thrown

Anish9901:
> I noticed that you are not using the correct error class when throwing up an error. Each error type should also have a unique error code so that the frontend can use the code to figure out the error that is thrown

Actually, the `mathesar.api.exceptions.validation_exceptions` module was not added to the codebase while I worked on this issue hence the wrong error class. Sorry for the delay.

mathesar/api/serializers/constraints.py
Anish9901: Since the `columns` parameter is a primary key-related field instead of a List field, it does not allow to have `allow_empty=False`. 
See: https://www.django-rest-framework.org/api-guide/relations/#primarykeyrelatedfield

mathesar/api/serializers/constraints.py
```diff
@@ -121,6 +134,8 @@ class Meta:
 
     def create(self, validated_data):
         serializer = self.get_serializer_class(self.get_mapping_field(validated_data))
+        if serializer is None:
+            raise UnknownDatabaseTypeIdentifier(db_type_id=None)
```
Anish9901, line 138: @silentninja I am getting an error: Error Response does not conform to the api spec. Please handle the exception properly 

During handling of the above exception ({'code': 4408, 'message': 'Unknown database type identifier None.', 'field': None, 'details': None}), another exception occurred

mathesar/api/serializers/constraints.py
```diff
@@ -129,3 +144,13 @@ def get_mapping_field(self, data):
...
         return constraint_type
+
+    def run_validation(self, data):
+        columns = data.get('columns', None)
+        if columns == []:
+            message = 'Columns field cannot be empty'
+            raise MathesarValidationException(
+                ValidationError,
+                message=message
+            )
```
Anish9901, line 155: @silentninja Also right here I am getting an error: 'ReturnList' object has no attribute 'get'
Is this occurring due to the constraint viewset?

mathesar/api/serializers/constraints.py
silentninja: I am not getting that error when I try to send an invalid `type`. Can you write some test cases to highlight the bug?

mathesar/api/serializers/constraints.py
silentninja: I am not able to reproduce this bug. I tried sending a constraints request with an empty column field and got the correct error response. Can you write some test cases to highlight the bugs

Anish9901: @silentninja The error seems to occur only while sending the request through the DRF API client. The response in the test seems correct.

Anish9901: @silentninja The error seems to occur only while sending the request through the DRF API client. The response in the test seems correct.

mathesar/tests/api/test_constraint_api.py
```diff
@@ -423,3 +423,35 @@ def test_drop_nonexistent_table(client):
...
+        'type': 'foo',
+        'columns': [1]
+    }
+    response = client.post(
+        f'/api/db/v0/tables/{table.id}/constraints/', data
+    )
+    response_data = response.json()[0]
+    assert response.status_code == 400
+    assert response_data['code'] == ErrorCodes.UnknownDBType.value
+    assert 'Unknown database type identifier' in response_data['message']
```
Anish9901, line 457: This test case is failing and the response_data for the sent request is `{'code': None, 'message': '', 'details': {'exception': "{'code': 'Unknown database type identifier foo.', 'message': 'None', 'field': None, 'details': None}"}}`

mathesar/api/serializers/constraints.py
Anish9901: Please see: https://github.com/centerofci/mathesar/pull/1406/files#r947495499

Anish9901:
@silentninja I am also thinking of fixing the following bugs in a later PR as I haven't figured out how to handle the [KeyError](https://github.com/centerofci/mathesar/pull/1406#issuecomment-1198305111) yet. I am changing the PR description accordingly.
- Non-existent column id, e.g.:

    ```json
    {"type":"unique","columns":[9999]}
    ```
and
- Attempting to set a unique constraint by supplying the id of a column that exists in a different table from the table whose id is supplied in the URL.

mathesar/api/serializers/constraints.py
silentninja: > error: Error Response does not conform to the api spec. Please handle the exception properly

It is a warning. You are getting it because the error object is incorrect and this is happening because a `ValidationException` is used inside a `create` method. Please use override the `validate` method and raise the exception there.

mathesar/api/serializers/constraints.py
silentninja: @Anish9901 Can you see if the frontend is facing that issue, it could be due to the content type of the request.

silentninja:
> I am also thinking of fixing the following bugs in a later PR as they are very closely related and I haven't figured out how to handle the https://github.com/centerofci/mathesar/pull/1406#issuecomment-1198305111 yet

Create an issue for it before you start working on it. I can point you to the source of the problem.

mathesar/api/serializers/constraints.py
Anish9901: Yes, it's a frontend-specific issue, I'll create an issue for it.

Anish9901:
Ah, it didn't occur to me that I should define specific exception classes instead of already written once.

silentninja (approved):

silentninja:
Thanks for the PR @Anish9901, I made some changes to error classes as it was a small fix and merged the PR.

PR #1458, opened 2022-07-12
Datetime grouping by parts

### LINKED ISSUE #429: Implement grouping options for date & time types
## Problem
Users may want to group their records with date and time fields by common intervals such as hours, months, quarters, years, etc. The appropriate grouping for a field will vary depend on the kind of data stored in that field. For example, if there are years ranging from 1500-2000 in the data, the user might want to group dates by century, but if the dates only range from 2008-2010, the user may want to group by month or quarter instead and centuries are useless.

We need to provide both useful grouping intervals for a given column and the ability to group by them.

## Solution
- We should support grouping by all valid date parts supported by Postgres' `EXTRACT` function. [Documentation](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT).
- We should provide useful grouping intervals for a given column in the Column API. 
- We should extend our record grouping mechanism to accept date parts for columns of Date & Time types. We should only accept valid date parts based on the type of column (e.g. we should not accept `century` for `TIME` columns).

This involves:
- Implementing the grouping in the backend
- Updating the `/api/v0/databases/<id>/types/` endpoint to store available grouping on this type

## Additional context
- Please see these PRs for code related to grouping:
  - #315
  - #353
- Complete implementation of this issue is blocked by implementation of Date & Time types.
  - #424 
  - #425 
  - #426
 - #557 provides some context on storing groups.

mathemancer:
Fixes #429 

This adds grouping by year, month, or day for any date/time types that include those parts. This also adds a separate grouping method that lets one group by any extractable date part (see below for details).

**Technical details**

### Method 1: simplified truncating

The first, simplified, grouping method utilizes the `preproc` functionality of grouping, and simply specifies some appropriate preprocessing functions to accomplish the desired part extraction. Note that I went with truncating rather than proper part extraction, so

```[2022-01-01, 2022-01-02, 2023-01-01, 2023-02-01]```

Would produce 2 groups when grouped by year, 3 groups when grouped by month, and then 4 groups grouped by day. This is in contrast to 2 groups by year, 2 groups by month, and 2 groups by day, which would be the effect of actually extracting date parts and grouping by those. I deemed the former logic to be more likely to be the intent of most users.

This should be called with a request like:

```
GET http://localhost:8000/api/db/v0/tables/7/records/?grouping={"columns": [23], "mode": "distinct", "preproc": "truncate_to_year"}
```
or
```
GET http://localhost:8000/api/db/v0/tables/7/records/?grouping={"columns": [23], "mode": "distinct", "preproc": "truncate_to_month"}
```
or
```
GET http://localhost:8000/api/db/v0/tables/7/records/?grouping={"columns": [23], "mode": "distinct", "preproc": "truncate_to_day"}
```

### Method 2: Extraction

This grouping method actually extracts a date (or time or interval) part based on a given parameter, and groups based only on that part. So,

```[2022-01-01, 2022-01-02, 2023-01-01, 2023-02-01]```

would produce 2 groups by year, 2 groups by month, and 2 groups by day.

This method can use any extractable date or time part listed in the [documentation](https://www.postgresql.org/docs/13/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT)

It should be called with a request like:

```
GET http://localhost:8000/api/db/v0/tables/7/records/?grouping={"columns": [23], "mode": "extract", "extract_field": "year"}
```

`extract_field` is the parameter giving the desired date part for grouping.

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
Would it be easy to add support for additional groupings? The original ticket called for additional groupings (e.g. decade, century, quarter, anything supported by [the `EXTRACT` function](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT)).

mathemancer:
> Would it be easy to add support for additional groupings? The original ticket called for additional groupings (e.g. decade, century, quarter, anything supported by [the `EXTRACT` function](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT)).

Yes, (but). The problems with that are two-fold.

First, we don't have (at this moment) a way to encode acceptable string arguments to a given function in the function info endpoint. Thus, we can't have (for example) a 2-argument function to describe both the desire to extract a date part, and _which_ date part to extract without the front end needing to hard-code every string that it wants to be able to extract. After talking a bit with @dmos62 , we decided to avoid creating a separate `DBFunction` for each date part, since we'd need to remove them all later once we have the descriptive infrastructure more fleshed-out.

The second problem is that simply extracting a date part and then grouping on it is a bit unintuitive, and less likely to be useful than the current truncating logic. In case my example from the description was unclear, extracting (for example) the month from a date and grouping on that will put all dates that happen to occur in December in the same group, regardless of their year. I.e., in the current logic, a "month" grouping is actually a year-month pair grouping.

As an interim step, I'll add a 2-arg extraction version so _if_ we want to hard-code the various strings describing date parts, and get the behavior described in the second paragraph, the functionality will be there. For the current cycle, I recommend the UI call the pre-fab year, month, and day truncating functions since those make more sense for the report generation described in the storyboard.

kgodey:
> The second problem is that simply extracting a date part and then grouping on it is a bit unintuitive

I agree with your decision here, I don't think we should naively group by date part.

I think other common groupings that users may want are:
- Second
- Minute
- Hour
- Quarter
- Decade

I assume those will be handled in subsequent PRs (maybe not in this cycle)?

mathemancer:
> > The second problem is that simply extracting a date part and then grouping on it is a bit unintuitive
> 
> I agree with your decision here, I don't think we should naively group by date part.
> 
> I think other common groupings that users may want are:
> 
> * Second
> * Minute
> * Hour
> * Quarter
> * Decade
> 
> I assume those will be handled in subsequent PRs (maybe not in this cycle)?

Long run, I want to improve the function description machinery to handle describing acceptable string arguments, then redo most of this work.

kgodey:
> Long run, I want to improve the function description machinery to handle describing acceptable string arguments, then redo most of this work. This would enable those.

Can you create a ticket to track this work?

PR #1472, opened 2022-07-19
Split table/Move columns without dropping the original Table

### LINKED ISSUE #1436: Add support for Moving columns to a new Table without dropping the original table
## Problem
Existing `extract_columns_from_table` and  `move_columns_between_related_tables` drops the original table and create a new table with remaining columns, which makes the table lose all its properties like constraints. 
## Proposed solution
Make sure the properties of a table are not lost either by not deleting the table or by properly migrating all the properties of the old table to the new table.

silentninja:
Fixes #1436 

- This PR extends upon #1430 and fixes the split_table and move columns functions to not drop the original table, thereby not losing the table's properties
- The PR also fixes the `split_table` and `move_columns` functions to use  `attnum` for columns and `oid` for tables instead of table and column names

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@silentninja It doesn't seem like this is currently working. Is that assessment correct? If so, what sort of feedback are you looking for from the review you requested?

silentninja:
That's right @mathemancer I accidentally opened it before marking it as a draft.

db/tables/operations/split.py
```diff
@@ -69,50 +58,43 @@ def _create_split_insert_stmt(old_table, extracted_table, extracted_columns, rem
...
-def extract_columns_from_table(old_table_name, extracted_column_names, extracted_table_name, remainder_table_name, schema, engine, drop_original_table=False):
+def extract_columns_from_table(old_table_oid, extracted_column_attnums, extracted_table_name, schema, engine,):
+    old_table_name = reflect_table_from_oid(old_table_oid, engine).name
     old_table = reflect_table(old_table_name, schema, engine)
     old_columns = (MathesarColumn.from_column(col) for col in old_table.columns)
     old_non_default_columns = [
         col for col in old_columns if not col.is_default
     ]
+    extracted_column_names = get_columns_name_from_attnums(old_table_oid, extracted_column_attnums, engine)
     extracted_columns, remainder_columns = _split_column_list(
```
mathemancer, line 79: `remainder_columns` is no longer accessed. I think you should consider removing the `_split_column_list` function, or simplifying it since there's no reason to create the list of remainder columns and pass it around with the new way of doing things.

db/tables/operations/split.py
```diff
@@ -17,33 +19,24 @@ def _split_column_list(columns_, extracted_column_names):
...
+    remainder_table_with_fk_key = reflect_table(remainder_table_name, schema, engine)
+    return extracted_table, remainder_table_with_fk_key, fk_column_name
 
 
-def _create_split_insert_stmt(old_table, extracted_table, extracted_columns, remainder_table, remainder_columns, remainder_fk_name):
+def _create_split_insert_stmt(old_table, extracted_table, extracted_columns, remainder_fk_name):
     SPLIT_ID = f"{constants.MATHESAR_PREFIX}_split_column_alias"
     extracted_column_names = [col.name for col in extracted_columns]
-    remainder_column_names = [col.name for col in remainder_columns]
     split_cte = select(
```
mathemancer, line 40: We should create a bug ticket.The creation of this CTE will fail in cases where the extracted columns aren't orderable (e.g., JSON). I realize this bug is carried over from my own work.

db/tests/tables/operations/test_split.py
```diff
@@ -22,18 +24,17 @@ def test_extract_columns_from_table_creates_tables(engine_with_roster, roster_ta
     t_dict = metadata.tables
     assert (
         f"{schema}.{teachers_table_name}" in t_dict
-        and f"{schema}.{roster_no_teachers_table_name}" in t_dict
     )
 
 
 def test_extract_columns_from_table_sets_up_one_fkey(extracted_remainder_roster):
-    extracted, remainder, _, _, _ = extracted_remainder_roster
+    extracted, remainder, _, _ = extracted_remainder_roster
```
mathemancer, line 31: `extracted` isn't accessed. While I'm not too fussy about that in tests, I think it's a bit confusing given that you're using `_` for other unaccessed return values.

db/tests/tables/operations/test_split.py
```diff
@@ -42,15 +43,15 @@ def test_extract_columns_from_table_sets_correct_reference(extracted_remainder_r
 
 
 def test_extract_columns_from_table_sets_correct_fkey(extracted_remainder_roster, roster_fkey_col):
-    extracted, remainder, _, _, _ = extracted_remainder_roster
+    extracted, remainder, _, _ = extracted_remainder_roster
```
mathemancer, line 46: `extracted` isn't accessed.

mathemancer (changes requested):
Overall, I think this looks good, but I have some requests:

- Please create an appropriate TODO or issue for the problem regarding assuming the appropriate join to use when moving the columns. This will need to be handled delicately, and probably require some design-level work to get right.
- I also noted some other bugs carried over from the prototype code; we should create issues for those as well.
- This PR seems to introduce a regression in the `db` library by removing the table-merging logic. Am I missing something? If not, I don't think we should remove that functionality from the library.
- See my specific note regarding `remainder_columns`.

db/tables/operations/split.py
silentninja: Created #1490 to track it

db/tables/operations/move_columns.py
```diff
@@ -19,76 +21,93 @@ def _find_table_relationship(table_one, table_two):
         if fkey_constraint.referred_table == table_one
     ]
     if one_referencing_two and not two_referencing_one:
-        relationship = {"referencing": table_one, "referenced": table_two}
+        relationship = {"referencing": table_one, "referenced": table_two, "constraint": one_referencing_two[0]}
     elif two_referencing_one and not one_referencing_two:
-        relationship = {"referencing": table_two, "referenced": table_one}
+        relationship = {"referencing": table_two, "referenced": table_one, "constraint": two_referencing_one[0]}
```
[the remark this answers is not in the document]
silentninja: Tracked by #1491

db/tables/operations/split.py
silentninja: Tracked by #1490

silentninja:
@mathemancer Retaining tables is messy compared to creating a new remainder table. I forgot to consider the situation when moving columns from the referrer table to a referent table could lead to data loss in case the moving column is not unique, in my initial implementation(before [b5bab71](https://github.com/centerofci/mathesar/pull/1472/commits/b5bab71ff170b4dee91e4441a1ba9c8de15694c6)]. I fixed it with [407ff28](https://github.com/centerofci/mathesar/pull/1472/commits/407ff2844b148d45ffba23ec4f6636b8da7892c0).

mathemancer (approved):
Okay, let's get this merged. Nice work!

PR #1486, opened 2022-07-28
First time setup script

### LINKED ISSUE #874: Makefile tasks for doing common tasks like first-time setup
## Problem
We have some sequences of commands that every developer has to run at one point or another. For example, for first time setup you run:

```
> cp .env.example .env
> docker-compose up -d
> docker exec mathesar_service python manage.py migrate
> docker exec -it mathesar_service python install.py
```

Last two might depend on docker-compose having finished setting up.

## Proposed solution
Would be nice to have a way to perform such tasks by running a single command, or just having a way not to have to remember the specific python commands or the container name, or that the `.env` file has to be instantiated.

## Additional context
Another task to automate would be reseting/purging development environment's state in case it got corrupted.

abhishak3:
Fixes #874 

**Technical details**
replaced all the commands used for first-time setup with a single "make init" command.

"make init" command is defined in the Makefile which executes a bash script "initial_setup.sh" with root privileges. The reason for running the script with root privileges is that the docker needs root privileges to run, otherwise it shows a "Permission denied" error. This error is user-specific and depends on the way docker is configured on their system.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


initial_setup.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# runs docker compose as daemon process
+docker-compose up -d
+error_check
+
+# loop until http://0.0.0.0:8000/ is not active and
+# timeout does not occurs
+while [[ ${status_code} -ne 302 && ${SECONDS} -lt ${TIMEOUT_TIME} ]];
+do
+    status_code=$(curl --write-out %{http_code} --silent --output /dev/null ${URL})
```
dmos62, line 29: What's `http_code` here? Is that a typo for `status_code`?

initial_setup.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# if timeout occurs
+if [[ ${SECONDS} -ge ${TIMEOUT_TIME} ]]; then
+    echo "TIMEOUT OCCURRED"
+    exit
+fi
+
+# run migrations
+docker exec mathesar_service python manage.py migrate
+error_check
```
dmos62, line 40: These error/panic checks don't provide any feedback about which command failed. What I often do is something like `cmd_to_run || do_this_in_case_of_failure "this parameter can provide debugging information"`.

initial_setup.sh
```diff
@@ -0,0 +1,45 @@
...
+        echo "FAILED"
+        exit
+    fi
+}
+
+# instantiate .env file
+cp ${ENV_EXAMPLE} ${ENV_ORIG}
+
+# runs docker compose as daemon process
+docker-compose up -d
```
dmos62, line 22: I think it's not good to leave a daemon running after the one-time initialization is finished. Or, we could print a notice to STDOUT that a docker-compose daemon is left lingering.

initial_setup.sh
dmos62: To provide more feedback to a user, it might be good to print notices about each action being performed as they're being performed. That would also take care of the "where did this panic happen" problem.

Makefile
```diff
@@ -7,11 +7,16 @@ SPHINXOPTS    ?=
...
 SOURCEDIR     = docs
 BUILDDIR      = docs/_build
+INITSCRIPT	  = initial_setup.sh
 
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+# Initial setup
+init:
```
dmos62, line 17: `initial_setup` might be a better name. More specific.

Makefile
```diff
@@ -7,11 +7,16 @@ SPHINXOPTS    ?=
...
+	@sudo bash "$(INITSCRIPT)"
```
dmos62, line 18: I think `sudo` is not a good idea here. Docker can be setup to not need superuser rights (that's how I set it up, for example), and even if it's not, it's better if the user prepends his script/make call with a `sudo` himself. `docker-compose` will tell the user if he doesn't have sufficient permissions.

dmos62 (changes requested):
Hey @abhishak3, thanks for the PR! I requested some changes.

initial_setup.sh
abhishak3: Refer to [curl's man page](https://man7.org/linux/man-pages/man1/curl.1.html). See the `--write-out` option.
>**http_code**
> The numerical response code that was found in the last retrieved HTTP(S) or FTP(s) transfer.

initial_setup.sh
abhishak3: These error checks don't hide the errors. It's just the way of exiting the script if there's an error so that further lines won't execute.

initial_setup.sh
abhishak3: I think it's best not to start the docker-compose as a daemon. Starting it as a  foreground process will be more verbose which will be more helpful in case of any unexpected error.
So, maybe we can remove the `-d` option?

Makefile
abhishak3: Yeah, I was confused on this one. Thanks for clearing it out for me. Will do that.

abhishak3:
@dmos62 I would like to know what should be the timeout limit for the script. For the first-time setup, it can take several minutes to hours depending on the system. So, should we even keep the timeout or completely remove it?

initial_setup.sh
dmos62: I realise that. My concern is about not immediately knowing which script-line/command failed.

initial_setup.sh
dmos62: Would you run it in the background with its output directed to stdout? Maybe just request that `docker-compose up` be already running for this script to proceed.

dmos62:
@abhishak3 hm, I'm not very familiar with the container startup process, but I was under the impression that you don't actually need Django ready to serve requests to perform migrations and run the installation script.

Try it out without the timeout. If it turns out to be necessary, I'm guessing that a 60 second timeout is good enough for now.

abhishak3:
@dmos62 I've made the requested changes. Please check 'em out.

initial_setup.sh
```diff
@@ -0,0 +1,34 @@
...
+URL="http://0.0.0.0:8000/"
+status_code=000
+ENV_EXAMPLE=".env.example"
+ENV_ORIG=".env"
+#TIMEOUT_TIME=60
+
+exit_on_error () {
+    echo "FAILED AT: $@"
+    exit
+}
```
dmos62, line 12: This might be simpler:

```
function wrap {
  # Echo what the command is before running it
  echo "$@"
  # Run the passed command
  eval "$@"
  local eval_status=$?
  # Enable panicing this function's parent in case of non-zero exit
  set -e
  # Do non-zero exit, if eval's return was non-zero
  [ $eval_status -eq 0 ] || (echo "failed" && exit 1)
  # Disable panicing this function's parent in case of non-zero exit
  set +e
}

```

Usage: 

```
wrap docker exec mathesar_service sh -c "python manage.py migrate && python install.py"
```

`wrap` might not be the best name, feel free to change it.

initial_setup.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# instantiate .env file
+cp ${ENV_EXAMPLE} ${ENV_ORIG}
+
+# runs docker compose as daemon process
+docker-compose up -d || exit_on_error "docker-compose up -d"
+
+#while [[ ${status_code} -ne 302 && ${SECONDS} -lt ${TIMEOUT_TIME} ]];
```
dmos62, line 20: When I suggested disabling the timeout, I actually meant to suggest not to wait for the HTTP server to start serving. I think you can start running migrations and the install script much earlier.

dmos62 (changes requested):

initial_setup.sh
abhishak3: I cannot run migrations and install scripts until Django and all other requirements have been installed from the `requirements.txt` and `requirements-dev.txt` in the container.
According to my understanding, Django will not take much time to start but the installation of dependencies does. That's why I am waiting for Django to start the HTTP to ensure that the container is up and running with all dependencies installed.

dmos62: You might consider something like [pip3 -vvv freeze -r requirements.txt | grep "not installed"](https://stackoverflow.com/a/64949020) to check if pip dependencies have been installed.

If that doesn't work, checking the HTTP service is acceptable. Django server takes a long time to get ready on my machine, that's why I'm looking for a closer approximation of when the container is ready for migrations.

seancolsen (changes requested):
I think GitHub automatically requested a review from me for this PR because it modifies front end files within `mathesar_ui`. @abhishak3 Can you remove the changes to `package-lock.json` from this PR? That file gets changed sometimes by Docker when starting up, but we shouldn't put those changes in this PR.

kgodey:
Assigning this to @mathemancer since @dmos62 is out sick.

abhishak3:
> I think GitHub automatically requested a review from me for this PR because it modifies front end files within `mathesar_ui`. @abhishak3 Can you remove the changes to `package-lock.json` from this PR? That file gets changed sometimes by Docker when starting up, but we shouldn't put those changes in this PR.

removed changes to the `package-lock.json` from the PR.

dmos62:
@abhishak3 thanks for the contribution! I'm merging this in. I'll make a few changes to the blocking conditions in a separate PR. I'll post a link.

dmos62 (approved):

Issue #1410, opened 2022-06-01
Implement JSON filtering


Jinxiao0302:

seancolsen:
Closing this since it's not something we're planning to do soon and there's not much detail in this ticket.

PR #1493, opened 2022-08-02
Implement JSON array filtering functions

Jinxiao0302:
Related to  #1410

**Technical details**
Added a DBFunction for getting JSON array length;
To be done: Add comparaing functions that nest array length function.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/functions/hints.py
```diff
@@ -121,9 +121,16 @@ def _is_hint_applicable_to_types(hint):
...
 literal = _make_hint("literal")
 
+
 json = _make_hint("json")
 
 
+json_array = _make_hint("json_array")
+
+
+json_object = _make_hint("json_object")
```
dmos62, line 131: You'll want to tell Mathesar how these hints relate to actual types. Look at `db/types/hintsets.py`. Note this line:

```
    _add_to_db_type_hintsets(categories.JSON_TYPES, (hints.json,))
```

This says that all database types defined as `categories.JSON_TYPES` are associated with the hints `(hints.json,)`.

You'll probably want to make changes to this. And, to the related categories in the `db.types.categories` as as well, since currently `categories.JSON_TYPES` is a catch-all category for everything JSON, and you probably want something more granular.

Note how the `categories` sets define a sort of hierarchy, where for example time-related types are made up of point-in-time as well as duration types, those in turn are made up of other more granular categories, etc. You might want something like this for JSON, though I wouldn't invest too much time into that right now: the mentioned hierarchy was borne out of necessity.

By the way, if this wasn't clear before (probably not), the way Mathesar finds what hints are associated with what UI type, is it takes the intersection of the UI type's DB types' hints (a UI type is effectively a set of DB types, and each of those have a set of hints, as seen in `db/types/hintsets.py`); and, the way it checks whether a filter is applyable to a column of a given UI type, is that it checks if the hints of the first parameter of that filter are satisfied by the hints of the UI type. "Satisfied" here is intentionally ambiguous, since we want the freedom to evolve how this works, but at the moment this mostly means that the first param hint-set must be a subset of the UI type hint-set.

Oof, sorry for the wall of text. Thumbs up for braving through :+1::+1:, feel free to reach out.

db/functions/base.py
```diff
@@ -306,6 +307,19 @@ class ToLowercase(DBFunction):
...
         return func.lower(string)
 
+class ArrayLength(DBFunction):
+    id = 'array_length'
+    name = 'length'
+    hints = tuple([
+        hints.returns(hints.comparable),
+        hints.parameter_count(1),
+        hints.parameter(0, hints.json_array),
+        hints.mathesar_filter,
```
dmos62, line 317: This isn't actually a filter. You'll probably use this function to compose a `DBFunctionPacked`-based filter and you'll give that a `mathesar_filter` hint, but giving this function here a `mathesar_filter` will result in unexpected behaviour.

dmos62 (commented):
Regarding your question about how to use the `ArrayLength` to compose a filter like `json array length is less than` (I presume that that's a valid approximation of your goal), create a `DBFunctionPacked` subclass inside `db.types.custom.json_array`. `DBFunctionPacked` (and the whole of `hints.mathesar_filter` mechanism, really) is there to bypass the lack of composition in our current UI filtering system.

db/functions/base.py
Jinxiao0302: Thanks for your very helpful explanation! @dmos62 , I added a DBFunctionPacked function using ArrayLength in the `db/types/custom/json_array` and also moved the ArrayLength into that file. Also, I added the json array into categories.

mathemancer (changes requested):
Hey @Jinxiao0302 , as noted through other channels I'd like you to add some tests before we merge this PR. Thank you, it's looking good so far!

pavish (approved):
The changes look good to me. As @mathemancer mentioned, we can merge this in once we have tests. I'm approving from my end.

mathemancer (approved):
Okay, I think this looks good. Let's get it merged!

PR #1505, opened 2022-08-07
Constraints based column icons

### LINKED ISSUE #1485: Set special column icons based on constraints
## Current behavior

- The column header displays an icon to the left of the column name.
- The icon changes depending on the column type.

## Desired behavior

- The icon _also_ changes depending on the _constraints_ associated with the column.
- If the column has a PK constraint, then it gets a [key](https://fontawesome.com/icons/key?s=solid) icon.
- If the column has a FK constraint, then it gets a [link](https://fontawesome.com/icons/link?s=solid) icon.
- Otherwise, the existing logic applies, giving the column a type-dependent icon.

rajatvijay:
Fixes #1485

This PR enhances the table column headers to show icons based on constraints (if any) left to the column name. 

**Technical details**

This PR touches two core parts
1. Enhances the `ColumnName` cell to take `ProcessedColumn` as prop instead of just `Column`, this enables it to be aware about the constraints on this column. 
2. Enhances the `getColumnIconProps` utility to take into considerations the FK and PK constraints on the column in deciding the icon

**Screenshots**

Screenshot 2022-08-07 at 5 02 42 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/components/ColumnName.svelte
```diff
@@ -1,16 +1,16 @@
 <script lang="ts">
-  import type { Column } from '@mathesar/api/tables/columns';
+  import type { ProcessedColumn } from '@mathesar/stores/table-data/processedColumns';
   import { getColumnIconProps } from './cell/utils';
   import TypeIcon from './TypeIcon.svelte';
 
-  export let column: Column;
+  export let column: ProcessedColumn;
```
rajatvijay, line 6: @pavish I saw you branch `data_explorer_ui_2` and realised that this is the only place where it will have actual conflict. Please take a look at this and let me know if you want to take this approach or the one in your branch. 

Explicitly mentioning the different here: 

This PR
`export let column: ProcessedColumn;`

Your branch
`export let column: CellColumnLike & { name: string };`

PS: Not that we will need all the constraints data here

pavish: We cannot use `ProcessedColumn` here because this component is common to Tables & Queries, and queries have a very different column structure compared to tables.

It would be better to extend the type or add additional props to pass down to `getColumnIconProps`

mathesar_ui/src/components/cell/utils.ts
```diff
@@ -70,9 +72,28 @@ export function getDbTypeBasedInputCap(
...
+  );
+
+  if (hasFKConstraints) {
+    return iconTableLink;
+  }
+
+  if (hasPKConstraint) {
+    return iconConstraint;
+  }
+
```
pavish, line 94: As mentioned on the previous comment, we shouldn't be passing down `ProcessedColumn` to the `Cell` and it's related components and utils. Please refer `getCellCap` in the same file which gets `linksToTable` as an optional argument to determine if there is a link present.

It is better to accept an optional argument in the method, for eg,. `linkType?: 'inward' | 'outward'` (or you could come up with a better term), and use that argument to determine the icon.

pavish (changes requested):

seancolsen (approved):
These These changes work for me, so I'm approving. Thanks, @rajatvijay! I like your I'm ready to merge once the concerns from @pavish are addressed.

For what it's worth, on a call I recommended to Rajat that he change `getColumnIconProps` to accept a `ProcessedColumn`, so you can blame me! 😬

Personally, I find types like `CellColumnLike & { name: string };` pretty hard to read. I'd love it if we could come up with more descriptive names for these types. Maybe we need some kind of `AbstractColumn` or something? But I don't want to hold up _this_ PR for the sake of cleaning up those types and names. I feel like there's still a lot of dust to settle with all the changes we're making, and I'm hoping we can do some refactoring to address readability once we can see the big picture more clearly after all the features in this cycle are completed.

I added a couple very minor commits...

- In e4b2ef6211e7cdffc46de92a3ada480d16b25523, I adjusted the logic slightly so that multi-column foreign key constraints will not affect the column icons. The decision of whether or not to consider multi-column FK constraints here feels very borderline to me, but I have a hunch it might be wise if we disregard them for the time being. Currently, our approach to multi-column FK constraints is broadly: "we don't support them within the UI". Our UX design for them is basically undefined, so my approach has been to ignore them. Apologies that I didn't mention this detail in the ticket.

- In fee6e76c60a06b0d0f1eb5c772696195f8876c14, I made a very small nitpick adjustment for readability. Here's my rationale... Before my change, the type of `hasFKConstraints` was `Constraint | undefined` -- but when reading the variable name, I'd expect its type to be `boolean`. Not a big deal.

rajatvijay:
Closing this since it the fixes requires major changes after discussing with Pavish

PR #1506, opened 2022-08-08
Implementation for column mappings

### LINKED ISSUE #1508: Add support for column mappings for importing records from one table to another.
## Description

To import records from one table to another having different column names, types, etc, we need the ability to specify which column of the temp/import table should map to which specific columns of the target table in which the records are to be imported.  

## Expected behavior

To achieve the above we need support for column mappings. Which will be set by the users and sent to the backend with a request to the `/existing_import` like this:
```
{
    "import_target": <target_table_id>,
    "data_files": [<data_file_id>],
    "mappings": [
                    [<temp_table_col1_id>,<target_table_col1_id>],
                    [<temp_table_col2_id>,<target_table_col2_id>]
                ]
}
```

Anish9901:
Fixes #1508 

This PR introduces the implementation of column mappings which will be used as a part of the import process while the records are to be imported from one table to another so that a user can exactly specify which column from the import file/temp_table should map to which specific columns of the target table.

**Technical details**
```
POST /api/db/v0/tables/<temp_table_id>/existing_import/
```
**Request**
```
{
    "import_target": <target_table_id>,
    "data_files": [<data_file_id>],
    "mappings": [
                    [<temp_table_col1_id>,<target_table_col1_id>],
                    [<temp_table_col2_id>,<target_table_col2_id>]
                ]
}
```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/records/operations/insert.py
```diff
@@ -76,10 +76,15 @@ def insert_records_from_csv(table, engine, csv_filepath, column_names, header, d
...
+def insert_from_select(from_table, target_table, engine, sa_col_mappings=None):
+    if sa_col_mappings:
+        # unzipping column mappings into list of respective tables
+        from_table_col_list, target_table_col_list = zip(*sa_col_mappings)
+    else:
+        from_table_col_list = [col for col in from_table.c if not MathesarColumn.from_column(col).is_default]
+        target_table_col_list = [col for col in target_table.c if not MathesarColumn.from_column(col).is_default]
     with engine.begin() as conn:
-        sel = select([col for col in from_table.c if not MathesarColumn.from_column(col).is_default])
+        sel = select(from_table_col_list)
```
Anish9901, line 87: @mathemancer during the execution of this part of the code the select statement generates a warning when the `from_table_col_list` from the if block is passed, 
SAWarning: SELECT statement has a cartesian product between FROM element(s) "public.books2" and FROM element "public.books2".  Apply join condition(s) between each element to resolve.
This then causes an error in the formation of the SQL statement to be executed:
```sql
ERROR:  table name "books2" specified more than once
STATEMENT:  INSERT INTO public.books (title, author) SELECT public.books2.name, public.books2.author_name FROM public.books2, public.books2
```
Notice the extra `public.books2` at the end of the SQL statement.

The `from_table_col_list` from the else block works as expected.

mathesar/models/base.py
```diff
@@ -434,9 +434,13 @@ def insert_records_to_existing_table(self, existing_table, data_files, mappings=
         from_table = self._sa_table
         target_table = existing_table._sa_table
         engine = self._sa_engine
+        if mappings:
+            sa_col_mappings = [[from_col._sa_column, target_col._sa_column] for from_col, target_col in mappings]
```
mathemancer, line 438: Each of these `_sa_column` attributes comes with its own SQLAlchemy `Table` instance. Just pass the names instead (see the other comment)

db/records/operations/insert.py
mathemancer: Each column from the `Column` model carries its own SQLAlchemy `Table` instance. This is a problem, since SQLAlchemy can't determine automatically that those table objects refer to the same underlying database table. Thus, it generates a `FROM` statement for each.

You should use the names from the columns and get them from the passed table instance instead (as you do in the `else` block).

mathemancer (commented):
I added a couple of hints to fix the problem you're having with the cross product join.

db/records/operations/insert.py
Anish9901: That makes sense, I was wondering why that behavior was happening, as in printing the column instance also showed no change `(Column('name', TEXT(), table=<books2>), Column('author_name', TEXT(), table=<books2>))`.

db/records/operations/insert.py
```diff
@@ -76,10 +76,14 @@ def insert_records_from_csv(table, engine, csv_filepath, column_names, header, d
                     cursor.copy_expert(copy_sql, temp_file)
 
 
-def insert_from_select(from_table, target_table, engine, mappings=None):
-    target_table_col_list = [col for col in target_table.c if not MathesarColumn.from_column(col).is_default]
+def insert_from_select(from_table, target_table, engine, col_mappings=None):
+    if col_mappings:
+        from_table_col_list, target_table_col_list = zip(*[(from_table.c[from_col], target_table.c[target_col]) for from_col, target_col in col_mappings])
+    else:
+        from_table_col_list, target_table_col_list = zip(*[(from_col, target_col) for from_col, target_col in zip(from_table.c, target_table.c) if not MathesarColumn.from_column(from_col).is_default and not MathesarColumn.from_column(target_col).is_default])
```
mathemancer, line 83:
```suggestion
        from_table_col_list = [
            col for col in from_table.columns
            if not MathesarColumn.from_column(col).is_default
        ]
        target_table_col_list = [
            col for col in target_table.columns
            if not MathesarColumn.from_column(col).is_default
        ]
```

Why tie these two together? I'm not sure the dependency implied by only avoiding mapping a column if both ends of the map are default is appropriate.

db/records/operations/insert.py
```diff
@@ -76,10 +76,14 @@ def insert_records_from_csv(table, engine, csv_filepath, column_names, header, d
                     cursor.copy_expert(copy_sql, temp_file)
 
 
-def insert_from_select(from_table, target_table, engine, mappings=None):
-    target_table_col_list = [col for col in target_table.c if not MathesarColumn.from_column(col).is_default]
+def insert_from_select(from_table, target_table, engine, col_mappings=None):
+    if col_mappings:
+        from_table_col_list, target_table_col_list = zip(*[(from_table.c[from_col], target_table.c[target_col]) for from_col, target_col in col_mappings])
```
mathemancer, line 81:
```suggestion
        from_table_col_list, target_table_col_list = zip(
            *[
                (from_table.c[from_col], target_table.c[target_col])
                for from_col, target_col in col_mappings
            ]
        )
```
I find this a little easier to read.

mathemancer (changes requested):
Overall, looking pretty good. I have a request in the logic of the list comprehension when no mapping is passed.

mathemancer (approved):
Okay, I think this is looking good.

PR #1504, opened 2022-08-05
Navigation

seancolsen:
## Summary

- This PR works towards implementing the [Navigation specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/navigation.md).
- All work required to implement the spec should either be: (a) in this PR, (b) tracked in a separate named feature, or (c) tracked within a smaller ticket as written below, ready to create.
- After merging this PR and creating the tickets spec'ed below, I'd like to mark "Navigation" as 100% complete on the features spreadsheet, because the smaller ticket should be lower-priority tasks that might not necessarily get completed this cycle.

## In this PR

- Routing changes, both server-side and client-side.
- New header with breadcrumb-like navigation
- No more tabs (lots of code ripped out!)
- Table Page doesn't put filter/sort/group/pagination into the URL (it will be added back via ticket)
- Data Explorer route is moved to `http://localhost:8000/mathesar_tables/1/data-explorer`.

    @pavish This will need more work. For example, the Data Explorer page needs to be set within LayoutWithHeader so that it gets the AppHeader. I didn't want to touch the data explorer files within this PR because those edits would introduce merge conflicts with your work. I'd like you to give specific attention to the Data Explorer routing, ideally after merging this PR. But if this PR needs changes before merge to address Data Explorer regressions, I'd love it if you could push more commits here.

- Record Page is added at routes like: `localhost:8000/mathesar_tables/1/1/99`. No links point here yet, but you'll need to visit this route to fully review the header functionality because it's the only route which displays the full breadcrumb path.

## To be done via separate features

- Page content at `http://localhost:8000/mathesar_tables/` will be completed within the **Database Page** feature
- Page content at `http://localhost:8000/mathesar_tables/1/` will be completed within the **Schema Page** feature. Notable problems with the current page within this PR:
    - No way to create a new empty table
    - No loading indicator when tables are fetching
    - No way to add a new Exploration or navigate to the Data Explorer.
- Anything ugly will be made pretty within the **Styling** feature.

## To be done via smaller tickets

<details>
<summary><h3>Improve BreadcrumbSelector component</h3></summary>

`cat | gh issue create -F - -t 'Improve BreadcrumbSelector component' -m 'Cycle 3: Demo Video' -l "status: ready" -l "type: enhancement" -l "work: frontend" -l "restricted: new maintainers"`

> ## Current behavior
> 
> - This is the `BreadcrumbSelector` component:
> 
>     image
> 
[...]

PR #1517, opened 2022-08-11
Reinstate functionality to synchronize the URL with the table's grouping/sorting/filtering/pagination

seancolsen:
## Before

- The synchronization of a table's grouping/sorting/filtering/pagination with the URL was broken as part of #1504

## After

- Writing the URL works -- the URL updates when grouping/sorting/filtering/pagination changes
- Reading the URL works -- grouping/sorting/filtering/pagination is set on first page load based on the URL
- It uses the URL hash now. That seemed more appropriate than a query param. I can change it back if needed.
- It disappears entirely when the grouping/sorting/filtering/pagination all have default values. I'm excited about this! Simpler URLs!
- It uses more declarative logic whereas we were previously doing this somewhat more imperatively.
- I also did some refactoring along the way:
    - I removed TabularType because it seems like we'll probably never use it
    - I simplified TabularDataProps a bit

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/AppTypes.ts
+++ b/mathesar_ui/src/AppTypes.ts
@@ -16,11 +16,6 @@ export interface SchemaEntry extends DBObjectEntry {
   has_dependencies: boolean;
 }
 
-export interface ViewEntry extends DBObjectEntry {
-  // TODO: Temporary, update when view endpoints are ready.
-  columns: { name: string }[];
-}
-
 export interface SchemaResponse extends SchemaEntry, TreeItem {
   tables: DBObjectEntry[];
 }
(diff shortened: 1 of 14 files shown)


pavish (approved):
@seancolsen The changes look good to me.

> It uses the URL hash now. That seemed more appropriate than a query param.

Hashes are not visible to the server. Here are some reasons why that would affect us:
* They are also not persisted when a user session ends and the user logs in again, because the redirection is handled by the server.
* We may want the server to be aware of the filter/group/sort information for sharing, or any preloading in the future.

Since we'd probably have to revisit this when we're working on the sharing functionality, I'm approving this PR now.

PR #1538, opened 2022-08-18
Fix DB Query joining for demo

### LINKED ISSUE #1537: DB Query joining for demo video broken
## Description

Under certain circumstances, the auto-joining behavior of the DB query infrastructure doesn't work. The error thrown is:

```
NoReferencedTableError: 
Foreign key associated with column 'Publications.Publisher id' could not find table 'Library Management.Publishers' with which to generate a foreign key to target column 'id'
```

## Expected behavior

The joining should always work.

## To Reproduce

Using the library data set, Create a DB query that would create an overdue book report. In particular, you should join the `"Checkouts"` table with the `"Publications"` table via the `"Items"` table. Note that when you try to get records from such a query, it will throw the above error.

## Additional context

It's important to solve this, since the described join is part of our demo video.

mathemancer:
Fixes #1537 

This solves the bug by passing OIDs and attnums around rather than SQLAlchemy representations of DB objects.

**Technical details**

When we pass SQLAlchemy objects around, a number of problems can occur:
- They can be out of sync with the DB state
- They can be completely detached from the DB state
- They can be unaware of each other if they don't share their metadata instance

I'm starting to think we should treat SQLAlchemy class instances representing DB objects as radioactive, and generally avoid letting them live any longer than absolutely necessary.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/queries/base.py
+++ b/db/queries/base.py
@@ -35,13 +35,14 @@ class DBQuery:
         # then apply temporary transforms on it, like how you can apply
         # temporary transforms to a table when in a table view.
         return records_select.get_records(
-            table=self.sa_relation,
-            **kwargs,
+            table=self.transformed_relation, engine=self.engine, **kwargs,
         )
 
     # mirrors a method in db.records.operations.select
     def get_count(self, **kwargs):
-        return records_select.get_count(table=self.sa_relation, **kwargs)
+        return records_select.get_count(
+            table=self.transformed_relation, engine=self.engine, **kwargs,
+        )
 
     @property
     def sa_output_columns(self):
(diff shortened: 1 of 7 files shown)


silentninja:
It works for me, I am able to do transitive summaries now.

silentninja (approved):

silentninja:
@mathemancer Looks like I approved it too soon. The foreign key summary preview is still failing

db/queries/base.py
```diff
@@ -1,203 +1,145 @@
...
+                left_col = left.columns[_get_column_name(jp[0][0], jp[0][1])]
+                right_col = right.columns[_get_column_name(jp[1][0], jp[1][1])]
+                from_clause = from_clause.join(
+                    right, onclause=left_col == right_col, isouter=True,
+                )
+
+            return right.columns[col_name].label(col.alias)
+
+        stmt = select(
+            [_process_initial_column(col) for col in self.initial_columns]
```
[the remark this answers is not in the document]
mathemancer: The inner function can't return the join statement. It has to build on the same global cache. Otherwise you end up with duplicates (or you need another piece of logic to remove them). This is why I went with the kind of "hybrid approach" with the side effect for the join, and the return for the column. I suppose I could make both side effects (i.e., also build the select column list via a side effect), but I'd rather minimize mutating things all over the place.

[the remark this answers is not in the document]
kgodey: A comment explaining that in the codebase might be helpful for the future.

mathemancer:
> @mathemancer Looks like I approved it too soon. The foreign key summary preview is still failing. Please take a look at
> 
> https://github.com/centerofci/mathesar/blob/38c2c6e46c90962c28e3ecff12479727958562cd/mathesar/tests/api/test_record_api.py#L711
> 
> I am still getting the `Foreign key table not found error`

I can't replicate this. I've uncommented the according line in he test and updated it accordingly, and it seems to pass. What am I missing?

silentninja:
The query part seems to be working now, looks like docker was still using the old code. We can merge it now

I don't know how the test case passed because it was still expecting the old template while the backend was sending a different template because the correct template is now uncommented.

I fixed the test cases and they seem to be passing now.

silentninja (approved):
Thanks for the PR, looks good to me.

db/queries/base.py
```diff
@@ -1,203 +1,145 @@
...
+            transformations=None,
+            name=None,
     ):
-        assert isinstance(base_table, SATable)
-        self.base_table = base_table
+        self.base_table_oid = base_table_oid
         for initial_col in initial_columns:
             assert isinstance(initial_col, InitialColumn)
         self.initial_columns = initial_columns
+        self.engine = engine
```
dmos62, line 23: I somehow feel uneasy about engine being an attribute of `DBQuery`. Can't articulate further. Did you make this change just for convenience, or was it some overarching principle?

db/queries/base.py
```diff
@@ -1,203 +1,145 @@
...
 
     # mirrors a method in db.records.operations.select
     def get_records(self, **kwargs):
-        # NOTE how through this method you can perform a second batch of transformations.
-        # this reflects fact that we can form a query, and then apply temporary transforms on it,
-        # like how you can apply temporary transforms to a table when in a table view.
+        # NOTE how through this method you can perform a second batch of
+        # transformations.  this reflects fact that we can form a query, and
+        # then apply temporary transforms on it, like how you can apply
+        # temporary transforms to a table when in a table view.
```
dmos62, line 32: I noticed that you reformatted comment lines to something like 80 characters. I've been using 100 character lines for comment blocks (and 80 for code). I thought I got that from PEP-8, but a quick glance didn't reveal such a recommendation there. Should I format comments to 80 chars?

dmos62 (commented):
Not requesting changes. Have questions and observations though.

db/queries/base.py
mathemancer: I don't think it matters. Emacs does this automatically when I edit functions under certain circumstances, and I didn't take the time to undo those changes. Generally, I don't think this sort of formatting question is for humans, so I kind of let my editor do what it wants.

db/queries/base.py
mathemancer: I realized that the changes resulted in every public method of the class requiring an engine, and since we have a global engine anyway, it seemed reasonable to just instantiate with it. I didn't see a use case for passing different engines to different methods of the same query, since they should all use the same connection string for a given "conceptual" query, anyway.

db/tests/queries/test_initial_relation.py
```diff
@@ -4,98 +4,117 @@
...
-                    right_column=uni_table.c.id,
-                ),
+                [
+                    (art_oid, get_attnum(art_oid, 'primary_author', engine)),
+                    (acad_oid, get_attnum(acad_oid, 'id', engine)),
+                ],
+                [
+                    (acad_oid, get_attnum(acad_oid, 'institution', engine)),
+                    (uni_oid, get_attnum(uni_oid, 'id', engine)),
+                ]
```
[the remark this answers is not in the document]
mathemancer: Point taken. I'm not opposed to iterating on giving meaningful names (via classes) to various data structures now that things seem to be working. A `JoinParam` instance can as easily be a pair of columns represented by `(oid, attnum)` pairs rather than SA columns, I suppose.

Issue #473, opened 2021-07-22
General styling updates


kgodey:
## Problem

Currently, we've been focusing Mathesar design work on functionality and user interactions. The visual style is pretty basic and should be improved.

## Proposed solution

We should update the design that we have for Mathesar so far with improved visual styles. There aren't a lot of hard requirements, and this issue involves taking some time to focus on how Mathesar looks.

### The solution requires the following aspects of the UI to be defined:
- The different themeable attributes that can be modified for the theme (e.g., background color, table borders)
- UI layout definitions: Define how the structure that supports the interface components should be styled and how they should behave in different states and resolutions (e.g., visible, hidden, minimized)
- UI component definitions: Define how the various interactive components should look like and behave in different states (e.g., empty, active, error)

## Additional context

- More color and contrast would be great!
- We should design Mathesar to be theme-able in the future so that people can change the colors for a different look.

## Components to style
- [x] Top navigation
- [x] Table and Table Toolbar
- [x] Schema and Database Pages
- [x] Data Explorer
- [ ] Record Page
- [ ] Data Import
- [ ] Table Inspector
- [ ] Icons and Buttons
- [ ] Actions Pane
- [ ] Record Selector

kgodey:
We plan to do this after the design for the [11. User Management](https://github.com/centerofci/mathesar/milestone/17) milestone, so marking as blocked until then.

kgodey:
I'm going to close this issue since we're covering this the various "demo readiness" issues like #1604.

PR #1543, opened 2022-08-18
Style Table

ghislaineguerin:
Related to #473

This PR updates styling for:

- Table headers
- Icon + Name contrast in making the icon color slightly lighter
- Status pane

Notes:
- A minimum width CSS property for dropdown toggle was deleted since it caused the "Add Column" button icon to be misaligned.
- This PR partially replaces some css properties values with variables. Replacing the variables is an ongoing task and will be addressed as part of general styling updates.


mathesar_ui/src/components/column/ColumnName.svelte
```diff
@@ -31,4 +31,12 @@
...
-<NameWithIcon on:click {icon} {isLoading}>{column.name}</NameWithIcon>
+<div class="column-name">
+  <NameWithIcon on:click {icon} {isLoading}>{column.name}</NameWithIcon>
+</div>
+
+<style>
+  .column-name {
+    padding: 0 0.5rem;
+  }
+</style>
```
pavish, line 42: ColumnName is a common used in several places throughout the applications. There are other components such as TableColumn and SelectableColumn which compose over it and are used in other places.

Please be mindful when modifying the dom structure and styling of such components. Here are some of the places which are affected by this change.

Record selector:
Screenshot 2022-08-22 at 11 03 05 PM

Data explorer column selection pane:
Screenshot 2022-08-22 at 11 04 14 PM

Data explorer result header:
Screenshot 2022-08-22 at 11 19 09 PM

Notice the issues with uneven padding and text overflow. Another place where this component is used while creating an unique constraint.

pavish (changes requested):
@ghislaineguerin The changes look fine to me with respect to the table. I've added a couple comments where they affect other places where the common components are used.

Since the regressions are only in a few places, I'd prefer fixing them in this PR itself. But if you'd like to handle them incrementally, let me know and we could merge this in.

mathesar_ui/src/components/column/ColumnName.svelte
ghislaineguerin: @pavish I've removed the extra HTML and restored the padding at the cell level. However, in data explorer header I see that the button element is used, which adds padding, but it is not being used for table header, so we end up with additional padding around the result table header.

mathesar_ui/src/components/sheet/Sheet.svelte
```diff
@@ -93,6 +93,7 @@
     bottom: 0;
     display: flex;
     flex-direction: column;
+    background: var(--color-gray-lighter);
```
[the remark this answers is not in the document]
ghislaineguerin: @pavish I've removed this.

mathesar_ui/src/components/column/ColumnName.svelte
pavish: Button is the right element to use for headers that are actionable. We should also be using a button for the table since it has a click action. It's a regression from one of the recent PRs, I've added a commit https://github.com/centerofci/mathesar/pull/1543/commits/76e4fa7b45c7a3a7dedf8ae37e829fb0ba4884a3 to address that.

Notice that this commit removes padding from the SheetHeader and moves it to the table's header cell.

Reason: The components within `sheet/` are meant to be headless. You can notice the other components within `sheet/` to not contain any elements, and just have a slot. In some components like the SheetHeader, we need to handle events such as scroll so they contain a basic wrapper. These components should not modify the layout of their children i.e. should not contain properties like padding, margin etc.,

pavish (approved):

PR #1548, opened 2022-08-22
Handle constraint violations while import

### LINKED ISSUE #1574: Raise specific exceptions for the errors caused by constraint violations during import.
## Description
Data imported from a CSV/TSV can be inconsistent with the schema of the target table to which the data has to be imported, to acknowledge and handle these inconsistencies we need to raise specific exceptions for the various errors encountered by the database during the import process.

## Expected behavior
Raise specific exceptions for the errors caused by constraint violations during import.

Anish9901:
Fixes #1574 

**Technical details**

This PR adds multiple exceptions for handling various DB layer errors that may occur during the import process. 
The exceptions added in this PR handle the following errors:
- Null Constraint not met
- Foreign key constraint not met
- Exclusion constraint not met
- Uniqueness constraint not met
- Data type mismatch
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/records/operations/insert.py
```diff
@@ -98,6 +100,26 @@ def insert_from_select(from_table, target_table, engine, col_mappings=None):
...
+            else:
+                raise e
+        except ProgrammingError as e:
+            if type(e.orig) == DatatypeMismatch:
+                raise TypeMismatchError
+            else:
+                raise e
         except Exception as e:
             # ToDo raise specific exceptions
             raise e
```
mathemancer, line 125: I think this is redundant, correct? I.e., if you don't catch the exception, it will be raised anyway.

mathesar/api/db/viewsets/tables.py
```diff
@@ -208,6 +209,46 @@ def existing_import(self, request, pk=None):
...
+                status_code=status.HTTP_400_BAD_REQUEST
+            )
+        except ProgrammingError as e:
+            raise database_base_api_exceptions.ProgrammingAPIException(
+                e,
+                status_code=status.HTTP_400_BAD_REQUEST
+            )
         except Exception as e:
             # ToDo raise specific exceptions.
             raise e
```
mathemancer, line 254: This is redundant now. The error will be raised if the exception doesn't match any `except` block above.

mathesar/api/exceptions/database_exceptions/exceptions.py
```diff
@@ -234,3 +234,77 @@ def __init__(
...
+    """ Exception raised when trying to insert a non castable datatype value to a column with certain datatype"""
+    error_code = ErrorCodes.TypeMismatchViolation.value
+
+    def __init__(
+            self,
+            exception,
+            message=None,
+            field=None,
+            details=None,
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR
```
mathemancer, line 249: I think all of these error codes would be more appropriate as status code 400, since the request shouldn't be repeated in the same form (you'll just get the same violation again). 500 is more appropriate for "maybe trying again in a minute will work". @silentninja may have an opinion on this, though.

mathemancer (approved):
Looking good overall, but I have some changes to request. See the specific comments for details.

mathemancer (changes requested):
The previous review was intended to be of the "request changes" type, sorry.

db/records/operations/insert.py
Anish9901: Yes, That is true, But can we be sure that all the unhandled exceptions will get caught in `IntegrityError` and `ProgrammingError`?

mathesar/api/exceptions/database_exceptions/exceptions.py
Anish9901: I was also curious about that, as all the exceptions in this file raise 500s and are then overridden to raise a 400 when the actual error/exception occurs.

silentninja: Most of the error classes are written as `Internal Server Error` because 
- Error classes are meant to be reused when the same error can happen in multiple places
- It is assumed the actual reason is due to some bug with the program and not due to the request sent by the user.

We then convert it to a 400 error when we find out the actual reason is due to an incorrect request sent by the user. If an exception is always known to occur due to incorrect data sent by the user, setting the default status code as `400` makes sense.

Anish9901: @silentninja That Makes sense, does that mean that `NotNullImportViolationAPIException` and `UniqueImportViolationAPIException` should produce an error code 400 as they are known to get raised only during import, and the rest of the exceptions should produce a 500 as they are more general and may occur at different places?

mathesar/api/db/viewsets/tables.py
Anish9901: Please refer to my comment above.

mathesar/api/exceptions/error_codes.py
```diff
@@ -16,18 +16,23 @@ class ErrorCodes(Enum):
...
     RaiseException = 4202
+    TypeMismatchViolation = 4214
     UndefinedFunction = 4207
     UniqueViolation = 4208
     UnsupportedType = 4209
 
     # Data Imports error code
     InvalidTableError = 4301
+    NotNullImportViolation = 4302
+    UniqueImportViolation = 4303
```
Anish9901, line 35: @mathemancer Can you please verify that these error codes are valid and can be added?

db/records/operations/insert.py
mathemancer: My point is that this line checks, and then _raises_ the exception if it's unhandled. It doesn't actually handle it. If this `except` block were missing, the behavior of the function would be precisely to raise the exception `e` under these circumstances.

mathesar/api/exceptions/error_codes.py
mathemancer: As far as I can tell, these aren't conflicting with anything.

db/records/operations/insert.py
Anish9901: Oh, I see your point, refactoring it.

mathesar/api/exceptions/database_exceptions/exceptions.py
silentninja: Wouldn't it be better to use the existing `mathesar.api.exceptions.database_exceptions.exceptions.NotNullViolationAPIException` instead of `NotNullImportViolationAPIException`. Both seem to be pointing to the same postgres Exception

Anish9901: > Wouldn't it be better to use the existing `mathesar.api.exceptions.database_exceptions.exceptions.NotNullViolationAPIException` instead of `NotNullImportViolationAPIException`. Both seem to be pointing to the same postgres Exception

Yes but, the reason I decided to make new exception classes i.e. `NotNullImportViolationAPIException` and `UniqueImportViolationAPIException` because `NotNullViolationAPIException` and `UniqueViolationAPIException` were getting thrown when trying to set a column constraint on a column consisting invalid values. However, in this case, even though the Postgres exceptions are identical the reason/way they occur are different i.e they are made to be thrown when trying to insert invalid values in a column with these constraints already set.

silentninja: > NotNullViolationAPIException and UniqueViolationAPIException were getting thrown when trying to set a column constraint on a column consisting invalid values

That makes sense @Anish9901 but I noticed that `NotNullViolationAPIException` is thrown when creating a new record 
https://github.com/centerofci/mathesar/blob/c03a2b0e197f2329b227b8aa1e96a06417070f5b/mathesar/api/serializers/records.py#L33
So it looks like they are the same type of error, correct me if I am wrong.

Anish9901: Um, you are right I based my decision on the docstring of the exception class, I didn't notice the `NotNullViolationAPIException` was getting thrown while adding an invalid record.

mathemancer (approved):
I think this is looking good. Thanks for your work on this, @Anish9901 !

PR #1565, opened 2022-08-25
Add test for json object sorting as well as json array contains function

Jinxiao0302:
1. Added test cases for json object type sorting functions. The sorting results is displayed at `db/tests/records/operations/test_sort.py/test_json_sort_object`
2. Implemented the filtering functions of "json array contains" and "json array is not empty". Contain function is based on the sqlalchemy function ` jsonb_contains(jsonb, jsonb)`, which transforms input strings into jsonb type. The useage of either array ot string is to be discussed.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/types/custom/json_array.py
```diff
@@ -163,3 +163,37 @@ def unpack(self):
             ArrayLength([param0]),
             param1,
         ])
+
+
+class NotEmpty(DBFunctionPacked):
```
dmos62, line 168: I'd like the python class name to reflect that this function is about JSON, similarly to how the id string makes that clear.

Same goes for all function classes that are meant to operate only on JSON values.

dmos62 (changes requested):

mathesar_ui/src/stores/abstract-types/operations/filtering.ts
```diff
@@ -97,32 +97,46 @@ const filterResponse: AbstractTypeFilterDefinitionResponse[] = [
...
+    name: 'number of elements <=',
     uiTypeParameterMap: {
       [abstractTypeCategory.Array]: [abstractTypeCategory.Number],
     },
   },
+  {
+    id: 'json_array_not_empty',
+    name: 'array is not empty',
+    uiTypeParameterMap: {
+      [abstractTypeCategory.Array]: [abstractTypeCategory.Number],
```
pavish, line 130: For this filter, we do not expect any user input. So, this should be:
```suggestion
      [abstractTypeCategory.Array]: [],
```

pavish (changes requested):
@Jinxiao0302 Looks good on the frontend part. I have one small suggested change.

pavish (approved):
Approving from the frontend side of changes.

@Jinxiao0302 Nice work, overall!

mathesar_ui/src/stores/abstract-types/operations/filtering.ts
Jinxiao0302: Thanks for your suggestions, Pavish! I have made that change.

mathemancer (approved):
Okay, this is looking good. Thank you for your work on this, @Jinxiao0302 !

db/functions/base.py
```diff
@@ -86,6 +86,19 @@ def to_sa_expression(primitive):
         return literal(primitive)
 
 
+class Noop(DBFunction):
```
dmos62, line 89: Since `Noop` is so similar to `Literal`, could you explain in a docstring what it is necessary? I presume it's because `Literal` creates an SQLAlchemy expression like `literal(subexpression)`, which in some way doesn't play well with the subexpression being a list of literals? Thanks!

dmos62 (changes requested):

db/functions/base.py
Jinxiao0302: Thanks for your review! Since we are using array as a input in `JsonArrayContains`, it is having trouble passing the filtering tests because of the literal() function. Thus, we need another function without literal wrapping to have that porperly implemented.

mathemancer: @Jinxiao0302 Given the confusion, I do think it makes sense to add a docstring indicating the reason for this DBFunction.
- The `Literal` DB function produces a `literal` SQLAlchemy wrapper which doesn't play nicely with the type conversion between python classes and db types in `psycopg2`. 

@dmos62 Long-term, we need to reassess how we're dealing with literal (in the general sense) values in the DBFunction subclasses. Wrapping an arg in `literal` should probably be avoided, except in situations where we actually want to produce a column of literals in the produced query. I don't think it should ever be used for the comparand in a filter.

mathemancer (changes requested):
Hey, @Jinxiao0302, I think all that remains before merging this is to add the requested docstring to the `Noop` class to avoid future confusion.

Thank you!

Jinxiao0302:
Completed required changes.

kgodey:
@Jinxiao0302 I disabled auto-merge because the PR won't merge until @mathemancer has approved it (since he requested changes earlier). I'm assigning this back to him to review.

mathemancer (approved):

mathemancer:
@dmos62 I'm happy with this now. Feel free to merge after you re-review.

dmos62 (approved):

dmos62:
Thanks for the contribution @Jinxiao0302!

PR #1530, opened 2022-08-16
Cell Selection for table inspector

rajatvijay:
Provide the feature of selecting cells to build the base of the table inspector. 
1. The user can select a single cell by directly clicking on it. The UI of a single selected cell and that of the active cell are exactly the same. 
2. The user can select multiple jointed cells using click and drag
3. The user can also toggle column selection across all rows by clicking on the column header
4. The user can also toggle row selection across all columns by clicking on the row header
5. Row and Column header are highlighted when there is at least one cell selected that belongs to that [row, column]
combination. 

**TODOs**
- [ ] Cell selection feedback box while the user is dragging to select multiple cells.

**Technical details**
1. Removed checkboxes used to select rows. NOTE: This also disabled the functionality of deleting rows for now. This will be up again when the table inspector is in place. 

<details>

<summary>2. Click + drag to selected multiple jointed cells. </summary>

Changes to existing components
1. CellFabric, CellWrapper supports a new prop isSelectedInRange passed as-is to cell type components.
2. Each cell type component now supports a prop isSelectedInRange.
3. isSelectedInRange is purely for coloring the background of the cell only when its selected with atleast one more cell.

Selection Store
1. There is a new store for storing selected cells
2. Selection process starts when an activate event is received for a cell.
3. Every cell which received mouseenter event will be included in the same selection
4. Selection process terminates when the document recieved mousedown event.
5. Selected cell is represented by `rowIndex-columnId` and is stored in a set.

</details>

**Screenshots**

https://www.loom.com/share/027b0e92be42403db01515cb9949d167

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


[...]

PR #1501, opened 2022-08-04
Refactor icon storage - Isolate dependency on font-awesome

rajatvijay:
Fixes #1187

This PR accumulates all the icons into two places so that the direct import from font-awesome is removed from throughout the codebase except from inside the `icons.ts` files and the component stories. These two palces are: 
1. All of the icons used inside the components library are now imported from `component-library/common/icons.ts`
2. All of the icons used inside the core application code (`src/` except the component-library) are now imported from `src/icons.ts`

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1568, opened 2022-08-25
Table inspector/table properties

rajatvijay:
Implements the UI of table inspector and table properties within it including: 
1. Table name (rename functionality)
2. Table link (create/edit functionality)

**Technical details**

1. Table inspector can now only be toggled using the icon button in the table toolbar
2. Created a new EdtiableTextWithActions that enables inline editing with support for "Save" & "Cancel" actions

**TODOs**
- [ ] Make table inspector visible as soon as the table is visible
- [ ] Column Properties
- [ ] Table actions
- [ ] AIs from #1530 

**Screenshots**

https://www.loom.com/share/4a6d8e97cbde42ac80
Screenshot 2022-08-25 at 5 06 47 PM
bd7f90194f208b

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
Apologies for this litany of a PR review! Many of these points are small nit-picks which didn't necessarily _need_ to be done -- but I'm taking some extra time here to help you get up to speed in our codebase and ensure all of our styles are in alignment. Reviews generally won't be nearly this thorough once you have more momentum :grin:. If you have concerns with any of the changes I made, feel free to push back. We can easily revert some of my commits if necessary.

## Critique via commits

I added a bunch of small commits, taking care to isolate each concept. Below I've added some additional comments to most of the commits. Please review each and let me know if you have any questions or concerns.

- 6d48fbd9 Fix spelling mistakes

    I recommend installing the VS Code extension `streetsidesoftware.code-spell-checker` to prevent problems like this in the future.

- 9bda4acd Remove question in code comment

    Yep -- that look like a good place for it.

- a9a2d7fa Use semantic names for icons

    In reviewing #1501, I was hoping that I adequately explained our system of categorizing icons, but perhaps I wasn't clear enough. A "toolbox" is not a concept within Mathesar, so `iconToolbox` doesn't make sense as an icon name. I changed the icon name to `iconTableInspector` because "Table Inspector" _is_ a concept within our product. Indeed the specs require a slightly different icon here (which I can see Font Awesome doesn't have). So once we procure a suitable icon, then we can change `iconTableInspector` to point to something like `iconRightSidePanel` or similar. If this is still confusing, let me know.

- fd33222f Sort icons lines

    This is not strictly necessarily, but I've found that maintaining deterministic ordering within long blocks of lines can slightly reduce the likelihood of merge conflicts.

- 0505ea32 Add key to {#each} block

    Keys on `{#each}` blocks help performance and are thus generally good practice when possible. See https://svelte.dev/docs#template-syntax-each

- 1d7dcc35 Use function instead of const

    The `const` vs `function` question is one of those minor stylistic choices which I think are best applied consistently throughout a codebase to aid readability. Our codebase currently uses `function` basically everywhere, so I think it'd be wise to stick with that style, at least for the time being. We don't have a [code standard](https://wiki.mathesar.org/en/engineering/standards/frontend) requiring it, but I've just [proposed one](https://github.com/centerofci/mathesar-wiki/pull/57).

- c2c381cf Use plural variable name for array type

- 11cfe221 Improve code comment formatting

- 63f66d4c Treat acronyms as words in PascalCase

    We do have a [standard](https://wiki.mathesar.org/en/engineering/standards/frontend#naming-conventions) about this one.

- 57d8104d Make stringified UiError show its messages

    Without this change `String(new UiError(["oh no!"]))` gives `"Error"` when it should give `"oh no!"`.

- a1f8211f Don't assume that only UiError will ever be thrown

    Before this change, the `EditableTextWithActions` expected the `onChange` to potentially throw an error -- but only of type `UiError`. It couldn't handle other errors like `Error` because they wouldn't have the `errorMessages` property. I can see that the consuming component addressed that limitation by only throwing a `UiError` -- but that's a bit of a brittle pattern because that subtle limitation isn't reflected well to consumers of `EditableTextWithActions`.

- f887a381 Use `await` instead of `.then`

    Similar to 1d7dcc35, I think it would be nice if we could consistently use `await` so I've
    [proposed a standard](https://github.com/centerofci/mathesar-wiki/pull/57) for it.

- ed0dd3ee Remove unnecessary handler function

    Svelte has three mechanisms for handling upwards data flow from child to parent:

    - (A) Declaratively via `bind:value`
    - (B) Imperatively via passing a function as a prop.
    - (C) Imperatively via dispatching events like `on:change={handleValueChange}`.
    
    Generally it's best to use (A) whenever possible because it's the most concise and idiomatic. Sometimes we need more control so we resort to imperative code with (B) or (C).

- 2ae85996 Improve inline component documentation

    - Using `@component` for code comments like this is nice because it makes the code appear in intellisense wherever the component is used.
    - I removed the documentation for `onChange` because I don't think it adds any value beyond the type of the prop. If we _do_ want to document props, that documentation should go directly above the prop with a JS Doc comment.

- 50158fb5 Use more appropriate prop name

    Some people might expect `onChange` to be called when the user blurs the input. I think `onSubmit` is less ambiguous.

- 4869428d Simplify error logic, show validation while typing

    This makes the code a bit simpler too.

- 327642d8 Remove UiError class non longer used

    With the new pattern employed in the previous commit, we no longer need `UiError`.
    
    I'm not entirely opposed to adding a class like `UiError` at some point. But I would be a little surprised if we encounter a use-case which wouldn't be better solved by other patterns that don't rely on differentiating error types. In general, I try to avoid writing code that throws errors, preferring instead to use return values because TypeScript will complain if we don't handle them appropriately. Of course sometimes we need to use errors, but I generally considering using errors only as a last resort because they're more prone to bugs when when refactoring.

- 2c1aca58 Prevent saving an unchanged or invalid value

- c9ddf784 Move padding value from JS to CSS

    Was there a reason you used a JS variable and used `style` here?

- 1d307e94 Use CSS variables for common style values

    I did a little refactoring in other components here too, collecting all the error text colors into one place.

- ba6a140b Utilize CancelOrProceedButtonPair component

    This change reduces some code and has the added benefit of ensuring that our UI is consistent. For example, users will be less likely to hit the wrong button if "cancel" is always on the left (i.e. go _back_) and "proceed" is always on the right (i.e. go _forward_). Plus the buttons get icons.

- 55f973f3 Use appropriate alias for imports

    The `@mathesar-component-library-dir` alias should only be used _inside the component library_. I know this is confusing. I didn't make these rules.

- d636c539 Don't import from parent

- e28c9918 Remove customization of collapsible header

    Ok this one is a little tricky.

    I removed the logic that un-sets the border and padding of the collapsible header. Here are more thoughts...

    Your code comment (which I removed) in `TableGroupCollapsible` says:

    > This is leading to unwanted padding and border when collapsible is used in isolation

    I don't think that's true. For example, (after this commit) if I comment out that CSS, it doesn't eliminate the padding and border. Even though it's using `:global`, the selector is still nested inside `.table-group` so it shouldn't apply to your feature. That said, I don't want to have `:global` unless absolutely necessary, and that CSS declaration doesn't seem to actually be doing anything. I'm a little confused why that CSS exists, but basically it's Pavish's problem since the Query Builder is still very much in flux. I wouldn't be surprised if that CSS is just cruft that he's already removed in a local branch of his. Or maybe it's important for some reason that I don't understand.

    The reason the collapsible header has padding and border is because `Collapsible.svelte` wraps that content inside a `<Button>`. The _Button_ is applying that padding and border. I don't really like the approach of overriding it with `noStyle`, so I've removed that. Collapsible should potentially be adapted to allow more configuration of its heading, but I don't want to put too much time into thinking about it before Ghislaine has had a chance to look at it. I think you were following the wire frame as the design, which is to be taken very loosely. Ghislaine might want to go a different direction with the styling here, so I'd rather keep the code as simple as possible for now, leaving the door open to figuring out the best style once we're further along with the Table Inspector.

- ed4eebc5 Make table inspector sections open by default

    The spec is not clear on this, but it seems to me like they should be open by default.

- 16893b50 Remove unnecessary div elements

    Aside from simplifying the code, one reason to remove the `div` here is do adhere to this [code standard](https://wiki.mathesar.org/en/engineering/standards/frontend#build-components-for-valid-dom-nesting) about valid DOM nesting. The header nests its content in a `<button>`. Technically, it's invalid HTML to put a `<div>` inside a `<button>` (though certainly we have other code that breaks this rule, and I've only seen problems with invalid nesting in very rare circumstances).


## Additional critique, not reflected in my changes

- In some code that I removed (in ba6a140b7ff3b21e84480118ea9a2eb669f95756) I see you used `gap` for flexbox which is [not very well supported yet](https://caniuse.com/flexbox-gap). As a team, we haven't established a clear rule-of-thumb for determining whether a browser feature is supported widely-enough to use. We can work towards some kind of rule if you think it would be helpful.

- In `Collapsible.svelte` you had this code:

    ```
    class={`collapsible-header ${noStyle && 'no-style'}`}
    ```

    I ended up removing the `noStyle` bit for other reasons, but I wanted to highlight a problem with this code if we were to keep it. When `noStyle` is `false`, `class` ends up becoming `"collapsible-header false"`, i.e. the element ends up with a `"false" class applied (not what we want).

    The proper idiomatic way to do this with Svelte is to use [class:name](https://svelte.dev/docs#template-syntax-element-directives-class-name) as follows:

    ```
    class="collapsible-header"
    class:no-style={noStyle}
    ```

    You can add as many `class:name` attributes to an element as you like, but you can't ([yet](https://github.com/sveltejs/svelte/issues/7170)) add more than one `class` attribute.

kgodey:
I tested this on staging and have two issues to report:
- The table inspector doesn't open by default when you go to a new table (which the spec says it should, IIRC)
- There's a heading called "Links" but what it's actually displaying is constraints.

Issue #1580, opened 2022-08-29
Improve Record Page


seancolsen:
- Record
    - [ ] Allow deleting the record
    - [x] ~~Display the record summary in the `<h1>` element and the `<title>` element.~~ (covered in #1600)
- Direct fields
    - [ ] Add a status bar at the top to let users know their data is saved (like we do in the sheet)
    - [ ] Fix updating boolean field
    - [ ] Show loading status when updating an FK field
    - [ ] Focus next input when pressing Enter (unless it's a `textarea`)
    - [ ] Allow user to set values to NULL
    - [ ] Display a tooltip like "Press Tab to Save" or "Press Enter to Save" when focused
- Table widget
    - [ ] Don't show page size input within the pagination widget (so that user can't modify it)
    - [ ] Make it collapsible
    - [ ] Don't show Pagination UI when there is only one page
    - [ ] Allow resizing columns
    - [ ] Header gets cut off while horizontal scrolling within each table widget

seancolsen:
I'm punting this to "Live demo" because I don't think any of these improvements should block the demo video. I might subdivide these tasks into separate milestones later once we get to Cycle 5.

kgodey:
Moved this back to "Demo video" because a lot of these things came up in usability testing. @seancolsen feel free to move things that aren't relevant until the live demo to a separate issue.

seancolsen:
> a lot of these things came up in usability testing

@kgodey Can you be more specific? Which of the tasks listed above are relevant to the demo video?

kgodey:
@seancolsen I saw that you had commented on a bunch of feedback that came up and said it was tracked in this issue. Please see your notes from [the sections under Screens 26 and 27 here](https://www.figma.com/file/F0FmNaNz8hvrgxPax3Bix1/Cycle-3-reference?node-id=856%3A624).

seancolsen:
@kgodey and I chatted about this. I'll be splintering off sub-tasks from this ticket into dedicated tasks for the live demo

seancolsen:
I'm splitting this into smaller tickets to set more granular priorities

- 1843
- 1842
- 1844
- 1845
- 1846
- 1847

PR #1581, opened 2022-08-30
Record Page, first pass

seancolsen:
This PR implements the [Record Page feature](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/record-page.md), except for some outstanding tasks as noted in #1580.

Feel free to give this a thorough review from a user's perspective and raise any shortcomings not already tracked in that outstanding tasks ticket.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/systems/table-view/TableView.svelte
```diff
@@ -1,39 +1,41 @@
...
   import Body from './Body.svelte';
   import Header from './header/Header.svelte';
   import StatusPane from './StatusPane.svelte';
   import TableInspector from './table-inspector/TableInspector.svelte';
 
-  export let database: Database;
-  export let schema: SchemaEntry;
-  export let tabularData: TabularData;
-  export let table: TableEntry;
+  const tabularData = getTabularDataStoreFromContext();
```
pavish, line 14: Is there an advantage to getting tabularData from context here, instead of a prop?

Since TableView is a system on it's own and tabularData is the primary data source for this system, it is more intuitive to use a prop instead unless there's a good reason not to.

mathesar_ui/src/systems/table-view/Body.svelte
```diff
@@ -58,22 +60,28 @@
...
+      {#each items as item (item.key)}
+        {#if $displayableRecords[item.index]}
+          <Row style={item.style} bind:row={$displayableRecords[item.index]} />
+        {/if}
+      {/each}
+    </SheetVirtualRows>
+  {:else}
+    {#each $displayableRecords as displayableRecord}
+      <Row
+        style={{ position: 'relative', height: rowHeightPx }}
```
pavish, line 82: Height of the row should be calculated using `getItemSize` method or rows with different heights would be misaligned eg., Group headers.

mathesar_ui/src/pages/record/TableWidget.svelte
```diff
@@ -0,0 +1,99 @@
...
+    // below.
+    undefined as unknown as TabularData,
+  );
+
+  export let recordId: number;
+  export let table: Pick<TableEntry, 'id' | 'name'>;
+  export let fkColumn: Pick<Column, 'id' | 'name'>;
+
+  $: meta = new Meta({
+    pagination: new Pagination({ size: 10 }),
```
pavish, line 28: I propose we hide the size selector in pagination for table widgets and limit the rows to be 10 always. Rendering a large number of rows in the record page may not be ideal since we do not use a virtual list here and using a virtual list would not be possible without having a scrollbar within each table widget.

mathesar_ui/src/pages/record/Widgets.svelte
```diff
@@ -0,0 +1,46 @@
...
+      fkColumn: {
+        id: joinableTable.jp_path[0].slice(-1)[0],
+        name:
+          columnNameMap.get(joinableTable.jp_path[0].slice(-1)[0]) ??
+          '(unknown column)',
+      },
+    }));
+</script>
+
+{#each tableWidgetInputs as { table, fkColumn }}
```
pavish, line 36: A key is required here for better performance.

pavish (approved):
@seancolsen Looks good! Glad to see the record page shaping up well.

I had a few questions & suggested changes. I'll leave it upto you to handle them in the PR or a later one. I'm approving this PR, feel free to merge once you've read the reviews.

pavish (changes requested):
I just noticed a bug in the PR where the Sheet component isn't visible in data explorer, making the data explorer unusable. I think it's important to fix it in this PR.

mathesar_ui/src/systems/table-view/TableView.svelte
seancolsen: Before this PR, `TableView.svelte` was setting `tabularData` in context. After this PR, it's _getting_ `tabularData` from context. I made that change because I moved `ActionsPane` outside of `TableView` and `ActionsPane` already depended on `tabularData` being in context. Moving the location at which the context is set seemed easier than chasing down all the context dependencies within `ActionsPane`. I would be open to establishing a standard that _"the top-level component of a system does not depend on context"_, but it's not how I've been operating thus far.

mathesar_ui/src/systems/table-view/Body.svelte
seancolsen: Thanks. Fixed in c3bbbc14d3d34301cc14499b9d7a43fc0005e947

mathesar_ui/src/pages/record/TableWidget.svelte
seancolsen: That's good. I like that idea. I've updated #1580 to include this

mathesar_ui/src/pages/record/Widgets.svelte
seancolsen: Thanks. Fixed in 9a54f6740d59408deab3deceea9658fec4ea7064.

FWIW, I went looking for a way to lint away this problem and found [this sad thread rejecting the idea](https://github.com/sveltejs/eslint-plugin-svelte3/issues/42). Just like the OP on that thread, I'm accustomed to the Vue linter that complains about missing keys.

seancolsen:
@pavish Ready for re-review.

pavish (approved):
@seancolsen Looks good

PR #1589, opened 2022-09-01
Switch query validation exception to use our error infrastructure

dmos62:
My introduced exception classes seem more generic than the rest. I think it works well in this case, but I'm open to other suggestions.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja (changes requested):
I have a few concerns
1. Why are the validations done at the model instead of the serializer? It would add comments about it. (While I don't find DRFs duplication of the validation layer to be correct, it is a convention that we are following so it would be better to add comments on why you are bypassing it)
2. Have you considered passing the validations to the DRF serializers instead of doing manual validation?
3. Most of the validation seems to overlap with DRF validations. In that case, it wouldn't it be good to use existing [DRF error codes](https://github.com/FutureMind/drf-friendly-errors#default-field-error-codes)
4. `field` property is not being passed to the exceptions, front end won't accurately point to the error source that is causing the error without it.

dmos62:
> I have a few concerns
> 1. Why are the validations done at the model instead of the serializer? It would be good to add comments about it. (While I don't find DRF duplicating the validation layer to be convincing, it is a convention that we are following so it would be better to have valid reasons and add comments on why you are bypassing it)
> 2. Have you considered passing the validations to the DRF serializers instead of doing manual validation?

These validations describe a "globally" valid model, as opposed to a valid model in the context of JSON deserialization.

They also define and document the kind of data structures the model uses. It's good to have that together with the rest of the model's definition.

Also, if validation were only done in DRF, that would mean that you could instantiate models with invalid properties. A large portion of tests for DE are not HTTP tests, but tests where models are instantiated directly. It's useful to have validation work there too.

> 3. Most of the validation seems to overlap with DRF validations. In that case, it wouldn't it be good to use existing [DRF error codes](https://github.com/FutureMind/drf-friendly-errors#default-field-error-codes)

I wasn't aware that DRF has error codes for its validators. That's a sensible suggestion. I'll make this change.

> 4. Frontend uses the `field` property to accurately point to the error source. Most of the Exceptions are not being passed the correct `field` property. I made specific comments to give a better picture

I was working under the assumption that the model doesn't know the serializer's field mapping, but i now realize that since we're using serializers that automatically mimick the model, we can just use model's field names. I'll make this change.

dmos62:
@silentninja by the way, you asked for code comments about why validation is done during model construction and not deserialization. I feel like explaining that in code would be weird in this case, since it's a general concept, and it's something more along the lines of code standards or just general architecture, so maybe let's put this in our individual notes and bring it up when we have our 1:1 or team discussion on the topic?

dmos62:
Finally got around to making changes.

I examined the drf-friendly error codes you linked: https://github.com/FutureMind/drf-friendly-errors#default-field-error-codes

They don't map to the things I'm throwing exceptions on. To illustrate, I'll list the exceptions I'm throwing and I'll tick whether or not they're part of the error code list you posted:

| Exception | Included in drf-friendly error code list |
| --- | --- |
| must be list | x |
| list must contain only dicts |  |
| dict doesn't contain following obligatory keys |  |
| dict contains unexpected keys |  |
| list must contain only 2-item lists |  |
| list must contian only integer column ids |  |
| list must contain only dicts that contain these keys |  |

Also, the error codes that _are_ present in that list seem to be too different to what we need.

@silentninja are you ok with using our own error codes in this case?

silentninja (approved):
The changes look good. The test case `test_query_with_initial_column_with_bad_jp_path` seems to be wrong when compared with the validation rules, because the `jp_path` field seems to be optional while the test tries to assert its requirement. Please fix it and merge

Also, I noticed that you were using pre-defined values instead of fetching them from the actual object in https://github.com/centerofci/mathesar/blob/f704629a8acdde63a6a06865c0805a4553f0b9af/mathesar/tests/api/query/test_query_api.py#L111
and https://github.com/centerofci/mathesar/blob/f704629a8acdde63a6a06865c0805a4553f0b9af/mathesar/tests/api/query/test_query_api.py#L90

It's hard to know what the values refer to and sometimes the column object might be having a different id when test cases run sequentially, it would be better to get the Django object and use its id field instead.

dmos62:
@silentninja thanks for the review. The `test_query_with_initial_column_with_bad_jp_path` is actually testing that the `jp_path` json blob is a seqeuence of integer tuples. The integers that represent Django ids are mocks that don't attempt to map to an actual Django column. We don't actually check here that the column ids refer to actual columns, but a reader doesn't know that, so it's confusing. I'll add comments.

dmos62:
Interestingly, [this change](https://github.com/centerofci/mathesar/pull/1589/commits/90e05796688cd9346d6e0d1a0ad380086134c4cc) to the initial migration seems to be necessary, since I renamed the validator functions.

By the way, the test was failing because I was calling a validator factory (from within another validator) as if it was a validator and that was silently messing up validation: https://github.com/centerofci/mathesar/pull/1589/commits/3c33e4a4b2d18f3d2f8859a870cf34e720eddfc8#diff-7c20daec2eb3fa4c134c0436f8e25ffddce46777de6a23364aef737fdfe4df77R57

silentninja:
There is a problem with duplicate migration being generated for the fields in the `UIQuery` model that contain custom validators for example `initial_column`. Please create a fix for it or an issue to track it.

silentninja:
> We don't actually check here that the column ids refer to actual columns, but a reader doesn't know that, so it's confusing. I'll add comments.
 
 But aren't they supposed to be actual column id even though we don't have a validation on the backend for now? Or can these ids also refer to something else?
 If it is the former, using the actual column would be better as it is straight to the point instead of having a comment

dmos62:
@silentninja It is weird that we can create a legal query with random integers instead of actual column Django IDs. Our validation logic here only cares about the data structure and not the semantics. If we fixed that, the confusion would go away. We can create an issue to track this.

Issue #1532, opened 2022-08-16
Improve logic for setting selected cells


seancolsen:
In #1530 @rajatvijay said:

> Every cell which received mouseenter event will be included in the same selection

This logic handles the simplest case well, but it fails to handle some interesting edge cases in a manner consistent with spreadsheets.

## Reproduce

1. View a table with a PK column containing sequential numbers.
1. Begin dragging on cell 5 to start selecting cells.
1. Drag down to cell 10, but don't stop yet.
1. Now drag all the way back up to cell 1 and release.
1. Expect cells 1 through 5 to be selected.
1. Observe cells 1 through 10 to be selected.

PR #1591, opened 2022-09-05
Table inspector/column properties

rajatvijay:
Table Inspector Sub modules
1. Table actions
2. Column mode
3. Record mode
4. Table inspector to open when the table view is loaded or updated

**NOTE**
Added a few comments on the PR for asking questions

**Screenshots**

https://www.loom.com/share/23218956e9f247ba853c33027bf656be

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/table-data/selection.ts
```diff
@@ -53,10 +62,13 @@ export class Selection {
 
   selectedCells: WritableSet<string>;
 
+  freezeSelection: boolean;
```
seancolsen, line 65: Can you help me understand what `freezeSelection` is and why we need it?

mathesar_ui/src/stores/table-data/tabularData.ts
```diff
@@ -99,6 +99,9 @@ export class TabularData {
     this.columnsDataStore.on('columnPatched', async () => {
       await this.recordsData.fetch();
     });
+    this.columnsDataStore.on('columnsFetched', async () => {
+      this.display.isTableInspectorVisible.set(true);
+    });
```
seancolsen, line 104: Why do we need to set the table inspector to be visible after fetching the columns?

mathesar_ui/src/utils/errors.ts
```diff
@@ -76,4 +76,16 @@ export class ApiMultiError extends Error {
   }
 }
 
+/**
+ * This is meant to be a build time error
+ * Generally use to make sure of exhaustive conditions handlint
+ */
+export class UnhandledError extends Error {
+  constructor(value: never, message: string) {
```
seancolsen, line 84: I can see the usefulness of `UnhandledError` so I'm okay with keeping it.

But why does `value` need to be `never` here? More broadly, I don't see the usefulness of the `value` argument all. I might want to throw an `UnhandledError` even if I don't have a value. Or I might have an object as a value. From the calling site, it's not clear what happens with `value` when I pass it into the constructor. Witt the code as-is, if I pass in an object, the error message will display "[object Object]" but that behavior is not clear from the calling site.

I'd be inclined to do:

```ts
export class UnhandledError extends Error {
  constructor(message: string) {
    super(message);
    this.name = "Unhandled Error";
  }
}
```

That design forces the caller to stringify their value (as they see fit, depending on the type of the value) before instantiating the `UnhandledError` object. It also sets the `name` property on the instance so that `String(new UnhandledError("Foo"));` will give `"Unhandled Error: Foo"` instead of `"Error: Unhandled Error: Foo."`

seancolsen (changes requested):
- The "Data Type" UI is not displaying correctly, but it [seems](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$UK6eTVmuW2_xoHUl1bXYiv9ZIDT47kALhKEudYJXerw?via=matrix.mathesar.org&via=matrix.org) that you're already working on a fix for that. I'd like to get that in this PR before we merge.

- _For the following points, you can address them in this PR or subsequent ones -- whichever you prefer._

- If I select two vertically adjacent cells, I expect the "Column" tab to show me the UI for that column. Instead it is empty.

- If I select two horizontally adjacent cells, I expect the "Column" tab to show me properties and actions for both columns, instead it is empty.

- If I select two horizontally adjacent cells, I see a button within the Record tab labelled "Delete 2 Records". This should be "Delete 1 Record" because only one row is selected.

- If the contents of the table inspector are tall enough, they'll force the entire page to scroll. Instead, we should be scrolling only the content below the tabs within the table inspector.

mathesar_ui/src/stores/table-data/selection.ts
rajatvijay: The idea here is to not let the user change the selection/make a new selection when this boolean is true. This is helpful when there is a long-running action on a particular selection and we don't want the user to update the selection until the action is finished.

mathesar_ui/src/stores/table-data/tabularData.ts
rajatvijay: This is to make the inspector visible as soon as the table view is opened.

mathesar_ui/src/utils/errors.ts
rajatvijay: @seancolsen `value` being `never` is what makes this error class useful. For example check this case here: 

```typescript
let mode: 'read' | 'edit' = 'read';
  function toggleMode(): undefined {
    switch (mode) {
      case 'read':
        mode = 'edit';
        break;
      case 'edit':
        mode = 'read';
        break;
      default:
        throw new UnhandledError(mode, 'ColumnType');
    }
    return undefined;
  }
```

If tomorrow someones adds a new possible value for `mode` and forgets to handle it inside the `toggleMode` function, the type of `mode` in line `throw new UnhandledError(mode, 'ColumnType');` will not be `never` which is expected hence forcing TS to show build time error.

rajatvijay:
@seancolsen 

> The "Data Type" UI is not displaying correctly, but it [seems](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.or
g/$UK6eTVmuW2_xoHUl1bXYiv9ZIDT47kALhKEudYJXerw?via=matrix.mathesar.org&via=matrix.org) that you're already working on a fix for that.

is fixed 

Screenshot 2022-09-07 at 7 18 08 AM

rajatvijay:
> * If I select two horizontally adjacent cells, I expect the "Column" tab to show me properties and actions for both columns, instead it is empty.

In accordance with the [specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/table-inspector.md#actions-for-multiple-components-and-selection-status) only Column Extraction is supported when multiple columns are selected.

rajatvijay:
Pushed the following changes

1. 92c27de The "Data Type" UI is not displaying correctly, but it [seems](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$UK6eTVmuW2_xoHUl1bXYiv9ZIDT47kALhKEudYJXerw?via=matrix.mathesar.org&via=matrix.org) that you're already working on a fix for that.

image

2. 527be71 Fix for #1532
3. 42b25d9 CellWrapper to use CellBackground to style selected c
ell instead of a dedicated class. Fix asked here: https://github.com/centerofci/mathesar/pull/1530#issuecomment-1216855529
2. ad4411b If I select two vertically adjacent cells, I expect the "Column" tab to show me the UI for that column. Instead it is empty.
4. 0d91fa7 If I select two horizontally adjacent cells, I see a button within the Record tab labelled "Delete 2 Records". This should be "Delete 1 Record" because only one row is selected.

NOTE: 
> * If I select two horizontally adjacent cells, I expect the "Column" tab to show me properties and actions for both columns, instead it is empty.

In accordance with the [specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/table-inspector.md#actions-for-multiple-components-and-selection-status) only Column Extraction is supported when multiple columns are selected.

kgodey:
@rajatvijay we should only have one `status` label on an issue/PR at a time. I updated the labels and assignees on this issue.

seancolsen (approved):

mathesar_ui/src/pages/data-explorer/DataExplorerPage.svelte
```diff
@@ -10,11 +10,13 @@
...
 
+  const routeMeta = meta();
+
   $: ({ query } = queryManager);
 
-  function gotoSchema() {
+  function goBackWithFallbackToSchema() {
     const schemaURL = getSchemaPageUrl(database.name, schema.id);
-    router.goto(schemaURL);
+    router.goto($routeMeta.from || schemaURL);
```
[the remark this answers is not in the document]
seancolsen: > is this the right way to handle back navigation

Maybe? I'm not sure. Basically, I don't think we should ever be doing this because I can't imagine a scenario where I think this type of contrived "back" button would be a good design. I opened #1601 requesting that we remove that button altogether, and in a sync call, Pavish, Kriti, and Ghislaine agreed. In place of that button, we'll have a "View Results" _link_ (not button) which only displays while you're editing a saved exploration.

Because of the slight change in design, I've reverted your changes to this file. @pavish will handle the remaining changes needed within `DataExplorerPage` and that work can be considered independent of your work within this PR.

mathesar_ui/src/stores/table-data/selection.ts
```diff
@@ -5,6 +5,7 @@ import type { Row, RecordsData } from './records';
 import type { ColumnsDataStore } from './columns';
 
 const DEFAULT_ROW_INDEX = 0;
+const ROW_COLUMN_SEPARATOR = '-';
```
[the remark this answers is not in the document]
seancolsen: Perhaps. But that sort of unification is only the tip of the iceberg which represents some larger refactoring that I think is coming due soon within this code. I'd suggest we forge ahead amidst this cruft and staving off the work of managing small tickets like that until we have the energy to engage with some higher-level improvements.

mathesar_ui/src/systems/table-view/row/RowCell.svelte
```diff
@@ -53,8 +53,20 @@
...
+   * that active cell and selected cell does not remain in sync when using keyboard
+   * this boolean is true even when atleast one cell is selected
+   * to differentiate between different active and selected cell
+   * using blue background styling for selected cell
+   * and blue border styling for active cell
+   * The above bug can be fixed when following two conditions are met
+   * 1. We are working on keyboard accessbility of the application.
+   * 2. selectedCells and activeCell are merged in a single store.
+   */
+  $: isSelectedInRange = isCellSelected($selectedCells, row, column);
```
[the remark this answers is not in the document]
seancolsen: This seems fine for now. Though I did clean up your code comment, making it easier to read in 01db5b18a8b8c5b0b568bb9ef1aff650b8c11874

mathesar_ui/src/stores/table-data/selection.ts
seancolsen: Cool. Makes sense. Thanks

seancolsen:
@rajatvijay This looks good for merge now :grin: 

I have some more follow-up comments to make when I return to work later, some of which I'll handle in separate tickets.

mathesar_ui/src/stores/table-data/tabularData.ts
seancolsen: I opened #1616 to propose deleting these lines. We can continue the discussion there if there's a reason to keep the lines that I'm not seeing.

mathesar_ui/src/stores/table-data/columns.ts
```diff
@@ -113,6 +114,7 @@ export class ColumnsDataStore
       };
       this.set(storeData);
       this.fetchCallback?.(storeData);
+      await this.dispatch('columnsFetched', storeData.columns);
```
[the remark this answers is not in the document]
seancolsen: It's fine to dispatch this event here, but in #1616, I've proposed that we remove the listener to the event. 

> I personally feel having this event system is super difficult to manage and reason about.

This is a pattern that @pavish designed. I agree that it's difficult to manage and reason about, and I have some other qualms about it too, which Pavish and I have debated on several occasions.

For _the particular use-case in this PR_, it doesn't seem like it's necessary to use this event system at all (as I've described in #1616).

But for other use-cases we do need _some system_ to keep our stores in sync when one change necessitates changes somewhere else. For example, if the user deletes a column, we need to make sure we remove references to it from our filtering/sorting/grouping stores. The pattern that I had proposed to Pavish was that we handle those "side effects" in a more manual manner. The calling context would be responsible for figuring out _what else_ needs to happen when we delete a column for example. That pattern I proposed is a little simpler and easier to reason about, but it's also more error-prone. Perhaps our UI has two wildly different mechanisms to delete a column. Then we'd need to make sure that each of them behaves in a similar manner, taking care to "clean up" after the column is deleted. I've come around to this `EventHandler` pattern, though it still doesn't feel optimal to me.

I'd be very curious to hear what sort of patterns you might prefer using here. This ties in to some of the conversation we had at our front end team meeting where I wondered out-loud whether a tool like [TanStack](https://tanstack.com/query/v4) might help us with some of these rough spots in our code. If I were to spend time trying to find a better pattern than our current `EventHandler` pattern, looking at TanStack (and similar libraries) would probably be the first direction I'd head. Since you've used React Query a bit, you might be able to offer some good insight here for us!

mathesar_ui/src/utils/errors.ts
seancolsen: _Ohhhhh_ -- I get it now. 💡 Cool!!

Very clever. I like this!

The name `UnhandledError` threw me off because it's so general. I've opened #1618 to propose some minor clean up which improves the name.

PR #1592, opened 2022-09-05
Data Explorer Styling

ghislaineguerin:
Related to #473 

This PR addresses styling issues for Data Explorer.
- It updates the query builder page header to look like the one in tables.
- It replaces the colors to use variables instead.

### DIFF
--- a/mathesar_ui/src/components/EditableTitle.svelte
+++ b/mathesar_ui/src/components/EditableTitle.svelte
@@ -4,15 +4,11 @@
   export let value = '';
   export let size = 1;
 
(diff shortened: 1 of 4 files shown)


pavish (approved):
@ghislaineguerin Looks great!

I added a small commit https://github.com/centerofci/mathesar/pull/1592/commits/6a60961d4c09e2b1d595d43c7b1ddd00c1d59919 which removes some unused css. The `.header .type` is unused because there's no element with class `type`. I also added a small fix for the editable title placement, which is a styling issue that existed earlier to the PR.

PR #1624, opened 2022-09-10
Fix records endpoint's DELETE behavior

### LINKED ISSUE #1613: API should respond with HTTP 404 status when attempting to delete a record that does not exist
## Reproduce

1. `DELETE` `http://localhost:8000/api/db/v0/tables/1/records/0/`
1. Expect response to have HTTP status code 404 and a body with an error message indicating the record could not be found.
1. Observe response to have HTTP status code 204

Anish9901:
Fixes #1613 

While trying to delete a non-existing record, the records endpoint responded with a 204(No content) instead of a 404(Not found).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/api/db/viewsets/records.py
+++ b/mathesar/api/db/viewsets/records.py
@@ -5,6 +5,7 @@ from rest_framework.response import Response
 from sqlalchemy_filters.exceptions import BadSortFormat, SortFieldNotFound
 
 import mathesar.api.exceptions.database_exceptions.exceptions as database_api_exceptions
+import mathesar.api.exceptions.generic_exceptions.base_exceptions as generic_api_exceptions
 from db.functions.exceptions import (
     BadDBFunctionFormat, ReferencedColumnsDontExist, UnknownDBFunctionID,
 )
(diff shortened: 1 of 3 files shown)


mathesar/api/db/viewsets/records.py
```diff
@@ -199,6 +199,8 @@ def partial_update(self, request, pk=None, table_pk=None):
 
     def destroy(self, request, pk=None, table_pk=None):
         table = get_table_or_404(table_pk)
+        if table.get_record(pk) is None:
+            raise NotFound
```
silentninja, line 203: `404` could be due to a missing Table or it could be due to a missing record. Even though the `NotFound` exception class could work in this context because the frontend does not care about the object that is missing right now, it would be better to use `NotFoundAPIException` as it would enable the backend to send a more useful error through the `error_code` property of the `NotFoundAPIException` on top of the `404 status code`

On a side note, `get_table_or_404` should be using `NotFoundAPIException` but it is still using `NotFound` because it was created before the exception handler was created and refactoring it is not a priority atm

silentninja (changes requested):

mathesar/api/db/viewsets/records.py
Anish9901: Ah, I did raise the `NotFoundAPIException` before(in my previous commit). I'll revert it.

Anish9901: I also found out `NotFound` getting raised here: https://github.com/centerofci/mathesar/blob/091ae33d4e354d80e9872f06d77649b548a3acb9/mathesar/api/db/viewsets/records.py#L123-L124

and 

https://github.com/centerofci/mathesar/blob/a45c557085c609aef12eaf38cc9601d8153540a6/mathesar/api/db/viewsets/columns.py#L172-L173

Issue #1604, opened 2022-09-07
2022-09-07 Demo Readiness meta issue


kgodey:
### Design - @ghislaineguerin 
- [ ] https://github.com/centerofci/mathesar/issues/1640
- [ ] https://github.com/centerofci/mathesar/issues/1641
- [ ] https://github.com/centerofci/mathesar/issues/1642

### Frontend - @pavish 
- [ ] #1647
- [ ] https://github.com/centerofci/mathesar/issues/1595

### Frontend - @rajatvijay
- [ ] #1650
- [x] Open Db page when Mathesar is opened
- [ ] https://github.com/centerofci/mathesar/issues/1648
- [ ] https://github.com/centerofci/mathesar/issues/1649

### Backend - @mathemancer 
- [x] Database page
    - [x] expose object descriptions (schemata, read/write)
    - [x] add table count and exploration count to schema list endpoint
- [x] Schema page
    - [x] expose object descriptions (tables, read/write)
    - [x] add table count and exploration count to schema detail endpoint
- [x] Data import page
    - [x] potential API work for removing column

### Notes on navigation design
We should redesign navigation once we've gone through the entire storyboard, but here are the notes we've come up with so far:
- Add current page to breadcrumb
- On the schema page, it's not clear how to get to other schemas
- Table features are not discoverable
- The schema icon + name should do what the little triangle does right now
- Database name in header is confusing, we should have some other solution for orienting the user to which Mathesar installation they are in

seancolsen:
> The schema icon + name should do what the little triangle does right now

Can someone clarify this point?

kgodey:
@seancolsen We thought the small triangle with the magnifying glass made a very small click target for a probably-frequent operation (navigating somewhere else).
<img width="286" alt="Screen Shot 2022-09-09 at 8 56 00 AM" src="https://user-images.githubusercontent.com/287034/189355074-c45286e0-b2be-4a88-87ac-687fab7ea5c5.png">

We wanted to get rid of the triangle and instead transfer opening the navigation menu to a larger click target. We have not come up with a specific design solution yet.

seancolsen:
> small click target for a probably-frequent operation

Makes sense. I'd like to be looped in to review that design work.

pavish:
I moved all data import related tasks to me, from @ghislaineguerin, as I'm already making those changes as part of my PR.

kgodey:
[...]

PR #1630, opened 2022-09-13
Schema description

rajatvijay:
This PR adds schema description while adding and editing a schema as per #1604

**Technical details**

This PR also introduces a new component called `WithViewMoreOrLessButton` which can be used to toggle between a completed and truncated text. 
Example Code: 
```typescript
<WithViewMoreOrLessButton
        content={schema.description}
        maxLength={MAX_DESCRIPTION_LENGTH_IN_LESS_VIEW}
      >
        <p class="description" slot="truncatedText" let:text>{text}</p>
        <p class="description" slot="completeText" let:text>{text}</p>
      </WithViewMoreOrLessButton>
```

**Screenshots**

Demo Loom: https://www.loom.com/share/2c448fd0eb9d4c4da1866ec2dcd2c3ac

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/component-library/with-view-more-or-less-button/WithViewMoreOrLessButton.svelte
```diff
@@ -0,0 +1,50 @@
...
+  <slot name="completeText" text={completeText} />
+  {#if showExpandCollapseButton}
+    <span
+      role="button"
+      class="expand-collapse-button"
+      on:click={showTruncatedView}
+    >
+      View Less
+    </span>
+  {/if}
```
pavish, line 30: Move the showTruncatedView/showCompleteView logic out of the parent `if` block.

Reason: This span has a role of a button and so we need to allow the browser to be able to focus on the element. When the parent 'if' invalidates, it removes this span and renders the span from the 'else' block, which is a different element. This will lead to losing the focus on the element, which is an accessibility concern.

mathesar_ui/src/pages/database/AddEditSchemaModalForm.svelte
```diff
@@ -0,0 +1,137 @@
...
+    CancelOrProceedButtonPair,
+    ControlledModal,
+    TextInput,
+  } from '@mathesar-component-library';
+  import { toast } from '@mathesar/stores/toast';
+  import TextArea from '@mathesar/component-library/text-area/TextArea.svelte';
+
+  export let controller: ModalController;
+  export let getNameValidationErrors: (name: string) => string[];
+  export let getDescriptionValidationErrors: (description: string) => string[];
```
pavish, line 14: Since we're not validating description, why do we need this additional prop and code around it?

I think we can remove the unnecessary code blocks in this component surrounding description validation.

mathesar_ui/src/component-library/with-view-more-or-less-button/WithViewMoreOrLessButton.svelte
```diff
@@ -0,0 +1,50 @@
+<script lang="ts">
+  export let content: string;
+  export let maxLength: number;
```
pavish, line 3: I'm curious as to why we are implementing this in js instead of css. This will result in the view looking differently with different text & languages.

A better option would be to truncate based on number of lines:
* Get `maxLines` as a prop.
* Use [webkit-line-clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) to truncate multi-line text using the `maxLines` prop. We currently use this for displaying multi-line text when a cell becomes active.
   - There are issues with this property, but it'll ensure the UI doesn't look different based on different text & languages.

This isn't needed for the live demo and no change is needed on this PR. We may need it for the release with localization. Please create an issue to track this.

mathesar_ui/src/component-library/with-view-more-or-less-button/WithViewMoreOrLessButton.svelte
```diff
@@ -0,0 +1,50 @@
...
+    <span
+      role="button"
+      class="expand-collapse-button"
+      on:click={showTruncatedView}
+    >
+      View Less
+    </span>
+  {/if}
+{:else}
+  <slot name="truncatedText" text={`${truncatedText}...`} />
```
pavish, line 32: This will always show the ellipsis even if the description is shorter than the maxLength. The ellipsis should not be displayed in that scenario.

pavish (changes requested):
@rajatvijay Looks good. I have a few suggested changes.

mathesar_ui/src/component-library/with-view-more-or-less-button/WithViewMoreOrLessButton.svelte
kgodey: The new design in Figma uses a tooltip instead of a "view more" button, so it might be easiest to remove this feature for now.

rajatvijay: Got it removing this component completely for now

mathesar_ui/src/pages/database/AddEditSchemaModalForm.svelte
rajatvijay: I kept it if later we want to add some validations but I agree that's just more mantainance for the code you MIGHT need later. Hence removed it

pavish (approved):
@rajatvijay Looks good!

PR #1710, opened 2022-09-21
Selected and active cell sync & Table inspector should not scroll the table view

### LINKED ISSUE #1534: Keep active cell and selected cells in sync
We need to make sure the active cell and the selected cells don't get out of sync when the user uses the keyboard to move the active cell. For example, if you begin with one cell selected and then press `DownArrow`, you'll have one cell selected and a different cell active which is very weird. I propose the following UX:

> Any time the active cell is moved via keyboard, the set of selected cells is reset to be a set containing only the active cell.

The above UX is similar to LibreOffice Calc and Google Sheets with a minor difference. In those spreadsheet applications, using `Tab` and `Shift+Tab` moves the active cell _within the range of selected cells_ (without modifying the selection). That's kind of neat, but I don't think it's worth it for us to implement, given the additional complexity.

Additionally, if I have one cell selected, and I click somewhere else on the page (un-setting the active cell), then the selected cells should be reset to an empty set.

### LINKED ISSUE #1612: Vertically scroll content of table inspector instead of viewport
If the contents of the table inspector are tall enough, they'll force the entire page to scroll. Instead, we should be scrolling only the content below the tabs within the table inspector

### LINKED ISSUE #1621: Visual distinction between Table Inspector and table vanishes when inspector overlaps cells
## Reproduce

- When the sheet does not consume the full width of its allocated space on the page, I see a drop shadow on the left of the Table Inspector

- But when the Table Inspector overlaps the sheet cells, that drop shadow disappears.

Peek 2022-09-09 14-48

CC @rajatvijay

### LINKED ISSUE #1625: Hide selection background when only one cell is selected
## Current behavior

- A blue background displays on all cells that are selected -- even when only one cell is selected.
- The background displays during edit mode too, which is particularly weird.

## Desired behavior

- The selection cell background is hidden when any of the following conditions are true:

    - only one cell is selected
    - The cell is in edit mode

    (Note that a cell can be in edit mode even when multiple cells are selected, so it's important to use both the above criteria)

CC @rajatvijay

rajatvijay:
Fixes #1612
Fixes #1534
Fixes #1625
Fixes #1621

This PR tackles three items
1. Vertically scroll the content of table inspector instead of viewport
2. Keep active cells and selected cells in sync when using the keyboard
3. Hide selection background when only one cell is selected
4. Visual distinction between Table Inspector and table vanishes when inspector overlaps cells

**Screenshots**
https://www.loom.com/share/174c7aa1a33740919e613bbd5bfa01d4
Screenshot 2022-09-22 at 2 03 52 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/table-data/selection.ts
```diff
@@ -79,6 +87,30 @@ export class Selection {
     document.addEventListener('mouseup', () => {
       this.onEndSelection();
     });
+
+    // Keep active cell and selected cell in sync
+    this.display.activeCell.subscribe((activeCell) => {
```
seancolsen, line 92: The `subscribe` function returns an `Unsubscriber` function, and to avoid memory leaks it's best-practice to find an appropriate place to call that Unsubscriber function. [Here](https://github.com/centerofci/mathesar/blob/e0ba80ba5abc495698f6f808dbbdf167784551a2/mathesar_ui/src/stores/table-data/records.ts#L307) is an example of how we store that Unsubscriber function and then [call it](https://github.com/centerofci/mathesar/blob/e0ba80ba5abc495698f6f808dbbdf167784551a2/mathesar_ui/src/stores/table-data/records.ts#L653) from a destroy method that [gets called](https://github.com/centerofci/mathesar/blob/e0ba80ba5abc495698f6f808dbbdf167784551a2/mathesar_ui/src/stores/table-data/tabularData.ts#L119) from `TabularData.destroy`. Can you implement the same pattern here? You'll need a new property on `Selection` to store the Unsubscriber function. Then you'll need a `destroy` method. Then you'll need to call `this.selection.destroy()` from `TabularData.destroy`

seancolsen (changes requested):
This looks great and works great! I'm requesting one small change to follow best practice in avoiding memory leaks.

mathesar_ui/src/stores/table-data/selection.ts
rajatvijay: @seancolsen makes sense. Done.

seancolsen (approved):
Perfect!

Issue #802, opened 2021-11-08
Prefetch SqlAlchemy property objects when querying Django Objects


silentninja:
## Problem
Django objects are a thin layer over the reflected database objects, so many fields are just properties that point to an equivalent SqlAlchemy object. The problem is that those underlying objects are generated only when an individual Django object is accessed, resulting in an N+1 query when fetching a list of Django objects.

## Proposed solution
Property fields that do not belong to Django but can be fetched in bulk should be pre-fetched when the Django objects are queried. We could make use of this [library](https://github.com/ionelmc/django-prefetch) to mimic the Django prefetch_related method 

## Additional Context

List of affected objects to be prefetched(Work in progress)|

#### TableSerializer
  - [x] #2883
  - [x] Prefetching Column Name and type
  
#### Column Serializer
 - [x] Prefetching Column Name and type
 
#### Schema Serializer
 - [ ] num_tables field
 - [ ] num_queries field 
 
 #### Constraint Serializer
 - [ ] #1764

mathemancer:
Do you have some examples of where this library would improve things?  Most of the attributes that we loop through (e.g., when listing tables) aren't related to the model via some foreign key relationship.  To solve the problem, we'd need to write a (very complicated) fetcher that could get, e.g., all column details for all columns of each table in the page.  I.e., the column details won't be stored in the (eventual) column model, so they'll need to be queried for each column in real-time anyway.  (Or am I missing something?)

silentninja:
Let's take the `tables/` get api. We fetch all the tables from the django database and the relevant _sa_table is reflected each time the property of the Django table is used(caching reduces this to reflect on first access only). But this could be made better by prefetching when querying for a list of tables using the library.

` _sa_tables = Prefetcher(
            filter = lambda ids: reflect_tables_from_oid(ids),
            reverse_mapper = lambda sa_table: [sa_table.id],
            decorator = lambda table, sa_table=(): setattr(table, '_sa_table', table),
mapper = lambda table: table.oid
        )
`

kgodey:
I think this sounds like a good idea. @silentninja I think you should spec out the specific queries you see us using the library for so that we can evaluate it more thoroughly before implementation.

kgodey:
Closing this since it's not going to be relevant with upcoming architectural changes.

PR #1712, opened 2022-09-21
Prefetch column properties when fetching list of columns

silentninja:
Related to #802 

1. This PR fetches column names in bulk when listing columns instead of making a DB call for each column
2. Also makes use of a single table instance when listing a set of columns belonging to a single table

### Additional context
There are a few other column properties(valid_target_types, default_dict) that are not prefetched in this PR because I would have to spend some time looking for a proper way to prefetch those properties. I would be fixing that in a different PR.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/columns/operations/select.py
```diff
@@ -95,14 +95,17 @@ def get_columns_name_from_tables(table_oids, engine, connection_to_use=None, fet
     return column_names
 
 
-def get_columns_name_from_attnums(table_oid, attnums, engine, connection_to_use=None):
+def get_columns_name_from_attnums(table_oid, attnums, engine, connection_to_use=None, fetch_as_map=False):
```
dmos62, line 98: I think that changing a function's behaviour based on a boolean switch has a negative effect on refactoring later. It also leads to bigger, more complicated methods. Could you split this behaviour into its own method? Something like `get_column_name_from_attnums_as_map`? Thanks!

dmos62 (changes requested):

db/columns/operations/select.py
silentninja: I get your concern but I think We can break the function when the function gets more complicated. We already have too many of these conversion functions atm.

dmos62: > We can break the function when it gets more complicated

That's a bad idea, because this also affects refactoring. Once you want to break it up, you'll have to walk through its usages (it's a public function) and split them up into whether they're using this version of the function or that version.

So it's not only about the function being complicated, but also about what it does to refactoring. Plus, when you're reading the usage of this method, it's another cognitive step to check what state the boolean switch is in.

> We already have too many of these conversion functions already.

You're not reducing the number of branches with similar-ish code. Due to above reasons, it's better to have functions than switches.

By the way, if you want to get some code reuse, wrap the common code into its own private function:

```
    statement = _get_columns_name_from_attnums([table_oid], attnums, engine, connection_to_use=None)
    column_names_tuple = execute_statement(engine, statement, connection_to_use).fetchall()
```

Maybe `_get_column_names_tuple(table_oid, attnums)`. That will go a longer way than a switch.

mathesar/models/base.py
```diff
@@ -204,13 +207,17 @@ def clear_name_cache(self):
 
 class ColumnNamePrefetcher(Prefetcher):
     def filter(self, column_attnums, columns):
-        pass
+        if len(columns) < 1:
+            return []
+        table = list(columns)[0].table
+        return get_columns_name_from_attnums(table.oid, column_attnums, table._sa_engine, fetch_as_map=True)
```
dmos62, line 213: This is probably meant to be `get_column_attnum_from_names_as_map`.

dmos62 (changes requested):

dmos62 (approved):

silentninja:
The refactor looks good, thanks. I have some comments on the but I think we can refactor it later after the next backend meeting

PR #1723, opened 2022-09-23
Specify types for DBFunction call returns

### LINKED ISSUE #1692: Query RPC endpoint results in a 500 when summarization when preproc functions are used.
## Description
* Using the Query RPC endpoint, run a query with an `initial_column` which contains preproc functions. Eg., `author_website` column which can utilize the preproc function `extract_uri_scheme`.
* Create a summarization transform with a grouping expression utilizing the preproc function.
* Notice that the server throws a 500.

## Example Request
```
{
   "base_table":6,
   "initial_columns":[
      {
         "alias":"author_website",
         "id":34,
         "display_name":"author_website"
      }
   ],
   "transformations":[
      {
         "type":"summarize",
         "spec":{
            "grouping_expressions":[
               {
                  "input_alias":"author_website",
                  "output_alias":"author_website",
                  "preproc":"extract_uri_scheme"
               }
            ],
            "aggregation_expressions":[]
         },
         "display_names":{}
      }
   ],
   "parameters":{
      "limit":100,
      "offset":0
   }
}
```

## Response - 500 from server
```
Environment:


Request Method: POST
Request URL: http://localhost:8000/api/db/v0/queries/run/

Django Version: 3.1.14
Python Version: 3.9.8
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'rest_framework',
 'django_filters',
 'django_property_filter',
 'mathesar']
Installed Middleware:



  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
  File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 466, in handle_exception
    response = exception_handler(exc, context)
  File "/code/mathesar/exception_handlers.py", line 53, in mathesar_exception_handler
    raise exc
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/code/mathesar/api/db/viewsets/queries.py", line 76, in run
    column_metadata = query.all_columns_description_map
  File "/code/mathesar/models/query.py", line 173, in all_columns_description_map
    return {
  File "/code/mathesar/models/query.py", line 174, in <dictcomp>
    alias: self._describe_query_column(sa_col)
  File "/code/mathesar/models/query.py", line 166, in _describe_query_column
    'type': sa_col.db_type.id,
  File "/code/db/columns/base.py", line 211, in db_type
    return get_db_type_enum_from_class(self.type.__class__)
  File "/code/db/types/operations/convert.py", line 37, in get_db_type_enum_from_class
    raise UnknownDbTypeId

Exception Type: UnknownDbTypeId at /api/db/v0/queries/run/
Exception Value: 

```

mathemancer:
Fixes #1692

This PR specifies the return type of various function calls involved in running DBFunctions.

**Technical details**

The problem in the linked issue is that the return type of SQL functions isn't easy to determine from SQLAlchemy. Many don't even return a type that we use in our app (i.e., they usually return a generic SQLAlchemy type rather than a PostgreSQL-specific SQLAlchemy type).

This PR is essentially a temporary fix. It will be quite difficult to maintain the needed specificity within the SQLAlchemy framework. This is because the return type of a given function call is determined at compile time if possible (compiling the Python into SQL). Often, this results in the wrong type being chosen, or no type at all being chosen for the column resulting from a function call. In the case of summarization, since the column often changes type, this is a particular problem.

To completely fix this, we'd need to move away from SQLAlchemy, or write some kind of machinery that reflects function info from the DB at runtime (i.e., when the function is called on the DB), then modifies the query result column types accordingly. When we do move away from SQLAlchemy, this will be easier, because we can simply request result types along with result data when we run the query, and report both back without worrying about modifying incorrectly-specified SQLAlchemy objects.

The most important addition of this PR is modifying the `db.functions.base.sa_call_sql_function` function to create an anonymous class describing the function and a specified return type on-the-fly as the function call is being specified. The result is that SQLAlchemy magically notices this class, and adds some metadata to the context in which it interprets the function call (and its response). This is pretty weak, since the return type can only be specified once in code using this method, but the return type of SQL functions is dependent on the argument type(s). At the moment, we don't have any functions for which this problem is relevant, but (for example) `SUM`ming a column isn't possible to properly specify in this way.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/functions/base.py
```diff
@@ -13,15 +13,46 @@
...
+            "sa_call_sql_function should be called with the return_type kwarg set"
+        )
+        # We can't use PostgresType since we don't want an engine here
+        return_type = TEXT
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="The GenericFunction")
+        # Creating this type registers the function (more importantly,
+        # its return type) with SQLAlchemy. **magic!!**
+        type(
```
dmos62, line 47: I wouldn't expect this `type` call to have side-effects. Shouldn't it be assigned to something?

db/functions/base.py
```diff
@@ -13,15 +13,46 @@
...
+    function_name: string giving a namespaced SQL function
+    *parameters:   these will be passed directly to the generated function.
+    return_type:   an SQLAlchemy type class
+    """
+    if return_type is None:
+        warnings.warn(
+            "sa_call_sql_function should be called with the return_type kwarg set"
+        )
+        # We can't use PostgresType since we don't want an engine here
+        return_type = TEXT
```
dmos62, line 41: > We can't use PostgresType since we don't want an engine here

I wonder what you mean by that. The types under `db.types` should completely replace all SQLAlchemy type usage in the rest of codebase. Sounds like you're running into some tech debt. I'd be willing to help solve this. Hoping this doesn't prove necessary, since feels like a serious regression.

db/tests/functions/test_func_imports.py
```diff
@@ -0,0 +1,6 @@
+from db.functions.known_db_functions import known_db_functions
+
+
+def test_all_functions_no_raw_func():
```
dmos62, line 4: Could you add a docstring explaining why we're forbidding raw `func` calls?

db/tests/types/test_email.py
```diff
@@ -3,22 +3,35 @@
...
 
 
 def test_domain_func_wrapper(engine_with_schema):
     engine, _ = engine_with_schema
-    sel = select(sa_call_sql_function(email.EMAIL_DOMAIN_NAME, text("'test@example.com'")))
+    sel = select(
+        sa_call_sql_function(
+            email.EMAIL_DOMAIN_NAME,
+            text("'test@example.com'"),
+            return_type=PostgresType.TEXT.get_sa_class(engine)
```
dmos62, line 18: Is this why you didn't want to use `PostgresType` in `sa_call_sql_function`? There is indeed no good reason why `PostgresType.get_sa_class` should require an engine. I solved this elsewhere... Let me see.

dmos62: There's `db.engine.dummy_engine` for that. See its docstring and its usage in `db/types/operations/convert.py::_get_sa_type_class_id_from_ischema_names`. Here's what you can do:

```diff
diff --git a/db/types/base.py b/db/types/base.py
index d3a65fec..593d1a06 100644
--- a/db/types/base.py
+++ b/db/types/base.py
@@ -4,6 +4,7 @@ from sqlalchemy import text, create_engine as sa_create_engine
 
 from db import constants
 from db.utils import OrderByIds
+from db.engine import get_dummy_engine
 
 
 class DatabaseType(OrderByIds):
@@ -17,13 +18,14 @@ class DatabaseType(OrderByIds):
         """
         return self.value
 
-    def get_sa_class(self, engine):
+    def get_sa_class(self):
         """
         Returns the SA class corresponding to this type or None if this type is not supported by
         provided engine, or if it's ignored (see is_ignored).
         """
+        dummy_engine = get_dummy_engine()
         if not self.is_ignored:
-            ischema_names = engine.dialect.ischema_names
+            ischema_names = dummy_engine.dialect.ischema_names
             return ischema_names.get(self.id)
 
     def is_available(self, engine, type_ids_on_database=None):
```

dmos62 (changes requested):

db/functions/base.py
mathemancer: That's why it's magic. It registers a type with the python interpreter, which gets picked up by SQLAlchemy.

I also find it arcane, hence the comment. Maybe I can extend that comment it a bit?

db/functions/base.py
mathemancer: We can't get the SQLAlchemy class associated with a type without an engine. To get an engine inside this function seemed pretty contrived. I agree that's a bit irritating, but I didn't want to do a major refactor to get the engine in here (it would have to be passed from the calling model to be proper in our current design, but how should that happen for a `DBFunction` call?). So, I went with the hack. 

I think it's okay, since I consider all code that uses SQLAlchemy types to be temporary at this point.

db/tests/functions/test_func_imports.py
mathemancer: Sure, that's a good idea.

db/functions/base.py
dmos62: Oh, that's what the comment above talks about. Magic indeed.

db/functions/base.py
mathemancer: Okay, I see your later comment. Let me try to get that working.

db/functions/base.py
dmos62: Now that I get it, the comment makes sense, but that took me a while. Might want to add a few more details or something. Not requesting this change though.

db/functions/base.py
mathemancer: Okay, I've looked into this. Long story short, I don't want to do the amount of refactoring that would be necessary to fix this preexisting problem in this PR. I think we should instead 
1. make an issue to note the problem, and
2. consider whether it's worth refactoring at all, since we are likely going to get rid of SQLAlchemy anyway, and this will mean reworking how we handle types.

dmos62 (approved):
Good stuff! Thanks for refactor!

Issue #1727, opened 2022-09-23
Improve performance when syncing Django objects with reflected db objects


silentninja:
## Problem
Reflecting and syncing up db objects into the Mathesar Django database leads to a lot of queries(around 28 queries on Django and 300 queries on the user database) 
## Proposed solution
Fetch and updates objects in bulk instead of a single query for each object

PR #1728, opened 2022-09-23
Improve performance when syncing DB objects into Django database

silentninja:
Related to #1727 

The PR improves performance when syncing 
- Schemas
- Tables
- Columns

**Technical details**
I also removed Deferring unique attnum constraint check on Column as it got carried when implementing `move_columns` by dropping the related tables, but since we are reusing the existing table, it isn't necessary anymore and deferring the constraint check causes issue when creating objects in bulk.

**Additional Context**
I will be adding Performance improvements for constraints in a separate PR

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/reflection.py
```diff
@@ -42,53 +44,101 @@ def reflect_databases():
...
+    schemas = []
     for oid in db_schema_oids:
-        schema, boolean = models.Schema.current_objects.get_or_create(oid=oid, database=database)
-    for schema in models.Schema.current_objects.all():
+        schema = models.Schema(oid=oid, database=database)
+        schemas.append(schema)
+    models.Schema.current_objects.bulk_create(schemas, ignore_conflicts=True)
+    deleted_schemas = []
+    for schema in models.Schema.current_objects.all().select_related('database'):
         if schema.database.name == database and schema.oid not in db_schema_oids:
```
dmos62, line 60: Sort of unrelated, but notice that we're checking for equality between a database's name and a database model instance here. Fixed this in my performance pr, but might want to change it here as well, in case that fix gets lost in merge conflicts.

mathesar/reflection.py
```diff
@@ -42,53 +44,101 @@ def reflect_databases():
...
-            defaults={'display_options': None})
-        if not created and column.display_options:
+    tables = []
+    for oid, schema_oid in db_table_oids:
+        schema = next(schema for schema in schemas if schema.oid == schema_oid)
+        table = models.Table(oid=oid, schema=schema)
+        tables.append(table)
+    models.Table.current_objects.bulk_create(tables, ignore_conflicts=True)
+    # Calling signals manually because bulk create does not emit any signals
+    models._create_table_settings(models.Table.current_objects.filter(settings__isnull=True))
```
dmos62, line 82: `_create_table_settings` is the signal callback, right? And `models.Table.current_objects.filter(settings__isnull=True)` are newly created tables? Would really like some descriptive variable names to immediately know what's what.

Btw, if we change what's listening for table creation signals, this piece of code would become out-of-sync with signal callbacks. Maintainability/debugging nightmare. Would also love a solution for that.

mathesar/reflection.py
```diff
@@ -42,53 +44,101 @@ def reflect_databases():
...
+    attnum_tuples = get_column_attnums_from_table(table_oids, engine)
+
+    _create_reflected_columns(attnum_tuples, tables)
+
+    _delete_stale_columns(attnum_tuples, tables)
+    # Manually trigger preview templates computation signal
+    for table in tables:
+        models._compute_preview_template(table)
+
+    _invalidate_columns_with_incorrect_display_options(tables)
```
dmos62, line 105: Love how these discreet pieces of logic are broken out!

mathesar/reflection.py
```diff
@@ -125,11 +175,13 @@ def reflect_new_table_constraints(table):
...
-            reflect_schemas_from_database(database.name)
-        for schema in models.Schema.current_objects.all():
-            reflect_tables_from_schema(schema)
-        for table in models.Table.current_objects.all():
-            reflect_columns_from_table(table)
-        reflect_constraints_from_database(database.name)
+        databases = models.Database.current_objects.filter(deleted=False)
+        for database in databases:
+            reflect_schemas_from_database(database)
+        # Prefetching queries make use of a single db engine, so we need to
```
dmos62, line 181: Is this comment meant to be here? Don't understand it.

dmos62 (changes requested):

mathesar/reflection.py
silentninja: > Btw, if we change what's listening for table creation signals, this piece of code would become out-of-sync with signal callbacks. Maintainability/debugging nightmare. Would also love a solution for that.

Its a known issue with Django, one solution I have used in the past dealt with using a custom bulk_create method so that it is not spread out. But lets visit it when when start using bulk_create in more than one place

silentninja: `_create_table_settings`  is usually called by a signal but can also be triggered manually. What name would you suggest for it?

`models.Table.current_objects.filter(settings__isnull=True)` is basically any table whose setting is `null`. Why would you want a descriptive name for it, the statement already says what it is doing

silentninja:
Assigning it to @mathemancer since @dmos62 is on leave.

mathemancer (approved):
I think this looks good, with the exception of a couple very minor requests.

Edit: Please merge after changing those function names.

db/tables/operations/select.py
```diff
@@ -75,12 +75,12 @@ def get_map_of_table_oid_to_schema_name_and_table_name(
     return table_oids_to_schema_names_and_table_names
 
 
-def get_table_oids_from_schema(schema_oid, engine, metadata):
+def get_table_oids_from_schema(schema_oids, engine, metadata):
```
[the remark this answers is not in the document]
silentninja: Thanks for catching it!

mathesar/reflection.py
dmos62: > lets visit it when when start using bulk_create in more than one place

My concern is that things will start to break "when when start using bulk_create in more than one place" and it will be hard to know why.

> _create_table_settings is usually called by a signal but can also be triggered manually. What name would you suggest for it?

`_table_creation_callback` would tell me what I need to know.

I'd probably have it in the `models.Table` namespace (either in its own file, which `Table` doesn't have, or a static class method) and have it call `_create_table_settings`.

> models.Table.current_objects.filter(settings__isnull=True) is basically any table whose setting is null.

`created_tables_that_didnt_trigger_signals` would tell me what I need to know here.

> I would like to understand the rationale behind setting a descriptive name for it because the statement already informs what it is doing

The comment gave me valuable hints, but I still need to parse the code.

silentninja: > My concern is that things will start to break "when when start using bulk_create in more than one place" and it will be hard to know why.

The problem is, it is not desirable to group all signals under one common callback function. For example in this case, `Table` has two signals `sync_table_constraints` and `create_table_settings`. Calling `sync_table_constraints` when creating Tables in bulk is not the desired behaviour as we create constraints in a separate call.

mathesar/reflection.py
silentninja: Thanks for the catch!

PR #1738, opened 2022-09-28
Improve UX for FK and PK cell interactions

### LINKED ISSUE #1689: Navigating from a Record Page to another Record Page doesn't work
I tested this using: https://staging.mathesar.org/mathesar_tables/5/44/1267

Mathesar_Home

I would expect that clicking where I highlighted would take me to the Record Page for the checkout. But it doesn't work, I instead have a mostly blank Items Record Page.

Screenshot 2022-09-16 at 17-49-01 Mathesar - Home

### LINKED ISSUE #1713: Improve UX for FK and PK cell interactions
- FK cell
    - Remove navigation link
    - Show a dropdown arrow on the right side, which opens the record selector
    - Remove the magnifying glass icon
    - Single click on active cell opens the record selector
    - Navigating to the linked record is performed via a context menu or a popover that displays on hover (like in Google Sheets)
- PK cell
    - Remove navigation link (place it in the row header instead)

seancolsen:
Fixes #1713
Fixes #1689

## Demo

https://user-images.githubusercontent.com/42411/192673946-ab3e3450-49b5-48b4-910c-5af33c08ea63.mp4

## Notes

- This PR introduces a minor regression for the Linked Record _Input_ (not Cell), where the magnifying glass icon to launch the record selector no longer appears. I plan to change it to use the dropdown arrow, but that work will be slightly easier after I've implemented the display of the record summary within the input. I've opened #1743 to track that work.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey @ghislaineguerin I invite you to look at the demo video in this PR and provide feedback on the design changes here, since these changes relate to a call we were all in. Keep in mind this is only for the PK and FK _cells_ -- it doesn't cover any of the behavior after the record selector is opened.

ghislaineguerin:
- I like what you did with the ID field and I agree that the record page link should remain visible at all times.
- I'm not so sure about right-aligning the ID field. Have we considered a separate column for the record page link?
- The FK cell interactions make sense. I'm wondering if the down arrow could mislead the user into thinking a dropdown menu will show up. Maybe we could use a different icon like a link icon or a magnifying glass icon?

seancolsen:
Thanks @ghislaineguerin.

- Nice idea with the separate column for the record page link instead of putting it in the PK cell. I'll work on that in this PR, and I'll put the PK value back to left-aligned.
- I'll try using the magnifying glass instead of the down arrow as well and see how that looks.

seancolsen:
I added some more commits.

- Record page link in the row header

    > I agree that the record page link should be moved out of the ID column. During the meeting, I think we discussed moving it to the row header,

    Right. I meant to mention this in my demo video. I initially went this direction of putting the record page link in the row header, but I didn't like it for the following reasons:

    - The link felt like it was more associated with the row _index_ than it was with the PK value. Already, I think some users will be confused by the difference of the two numbers, especially since they appear directly adjacent, and will sometimes be identical. Putting the link in the row header felt to me like it would increase this confusion.

    - I also wanted to address the concern of click areas being too small in some cases. Being that we're also using the row header to select rows, it was tough fitting those two click interactions into the same small space.

- Record page link in a separate column

    Ok, so I started working on this today, but I have some more concerns that I'd like to discuss before proceeding.

    - For starters, it's not super straightforward to add a new column to the table. Possible of course, but it's not a quick thing I can easily try out. I got about half way there with a couple hours of work.

    - More importantly, I'm wondering how we might want to display record page links within the data explorer. I think it's worth giving this some thought from a design perspective. Not that we'd want to implement that functionality right away. But it might actually make sense to use the PK column for the record page link in the data explorer, in which case it would be nice to stick with the same design for the table page.

- I flipped the alignment of the PK cell content, and I think it looks a lot better.

    image


I'd like to get this PR merged without fiddling with the record page link anymore because I have lots of other things to work on, and my sense is that putting it in the PK cell is "good enough" for now. Is that okay with you, @ghislaineguerin and @kgodey?

seancolsen:
Oh, and also, I'll keep the record selector trigger icon as a dowwards chevron for the time being since @kgodey prefers that.

kgodey:
> putting it in the PK cell is "good enough" for now.

I agree, especially since we're also going to have the record page link in the table inspector, which will help with discoverability. I do think the alignment looks better.

>  might actually make sense to use the PK column for the record page link in the data explorer

I'm not sure this will make sense, because I don't expect most queries will use the PK column. I don't think we need to worry about this concern, though.

pavish (approved):
@seancolsen The code looks good. I haven't put thought into the usability but it seems everyone is in agreement and I see no major concerns.

PR #1747, opened 2022-10-03
Improve Linked Record Input UX

### LINKED ISSUE #1743: Update Linked Record Input to follow new UX implemented for Linked Record Cell
- [ ] Display the dropdown arrow like we do in the cell
- [ ] Launch the record selector when clicking the input
- [ ] Focus the input when its containing label is clicked
- [ ] Give the user a way to clear the value
- [ ] Hyperlink to Record Page

seancolsen:
Fixes #1743

## Demo

https://user-images.githubusercontent.com/42411/196701177-ad8baff8-9a0b-44c4-8f51-b0e1bb073b16.mp4

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey and @ghislaineguerin if you'd like to weigh in on the UX changes here, take a look at the demo video.

kgodey:
@seancolsen I think all your UX choices here make sense. I would keep the single click behavior because it makes the record selector more discoverable and I think erring towards discoverability is good for our live demo goal. We can always change it after (external) usability testing if needed.

mathesar_ui/src/component-library/label/Label.svelte
```diff
@@ -3,18 +3,36 @@
...
   $: ({ inputId, disabled } = controller);
+
+  function handleClick() {
+    const inputElement = document.querySelector(`#${$inputId}`) as
+      | HTMLInputElement
+      | undefined;
+    // Why do we imperatively focus the input element when the DOM performs this
+    // behavior natively? Because we might need to associate a label with a
+    // custom input component that doesn't have that doesn't contain any native
+    // DOM elements with that behavior.
```
rajatvijay, line 20:
```suggestion
    // Why do we imperatively focus the input element when the DOM performs this
    // behavior natively? 
    // Because we might need to associate a label with a
    // custom input component that doesn't have or doesn't contain any native
    // DOM elements with that behavior.
```

mathesar_ui/src/component-library/label/Label.svelte
```diff
@@ -3,18 +3,36 @@
...
 <!--
   TODO use a better class name. We already had some "label" classes in use. We
   need to figure out how to avoid naming collisions here.
 -->
-<label for={$inputId} class="label-component" class:disabled={$disabled}>
+<label
+  for={$inputId}
+  class="label-component"
+  class:disabled={$disabled}
+  id={getLabelIdFromInputId($inputId)}
```
rajatvijay, line 33: just curios what is the use of giving the `id` prop to the label tag?

rajatvijay (approved):
nit: The code looks good to me a small input from a UX perspective: 

The new UI gives me an impression of a multi-select field. I would not be surprised if a non-technical user tried clicking on it expecting to link multiple items. 

Not sure about the reasons for removing the clear icon apart from making it look like the linked record cell. But I would also suggest 
1. Consider an external link icon(`fa-up-right-from-square`) instead of the current cross icon in the `LinkedRecord` Tag. This will also remove the need for an underline which currently looks weird since it's nested inside a colored tag that is nested inside an input element. 
2. Bringing back the `ClearCueButton` instead of the down chevron. If the single-click-to-clear link is not something that the user will need to use too often, we can also move that option inside the modal. 

cc: @kgodey @ghislaineguerin

mathesar_ui/src/component-library/label/Label.svelte
seancolsen: In the record page, I wanted a way to semantically associate the LinkedRecordInput elements (which are not actual `<input>` elements, but rather `<span>` elements with `role="listbox"`) with their corresponding form labels. Associating the label with the input improves a11y and also allow users to focus the input by clicking on the label. Because LinkedRecordInput is a fake input, I need to resort to using aria attributes instead of the standards approaches. So in `LinkedRecordInput.svelte`, I'm putting `aria-labelledby={getLabelIdFromInputId(id)}` on the `<span>`. At the other end, `Label.svelte`  sets `id={getLabelIdFromInputId($inputId)}` on the `<label>` element and the two id values match up, associating the label with the input. This was the best way I could find to do that. If you have better way in-mind, I'm open to changing this.

seancolsen:
@rajatvijay

> The new UI gives me an impression of a multi-select field. I would not be surprised if a non-technical user tried clicking on it expecting to link multiple items.

I agree this is a problem, though I'm not sure it'll be a _major_ problem. I'm open to finding ways to fix it.

_But_, your proposed changes are basically taking us back to the previous design. I liked that design initially, but I do have a moderate preference for the UX implemented in this PR over the UX before this PR. There are a lot of subtle details to consider that I think are difficult to efficiently discuss without hopping on a call. Being that you have some opinions and ideas, I think we ought to incorporate you into the next phase of design iteration for this UX. I'm sure it will continue to evolve.

PR #1793, opened 2022-10-11
DB Page re-design

### LINKED ISSUE #1648: Implement new Database Page design
[Design is on Figma](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=218%3A2)
This should also cover:
- [ ] Show count of tables and explorations
- [ ] Create schema - get description from user

### LINKED ISSUE #1740: Layout components to accommodate new designs
## Problem
In order to implement the new designs for the database page and schema page the first thing that we need to do is create components for common layouts. 

## Proposed solution
After looking at the designs here are the common layouts IMHO.

### LayoutWithHeader
This component will have two responsibilities: 
1. Adding a static header on the top.
2. Rendering the children inside the `main` with some `max-width` so that the layout does not spread across the width of the larger screens. This `max-width` behavior can be toggled using a prop since we might not need this restriction for certain pages for example the Table page. 

Screenshot 2022-09-28 at 2 55 44 PM

_NOTE: This will be purely a presentational component._

#### Can we reuse the existing `LayoutWithHeader` component?
We can use this existing component but this might need us to check the layout at other places too where this component is used and make sure no bugs are introduced there which might delay the delivery of the DB and schema page. I suggest we create a new component with the current needs and gradually start using at other places once we do their designs. 

#### Header
The header in itself will be a separate component that gets rendered inside the `LayoutWithHeader` component. I don't see the header requiring any props as of now but if we need this in the future, the `LayoutWithHeader` should support all of the header props too and pass them down to this component. 

### PageTitleAndMeta
Screenshot 2022-09-28 at 3 02 21 PM

This component will be responsible for the following things: 
1. Rendering the `type` of the entity. It will be of type `string` but for now `schema` / `database` will be the possible values. 
2. Rendering the `name` of the entity. So if the entity type is a `database` then it will be the name of the DB. 
3. Rendering the `icon` using the `icon` prop of type `IconProps`.
4. Rendering the `description`, if any. 
5. Rendering an action button using the `buttonProps` prop. 

Screenshot 2022-09-28 at 2 57 46 PM

_NOTE: This will be purely a presentational component._

rajatvijay:
Fixes #1740 
Fixes #1648 

**What this PR includes**
1. Implements the new designs for the DB page as mentioned here: https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=218%3A2
2. Created some common layout components to be used for schema page re-design too as mentioned here https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=604%3A3200

**Screenshots**
https://www.loom.com/share/abdeb8a295a944f78b79f1c920ff056d

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/component-library/dropdown-menu/DropdownMenu.svelte
```diff
@@ -14,7 +18,9 @@
     {#if icon}
       <Icon {...icon} />
     {/if}
-    <span class="label">{label}</span>
+    {#if label}
```
rajatvijay, line 21: made the `label` optional

mathesar_ui/src/component-library/dropdown/Dropdown.svelte
```diff
@@ -40,6 +40,7 @@
   $: tgClasses = calculateTriggerClass(triggerClass, showArrow);
 
   function toggle(e: Event) {
+    e.preventDefault();
```
rajatvijay, line 43: To make it work when wrapped with an anchor tag. Since the event is already not bubbling up and this component will always necessarily handle the click event, preventing the default behaviour should be fine

mathesar_ui/src/component-library/menu/ButtonMenuItem.svelte
```diff
@@ -7,6 +7,7 @@
 
   export let icon: IconProps | undefined = undefined;
   export let disabled = false;
+  export let danger = false;
```
rajatvijay, line 10: introduced a new danger mode to color the option like delete red.

mathesar_ui/src/component-library/tutorial/Tutorial.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
```
rajatvijay, line 1: Will create separate issues for writing stories for Alert and Tutorial component once the PR is merged

mathesar_ui/src/pages/database/DatabasePage.svelte
```diff
@@ -52,36 +61,97 @@
     targetSchema = schema;
     addEditModal.open();
   }
+
+  function deleteSchema(schema: SchemaEntry) {
```
rajatvijay, line 65: Moved this inside the DatabasePage component so that this component drives all the business logic and child components are only presentational

mathesar_ui/src/pages/database/DatabasePage.svelte
```diff
@@ -52,36 +61,97 @@
...
+      type: 'database',
+      icon: iconDatabase,
+    }}
+  >
+    <Button slot="action" on:click={addSchema} appearance="primary">
+      <Icon {...iconAddNew} />
+      Create Schema
     </Button>
-  </div>
+    <!-- TODO: Still fixing this... -->
```
rajatvijay, line 98: Fix coming in the next PR. Do not want to block the review on this

mathesar_ui/src/component-library/text-input/TextInput.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
```
pavish, line 1: Please rename this component differently, and revert back the name of BaseTextInput component to TextInput.

Our default input components should do no more than the basic operations.

mathesar_ui/src/components/SchemaName.svelte
```diff
@@ -4,6 +4,34 @@
...
+  :global(.container.is-locked .icon) {
+    background-color: var(--slate-800);
+  }
+
+  :global(.container .name) {
+    font-size: 1.428rem;
+    font-weight: normal;
+    color: var(--slate-800);
+    margin-left: 0.25rem;
+  }
```
pavish, line 36: `global` tag should not be on the outermost scope in a component. This will make all the styles here global eg., `.container .icon` will be a globally available css.

Instead change this to:
```
.container {
  :global(.icon) { ... }
}
```

pavish (changes requested):
I've not done a thorough review since this is assigned to Sean. I've commented on some top level issues that I skimmed through the PR.

mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.scss
```diff
@@ -1,4 +1,5 @@
 .cancel-or-proceed-button-pair {
   display: flex;
-  justify-content: space-between;
+  justify-content: end;
+  gap: 1rem;
```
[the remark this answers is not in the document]
pavish: We need to support browsers people currently use, and Safari desktop browser is in use by atleast 9.6% of users (based on [wikipedia](https://en.wikipedia.org/wiki/Usage_share_of_web_browsers)). 

Do not use `gap` or anything not supported by the last 2 versions of all the browsers. (IE can be ignored, since Edge fills for it).

mathesar_ui/src/component-library/text-input/TextInput.svelte
rajatvijay: Conclusion: Use the following options in the fallback

1. Use InputGroup to provide the functionality of prefix and suffix
2. Create a new component wrapper for TextInput with suffix capability

mathesar_ui/src/layouts/LayoutWithHeader2.svelte
```diff
@@ -0,0 +1,43 @@
...
+    height: 100vh;
+  }
+  .app-layout-header {
+    position: sticky;
+  }
+  .app-layout-content {
+    position: relative;
+  }
+  .app-layout-content.max-width {
+    max-width: var(--max-layout-width, 750px);
```
seancolsen, line 35: I'd like to use `rem` here instead of `px`.

mathesar_ui/src/component-library/tutorial/Tutorial.svelte
seancolsen: I wouldn't bother spending any time on that. I sometimes write stories for components because in certain situations speeds up the process of building lower-level components and testing their various scenarios in isolation. But writing the stories for the sake of documentation is something to defer until well after our first release I'd say.

mathesar_ui/src/pages/database/DatabasePage.svelte
seancolsen: Nice. That is better. Thanks!

mathesar_ui/src/component-library/button/Button.scss
```diff
@@ -23,43 +23,48 @@ button.btn {
...
   }
 
   &.btn-primary {
-    background: #39a0f5;
-    border-color: #39a0f5;
-    color: #fff;
+    background: var(--brand-500);
+    border-color: var(--brand-500);
+    color: var(--white);
+    font-weight: 500;
```
seancolsen, line 40: I would be cautious of using non-standards font weights like this (anything other than 400 or 700), so I'd suggest removing `font-weight: 500;`. Although perhaps you have some rationale that I don't understand here. My thinking is like this... Since we're using system fonts, I think it'll be difficult for us to preview the effect of non-standard font weights across everyone's different environment. In my experience, non-standard font weights can sometimes look a bit awkward, like, they end up being more bold that you think they'll be, or the subpixel rendering or kerning degrades slightly. If we were defining (and shipping) a variable font, I'd be in support of getting fancy with subtle font weights like this. But since we're using system fonts, I think we ought to stick to `font-weight` values of either `normal` (i.e. 400) or `bold` (i.e. 700).

mathesar_ui/src/component-library/dropdown/Dropdown.svelte
```diff
@@ -40,6 +40,13 @@
   $: tgClasses = calculateTriggerClass(triggerClass, showArrow);
 
   function toggle(e: Event) {
+    /**
+     * To make it work when wrapped with an anchor tag.
+     * Since the event is already not bubbling up
+     * and this component will always necessarily handle the click event,
+     * preventing the default behaviour should be fine
+     */
+    e.preventDefault();
```
seancolsen, line 49: This is a hack that's not great for accessibility. Ideally we shouldn't ever have a `<button>` element within an `<a>` element because that's invalid HTML. I opened #1840, proposing an alternate approach. Working on that ticket is low priority, but this concept would be good to keep in-mind for new code you write going forward.

mathesar_ui/src/components/NameAndDescInputModalForm.svelte
```diff
@@ -18,6 +22,8 @@
   import { toast } from '@mathesar/stores/toast';
   import TextArea from '@mathesar/component-library/text-area/TextArea.svelte';
 
+  export let helpText = '';
```
seancolsen, line 25: `helpText` would work better as a slot than as a prop. I see in the parent component you're passing a huge string into this prop. If you use a slot, you'll be able to move that text out of a string and put it directly into the template. This also makes it possible to put formatting in the help text.

mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.svelte
```diff
@@ -2,21 +2,15 @@
...
 
   const cancelButtonDefaults: ButtonDetails = {
     label: 'Cancel',
-    icon: iconCancel,
   };
 
   const proceedButtonDefaults: ButtonDetails = {
     label: 'Proceed',
-    icon: iconProceed,
   };
```
seancolsen, line 14: Can you add these icons back in as defaults? I know that Figma doesn't show icons for these buttons, but that's okay. I think it's probably just because Ghislaine didn't want to take the time to put icons in all the buttons.

mathesar_ui/src/App.svelte
```diff
@@ -98,7 +133,14 @@
 
     --page-padding: 1em;
 
-    color: var(--color-text);
+    --max-layout-width: 700px;
+    // Setting the header height here
+    // since when the header is fixed
+    // we can use this variable to add margin-top
+    // to the below header content container
+    --header-height: 60px;
```
seancolsen, line 141: I would prefer `rem` units here.

mathesar_ui/src/pages/database/DatabasePage.svelte
```diff
@@ -1,18 +1,26 @@
 <script lang="ts">
```
seancolsen, line 1: This file still seems a bit messy. For example, `hasOnlyPublicSchema` is unused, and there is a bunch of commented-out code. I notice that this PR is supposed to close #1648. Is there another ticket which will remain open to track the work that's still needed here?

mathesar_ui/src/components/SchemaName.svelte
seancolsen: I'd like to take this critique a step further...

First of all, I agree with @pavish. Yes, `.container :global(.icon)` is better than `:global(.container .icon)`.

But also, `.container :global(.icon)` is still using global CSS which we ought to be using only as a last resort. Global CSS breaks encapsulation which I think hinders maintainability.

I added 45ec13325aa77328b2a05e5389de23626e82e322 to demonstrate the approach I would take here.

1. I made the low level component, `NameWithIcon`, slightly more flexible. Now it accepts a prop to configure whether or not it renders its icon within a rounded box. It also accepts a new CSS variable, `--icon-opacity`.

1. The mid-level component, `SchemaName`, now sets the color of the icon based on whether the schema is locked. It sets the icon opacity but does _not_ set the font size because it doesn't know where it will be rendered.

1. The high-level component, `SchemaRow`, specifies the font size and the boxed icon.

This way all the concerns are separated to their various levels so that the lower level components can be reused and refactored without worrying about what happens outside them.

seancolsen (changes requested):
## High-level comments

- Nice work making the app look like Figma!

- I think you might be taking the Figma design a little _too_ literally. Not a big deal. This is always a tricky balance to establish. As we move forward, I suggest you take the Figma design a little more lightly. Don't be afraid to _doubt_ what you see in Figma, use your own design sense, and seek clarification for changes that you think could potentially just be artifacts of the speed of the design process.

## Questions

- Why do we have `LayoutWithHeader` and `LayoutWithHeader2`? Can't we adapt `LayoutWithHeader` to meet our needs for this PR?


## Things to handle before merge

- _Comments on the diff (noted separately)_

- The text color changes make the site navigation difficult to use on other pages. For example:

    image

- I'd like to revert button placement changes within CancelOrProceedButtonPair so that the cancel button goes back to the left side.
([Matrix discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$2JNCCypJzNK70nma7OoKkK15J6RCB6m6LGFbGsTYvmU?via=matrix.mathesar.org&via=matrix.org))

- I'd like to ensure that the CSS variables within the component library do not depend on CSS variables defined outside the component library. Although we don't intend to release the component library any time soon, this dependency problem still has consequences for our local development in the short term. For example, if you look at the TextInput component in Storybook, you'll see that it has no border. That's because it has `border: 1px solid var(--slate-200);` but `--slate-200` is only defined within `App.svelte` (outside the component library). Pavish [suggested](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$4DhBSPvspfeNhn1KWWk0wx7C56WGLOU8XcgW-EvqFTM?via=matrix.mathesar.org&via=matrix.org) we move all those CSS variable definitions into `variables.scss`.

- Throughout the codebase I see eight occurrences of `font-size: 1.142rem;`. That's weird for two reasons... First, I'd expect that, at this point, we'd be using a CSS variable like `--text-size-large`. Second, it's weird to me that we're setting a larger text size at such a low level in so many places. I [asked Ghislaine in Matrix](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$96IClUWvd7ziC5vyZtCPUDlJPs6T1DC7J94szLSpUpQ?via=matrix.mathesar.org&via=matrix.org) if this was her intention with the design, and she said no. I suspect that we'll probably want to just remove all those occurrences of `font-size: 1.142rem;`, keeping the default font size in those cases.

mathesar_ui/src/components/NameAndDescInputModalForm.svelte
rajatvijay: Create a good first here: #1858

mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.scss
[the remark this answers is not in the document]
rajatvijay: Removed here: bf4f022

mathesar_ui/src/layouts/LayoutWithHeader2.svelte
rajatvijay: Done 52eb3ef

mathesar_ui/src/component-library/button/Button.scss
rajatvijay: Done 52eb3ef

mathesar_ui/src/App.svelte
rajatvijay: Done 52eb3ef

mathesar_ui/src/components/AppSecondaryHeader.svelte
```diff
@@ -0,0 +1,38 @@
...
+  </div>
+</div>
+
+<style>
+  .container {
+    background-color: var(--slate-400);
+    padding: 2rem;
+  }
+  .content.max-width {
+    max-width: var(--max-layout-width, 700px);
```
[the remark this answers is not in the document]
rajatvijay: Done 52eb3ef

mathesar_ui/src/component-library/cancel-or-proceed-button-pair/CancelOrProceedButtonPair.svelte
rajatvijay: Done 4cd6563

mathesar_ui/src/pages/database/DatabasePage.svelte
rajatvijay: Done beb3d45

rajatvijay:
@seancolsen 

>Why do we have LayoutWithHeader and LayoutWithHeader2? Can't we adapt LayoutWithHeader to meet our needs for this PR?
As mentioned #1740 
Screenshot 2022-10-21 at 3 27 52 PM

I will suggest deleting the older layout component and removing the suffix `2` from the new layout component once all the design work is done. I will create a PR for this then. 

>I'd like to ensure that the CSS variables within the component library do not depend on CSS variables defined outside the component library. Although we don't intend to release the component library any time soon, this dependency problem still has consequences for our local development in the short term. For example, if you look at the TextInput component in Storybook, you'll see that it has no border. That's because it has border: 1px solid var(--slate-200); but --slate-200 is only defined within App.svelte (outside the component library). Pavish [suggested](https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$4DhBSPvspfeNhn1KWWk0wx7C56WGLOU8XcgW-EvqFTM?via=matrix.mathesar.org&via=matrix.org) we move all those CSS variable definitions into variables.scss.

I am tackling this in my schema re-design PR #1850  (Commit ID coming soon...)

>Throughout the codebase I see eight occurrences of font-size: 1.142rem;. That's weird for two reasons... First, I'd expect that, at this point, we'd be using a CSS variable like --text-size-large. Second, it's weird to me that we're setting a larger text size at such a low level in so many places. I [asked Ghislaine in Matrix](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$96IClUWvd7ziC5vyZtCPUDlJPs6T1DC7J94szLSpUpQ?via=matrix.mathesar.org&via=matrix.org) if this was her intention with the design, and she said no. I suspect that we'll probably want to just remove all those occurrences of font-size: 1.142rem;, keeping the default font size in those cases.
Done here: 52eb3ef

mathesar_ui/src/component-library/text-input/TextInput.svelte
rajatvijay: Done: 434b97

seancolsen: @rajatvijay, in this PR's diff, I'm still seeing some changes in `TextInput.svelte`. I'd expect that this PR would not have any changes in that file. Those classes `'input-element'` and `'text-input'` _do_ need to be these for styling purposes. For example, if you edit a schema's name, you'll see that the name input is not styled correctly.

seancolsen (changes requested):
@rajatvijay Your recent changes seem to resolve most (but not all) of the critique here.

- I've left a new comment on on sub-thread and resolved all the others.

    Additionally...

- I don't see any response to this point from my first review:

    > I'd like to revert button placement changes within CancelOrProceedButtonPair so that the cancel button goes back to the left side. ([Matrix discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$2JNCCypJzNK70nma7OoKkK15J6RCB6m6LGFbGsTYvmU?via=matrix.mathesar.org&via=matrix.org))


- Your changes in 52eb3efaa55f7088163f803c0c421afaa9ab3be3 addressed the _first_ "weird thing" I mentioned in my review comment, but not the _second_. I said:

    > It's weird to me that we're setting a larger text size at such a low level in so many places. I [asked Ghislaine in Matrix](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$96IClUWvd7ziC5vyZtCPUDlJPs6T1DC7J94szLSpUpQ?via=matrix.mathesar.org&via=matrix.org) if this was her intention with the design, and she said no.

    Can we remove those occurrences of `font-size: var(--text-size-large);`, keeping the default font size?

mathesar_ui/src/component-library/text-input/TextInput.svelte
rajatvijay: @seancolsen that was a bad revert on my part. Here is the latest commit for this: 1841f1c

rajatvijay:
>  * I've left a new comment on on sub-thread and resolved all the others.
>       Additionally...

replied to the comments.

 
>     * I don't see any response to this point from my first review:
>       > I'd like to revert button placement changes within CancelOrProceedButtonPair so that the cancel button goes back to the left side. ([Matrix discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$2JNCCypJzNK70nma7OoKkK15J6RCB6m6LGFbGsTYvmU?via=matrix.mathesar.org&via=matrix.org))

Got missed between all of the other comments: 3670b2d

>     * Your changes in [52eb3ef](https://github.com/centerofci/mathesar/commit/52eb3efaa55f7088163f803c0c421afaa9ab3be3) addressed the _first_ "weird thing" I mentioned in my review comment, but not the _second_. I said:
>       > It's weird to me that we're setting a larger text size at such a low level in so many places. I [asked Ghislaine in Matrix](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$96IClUWvd7ziC5vyZtCPUDlJPs6T1DC7J94szLSpUpQ?via=matrix.mathesar.org&via=matrix.org) if this was her intention with the design, and she said no.
>       
>       
>       Can we remove those occurrences of `font-size: var(--text-size-large);`, keeping the default font size?

Now I understood, that makes sense: 58252d4

seancolsen (approved):

PR #1810, opened 2022-10-12
make row order deterministic in more cases

### LINKED ISSUE #1786: Deterministic row ordering
## Current behavior

- The records API returns records in an arbitrary, nondeterministic order.
- I find it somewhat disorienting to refresh a page and see different records, even when no data has changed.

## Desired behavior

- Ordering by PK (asc) is appended to the sorting criteria automatically.
- We could do this either on the back end or on the front end.
- We could even choose to display the auto-added sorting condition to the user, though this would require some additional UX design work.

mathemancer:
Fixes #1786 

Now, we automatically append a last-step ordering by the primary key field(s) of a table when getting its records. It does the same for the results returned by a general query (i.e., an exploration).

**Technical details**

This does _not_ handle the case where both are true:

- The user does not specify a fully-determined order, and
- there is no primary key.

It would be possible to eke out a bit more ordering, since we could append all non-included columns to the requested `order_by` clause to make it more deterministic, but I chose not to, since this would have a performance impact, and any table created through Mathesar will have a primary key anyway. The performance impact would be due to sorting by non-indexed columns.

In case NO primary key can be successfully detected, we sort by the complete row set, unless the user specifies a sorting. This may have a performance impact, but should only occur on queries run through data explorer, in cases where no sorting is requested by the user.

**TODO**

For general queries without a detectable primary key, we should try to choose a reasonable join key if one exists and sort by that for determinism. This would improve performance over the sorting by all columns approach, but will get quite fiddly if we have to detect the join keys using SQLAlchemy.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@seancolsen The only review I'm requesting from you is that this solves the problem you described from the user perspective.

dmos62 (approved):

seancolsen (commented):
@mathemancer 

I'm loading the Publications table page with this PR and still seeing out-of-order rows.

image

Row 1 has an id value less than the id in row 2.

I can appreciate that -- due to the very nature of the nondeterminism here -- it's possible we are observing different behavior and you may in fact be seeing rows ordered by ids. So I'm not sure what else to do to aid in review/troubleshooting here. I'm happy to hop on a call if needed.

mathemancer:
Fascinating. I clicked on a couple tables to make sure the order looked good, but didn't notice anything. Now, with more info, I've found that I clicked the wrong tables to check. I've determined that:

- Authors, Publications, and Publishers are all consistently out of order on my machine.
- Checkouts, Items, and Patrons are all consistently properly ordered on my machine.

@seancolsen Is this the case on your machine as well?

mathemancer:
Moving to `draft` status while I figure out why the fix isn't working properly.

seancolsen:
@mathemancer

> - Authors, Publications, and Publishers are all consistently out of order on my machine.
> - Checkouts, Items, and Patrons are all consistently properly ordered on my machine.
> 
> Is this the case on your machine as well?

No. Currently all of that is true for me except that Items is out of order. I have wiped out my `.volumes` and rebuilt Mathesar a number of times since we began using this Library Management schema and I'm about 80% confident that I've observed different sorting behavior after rebuilding. Although I only have a rudimentary understanding of the inner-working of Postgres, I would expect to observe these ordering inconsistencies, given this excerpt from the Postgres [docs](https://www.postgresql.org/docs/current/queries-order.html) (emphasis added):

> The actual order in that case will depend on the scan and join plan types **and the order on disk**, but it must not be relied on

My hunch is that when we load the library data, it gets placed on-disk with an ordering subject to the fragmentation of other data on-disk at that point, though also I understand even less about how SSDs work nowadays. Just a hunch. I'm about 60% confident that ordering is consistent after I re-build, but re-building seems to shuffle the ordering somewhat.

mathemancer:
> . I'm about 60% confident that ordering is consistent after I re-build, but re-building seems to shuffle the ordering somewhat.

This is my experience as well. I finally figured out the problem. Unbeknownst to me, we'd added another way to get records from a table so we can join previews in. I fixed the ordering (and the problem you'd noticed was already fixed) on a method for getting table records that's not used in most cases any more. The downside is, I'm now trying to solve a much more complicated problem of providing a default ordering on query results (where a primary key is often not present). The plus side is, when I succeed, query results will also be consistently ordered.

mathemancer:
@seancolsen @dmos62 This needs another review, since I had to make some major changes.

As noted above, we're using the querying infrastructure for many (but not all) table row requests now, and I'd neglected that part of things. Unfortunately, adding default sorting to general queries makes this PR quite a bit more complicated.

mathemancer:
This time, I checked all tables to make sure they were properly ordered.

db/columns/utils.py
```diff
@@ -41,3 +41,24 @@ def init_mathesar_table_column_list_with_defaults(column_list):
             if c.name == constants.ID:
                 c.autoincrement = False
     return default_columns + given_columns
+
+
+def get_column_obj_from_relation(relation, column):
+    try:
+        column = find_column_by_name_in_relation(relation, column)
```
dmos62, line 48: Why is the `col_name_string` parameter being passed a `column`? `find_column_by_name_in_relation` seems to expect a string as that parameter.

db/columns/utils.py
```diff
@@ -41,3 +41,24 @@ def init_mathesar_table_column_list_with_defaults(column_list):
...
+    except AttributeError:
+        column = relation.columns[column.name]
+
+    return column
+
+
+# TODO deal with quotes; still better than the default
+def find_column_by_name_in_relation(relation, col_name_string):
```
dmos62, line 56: Hm, seems a bit weird that we have two methods looking up column names, one delegating to the other, but then also both primarily attempting to do column indexing (`relation.columns[...]`). I'm having trouble understanding the intent and the challenges faced here.

db/tests/queries/test_initial_relation.py
```diff
@@ -41,6 +42,8 @@ def test_shallow_link(shallow_link_dbquery):
     assert records == [(1, 'uni1'), (2, 'uni1'), (3, 'uni2')]
 
 
+# TODO determine why this is failing when all run, but not the individual file
+@pytest.mark.skipif
```
dmos62, line 46: Hm, seems serious. We're talking about fixture contamination, right? Each test passes in isolation, but when they are all run in order this test fails, or did I misunderstand?

mathesar/tests/api/test_record_api.py
```diff
@@ -241,7 +241,7 @@ def test_record_list_sort(create_patents_table, client):
     assert len(response_data['results']) == 50
 
     assert mock_get.call_args is not None
-    assert mock_get.call_args[1]['order_by'] == order_by
+    assert mock_get.call_args[1]['order_by'][:len(order_by)] == order_by
```
dmos62, line 244: Seems like an implementation detail. We do similar mocking in a few places and I'm not certain that it's beneficial. Just a note.

dmos62 (changes requested):

db/columns/utils.py
```diff
@@ -41,3 +41,24 @@ def init_mathesar_table_column_list_with_defaults(column_list):
             if c.name == constants.ID:
                 c.autoincrement = False
     return default_columns + given_columns
+
+
+def get_column_obj_from_relation(relation, column):
```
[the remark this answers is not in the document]
dmos62: Oh wait, `MathesarColumn` is actually in the `db` namespace. But, you could argue that it should be interchangeable with SA `Column` and hence the same in these contexts. My main point is still the same though.

db/records/operations/sort.py
```diff
@@ -0,0 +1,59 @@
+from sqlalchemy import select
+from db.columns.utils import get_column_obj_from_relation
+from db.records.exceptions import BadSortFormat, SortFieldNotFound
+
+
+def get_default_order_by(relation, order_by=[]):
+    # Set default ordering if none was requested
+    pkey = getattr(relation, 'primary_key', None)
+    pk_cols = getattr(pkey, 'columns', pkey)
```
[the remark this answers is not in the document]
mathemancer: The issue is that some `primary_key` attributes are of the `PrimaryKeyConstraint` type from SA, which has a `columns attribute. Others are of the `ColumnCollection` type. This logic coerces either type to a `ColumnCollection` (or some other iterable collection; there are actually a few options). The other alternative is that `pk_cols` ends up as `None` (if `pkey` defaults to `None`). Thus, when you call `set` below, it's always on a collection. No magic needed.

db/columns/utils.py
mathemancer: That's why it's in a `try` block. This function needs to deal with a `Column`, a `MathesarColumn`, or a (possibly namespaced) column name string.

db/columns/utils.py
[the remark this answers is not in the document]
mathemancer: I think I do want to reconsider these names. I know there's some type fuzziness in parts of this logic.

db/tests/queries/test_initial_relation.py
mathemancer: Exactly

db/columns/utils.py
mathemancer: The outer function finds a column in the relation whether it's a `Column`, `MathesarColumn`, or just a name string. It first delegates finding by string to the inner function, since that needs to handle different levels of namespacing (could be schema.table.name, table.name, or just name) and it's a bit complicated. The inner function will return an `AttributeError` when it's not passed a string (or something string-like enough to have a `split` method). In that case, the outer function tries to just get the table-associated version of column by looking for one with the name the same as whatever object it's been passed.

seancolsen (approved):
These changes work for me now, from a user's perspective. I haven't look at the diff though.

dmos62 (approved):
Thanks for making the logic clear! Appreciate it!

db/columns/utils.py
dmos62: We had a conversation about this code, but it didn't really dawn on me until later that day when I was daydreaming that I would try to avoid having a routine whose input can be one of multiple different data structures (here that's `MathesarColumn`, SA Column or a column name string). I would consider the fact, that same data is being passed around in different forms, as the problem. Also, if a single routine accepts different forms of input, its number of states increases and it's less predictable (same goes for output, btw). I'd say that the simpler and more efficient we can make the data structures being passed around the better. Off topic, but reminds me of those programming maxims that say that solving a problem is about finding the data structure, not the algorithm. This might also relate to those conversations we've been having about duct typing and type annotations.

PR #1829, opened 2022-10-16
Fixes: #1772 Close context menu on Esc

### LINKED ISSUE #1772: Close context menu on Esc
## Current behavior

- We have a custom context menu that opens when right-clicking on a table cell.
- The only way to close it is to click somewhere outside the menu.

## Desired behavior

- Pressing the Esc key closes the context menu

lakshya-dhariwal:
Fixes #1772

**Technical details**
Added an key down event listener to window using svelte:window, triggers a check if key is escape . If it is it calls an existing close function that closes the context menu.

Testing: Manually tested if feature works or not. Then ran lint and eslint commands (for specific file , as mentioned in docs) to check other issues.

## Checklist
 
  index.md`).
  visible errors.

### DIFF
--- a/mathesar_ui/src/component-library/context-menu/ContextMenu.svelte
+++ b/mathesar_ui/src/component-library/context-menu/ContextMenu.svelte
@@ -71,7 +71,15 @@
       close();
     }
   }
+
+  function handleKeydown(event: KeyboardEvent) {
+    if (event.key === 'Escape' && isVisible) {
+      close();
+    }
+  }
 </script>
+
+<svelte:window on:keydown={handleKeydown} />
 
 <div bind:this={element} class="context-menu-wrapper">
   {#if isVisible}


seancolsen:
Thanks for this PR, @lakshya-dhariwal.

Your changes worked great.

I also added 76ccdfa03e5a3f1806d5989beebee7fd4df3ca2f to make some small improvements...

- In your changes, the event listener got added to `window` for every context menu instance in the app, and we actually have quite a few context menu instances. There could be ~100 on the page at once. This is not terrible currently, but hypothetically it could become a performance issue to have so many event listeners that get called whenever the user presses Esc. With my change, the event listener only gets added when the context menu opens, and then it gets removed when the context menu closes.

- Sometimes we have multiple event listeners registered for actions on the Esc key at the same time. For example, if you double click to edit a cell, you can press Esc to leave edit mode. If you trigger the context menu while you're in edit mode, now Esc can do multiple things. Should it close the context menu? Should it leave edit mode? Should it do both? I would argue that is should close the context menu and leave the user in edit mode -- essentially moving back "one step" from the thing you did, but not all the way. Before my changes, pressing Esc in this situation would close the context menu _and_ leave edit mode. With my changes, you remain in edit mode here.

We welcome more PRs from you!

PR #1851, opened 2022-10-20
Data explorer performance

### LINKED ISSUE #1841: Improve performance when fetching records using DBQuery
## Problem

We are using DBQuery infrastructure to fetch records in the following API's
1. `/db/v0/queries/run/` 
2. `/tables/<id>/records/`

Both of these queries tend to be really slow especially when foreign key columns are involved.

Most of the queries issued to the Database are the metadata reflection related queries happening in `db.queries.base.DBQuery.initial_relation`

## Analysis

Taking a deeper look at the queries, it looks like the `db.tables.operations.select.reflect_table` used by `db.queries.base.DBQuery.initial_relation` is the cause of numerous reflection queries https://github.com/centerofci/mathesar/blob/bf2693ae020a6daa22b0afb188b19f39a8604fc5/db/queries/base.py#L114

I tried using a `cached metadata` and the function still ends up with duplicated reflection calls, I assume this is happening due to `extend_existing=True` used by the  https://github.com/centerofci/mathesar/blob/bf2693ae020a6daa22b0afb188b19f39a8604fc5/db/tables/operations/select.py#L19

## Proposed solution

We should consider setting `use_existing` when reflecting a `Table` or refactor the `db.queries.base.DBQuery.initial_relation` to fetch table names and columns directly from the database skipping SqlAlchemy objects.

silentninja:
Fixes #1841

I have added cached metadata to the function used within `DBQuery`. This improves the performance to a reasonable extent, especially when using multiple foreign key columns.

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/queries/base.py
+++ b/db/queries/base.py
@@ -16,6 +16,7 @@ class DBQuery:
             engine,
             transformations=None,
             name=None,
+            metadata=None
     ):
         self.base_table_oid = base_table_oid
         for initial_col in initial_columns:
@@ -24,6 +25,7 @@ class DBQuery:
         self.engine = engine
         self.transformations = transformations
         self.name = name
+        self.metadata = metadata
 
     # mirrors a method in db.records.operations.select
     def get_records(self, **kwargs):
@@ -97,8 +99,7 @@ class DBQuery:
 
     @property
     def initial_relation(self):
-        # TODO reuse metadata
-        metadata = get_empty_metadata()
+        metadata = self.metadata if self.metadata else get_empty_metadata()
         base_table = reflect_table_from_oid(
             self.base_table_oid, self.engine, metadata=metadata
         )
@@ -111,7 +112,7 @@ class DBQuery:
             We use the function-scoped metadata so all involved tables are aware
             of each other.
             """
-            return reflect_table_from_oid(oid, self.engine, metadata=metadata)
+            return reflect_table_from_oid(oid, self.engine, metadata=metadata, keep_existing=True)
 
         def _get_column_name(oid, attnum, metadata):
             return get_column_name_from_attnum(oid, attnum, self.engine, metadata=metadata)
--- a/db/tables/operations/select.py
+++ b/db/tables/operations/select.py
@@ -1,5 +1,5 @@
 from sqlalchemy import (
-    Table, select, join, inspect, and_, cast, func, Integer, literal, or_
+    Table, select, join, inspect, and_, cast, func, Integer, literal, or_,
 )
 from sqlalchemy.dialects.postgresql import JSONB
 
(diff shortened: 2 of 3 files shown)


db/queries/base.py
```diff
@@ -24,6 +25,7 @@ def __init__(
         self.engine = engine
         self.transformations = transformations
         self.name = name
+        self.metadata = metadata
```
dmos62, line 28: This can result in stale metadata; makes `DBQuery` stateful. If you create a `DBQuery` with metadata returned by `get_cached_metadata`, then call `reset_reflection`, and then do something with that `DBQuery` instance from earlier, it will still use the old metadata that was supposed to be invalidated by `reset_reflection`, but wasn't because `DBQuery` is effectively caching it. Could you make the `initial_relation` property a method like `get_initial_relation(metadata)`, and make sure that `DBQuery` operations always use up-to-date metadata?

dmos62 (changes requested):
Good work! I requested changes to how metadata is passed to `DBQuery`.

db/queries/base.py
silentninja: I am okay with passing `metadata` as a parameter but I am curious to know why `metadata` shouldn't be part of `DBQuery`. `DBQuery` is stateful, it has access to `engine` instance, shouldn't it be the same case with `metadata`?

I am suggesting that `DBQuery` should not be reused or we should be rethinking how the metadata caching actually works(should this be a singleton).

dmos62: Whether `engine` should be static in a `DBQuery` is debatable. Here it's sort of treated as a reference to a database, so in that sense it is static. Notice that we don't actually mutate engines and we don't expect them to be mutated, but we do mutate metadata. If you fetch a cached metadata instance, it's only meant to be valid at that point in time, so it shouldn't be cached again.

Not sure what you mean when you say that a DBQuery should be a singleton, or do you mean `MetaData`?

I don't think there's a problem with reusing `DBQuery`, unless it's stateful.

silentninja: > Not sure what you mean when you say that metadata should be a singleton.

We don't have a need for multiple `Metadata` objects in our case.

Usually, when dealing with caches, the cache object would be a singleton and the same instance would be accessed by the codebase, and if the cache has to be cleared, we would be calling a method of the object responsible for clearing the cache(in this case [sqlalchemy.schema.MetaData.clear()](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData)) this avoids passing metadata to each method

silentninja: Can we move this refactor to a different issue? Changing `initial_relation` to a function would trigger various refactors as other properties make use of this property

dmos62: If you're proposing a cache on the `db` namespace, I'd say that any kind of state is inappropriate in `db`. In `mathesar`, we still need multiple concurrent instances of metadata, because we can't yet reuse metadata everywhere. And, hopefully, we'll never have to, since that would be a lot of work.

If by singleton you mean having only a single instance of `MetaData` at any given time, that's not possible currently (in `mathesar` namespace) or not desirable (in `db` namespace`).

> [sqlalchemy.schema.MetaData.clear()](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData)

That's an interesting idea. Currently the cached metadata is either a `MetaData` or `None`, so you'd have to go from that to it always being `MetaData`.

> Can we move this refactor to a different issue?

I'm hesitant. I wouldn't want to run into this unexpected statefulness while working on something else. A bug involving this wouldn't be easy to find.

If you decide to postpone fixing this, I won't oppose it, but please create an issue to track this. Also, a warning in a code comment next to the `metadata` attribute might be helpful.

db/queries/base.py
```diff
@@ -16,6 +16,7 @@ def __init__(
             engine,
             transformations=None,
             name=None,
+            metadata=None
```
dmos62, line 19: If the refactor to remove `metadata` is postponed, this line has to be fixed now. `metadata` should never be `None`.

dmos62 (changes requested):

db/queries/base.py
silentninja: I made the fixes, can you review it again? #1877 tracks the proposed refactoring changes.

dmos62 (commented):
I don't have anything to add in a review. This adds a race condition, as I pointed out. I'm not giving approval, but you don't strictly need it.

silentninja:
Based on your previous review 
> This can result in stale metadata; makes DBQuery stateful. If you create a DBQuery with metadata returned by get_cached_metadata, then call reset_reflection, and then do something with that DBQuery instance from earlier, it will still use the old metadata that was supposed to be invalidated by reset_reflection, but wasn't because DBQuery is effectively caching it

I was thinking of 
```
db_query = DBQuery(metadata=get_cached_metadata())
db_query.do_something_that_uses_metadata()
reset_reflection()
db_query.do_something_that_uses_metadata()
```
which is rare given that `DBQuery` is mostly used as a read only API. Even in that case why would there be a race condition?

silentninja:
I didn't realise a race condition was also involved due to `cached_metadata` instance being shared between multiple threads.

silentninja:
I am merging this as I think making the requested changes would lead to refactors that could break something that is already working and doesn't seem to good use of the time as the risks seem to be something that could be easily dealt with. Moreover future refactors to the `db` namespace would eliminate the need for such refactors

PR #1856, opened 2022-10-20
Record selector bug fixes and UX improvements

### LINKED ISSUE #1552: Fix small bugs in Record Selector
- [ ] Test will all input types to identify more bugs
- [ ] Make boolean input look and function like other inputs
- [ ] Don't show loading spinner when nested selector is open
- [ ] Text input instead of textarea, especially in Chrome
- [ ] Blurring a number input with a value should not re-search
- [ ] Add help text to improve case where target table contains no rows
- [ ] Programmatically focus the first non-pk value on open.
- [ ] Programmatically focus re-focus the FK input after closing the record selector
- [ ] Remove border on LinkedRecordInput and other inputs like Date.
- [ ] Fix date-picker opening as soon as the modal opens, and searching with it doesn't work
- [ ] ColumnResizers for sheet need to have lower z-index than the modal and ColumnResizers within the record selector need to have higher z-index than the adjacent table cells (so that they display wider).
- [ ] Allow opening the record selector from the filters dialog dropdown while keeping the filters dialog open and underneath the record selector.
- [ ] All rows have tooltip "Create New Record and Select It". This tooltip should only apply to the ghost row.
- [ ] Column heading truncates column name too early. There's some unnecessary right padding.
- [ ] Allow negative numbers in NumberInput

### LINKED ISSUE #1553: Improve Record Selector UX
- Help content
    - [ ] Change title to "Pick a `<table_name>` Record"
    - [ ] Consider adding help text near the title that explains the affect the record selection will have within the parent context
    - [ ] Remove help text about keyboard actions, as hopefully the simpler UX now no longer necessitates it

- Creating a new record via the Record Selector
    - [ ] Headings to separate the new record for the existing records
    - [ ] Remove the "+NEW" thing because it's confusing and people think it's a button.
    - [ ] Add new column to the table which contains a "Pick" button for each row.
    - [ ] Experiment with putting the "Add new" UI below the results

- Nested selector
    - [x] New "stacked modal" layout (done in #1761)
    - [ ] Esc should close nested selector
    - [ ] Use a dark overlay instead of a light one, making it clearer that the background window is inactive.
    - [ ] Make the active FK cell peek through the overlay, improving aesthetics

- General
    - [ ] Automatically adjust column widths based on content
    - [ ] Hyperlink to table from within the title
    - [ ] #1871
    - [ ] Improve aesthetics during loading

- LinkedRecordInput
    - [ ] Disable hyperlinks
    - [ ] Don't open nested selector when clicking to clear value
    
- Results
    - [ ] Retain the results when the nested selector is open
    - [ ] Highlight matching cells using a background on the whole cell as well as a background on the matching substring text. Sometimes the matching text falls beyond the overflow and it's hard to tell which cells match and which don't.
    - [ ] vertical scrolling only for results
    - [ ] horizontal scrolling only for table
    - [ ] Use shadows to indicate when horizontal/vertical scrolling is possible
    - [ ] all cell should display as read-only. E.g. remove dropdown arrow for FK cells. 
    - [ ] merge the keyboard highlighting behavior with the mouse highlighting behavior

### LINKED ISSUE #1639: Creating a new record from the Record Selector does not select the record

### LINKED ISSUE #1683: Bug with primary keys in Record Selector
@pavish reports: 
> Searching for primary key, allows me to try adding a record even when a record with the primary key already exists. This results in an error, but that isn’t visible to the user.

seancolsen:
## This PR

- Should hopefully make the Record Selector **ready for live demo and release**
- Fixes #1552
- Fixes #1553
- Fixes #1639
- Fixes #1683

## Demo of some notable design changes

https://user-images.githubusercontent.com/42411/199802784-682655ca-7807-4617-85bb-1fa1153b6405.mp4

## Remaining issues to handle later

- `01. Live Demo`
    - [InvalidDatetimeFormat error when attempting to filter on a partially-entered date](https://github.com/centerofci/mathesar/issues/1890) (**back-end bug** affecting Record Selector and table filtering)

- `03. Post-Release Improvements`
    - [Keep records in store while fetching](https://github.com/centerofci/mathesar/issues/1893) (We need this so that the Record Selector table doesn't change widths while the search is loading)
    - [Allow keyboard navigation for complex inputs elements within the Record Selector](https://github.com/centerofci/mathesar/issues/1888)
    - [Re-focus table cell after opening record selector on that cell](https://github.com/centerofci/mathesar/issues/1892)
    - ["Today" in date picker is confusing in the context of the Record Selector](https://github.com/centerofci/mathesar/issues/1896)

- `2023 or later`
    - [Show cell-level errors after failure to create record via the record selector](https://github.com/centerofci/mathesar/issues/1882)
    - [Implement a focus trap for modals](https://github.com/centerofci/mathesar/issues/1886)
    - [Improve logic for displaying popovers on truncated cells in Record Selector](https://github.com/centerofci/mathesar/issues/1887)
    - [Highlight matching cells in temporal cell types within the Record Selector](https://github.com/centerofci/mathesar/issues/1889)
    - [Automatically scroll Record Selector results in a smart manner](https://github.com/centerofci/mathesar/issues/1891)
    - [Close all nested record selectors when navigating to a table from the record selector window title](https://github.com/centerofci/mathesar/issues/1894)
    - [Use bidirectional substring matching within the Record Selector fuzzy search algorithm](https://github.com/centerofci/mathesar/issues/1895)

## What's next

- After merging this PR, I'd like to invite the whole team to hammer at the Record Selector a bit to see if we can uncover any remaining bugs or UX issues. As such, I don't want to hold up this PR due to small things we can fix in that next phase of polish.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey and @ghislaineguerin You can look at the demo video here for a tour of the UX design changes to the Record Selector.

kgodey:
@seancolsen I don't think I'll be able to watch the video in full in the next couple of days (I don't yet have the necessary focus), but I skimmed through it and from what I see, the changes look great. I'm looking forward to playing around with it once it gets merged.

mathesar_ui/src/components/cell-fabric/data-types/boolean.ts
```diff
@@ -27,11 +29,11 @@ function getLabels(
...
+): SingleSelectCellExternalProps<boolean | null> {
   const labels = getLabels(column.display_options);
   return {
-    options: [true, false],
-    getLabel: (value?: boolean) => {
+    options: [null, true, false],
+    getLabel: (value?: boolean | null) => {
       if (isDefinedNonNullable(value)) {
         return value ? labels[0] : labels[1];
       }
```
pavish, line 39: Do we need to allow `null` here? If we do, we should represent Null using the NULL component instead of empty string.

While we can allow null, there's no way for any other components (checkbox, text boxes, textarea etc.,) to ever set null as a value, so this would be a special option allowed only by the Select menu within a cell.

seancolsen: > Do we need to allow `null` here?

The reason for this change is to improve the behavior of the input for boolean columns. Before this change, the input only had options for "true" and "false". The records selector needs to open with blank values for all inputs, and the user needs to be able to set a value and then set it _back_ to some sort of blank value after setting it. If you can think of a better way to accomplish what I'm after, then I'm open to changing the approach here. 

> we should represent Null using the NULL component instead of empty string.

I'm not sure about this. I think that would look pretty weird.

mathesar_ui/src/components/cell-fabric/data-types/components/linked-record/LinkedRecordInput.svelte
```diff
@@ -51,23 +56,40 @@
...
+  /**
+   * If this LinkedRecordInput in placed inside an AttachableDropdown, we want
+   * to tell the dropdown not to close when the user clicks within the Record
+   * Selector UI, so we set the modal element to "accompany" the dropdown.
+   */
+  function setRecordSelectorToAccompanyDropdown() {
+    const modal = document.querySelector<HTMLElement>('.modal-record-selector');
+    if (dropdownAccompanyingElements && modal) {
+      dropdownAccompanyingElements.add(modal);
+    }
```
pavish, line 71: The modal element that's added here is never removed. While the modal itself is removed from the DOM, it's reference will be held here until the parent AttachableDropdown is removed.

While this should cause no major problems, I'm certain we'd have to do similar operations in other places and it would be better if we ensure that we always remove stale references.

pavish (changes requested):
@seancolsen Looks good! Nice work!

I'll leave reviewing of the UX to @kgodey and @ghislaineguerin, which can probably be done post merge.

I noticed a couple bugs while doing a quick local test:
1. Regression: Array type column styling is broken. Open Data Explorer and summarize any column so that it results in an Array type column. Notice that the css modifications result in this column being messed up. **I'd like this to be fixed before merging the PR in**.
Screenshot 2022-11-10 at 11 59 15 PM
2. Selection is not reset after opening & closing record selector. Click on a column that opens Record selector, and use the keyboard to pick a row. Notice that the record selector is closed. Move the mouse around without clicking and notice that the cells are being selected when the mouse is moved, which is not desired. **Feel free to create an issue for this and handle it later if you do not wish to address it in this PR**.

    https://user-images.githubusercontent.com/11850603/201178024-e4aecf05-5acf-4b19-a79a-0adced15f43a.mov

mathesar_ui/src/components/cell-fabric/data-types/boolean.ts
pavish: >  I think that would look pretty weird.

I agree this would probably look weird in the context of record selector search. But we use the same component for the Cells in the table, and we represent nulls using the Null component in the Cell. The empty value looks odd in that context.

Either way, this is not a priority at the moment, and we can figure this out later.

pavish:
@seancolsen I noticed another bug:

https://user-images.githubusercontent.com/11850603/201179818-86b4c742-4402-42d1-aec4-d29ae9257998.mov

* Open the Record selector
* Use keyboard to navigate to a column which inturn opens up another record selector modal
* Pick a value by using the arrow keys and enter key, notice that it gets selected as expected, and the second modal closes
* Notice that the first modal now has filtered values based on the selection
* Use the arrow keys to highlight the row you'd like to pick and use enter key to select it
* Expect the row to get selected and the modal to close, instead observe that the second modal opens up again

I'll leave it upto you to decide if you'd like to fix this in this PR, or separately (in which case, please create an issue), or maybe later during the post-release improvements.

seancolsen:

- `1. Array type column styling` - This should be fixed in e4a159d45c6d9725da4017fbfd07b93b48ae74ab.

- `2. Selection is not reset` - I opened #1917 for this.

- `Bug with Enter key` - I opened #1918 for this.

mathesar_ui/src/components/cell-fabric/data-types/boolean.ts
seancolsen: I just tested a boolean column. The dropdown shows an empty option. When you select the empty option the cell value changes to `NULL` and displays stylized. I think that should suffice. It would be nice to hide the empty option of the column doesn't allow NULL, but I don't imagine that's a priority.

Though it's worth noting that I did uncover #1919 while checking this out, and I predict that's a regression from this PR too. But I think it's worth merging this in and fixing that subsequently.

mathesar_ui/src/components/cell-fabric/data-types/components/linked-record/LinkedRecordInput.svelte
seancolsen: > it would be better if we ensure that we always remove stale references.

Fair enough. I added 4f017206804fe84e816821e685ac00dbca78d277 to address this.

pavish (approved):

PR #1864, opened 2022-10-26
Cursor closed error hacky fix

### LINKED ISSUE #1549: Error "cursor already closed"
I've been getting this error from the various endpoints occasionally:

> InterfaceError at /api/db/v0/tables/19/columns/
>
> (psycopg2.InterfaceError) cursor already closed
> (Background on this error at: http://sqlalche.me/e/14/rvf5)

image

([Full error](https://rawgit.com/seancolsen/f9d5a5d856957272583e9e5af1a7e78c/raw/error-cursor-already-closed.html))

I can't reproduce it but it seems to happen when I'm submitting multiple API requests in parallel.

mathemancer:
Fixes #1549 

This adds logic to simply try again when the error in #1549 occurs.

**Technical details**

This is something of a hack. I added a middleware class that catches the specific exception, then sleeps for a second and tries again.
- Apparently, the underlying problem has to do with the way connection pooling works in SQLAlchemy.
- I don't think it's properly solvable without writing a custom connection pooler class or removing SQLAlchemy.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja (approved):
I approve of the fix proposed in the PR as I don't think it is worth spending time with SqlAlchemy 
 related issues but I feel like the underlying problem is happening because 
1. Large number of requests are sent to the database for a single call to Data Explorer API(as it does not reuse Metadata)
2. When we spam the Data Explorer API, multiple threads start using the same `connection`  and when the connection gets closed in a particular thread, other thread that happens to use the same connection throws up this error.

If the above theory is right I think the actual problem would be solved by #1851 as it reduces the number of database calls done by the Data Explorer API

mathemancer:
> If the above theory is right I think the actual problem would be solved by #1851 as it reduces the number of database calls done by the Data Explorer API

I think your theory is very close to correct, but I disagree that we can solve it for good by reducing the calls per call (especially if we have multiple users submitting requests to the same webservice). I do agree that will result in the problem only occurring under even heavier loads. I suspect it has to do more with requesting connections in quick succession on different threads more than sheer volume of connections requested (though those are obviously correlated).

seancolsen:
> If the above theory is right I think the actual problem would be solved by https://github.com/centerofci/mathesar/pull/1851

I'd also like to point out that I have been observing the "Cursor already closed" error _throughout the app__ (not just within the Data-Explorer-based reproduction steps listed in the ticket). As such, any fix that focuses on Data Explorer functionality alone (as it seems 1851 does, just from reading the PR title) would seem too narrow to me.

I've not yet t tried using the app with the code in this PR, but I'll pull master today and raise a new ticket if I spot the "Cursor already closed" error again.

PR #1865, opened 2022-10-26
Fix array aggregation for some custom types

### LINKED ISSUE #1798: Fix `array_agg` on money columns
`array_agg` on a money column looks pretty weird

image

dmos62:
Fixes #1798

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/types/custom/underlying_type.py
```diff
@@ -0,0 +1,29 @@
...
+    cases (e.g. array aggregation), and so, in those cases, we cast them to what we consider the
+    underlying type. Can also be thought of as the fallback type. For example, for URIs we fallback
+    to TEXT, and for Money we fallback to NUMERIC.
+
+    Note, doesn't use ABC, because its metaclasses clash with SQLAlchemy's metaclasses. Workaround
+    possible, but seemed burdensome.
+    """
+
+    @property
+    def underlying_type(self):
```
mathemancer, line 17: This property will behave differently based on whether the inheriting class is instantiated or not (if it doesn't define its own `underlying_type`); this is undesirable for something that's overridden as a class attribute in inheriting classes when (properly) used. E.g.:
```python
class MyType(UserDefinedType, HasUnderlyingType):
    # Forgot underlying_type
    
    def get_col_spec(self, **_):
        # This results in the type name being upper case when viewed.
        # Actual usage in the DB is case-insensitive.
        return DB_TYPE.upper()


MyType().underlying_type  # NoneType
MyType.underlying_type  # reference to property function
```

mathemancer (changes requested):
This PR is a perfect example of what I mean by bureaucracy.

In order to properly create a custom type, you have to do two things:
1. Add the underlying_type property, and also
2. Inherit the `HasUnderlyingType` mixin.

Only (1) is actually useful for functionality (except the `cast_to_underlying_type` method; more on that below). However, if you forget (2), then any part of the codebase that assumes you've inherited that won't work even though the actual attribute is there.

Then, to properly _use_ a custom type in situations where the underlying type matters, you have to:

0. Read the custom type class; somehow know that you have to read the mixin definition and do that too.
1. Implement logic to check whether the type is flagged with the mixin,
2. Know that the mixin provides a `cast_to_underlying_type` method (not visible from the custom type class itself, though `underlying_type` is), and then
3. Implement logic to branch based on (1) and use the method mentioned in (2).

This is instead of my suggested `getattr` solution which simply gets the attribute you want if it exists. I suppose a fair comparison would be:

0. Read the custom type class that gives a problem; notice the `underlying_type` attribute.
1. Get that attribute if it exists where needed.

Also, see my specific comments.

dmos62:
I agree that defining an attribute on a mixin and then overriding it on a subclass is somewhat obscure, but so is not formalizing the whole "cast to underlying type" thing. The former is preferable to me.

I attempted to write a response to individual points you make, but, to be honest, I think that part of your criticism is formulated in a haphazard way that is hard to respond to.

I'll instead illustrate how I see the cognitive steps required in understanding our code.

--------

Mixin code:

```
if isinstance(column_expr_type, HasUnderlyingType):
    column_expr = column_expr_type.cast_to_underlying_type(column_expr)
```

Steps:

1. Notice and check what `HasUnderlyingType` is
  - Read its definition and documentation; understand why it would be used, and how;
2. Understand that this branch signifies that this is a custom type that has this trait;
3. Recall from step 1 why we would use `cast_to_underlying_type`;
4. Understand that the branch updates `column_expr`.

---------

The plain attribute code:

```
if hasattr(column_expr_type, 'underlying_type'):
    underlying_type = getattr(column_expr_type, 'underlying_type')
    column_expr = cast(column_expr, underlying_type)
```

1. Notice and check what `underlying_type` is:
    - But, you don't know where to look: `column_expr_type` can be any SQLAlchemy type;
      - You'll probably do a grep and find the `underlying_type` attribute in some of our custom type classes;
    - Once you do find where the attribute is defined, where is its docstring?
      - If the custom type classes don't have a central class they subclass (something like `mathesar.types.custom.base.CustomType`), which they currently don't, you don't have a single place to document a new attribute;
    - If someone left a docstring, you're lucky, but the person who left it is upset, because he probably had to leave it in multiple places;
      - If no one left a docstring, you don't understand the purpose of this attribute;
2. Understand that this branch signifies that this is a custom type with `underlying_type` defined;
    - Notice that you don't yet understand why we're branching on it;
3. Try to understand why we're casting to underlying type;
    - You're unlikely to have a docstring or comment for this, because there isn't a central place where this piece of logic is defined;
4. Understand that the branch updates `column_expr`.

---------
    
I'll note that the no-branching version of your proposed code is worse. Namely:

```
column_expr = cast(column_expr, getattr(column_expr.type, 'underlying_type', column_expr.type))
```

Here the reader, in addition to steps outlined above, has to understand that casting to itself is a free no-op, otherwise he'll think this is wasteful. I also find the nesting excessive: I wish this was more vertically intense and horizontally relaxed.

---------

You'll notice that the differences I focus on are in where we define logic (contained vs distributed) and in how we are able to document it.

The benefits I outline apply to implementing a new custom class too, but, I'm almost hesitant to mention it, because, for this code, being able to read it easily is overwhelmingly more important. If we wanted to go the extra mile to make this easy to use for people implementing custom types (we probably don't), you'd use a `CustomType` class. Then, that central class would replace this mixin, maybe.

dmos62:
We talked this over in sync and I implemented some of the things we talked about. Not all: I stayed with a mixin. I considered a lot of options, like creating that `UserDefinedDomain` class and putting all the `underlying_type` stuff in there, but decided that this is just so similar that it's not worth the bother. Moving this to a central subclass is trivial in the future, if we ever want to.

db/types/custom/underlying_type.py
dmos62: Note, I played around with this a bit, and it seems that creating an abstract property (like I did) is the only way to enforce that a concrete attribute is implemented on a concrete subclass. Using `ABC` doesn't change that: it only helps enforce it. Maybe I missed something, but if I'm right the scenario you posted cannot be solved if you want an abstract class or instance property.

mathemancer (approved):
Alright, seems to work.

Issue #1434, opened 2022-06-14
Add support for specifying the foreign key column name created when splitting the table


silentninja:
## Problem
When splitting a table, columns are extracted from a table and moved into a new table and a foreign key that connects the table gets created. The foreign key column is created with a default column name, the user should be able to specify the name of the column.

## Proposed solution
Add an optional request parameter called foreign_key_column_name or similar which allows the requester to specify the name of the new FK column that will be created.

PR #1878, opened 2022-10-29
Optionally accept foreign key constraint to use when moving column 

### LINKED ISSUE #1491: Let the user choose the foreign key based on which to move the column to a different table
## Problem
 In cases where there are multiple foreign keys connecting the two tables, we need to be clear about which foreign key will be used when a column is moved between the tables.

## Proposed solution
The client should be able to choose the foreign key when moving a column to a different table

silentninja:
Fixes #1491

Allows the `/tables/<id>/move_columns/` API to optionally accept a parameter `relation_fk_constraint` whose value is the `id` of foreign key constraint which links the table that contains the columns and the table to which the columns have to be moved 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/tables/operations/move_columns.py
+++ b/db/tables/operations/move_columns.py
@@ -6,6 +6,7 @@ from db.columns.base import MathesarColumn
 from db.columns.operations.alter import batch_alter_table_drop_columns
 from db.columns.operations.create import bulk_create_mathesar_column
 from db.columns.operations.select import get_column_names_from_attnums
+from db.constraints.operations.select import get_constraint_from_oid
 from db.tables.operations.select import reflect_table_from_oid
 from db.metadata import get_empty_metadata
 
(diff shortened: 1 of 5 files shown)


mathemancer (changes requested):
Do we have a use-case for this at the moment? If not, I'd rather wait to do this work entirely:
- It further bakes in an assumption about Foreign keys being the way we specify joins that may not be valid by the time this PR is relevant
- It complicates the machinery, and I don't think that cost is acceptable without an actual use-case.

Am I missing something? I looked at the relevant issue and some surrounding context, but didn't see why this is relevant now.

As far as the code goes, I think it's fine, but I don't like the assumptions it bakes into the codebase w.r.t. foreign keys being the only way to specify the join.

silentninja:
> It further bakes in an assumption about Foreign keys being the way we specify joins that may not be valid by the time this PR is relevant

Can you elaborate? We are using foreign keys to join the table when moving the columns right now.


> It complicates the machinery, and I don't think that cost is acceptable without an actual use-case.

This feature is helpful 
- When there is a leftover column that should be normalized and a new foreign  column is introduced to the normalized table
- when there is a table that is created manually and has two foreign key columns that are pointing to the same table and needs a column that has to be moved

For example:
Items
 - Id: Int
 - Last Name: String
 - Reviewer: Int Foreign key to Author Table
 - Author: Int Foreign key to Author Table

By default the `Reviewer` column would be used by the actual column I would want to use would be the `Author` column

silentninja:
There is confusion about the priority of the issue this PR fixes. There were two issues related to Data Modeling tagged under the milestone `Live Demo` 

- [Add support for specifying the foreign key column name created when splitting the table #1434](https://github.com/centerofci/mathesar/issues/1434) which has the milestone `Live Demo`
- The above issue #1878 had live demo as its milestone(but the changes history in the issue was stating a change from Live demo to Post release) before I changed it to Post release.

seancolsen (commented):
I have not tested this, but here are some initial thoughts:

- I agree with @mathemancer that this is not a priority right now. We'll need to do front end work before this feature can offer any benefit to users, and I have a _lot_ of other front end work that feels higher-priority than this.
- When we _do_ implement this feature, I'd rather implement it as a breaking change, removing the `target_table` parameter from the API and using the `relation_fk_constraint` parameter instead. That feels like a more intuitive API to me.

In terms of project management, I'm not sure where that leaves us. Maybe @silentninja keeps this branch locally and resurrects it in a few months hoping that it doesn't accumulate too many merge conflicts?

kgodey:
> In terms of project management, I'm not sure where that leaves us. Maybe @silentninja keeps this branch locally and resurrects it in a few months hoping that it doesn't accumulate too many merge conflicts?

I think we should close the PR, we can always use it as a reference later when the related issue is deemed a priority. I'm not sure keeping the branch locally will have much value since the PR is pretty small, but that's up to @silentninja.

PR #1879, opened 2022-10-29
Allow partial summarizations in the API

dmos62:
Implements #1880, with the exception of a few edge cases. Builds on #1900 (see its description for overview of the mentioned edge cases).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
@mathemancer @pavish This PR is the backend infrastructure for the frontend to be able to submit a query with a not-fully-specified transform. More specifically, it's a way for the frontend to request a suggested grouping transformation. Frontend will submit a query (via `query/run` endpoint) with a special transformation that will effectively be the request for the suggested grouping, and the backend will return a query that has the "result" suggested transformation in place of the "request" transformation.

We decided to allow the frontend to request for a suggested transform via a regular `run` request, because that seemed easier for the frontend and because that removes one HTTP roundtrip, compared to a separate endpoint for suggesting a grouping (e.g. `query/suggest_grouping`).

I am now doubting this decision. This workflow is complicated and the necessary infrastructure on the backend seems to add more complexity than user benefit.

To be clear, the alternative is an endpoint like `query/suggest_grouping`, whose input is something like:

```json
{
  "base_table": x,
  "initial_columns": y,
  "user_selected_column": "initial_column_alias1",
}
```

Where `x` and `y` correspond to respective attributes in a query specification. The output would be a set of initial columns to group by.

I briefly described the negatives in our current approach from the perspective of the backend, but I don't have a clear picture of the pros and cons for the frontend.

@mathemancer Do you think that my implementation of infrastructure is optimal enough, given the requirements? Maybe I'm just doing it wrong.

@pavish Could you elaborate on what the downsides of using the alternative `query/suggest_grouping` approach would be for the frontend? If they are minor, I'd like to discard this PR and take the alternative approach. I basically want to make sure that adding this messy workflow is warranted.

I'll underline again that the alternative `query/suggest_grouping` approach is much, much easier to implement on the backend.

mathesar/models/query.py
```diff
@@ -1,13 +1,15 @@
 from functools import wraps
 from django.db import models
-from django.utils.functional import cached_property
+from mathesar.state.cached_property import cached_property
```
dmos62, line 3: This is an unrelated bug fix.

mathemancer (commented):
@dmos62 I think this structure makes sense _if we're planning to keep this flow long-term_ (i.e., the API accepts partially-specced queries, runs an inferred query, then returns results along with the fully-specced inferred query).

If we're planning to switch it out for the "ask for suggestion", then "run suggestion separately" flow, then I'd rather keep the parts knowing anything about processing or partially-specced queries out of the model entirely. The reason is that this will make the logic easier to separate into a different endpoint later.

All that said, as I'm looking through this, and noting one of your doc strings mentioning that processing can depend on other parts of the query, I'm starting to think the one-shot flow this PR implies is actually a good long-term solution, with the caveat that the processing is not enabled by default for performance and idempotency reasons (as you noted somewhere in a docstring). 

Reasons to keep long-term:
- Queries are transient. We're not saving something we've inferred without further intervention.
- Queries don't modify data. While this _can_ be a performance drain, this is mitigated if we only run processing when specified.
- This gives the backend the global view needed to  infer the desired query if possible (this is the most important point to me).

This assessment comes with a major caveat. I do think it should be the front end's responsibility to submit some parameter if they want the processing run, and this needs to be per-transform. The reason is that it's conceivable there would be a transform that's fully specified, but processing / inference would modify it (what if they _want_ to group by just one single column?).

dmos62:
> This assessment comes with a major caveat. I do think it should be the front end's responsibility to submit some parameter if they want the processing run, and this needs to be per-transform. The reason is that it's conceivable there would be a transform that's fully specified, but processing / inference would modify it (what if they want to group by just one single column?).

Actually, this already does that in a different way. Some transform types will be special "unprocessed transforms". They will trigger us processing them into a normal transform when the query is submitted via `query/run`, and they will cause an exception if they're in a persisted query and the query is being run.

In other words, when the frontend wants a grouping transform to be a default grouping, it'll use a transform like `unprocessed_default_group`, and when they just want to group, they'll just use `group`.

pavish:
@mathemancer

> I do think it should be the front end's responsibility to submit some parameter if they want the processing run, and this needs to be per-transform. 

I expect it to work as mentioned here. The frontend will submit additional paramter to do this.

@dmos62 

> Actually, this already does that in a different way. Some transform types will be special "unprocessed transforms". They will trigger us processing them into a normal transform when the query is submitted via query/run, and they will cause an exception if they're in a persisted query and the query is being run.

This seems to complicate things.
1. What should the frontend do to save the query once this is done i.e. when `processed_default_group` is returned and the user wants to save the query?
2. The entire structure is too different, and we'll have to render the same UI for two different request structures. Why is the internal backend model tied too much with the request/response structure? Can we not just add one additional parameter to group (since that's the only transform which will need guessing)?

dmos62:
@pavish as I see it, your points 1 though 3 are general inconveniences about the alternative being a two round-trip operation, as opposed to a one round-trip operation. What's the import flow problem you're referencing in 4?

> We cannot get rid of things being messy

We could if we were willing to stomach a two round-trip operation. I realise that there's a cost in UX, but it's limited. Are we making the right compromises?

> What should the frontend do to save the query once this is done i.e. when processed_default_group is returned and the user wants to save the query?

You can save it. It's `unprocessed` transforms that are forbidden on a saved query. They can be allowed on saved queries, but it's just that the pipeline required to support transform processing is much more complicated, so we'd prefer to isolate it to the `query/run` endpoint.

> the same UI for two different request structures

According to my suggestion you'll have three transforms for grouping: `regular`, `request for default`, and `default`. If the UX for `regular` and `default` is identical for the foreseeable future, we can reduce that to `regular` and `request for default`. I.e. we would process a `request for default` into a `regular` grouping transform, as opposed to a `default` grouping transform.

> The frontend will submit additional paramter to do this

This PR is based around the idea of doing control flow through transform types. Does that work for you?

> Why is the internal backend model tied too much with the request/response structure?

I imagined it would be helpful to the frontend (and the user) to know that a given transform is a default (partially specified by backend) as opposed to a user-specified transform. A request/response structure gives you that, without complicating the structure of already existing transforms.

pavish:
@dmos62 

I am pretty strong that this has to be handled at the backend. As the risk of sounding argumentative, here's my response:

> as I see it, your points 1 though 3 are general inconveniences about the alternative being a two round-trip operation, as opposed to a one round-trip operation
> We could if we were willing to stomach a two round-trip operation. I realise that there's a cost in UX, but it's limited. Are we making the right compromises?

Development wise, they are definitely not general inconveniences. It comes with high development cost with regards to state maintenance (it will take me atleast 3 to 4 days), associated with messy code on the frontend (and definitely more lines of changes compared to the 105ish diff in this PR), along with UX concerns. Considering that in terms of backend, we'd only have to deal with messy code, the benefit for the product as a whole lies in making it easier for the frontend.

Also, this is a _legitimate_ usecase. 'Perform an operation and guess the defaults', doesn't seem so out-of-the-ordinary to expect from an API. Our application is frontend heavy and the APIs are specifically for the frontend, I don't see a valid reason to move part of this logic to the frontend.

> This PR is based around the idea of doing control flow through transform types. Does that work for you?

I can make it work but I find that we seem to be over-engineering this. I am okay with a transform being complicated rather than the entire flow. From an API perspective, I submit a query object and the result transform type is different from what I sent, this is harder to maintain and understand for any API user rather than just the transform object changing internally.

> I imagined it would be helpful to the frontend (and the user) to know that a given transform is a default (partially specified by backend) as opposed to a user-specified transform. 

The request is sent from the frontend, and the frontend is explicitly requesting this operation and expecting the result. I don't think this solves any specific problem.

dmos62:
I could use some more details from you @pavish. I'll try and schedule a sync with you. I'll touch on a few of the things I'd like to talk about during sync.

> it will take me atleast 3 to 4 days [to implement the 2 round-trip approach]

That's not worth it then.

> the benefit for the product as a whole lies in making it easier for the frontend

I might be misunderstanding you here or taking it out of context, but it sounds like you're advocating for making it easier for the frontend at the expense of the backend; I think we should be looking at win-win solutions, since the product is only as good as its weakest part. Again, I may have misunderstood. Let's continue this during sync.

> this is a *legitimate* usecase

We're getting pretty meta here, but I think that talking about legitimacy of expectations is not productive. The relevant considerations I see are, in no particular order, technical debt on backend, technical debt on frontend and user experience.

> I submit a query object and the result transform type is different from what I sent, this is harder to maintain and understand for any API user rather than just the transform object changing internally

Interesting, my intuition is the opposite. More specifically, I find multiple data structures easier to reason about than a single "dynamic" data structure. We can do either, depending on what works better for you. Would be interesting to hear more about what you expect in situations like this (during sync).

dmos62:
Major changes decided on here https://github.com/centerofci/mathesar/issues/1880#issuecomment-1301052785. Will refactor or scuttle this PR.

dmos62:
Refactored to satisfy current requirements.

mathemancer (changes requested):
As discussed in our sync, I think the concept of `PossiblyPartialTransform` is unneeded and complicating rather than clarifying. There's no point to having this be a transform level concept. The concept should be at a higher level: A DBQuery has some transforms as well as some other pieces and parameters. Based on those parameters, it might need to modify some transforms, add some transforms, drop some transforms, etc. The transform itself doesn't need to include any concept of "partial".

dmos62:
@mathemancer that's an interesting point. Can we merge this to get closer to unblocking Pavish, and then I'll revisit this in PRs that build on this?

db/queries/operations/process.py
```diff
@@ -0,0 +1,24 @@
...
+        in enumerate(db_query.transformations)
+    )
+
+
+def _get_processed_transformation(db_query, ix, db_transformation):
+    if isinstance(db_transformation, PossiblyPartialTransform):
+        db_transformation = db_transformation.get_processed(
+            db_query, ix_in_transform_pipeline=ix
+        )
+    return db_transformation
```
mathemancer, line 24: This method bakes in a requirement for every transform that we'd want to process, along with an assumption about the needed input for processing. 
- We may want to process the same transform in multiple ways. 
- We may need different input for different transform types' processing, or for different ways to process the same transform. Etc.


This should just call something like the following method of the `DBQuery` class:
```python
    def finish_summarize_transforms(self):
        for transform, ix in enumerate(db_query.transformations):
            if isinstance(transform, Summarize):
                transform = finish_specifying_summarize_transform(self, ix, self.engine, self.metadata)
                break
        return transform
```

where the `finish_specifying_summarize_transform` is imported from the code in your other PR.

mathemancer:
> @mathemancer that's an interesting point. Can we merge this to get closer to unblocking Pavish, and then I'll revisit this in PRs that build on this?

I don't think that's a good idea. Building on this setup would be a mistake. I've put a suggestion in my newest review that simplifies the whole thing, and avoids some of the assumptions baked into the class you introduce here.

mathemancer (changes requested):
I added some more detail in specific comments in this review.

dmos62:
I don't think that building on this is a mistake, in that the frontend doesn't care in the short term. I presume you said that it's a mistake, because you're trying to avoid technical debt in the long term. My goal right now is to unblock frontend as quickly as I can. I am willing to take on this technical debt, and resolve it in my upcoming PRs. By upcoming PRs, I mean the next 2 (minimum) PRs: first one that will connect the infrastructure in my currently open PRs, and a second one that will expand the default summarization algo to be more general. The first one will unblock frontend, so I'll have more time to shuffle architecture about in the second one. Please reconsider merging this as soon as possible.

Btw, I fixed the failing test.

db/queries/operations/process.py
dmos62: Your proposal makes sense, especially since we don't have other partial transforms planned. Note that it's a little more work to refactor than just remove the `PossiblyPartialTransform` class: relevant tests are currently based around it.

dmos62:
@mathemancer got rid of most of the generalizations; tell me if that looks good.

@pavish the API is as we discussed: `summarize` transforms now only insist on the `base_grouping_column` field being defined, everything else is optional. When an alias is added to the `grouping_expressions` or the `aggregation_expressions` sequences by the backend, the output alias is the input alias with a suffix. The suffix is either `_grouped` or `_agged`.

db/queries/operations/process.py
```diff
@@ -0,0 +1,26 @@
...
+from db.transforms.operations.finish_specifying import finish_specifying_summarize_transform
+
+
+def get_transforms_with_summarizes_speced(db_query, engine, metadata):
+    """
+    Processes db_query's transformations and returns them. Resulting sequence is the db_query's
+    transform sequence, but each possibly partial transform is replaced with a transform that's the
+    result of processing it.
+
+    See PossiblyPartialTransform for more information.
```
mathemancer, line 11: This is out of date now.

mathemancer (approved):
I like it! Thanks for taking the time with all the back-and-forth.

I did find one out-of-date docstring.  Feel free to merge after fixing that.

PR #1898, opened 2022-11-04
Add some information to query column metadata

### LINKED ISSUE #1883: Additional information needed in column_metadata in Query run endpoint
We need the following additional information in the `column_metadata` property in Query run response. This is needed to show the information on the inspector panel based on the [new figma designs](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=1384%3A6672):
* Table name
* Base column name
* For aggregated columns, it should contain the respective input column alias

Previously, I was able to show the table name and base column name using the information from joinable_tables endpoint, but since we now have the inspector additionally within the Explorations page (where the joinable_tables call is unnecessary), I would like this information to come from the column_metadata.

cc: @dmos62 @mathemancer

dmos62:
Fixes #1883

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
These test case diffs should summarize changes to the API: https://github.com/centerofci/mathesar/pull/1898/files#diff-76c26665cde4f2b8a9df90c817174cccbc61ddac56388f41334fd3dd4a2c0ac1R97

Notice the difference in fields when `is_input_column` is true and false.

mathesar/tests/api/query/test_library_demo_reports.py
```diff
@@ -93,13 +93,19 @@ def check_overdue_books_columns(create_overdue_books_query, client):
...
         }, {
             'alias': 'email',
             'display_name': 'Patron Email',
             'type': 'mathesar_types.email',
             'type_options': None,
-            'display_options': None
+            'display_options': None,
+            'is_initial_column': True,
+            'base_table_name': 'Patrons',
+            'base_column_name': 'Email',
```
pavish, line 108: @dmos62 Can we change the names here to `input_table_name` and `input_column_name` so we don't confuse this terminology with the `base_table` we already have?

pavish (changes requested):

mathesar/models/query.py
```diff
@@ -194,13 +195,50 @@ def initial_columns_described(self):
...
+                    'summarization_input_alias': summarization_input_alias,
+                }
+        return output
+
+    def _get_summarization_input_alias_for_output_alias(self, output_alias):
+        db_transforms = self._db_transformations
+        if db_transforms:
+            for db_transform in db_transforms:
+                if isinstance(db_transform, SummarizeTransform):
+                    return db_transform.get_input_alias_for_output_alias(output_alias)
```
mathemancer, line 236: It would be better for the called function to simply return the appropriate input alias for any output alias. This is simpler, and easier to maintain long-term. The model should hot be branching based on type of transform here.

mathesar/models/query.py
```diff
@@ -194,13 +195,50 @@ def initial_columns_described(self):
...
-            'display_options': self._get_display_options_for_sa_col(sa_col),
+            'display_options': self._get_display_options_for_alias(alias),
+            'is_initial_column': is_initial_column,
         }
+        if is_initial_column:
+            initial_dj_column = _get_dj_column_for_initial_db_column(initial_db_column)
+            output = output | {
+                'base_column_name': initial_dj_column.name,
+                'base_table_name': initial_dj_column.table.name,
+            }
```
mathemancer, line 218: It would be better for these to exist, but be null if the column is not an initial column. This would be more extendable in the long run, and also more consistent at the API.

Otherwise, an API consumer can't tell the spec of what's returned without testing all possible inputs.

mathesar/models/query.py
```diff
@@ -194,13 +195,50 @@ def initial_columns_described(self):
...
+        else:
+            # NOTE currently only summarizations can "create" aliases; once that changes, this
+            # section should become more general.
+            summarization_input_alias = self._get_summarization_input_alias_for_output_alias(alias)
+            is_summarization_output_column = summarization_input_alias is not None
+            if is_summarization_output_column:
+                output = output | {
+                    'is_summarization_output_column': is_summarization_output_column,
+                    'summarization_input_alias': summarization_input_alias,
+                }
```
mathemancer, line 228: Same here. I'd prefer a logic that doesn't require the model to know anything about which columns have different input vs. output alias. It's also not great if the API consumer needs to handle these different schemas.

For a column which doesn't change, the input alias is the output alias.

mathemancer (changes requested):
There's a fair amount of logic in the model that should be moved down into the `DBQuery` class. I'm thinking specifically of the logic around whether you have a different input vs. output alias.

dmos62:
@mathemancer good insight; I'll refactor.

dmos62:
@pavish here's the new interface https://github.com/centerofci/mathesar/pull/1898/files#diff-76c26665cde4f2b8a9df90c817174cccbc61ddac56388f41334fd3dd4a2c0ac1R97

Applied Brent's suggestion to have each field always appear.

db/queries/base.py
```diff
@@ -176,4 +190,7 @@ def __init__(
 
     @property
     def is_base_column(self):
+        """
+        A base column is a column on a query's base table.
```
mathemancer, line 194: Using our terminology so far, it's more properly an _initial_ column (of the query) which is also in the base table. A non-query base table column shouldn't be included here.

mathemancer (approved):
Overall, looks good from a tech perspective. 

Regarding
> @mathemancer @pavish notice that a transform class by default defines that its output-input alias mapping is empty (meaning that output and input aliases are identical), and only the summarize mapping is non-empty. I think that that's technically not correct, since we allow all transforms to map their output to a different alias, but it seems that frontend only does it in the case of summarization. Does this cut corner make sense atm?

I'm less worried about the corner cut and more about the fact that then the front end needs to handle the nulls differently (i.e., it may have a need for a name for a column, then it's going to have to include logic for looking that up.) 

It's possible I'm missing something there.

I also noted a minor quibble about a docstring. Feel free to reword that, or not.

I'm approving regardless, since I consider my main problem (the null alias info) to be something more appropriate for @pavish to pass final judgement on than me.

pavish:
@mathemancer @dmos62 

> notice that a transform class by default defines that its output-input alias mapping is empty (meaning that output and input aliases are identical), and only the summarize mapping is non-empty. I think that that's technically not correct, since we allow all transforms to map their output to a different alias, but it seems that frontend only does it in the case of summarization. Does this cut corner make sense atm?

Only summarization currently changes the properties of the result column (eg., different type, separate display_name etc.,), and all other transformations produce a result column which has the same properties as the initial column. So I don't think this would be an issue.

> I consider my main problem (the null alias info) to be something more appropriate for @pavish to pass final judgement on than me.

The frontend is only going to read the `input_table_name` and `input_column_name` for `is_initial_column: True`. As long as this scenario is never null, there isn't any issue. For `is_initial_column: False`, the frontend will read `input_alias` and then get the values for that alias. 

I give my approval, if the PR satisfies what the above answers expect.

pavish (approved):

db/queries/base.py
dmos62: Good catch.

dmos62:
@pavish 

> The frontend is only going to read the input_table_name and input_column_name for is_initial_column: True. As long as this scenario is never null, there isn't any issue. For is_initial_column: False, the frontend will read input_alias and then get the values for that alias.
>
> I give my approval, if the PR satisfies what the above answers expect.

It does.

PR #1899, opened 2022-11-04
Fix alter_column() error behaviour

### LINKED ISSUE #1814: Show better error messages when data type conversion fails or when field size limit is too short for existing data
We're showing Python errors in the toast message.

Table_112_public_Mathesar

Screenshot 2022-10-12 at 13-20-09 Table 112 public Mathesar

Anish9901:
Fixes #1814 

**Technical details**

This PR handles the following errors while invalid type/type_options are passed in the request.
- Invalid type conversions.
- Field Size limit too small for existing data in the column.
- Field Size limit too large.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/exceptions.py
+++ b/db/columns/exceptions.py
@@ -40,3 +40,7 @@ class ExclusionError(Exception):
 
 class ColumnMappingsNotFound(Exception):
     pass
+
+
+class InvalidStringTruncation(Exception):
+    pass
--- a/db/columns/operations/alter.py
+++ b/db/columns/operations/alter.py
@@ -1,11 +1,11 @@
 from alembic.migration import MigrationContext
 from alembic.operations import Operations
 from sqlalchemy import DefaultClause, text, DDL, select
-from sqlalchemy.exc import DataError, InternalError
-from psycopg2.errors import InvalidTextRepresentation, InvalidParameterValue
+from sqlalchemy.exc import DataError, InternalError, ProgrammingError
+from psycopg2.errors import InvalidTextRepresentation, InvalidParameterValue, StringDataRightTruncation, RaiseException, SyntaxError
 
 from db.columns.defaults import NAME, NULLABLE
-from db.columns.exceptions import InvalidDefaultError, InvalidTypeError, InvalidTypeOptionError
+from db.columns.exceptions import InvalidDefaultError, InvalidTypeError, InvalidTypeOptionError, InvalidStringTruncation
 from db.columns.operations.select import (
     get_column_attnum_from_name, get_column_default, get_column_name_from_attnum,
 )
@@ -103,12 +103,22 @@ def retype_column(
     except DataError as e:
         if type(e.orig) == InvalidParameterValue:
             raise InvalidTypeOptionError
-        if type(e.orig) == InvalidTextRepresentation:
+        elif type(e.orig) == InvalidTextRepresentation:
+            raise InvalidTypeError
+        elif type(e.orig) == StringDataRightTruncation:
+            raise InvalidStringTruncation
+        else:
+            raise e.orig
+    except InternalError as e:
+        if type(e.orig) == RaiseException:
             raise InvalidTypeError
         else:
-            raise e
-    except InternalError as e:
-        raise e.orig
+            raise e.orig
+    except ProgrammingError as e:
+        if type(e.orig) == SyntaxError:
+            raise InvalidTypeOptionError
+        else:
+            raise e.orig
 
 
 def alter_column_type(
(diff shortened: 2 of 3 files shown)


db/columns/operations/alter.py
```diff
@@ -103,12 +103,22 @@ def retype_column(
...
-        raise e.orig
+        if type(e.orig) == RaiseException:
+            raise InvalidTypeError(column_name, new_type)
+        else:
+            raise e.orig
+    except ProgrammingError as e:
+        if type(e.orig) == SyntaxError:
+            raise InvalidTypeOptionError
+        else:
+            raise e.orig
```
mathemancer, line 121: I'm suspicious of this.
- How is this being produced?
- Is `SyntaxError` really precise enough?

db/columns/exceptions.py
```diff
@@ -40,3 +42,7 @@ class ExclusionError(Exception):
 
 class ColumnMappingsNotFound(Exception):
     pass
+
+
+class InvalidStringTruncation(Exception):
```
mathemancer, line 47: Is the custom exception necessary? Wouldn't it be better to just handle the underlying exception at the higher level?

db/columns/operations/alter.py
```diff
@@ -103,12 +103,22 @@ def retype_column(
     except DataError as e:
         if type(e.orig) == InvalidParameterValue:
             raise InvalidTypeOptionError
-        if type(e.orig) == InvalidTextRepresentation:
-            raise InvalidTypeError
+        elif type(e.orig) == InvalidTextRepresentation:
+            raise InvalidTypeError(column_name, new_type)
+        elif type(e.orig) == StringDataRightTruncation:
```
mathemancer, line 108: I think this is specific and understandable enough to just pass along rather than using a custom exception.

mathesar/api/db/viewsets/columns.py
```diff
@@ -92,7 +92,9 @@ def create(self, request, table_pk=None):
             except InvalidTypeError as e:
                 raise database_api_exceptions.InvalidTypeCastAPIException(
                     e,
-                    message='This type casting is invalid.',
+                    message=f'{e.column_name} cannot be casted to {e.new_type}.'
```
mathemancer, line 95:
```suggestion
                    message=f'{e.column_name} cannot be cast to {e.new_type}.'
```

This is the proper conjugation.

mathesar/api/db/viewsets/columns.py
```diff
@@ -182,6 +186,12 @@ def partial_update(self, request, pk=None, table_pk=None):
                     )
                 else:
                     raise base_api_exceptions.MathesarAPIException(e)
+            except InvalidStringTruncation as e:
+                raise database_api_exceptions.InvalidTypeOptionAPIException(
+                    e,
+                    message='Invalid string truncation length.',
```
mathemancer, line 192: It would be good to improve this message. Am I correct that the underlying error there is an attempt to cast some column of text-like values to a `VARCHAR(n)` or `CHAR(n)` where the length is too short? If so, that would be a more helpful message. (I.e., "the requested string length is too short" or the like)

mathesar/api/db/viewsets/columns.py
mathemancer, line 95: Also, it would be much clearer to construct the message outside of the call. You could either
- add a method to the exception to construct the message (since it's always the same), or
- put a method somewhere else convenient.

mathemancer (changes requested):
Grammar fix:
```
s/casted/cast/g
```

Also, see my other specific notes.

db/columns/operations/alter.py
Anish9901: `SyntaxError` occurs when we pass something off limits like 99999999 in the Field size limit.

db/columns/exceptions.py
Anish9901: You are right the higher-level DB-API exception is clear enough.

mathesar/api/db/viewsets/columns.py
Anish9901: Ok, I'll refactor that.

mathesar/api/db/viewsets/columns.py
Anish9901: > Am I correct that the underlying error there is an attempt to cast some column of text-like values to a VARCHAR(n) or CHAR(n) where the length is too short?

You are correct, I'll write a better message.

db/columns/operations/alter.py
mathemancer: My concern is whether some other thing can also cause a `SyntaxError`. After looking at the docs a bit (and what's reported by `psycopg2`), it doesn't seem like it's possible to do much better.

mathemancer (approved):
Okay, I think this looks good. Thank you!

Issue #1673, opened 2022-09-16
Users and permissions backend


mathemancer:
## Problem

Mathesar currently doesn't support any kind of authentication to use the API, and has not users/permission control (at the Mathesar level; the underlying DB could).

## Proposed solution

We will add users and permissions as per the spec described in the wiki [here](https://wiki.mathesar.org/en/product/specs/users-permissions) .

## Tasks
- [x] Make APIs require authentication - #1742
- [x] Add log in page - #1742
- [x] Add custom user model - #1751
- [x] Add Users CRUD API - #1753
- [x] Add list of permissions for user to Users API - #1774 
- [x] Allow permissions to be updated via API - #1774
- [x] Implement permissions for all APIs
    - [x] `/api/ui/v0/users/` - #1780 
    - [x] `/api/ui/v0/database_roles/` - #1902 
    - [x] `/api/ui/v0/schema_roles/` - #1916 
    - [x] `/api/ui/v0/databases/` - #2061
    - [x] `/api/db/v0/tables/` - #1962
    - [x] `/api/db/v0/queries/` - #1999 
    - [x] `/api/db/v0/links/` - #1997 
    - [x] `/api/db/v0/schemas/` - #1961 
    - [x] `/api/db/v0/databases/`- #2061
    - ~~[ ] `/api/db/v0/data_files/`~~ Doesn't make sense to add permission to Data files as they are not bound to any permissible object.
    - [x] `/api/db/v0/tables/{id}/records/` - #1998 
    - [x] `/api/db/v0/tables/{id}/settings/` - #1993 
    - [x] `/api/db/v0/tables/{id}/columns/` - #1992 
    - [x] `/api/db/v0/tables/{id}/constraints/` - #1963 
    - [x] Double check and make sure all APIs are covered
- [x] Implement reset password mechanism 
    - [x] Allow users to change their own passwords #2052 
    - [x] Allow admins to reset user passwords #2000 
    - [x] Require users to change password after first login #2071 
- [X] Implement logout view - Built in by Django
- [x] ~~Implement default permissions for `public` schema~~ - Not needed, refer https://github.com/centerofci/mathesar/pull/1916#discussion_r1023970448
- ~~[ ] Add additional user data to be saved to User model & API~~ - will be implemented on an as-needed basis in the future

kgodey:
@silentninja I realized that I forgot to put the type suggestions and previews endpoint in this checklist – do they need permissions?

silentninja:
[...]

PR #1902, opened 2022-11-04
Add Database Role permissions

silentninja:
Related to #1673

This PR introduces the following permissions for Database Roles(which are then used for granting granular permissions to the Database objects)

- Allows a superuser to create a `DatabaseRole` object
- Allows a user who has a `manager` role for a database to create `DatabaseRole` for other users thereby granting them permission for that database
- Allows a super user or a database manager to delete other `DatabaseRoles` 
- Allows a user who has any role related to the database to view other `DatabaseRoles` for that database

**Technical Details**

Note to the reviewer
- When a user sends a `POST` request to create a `DatabaseRole` it is not possible to know if they actually have the write access to that database without inspecting the payload. So the validation is possible only at the serializer level and this leads to throwing a `400` error instead of the `PermissionDenied` status code unless we customize it. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (changes requested):
@silentninja I find this PR confusing. All the logic for managing the database roles should live in the `DatabaseRoleAccessPolicy`. Reading through the current `DatabaseRoleAccessPolicy` doesn't give an accurate picture of what's actually allowed in the database roles API, and I'm not sure what the `DatabaseAccessPolicy` or changes to the user serializer have to do with the changes described in this PR. Additionally, there's not enough tests to evaluate that all the conditions work properly.

silentninja:
> I find this PR confusing
Sorry I didn't realise that some commits got rejected because I pushed them after I updated the branch using Github UI. These commits remove irrelevant comments and add a few additional test cases

> All the logic for managing the database roles should live in the DatabaseRoleAccessPolicy

The logic for creating a `Database Role` cannot be in the `DatabaseRoleAccessPolicy` because permissions for creating a `Database Role` for a Database are related to the `Database`.  In addition to that, there isn't much information for checking the permissions without inspecting the request body
So the question should "Can the user create a Database Role for the given Database"?

> List and retrieve should be separate from create. Create permissions should match destroy permissions.

`Destroy, partial_update, update` acts upon a resource based on the identifier in the URL while `list, create`  don't, they are two different things.

mathesar/api/ui/serializers/users.py
```diff
@@ -56,6 +58,8 @@ class Meta:
         model = DatabaseRole
         fields = ['id', 'user', 'database', 'role']
 
+    database = PermittedPkRelatedField(access_policy=DatabaseAccessPolicy, queryset=Database.current_objects.all())
```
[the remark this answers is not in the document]
silentninja: > I'm still confused about what it does here, though. The user serializer should only be used in the Users API. Why is it relevant to this PR?

The changes are for `DatabaseRoleSerializer` which was already part of the module `user`.

mathesar/tests/api/test_user_api.py
```diff
@@ -283,6 +283,17 @@ def test_database_role_create(client, user_bob):
     assert response_data['database'] == database.id
 
 
+def test_database_role_create_non_superuser(client_bob, user_bob):
```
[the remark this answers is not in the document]
silentninja: The commits that contain the test cases were rejected due to conflict with the upstream. I pushed them now, can you review again and see if any additional test cases should be added.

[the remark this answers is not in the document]
kgodey: Mostly looks good. It would ideal to add a tests for a scenario where user X had a viewer role on database A, manager role on database B, and no role on database C. We can then test that they can only see database roles for A & B (not C), and they can only create roles for B (and not A or C).

I'm not sure about the status of multiple database tests right now, but if it's easy to do, that would be good to add.

mathesar/api/ui/permissions/database_role.py
```diff
@@ -0,0 +1,44 @@
...
+from rest_access_policy import AccessPolicy
+
+from mathesar.models.base import Database
+from mathesar.models.users import DatabaseRole, Role
+
+
+class DatabaseRoleAccessPolicy(AccessPolicy):
+    statements = [
+        # Listing and retrieving a database role is allowed for everyone.
+        # We cannot restrict access for creating a Database Role object here because database for which the role is created can be known only by inspecting the request body
```
kgodey, line 11: Please wrap the comments so the lines are shorter.

kgodey: Also please ensure correct grammar (I see randomly capitalized words and no spaces between words and parentheses, for example)

kgodey (commented):
I think most of my concerns have been addressed. I'd like @mathemancer to take a look before merge since I don't trust my brain to do thorough code review right now. I did add a couple of comments.

mathesar/api/db/permissions/database.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class DatabaseAccessPolicy(AccessPolicy):
+
+    @classmethod
+    def scope_queryset(cls, request, qs):
+        if not request.user.is_superuser:
+            allowed_roles = (Role.MANAGER.value,)
+            if request.method.lower() == 'get':
+                allowed_roles = allowed_roles + (Role.EDITOR.value, Role.VIEWER.value)
```
mathemancer, line 14: Why would permissions be different for get? This seems like a bad idea.

mathemancer: Or rather, it seems like the roles are set up incorrectly.

mathemancer (changes requested):
I have a question about adding roles based on parsing the request method. This seems it shouldn't happen.

Also, see @kgodey 's comments about line length and grammar.

mathesar/api/db/permissions/database.py
silentninja: Anyone with an `Editor` or `Viewer` role can only view the related items, Create access is allowed for `Manager` only. I am curious to know why this is a bad idea.

kgodey (approved):
I'm approving this PR so that my previous review won't block auto-merge once @mathemancer reviews it.

mathesar/api/db/permissions/database.py
mathemancer: After talking synchronously with @silentninja , I'm satisfied that this makes sense.

mathemancer (approved):
I think this is fine after our sync.

PR #1905, opened 2022-11-07
Throw targeted error message for columns not longer present in the Query

### LINKED ISSUE #1596: Provide targeted error messages when tables & columns involved in an exploration are no longer present
* Create a new exploration with a base_table and some initial columns
* Delete some of the columns in the table directly
  - Notice that the API responds with a 500 error code
* Delete the table which is the base_table of the query
  - The query also gets deleted
  - We need to show these in the warning message while prompting for user confirmation
* Delete a linked table whose columns we use as initial_columns in the query
  - The API responds with a 500

Expected:
* The frontend needs to show the columns and highlight the ones that are no longer present
* The API response needs to provide both the column structure and targetted errors

silentninja:
Related to  #1596 

The PR aims to fix the errors related to two of the scenarios
- The `base_table` gets deleted - Frontend should check if the Table has dependencies and ask the user to fix the related query before deleting the Table.
- A column in `initial_columns` is deleted - Error is thrown along with the query and the `column id`. I am not able to find a way to get the `column` property as it needs the actual column object which is not possible once the column is deleted. 

**Technical Details**

The response for 
`GET` request to `/tables/<id>/ui_dependents/` is
```
{
    {
        'queries': [
            query_id: int
        ]

    }
```

The error response when an initial column is deleted is
```
 {
  'query': # Same as query object sent by the frontend
  'column_id': int //missing column id
}
```
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
@pavish Can you check if the response can be used for throwing the error message? The backend won't be able to return `output_column` or `column_metadata` as we need the actual `Column` object to construct those properties which is not possible with a deleted Column.

dmos62 (commented):
I don't have technical comments.

pavish:
@silentninja We will be able to construct the error message with this response. However, it looks like we can only show the first missing column based on your response. Is my assumption correct, or is there a way to get all the missing columns, when there are multiple `initial_columns` missing?

pavish:
@silentninja Is this waiting on me? I have provided my [comment above](https://github.com/centerofci/mathesar/pull/1905#issuecomment-1312766996), and I would like the response to return all the missing columns instead of the first one.

silentninja:
> We will be able to construct the error message with this response. However, it looks like we can only show the first missing column based on your response. Is my assumption correct, or is there a way to get all the missing columns, when there are multiple `initial_columns` missing?

You are right, the error response would contain only the first missing column. I see this PR as a temporary fix, long term I would want the user to figure out what to do with these dependents using the dependency API, which is a much better UX.

When a column has been deleted the query that uses the column does not make sense, in most cases queries are created with a context related to the columns. For example, If I had a query for "Books with a name starting with 'A'" and the column `name` gets deleted, then the query is meaningless. 

I would prefer a proper use case before going too deep into this as it would slightly complicate how things are done on the backend.

pavish:
@silentninja 

> I would prefer a proper use case before going too deep into this as it would slightly complicate how things are done on the backend.

I am confused by this. I agree that the query is meaningless, the idea is to show the user information for them to fix the query.

The UX we have is that when columns are missing, we indicate to the user that the columns are missing and we provide an option to them to delete those intial_columns from the query. The above error with only the first missing column would not satisfy the UX.

Reference: [Figma link](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=1616%3A26083)

dmos62:
@silentninja @pavish I feel like a sync call is warranted.

silentninja:
@pavish We agreed upon showing errors for a single column during the weekly meeting. If you are satisfied with the response can we merge this PR?

pavish:
@silentninja Yes please, I wasn't aware this was waiting on me.

pavish (approved):

Issue #1481, opened 2022-07-26
Update existing `install.py` script to walkthrough the user in setting up Mathesar


silentninja:
## Problem
We don't have an installation script to help walk through the user in setting up Mathesar. 

## Proposed solution
We should update our existing `install.py` to walk through the users in setting up Mathesar by
- Get credentials for Mathesar Database if none exists on the config file.
- Creating a super user
- Prompting the user to enter the database credentials for an external database
- Help the user create a new Database

silentninja:
Closed as it is duplicate of #2031

Issue #1906, opened 2022-11-07
Live demo checklist


mathemancer:
## Problem

This is a meta-issue to track things @mathemancer needs to get done for the live demo, in descending order of priority
- [x] Set up multitenancy at the DB level
- [x] Document how that works, ask other devs to try it.
- [x] Set up actual infrastructure for demo. (just use staging?)
- [x] Change demo username / password to `demo` / `demo` - @mathemancer 
- [x] Figure out how website --> live demo transition will work (login, etc.)
  - [x] Design for login page - @ghislaineguerin 
  - [x] https://github.com/centerofci/mathesar-website/issues/6 - @pavish 
- [x] Data set loading improvements
  - [x] Backend
    - [x] #2244
    - [x] https://github.com/centerofci/mathesar/pull/2247
  - [x] ~#2222 + frontend~ - not needed due to backend improvements
- [x] #1923 
- [x] #2225
- [x] #2223
- [x] Find and squish bugs that appear from multitenancy - @mathemancer 
  - [x] https://github.com/centerofci/mathesar/issues/2307 
- [x] Add cronjob to delete demo DBs regularly - @mathemancer 
- [x] Finalize test data sets. - @mathemancer
  - [x] Make Library data set more realistic.
  - [x] Add movie collection data set
  - [x] #2070 
  - [x] #2203
  - [x] Add default column display options, explorations, etc. to all data sets - @mathemancer 
- [x] Reset demo DBs after data sets are finalized

Data sets discussion: https://wiki.mathesar.org/en/meeting-notes/2022-11/2022-11-17-weekly#data-sets-discussion

seancolsen:
I'd like to request the following modifications to column names to make them simpler and more intuitive.

| Table | Current name | Desired new name |
| -- | -- | -- |
| Authors | Author First Name | First Name |
| Authors | Author Last Name | Last Name |
| Authors | Author Website | Website |
| Publications | Publication Year | Year |
| Publishers | Publisher | Name |

_(Instead of creating a new ticket, I'm posting this here at Brent's recommendation.)_

seancolsen:
As for "display options" (or perhaps "metadata" is more appropriate) to configure...

- It would be nice to customize the record summary template for both "Authors" and "Patrons" to show `{First Name} {Last Name}`.
- We might want to put a couple saved explorations in there to give the user a quick sense of what is the Data Explorer is capable of.
- Eventually, I'd like to customize all the column widths. But we don't have that functionality built yet.

[...]

PR #1907, opened 2022-11-07
Live Demo Multitenancy setup

mathemancer:
Related to #1906 

This sets up "invisible" multitenancy for our Live Demo. With this setup, each new session that's connected should receive its own separate copy of the test data sets (for now, just the Library Management data set).

**Technical details**

The multitenancy is set up off of the session ID. This means that connecting from different browsers will result in separate data sets. Same for private windows in the same browser (under most circumstances).
- This should _not_ be considered a secure way to keep data separate.
- This was the quickest way to get things working without implementing true multitenancy.
- A user that connects to the root address will be rerouted to a database with a key appended (the key is deterministically derived from their `sessionid` cookie once they're logged in).

The way database reflection works had to be changed. This change would be necessary anyway in the long run, so should be kept. The difference is we no longer "reflect" database information from the `config/settings.py` file on-the-fly. This is not recommended in the first place, and should have been avoided in the first place. Instead, we reflect database info on the first load of the App, and proceed from there. New databases can be (and are) added via the `Database` model.

### TODO

- Some tests made incorrect assumptions about how to handle multiple databases. Those tests had to be skipped for now.
- The new logic is largely untested.
  - I have more faith in people clicking around, given the current state of our code base.
  - I will add some tests for this behavior if I get time, but it's lower priority given the temporary nature of this solution.
- This may increase problems with state from `metadata`. I didn't notice any with my testing, but I can only test one `sessionid` at a time (for quick requests, anyway). A follow up PR will improve the metadata state handling in a way that should help.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


config/settings.py
```diff
@@ -89,6 +90,16 @@ def pipe_delim(pipe_string):
 }
 DATABASES[decouple_config('DJANGO_DATABASE_KEY')] = decouple_config('DJANGO_DATABASE_URL', cast=db_url)
 
+LIVE_DEMO = decouple_config('LIVE_DEMO', default=False, cast=bool)
```
kgodey, line 93: I'd recommend namespacing settings that are specific to Mathesar i.e. `MATHESAR_LIVE_DEMO`, since we may install other Django apps with generic setting names.

config/settings.py
```diff
@@ -89,6 +90,16 @@ def pipe_delim(pipe_string):
...
+TEST = decouple_config('TEST', default=False, cast=bool)
```
kgodey, line 94: Is this a default Django setting? If not, we should namespace it too.

db/install.py
```diff
@@ -7,18 +8,25 @@
...
         username, password, hostname, user_database, port
     )
-    install.install_mathesar_on_database(user_db_engine)
-    user_db_engine.dispose()
-    root_db_engine.dispose()
+    try:
+        user_db_engine.connect()
+        user_db_engine.dispose()
+        print(f"Database {user_database} already exists! Skipping...")
+    except OperationalError:
```
kgodey, line 19: Does this fix https://github.com/centerofci/mathesar/issues/1481?

mathesar/api/dj_filters.py
```diff
@@ -44,8 +77,11 @@ class Meta:
         fields = ['name', 'schema', 'created_at', 'updated_at', 'import_verified']
 
 
-class DatabaseFilter(PropertyFilterSet):
-    sort_by = OrderingFilter(
+class UIQueryFilter(PropertyFilterSet):
```
silentninja, line 80: Why are thse filters needed for multi-tenancy?

mathesar/tests/api/test_schema_api.py
```diff
@@ -49,6 +50,7 @@ def test_schema_list(client, patent_schema, MOD_engine_cache):
     )
 
 
+@pytest.mark.skip("Faulty DB handling assumptions; invalid")
```
kgodey, line 53: Please create an issue to fix and re-enable these at some point, might be a good `help wanted` issue.

kgodey (commented):
I'll let @silentninja take a full look at this, I just skimmed through the code and added some comments. 

I'm wondering if we need the DB-per-session functionality. Could we maintain a list of DBs in Ansible or in some script elsewhere and figure out which DB to send people to _outside_ of the product? I'm mainly trying to avoid building a bunch of live demo related middleware or conditionals directly in the product code. 

If this will take too long to figure out, we can create an issue for it instead of fixing it now.

mathesar/tests/api/test_schema_api.py
mathemancer: We need to rework the whole testing methodology around DBs. These tests being invalid is a symptom of an underlying problem. I'll make an issue for that, though.

db/install.py
mathemancer: Yes, actually.

config/settings.py
mathemancer: This appears to be set up by `pytest-django` automatically when we run the tests.

config/settings.py
mathemancer: Good point, will do.

silentninja (approved):
Nice work @mathemancer I didn't come across any bugs when testing out the PR locally. But there few issues that make me concerned

1. Leaking staging-related code into the product codebase - Are you planning to merge this PR into the master? If so we should think of abstracting staging-related code(maybe starting with a separate settings.py file). It should be okay for now but create an issue to track it.
2. Please update the documentation on how the Database objects are created and the steps to take when some one wants to connect to a new Database.

mathemancer:
> I'll let @silentninja take a full look at this, I just skimmed through the code and added some comments.
> 
> I'm wondering if we need the DB-per-session functionality. Could we maintain a list of DBs in Ansible or in some script elsewhere and figure out which DB to send people to _outside_ of the product? I'm mainly trying to avoid building a bunch of live demo related middleware or conditionals directly in the product code.
> 
> If this will take too long to figure out, we can create an issue for it instead of fixing it now.

I did consider avoiding changing the Mathesar service itself, but went with the current approach for the following reasons:
- speed of implementation and overall simplicity (biggest consideration)
- I couldn't figure out how to avoid some branching on `LIVE_DEMO` mode in the web service even if another abstraction were added (assuming one webservice; this would be avoidable by also having multiple web service installations).
- Most of the changes in this PR are actually general improvements aside from live demo concerns:
  - installation improvement
  - avoiding re-reading `config/setttings.py` on the running service.
- The _only_ changes which would be reverted later to completely take out live-demo specifics be:
  - remove the new `LiveDemoMiddleware`
  - remove the `demo` directory
  - take the conditional out of the initial page load template view.
  The last is the one I'm least comfortable with.

All that said, I actually think the Live Demo Mode should stay in the product long-term, since many data products come with a way to load sample data sets and play with them, and this enables that.

Given the feedback, I'm going to make a few changes, and re-request review from @silentninja. (You're welcome to re-review as well, @kgodey if you have time).

db/install.py
kgodey: Okay please mark #1481 a draft until you've updated it, @silentninja.

mathemancer:
Okay, the new version of this PR has the following changes from when you last reviewed:

- Moved the demo-specific middleware to `demo.middleware`
- Moved demo-specific settings (adding the middleware) to `demo.settings`
- Got rid of the `LIVE_DEMO` env var. Replaced it with using the `DJANGO_SETTINGS_MODULE` variable (more idiomatic, I think).

Now,  to get rid of all demo-specific logic, one need only delete the `demo` directory.

To _use_ the demo-specific logic, one should set `DJANGO_SETTINGS_MODULE=demo.settings` in their `.env` file.

silentninja:
Looks good, I don't have any concerns now.

silentninja (approved):

demo/settings.py
```diff
@@ -0,0 +1,5 @@
+from config.settings import *  # noqa
+
+MIDDLEWARE += [  # noqa
```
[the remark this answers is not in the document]
mathemancer: I'll go ahead and add it here. It'll have to be the caller's responsibility to handle the case where the variable is missing (i.e., when it's not `False`).

PR #1929, opened 2022-11-14
1701 Improve UX after columns are extracted to a new table

### LINKED ISSUE #1701: Improve UX after columns are extracted to a new table
## Current behavior

- We don't provide a great experience for the user after extracting columns to a new table.
- The new column ends up ordered at the end.
- The new column may be horizontally scrolled out of view (especially given that it's placed at the end of the table).
- The new column is not selected.
- The user doesn't have any indication of success, other than the modal closing.

## Desired behavior

- The new FK column is selected.
- The new FK is horizontally scrolled into view.

## Design questions

- Other than the "desired behavior" above, is there anything _else_ we want to do to communicate success to the user? Do we want to display a toast message?

## Other _eventually_ desired behavior, not covered by this ticket

(Mentioned here for the sake of completeness)

- Our [specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/column-extraction.md) also require the new FK to be positioned in the table where the old columns were. Notwithstanding the ambiguity arising from non-adjacent old columns, merely sorting the new column at all will already require some [formidable work](https://github.com/centerofci/mathesar/issues/1584). I don't want to attempt to address that within this ticket, but I'm highlighting this problem to make it clearer why we have the other problems.

AlekLefebvre:
Fixes #1701 

Once the columns are extracted to a new table and the current table is refreshed to display the new FK column, this PR retrieves the last column, selects it and scrolls to it.

**Technical details**
I re-used the scrollBasedOnActiveCell function to create scrollBasedOnSelection. A svelte tick() call is needed before calling it because it relies on the cell's html classes.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
This works great @AlekLefebvre! Great job figuring out how to do it!

I added be49840a36d7ecb98c65fdf8e770f3ff8654186e, changing the approach to getting that `ProcessedColumn` object that we need to feed into `toggleColumnSelection`.

You had this line (annotated by me):

```ts
//                                         ╭╌╌╮
const results = await Promise.all(followUps as FollowUpsTypes);
//                                         ╰╌╌╯
```

That `as` "type assertion" is a TypeScript construct that we generally want to use only as a last resort, and only at a pretty low level when we're fairly confident that nothing underneath us will change. We sacrifice type safety with `as`, so it can lead to bugs. For example, if the return type of `$tabularData.refresh()` were to change at some point in the future, we wouldn't see any type errors within this function because the `as` keyword tells TypeScript _"trust me, I'm certain that `followUps` will have type `FollowUpsTypes`"_. And as it turns out, the way you had defined `FollowUpsTypes` _already_ had an inconsistency with the return type of `$tabularData.refresh()`. You had used `ProcessedColumn` in the position where `refresh` was returning `Column`. That's a problem because `toggleColumnSelection` was expecting a `ProcessedColumn`. With your code we were passing a `Column` into `toggleColumnSelection` when it required a `ProcessedColumn`. With the current implementation of `toggleColumnSelection`, that was actually fine (because both `ProcessedColumn` and `Column` have the same `id` property), but were the implementation of `toggleColumnSelection` to change in the future, we could end up with a bug that's tricky to track down.

Your approach of getting the newly-created column from the `refresh` function was a good idea. You just didn't do it in a type-safe way. A type-safe way to do that would be like this...

```ts
// Initialize something empty that we can await. Tell TS to infer a wider type
// than it would by default. This way we can re-assign it.
let getNewTable: Promise<unknown> = Promise.resolve();
if (yadaYadaYada) {
  // Conditionally re-assign it to a different promise that we can also await.
  getNewTable = getTableFromStoreOrApi(response.extracted_table);
}
const refresh = $tabularData.refresh();

// Pass a tuple into `Promise.all` instead of an array. TS sees this as a tuple
// because `Promise.all` is typed in such a manner to see _literal_ arrays as
// tuples. Now TS correctly infers the type of `result` with no type assertions
// needed.
const result = await Promise.all([refresh, getNewTable]);
const refreshResult = result[0];
const refreshedColumns = refreshResult[0]; // has type `Column[] | undefined`
```

Or with nested destructuring we could further shorten those last three lines to...

```ts
const [[refreshedColumns]] = await Promise.all([refresh, getNewTable]);
```

But the above code would still fall short of what we need here. We need a `ProcessedColumn`, not a `Column`.

The approach I took was to read the `processedColumns` store that we already had in scope. In doing so, I learned that there is some funny behavior with Svelte here, hence my code comment. I'd like to learn more about that later, but I think this should suffice for now.

PR #1936, opened 2022-11-15
Add pretty DB naming for Live Demo

mathemancer:
Related to #1906 

This adds a function that deterministically generates an appropriate DB name from a user's `sessionid` cookie, and uses it for the auto-generated DBs in the live demo.

**Technical details**

The name has the form:
```
word1_word2_word3_word_4_word5
```
where `wordN` is a word chosen from a list in the source code. The current version maps session IDs uniformly(ish) into a space of 148<sup>5</sup> possibilities, making collisions unlikely but not impossible.

I had to keep the underscores (instead of spaces) and avoid capital letters in the generated names for now, since the DB names are passed to Postgres in a way that isn't safe for spaces or capital letters.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


demo/db_namer.py
```diff
@@ -0,0 +1,87 @@
+"""Utility function for generating database names from session ID cookies."""
+
+ADJECTIVES = [
```
kgodey, line 3: I think we should choose neutral words here (avoid things like `stupid`, `hideous`, `grisly`) and also avoid anything that could potentially represent an ethnicity (`asian`, `black`, `white`) because we could end up randomly generating unpleasant combinations. Perhaps including or entirely using nouns might make it easier (stuff like `car` or `fox` or `airplane`)?

kgodey (changes requested):
Logic seems fine, but we need to update the word list, see in-line comment.

Anish9901 (approved):
Looks good to me. I was wondering if we could've achieved the same result without converting the bytes to hex and reducing the chunk size by half.

mathemancer:
> Looks good to me. I was wondering if we could've achieved the same result without converting the bytes to hex and reducing the chunk size by half.

Do you know of an algorithmic way to convert a general string to decimal in an efficient way in python? I don't know of one that doesn't go through `bytes`, at least.

Anish9901:
> Do you know of an algorithmic way to convert a general string to decimal in an efficient way in python? I don't know of one that doesn't go through `bytes`, at least.

@mathemancer I think something like this could work:

```
session_dec = list(session_id.encode('utf-8'))
```

mathemancer:
> > Do you know of an algorithmic way to convert a general string to decimal in an efficient way in python? I don't know of one that doesn't go through `bytes`, at least.
> 
> @mathemancer I think something like this could work:
> 
> ```
> session_dec = list(session_id.encode('utf-8'))
> ```

That's a good idea, but:
- It _does_ go through bytes, and
- the decimal space isn't big enough. I suppose it might be possible to expand by performing some operation on chunks in the list (multiplying, adding, not sure). I'm not as confident that it maps "evenly" into the range of possible keys. The goal is to get a distinct string for each `sessionid`, and collisions are minimized if the mapping is even.

Anish9901:
> That's a good idea, but:
> - It _does_ go through bytes, and
> - the decimal space isn't big enough. I suppose it might be possible to expand by performing some operation on chunks in the list (multiplying, adding, not sure). I'm not as confident that it maps "evenly" into the range of possible keys. The goal is to get a distinct string for each `sessionid`, and collisions are minimized if the mapping is even.

You are right, I mistakenly thought the typecasting of the hex string was getting converted into decimal.

mathemancer:
@kgodey I created a new list of words with max length 3 from the list you turned me onto. The reason for shortening the words is that we get way more out of multiple words than we do from longer words when it comes to total entropy, and I found that this length gave a good combination of entropy and efficiency.

kgodey (approved):

Issue #202, opened 2021-06-07
Reset migrations


kgodey:
A reminder for us to reset migrations before the first release.

- [x] Reset migrations in the repo
- [x] Send email to developer mailing list with local instructions
- [x] Reset migrations on all servers tracked in Ansible
  - [x] `staging.mathesar.org` probably needs a reboot 
  - [x] `library.mathesar.org` can be removed entirely, since we're not using it

mathemancer:
@kgodey I've removed `library.mathesar.org` from the ansible playbooks in [this PR](https://github.com/centerofci/mathesar-ansible/pull/35) , but I figure I'll use it for testing the hosted prod setup since it already has a domain, etc. set up.

PR #1937, opened 2022-11-15
Removed Sphinx from codebase

kgodey:
We're not using Sphinx and it's causing an error for a contributor. So this PR removes Sphinx related commands from the Makefile and also moves mathesar setup commands from the Makefile to the docker `run.sh` script. So this PR completely removes our dependency on Makefile.  

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
We can integrate most of the Makefile functionality into docker-compose except for the step to copy the `.env` file. If we get rid of the Makefile, copying the `.env` file would have to be done by the user in addition to running docker-compose. 

I am leaning towards removing the Makefile and improving the documentation for copying the `env` file as it would give an idea of what is happening to the user(they would need to know about this if they want to connect to an existing database). Any thoughts on this?

kgodey:
I think it's fine to have the additional manual step of copying `.env` @silentninja

run.sh
```diff
@@ -22,4 +22,6 @@ else
 fi
 
 cd ..
+python manage.py migrate && python install.py
+python manage.py createsuperuser --no-input --username admin --email admin@example.com
```
kgodey, line 26: @silentninja Please note that is not idempotent and will fail if you've run it already; we should handle this somehow. I think `install.py` is fixed now, but if it will error we should handle that too.

silentninja: >  Please note that is not idempotent and will fail if you've run it already

The command would throw an error but it wouldn't crash the script. But in long run, we should definitely move it to a migration file and create a `superuser` as part of the migration step. I was planning to create a separate issue for it and fix it along with #202

silentninja: `install.py` should work similarly to Django migration once #1481 is fixed

silentninja:
> Right now I’m having a bit of trouble getting docker-compose to start everything but I’ll wait till the issue at hand is resolved until I bring up any other problems

@IyadElwy The Makefile was running the docker-compose command under the hood along with other setup commands which aren't really needed for the project and these stale dependencies have been removed in this PR. Can you try out this branch and see if works?

kgodey:
@silentninja Will you also be assigning a reviewer? The PR is still assigned to you.

silentninja:
> Will you also be assigning a reviewer? The PR is still assigned to you.

I have asked @dmos62 to do an additional review.

run.sh
kgodey: Please do make issues to track these things so we don't forget about them. As long as it doesn't crash the script, that's fine with me.

kgodey (commented):
I took a look at the code and it looks fine to me. I have a lot of other high-priority things that I can't delegate, so I'll delegate local testing of this PR to @dmos62.

IyadElwy:
@kgodey @silentninja Good news! The remove-sphinx branch works. Also, instead of using the makefile, I think copying the env file and running docker-compose is easier. So I think you should merge this branch with the master branch. I just came across another minor issue but it was easily fixed: Some files (Dockerfile, docker-compose, run.sh) have windows "\r" endings which aren't correctly interpreted when running docker-compose. The fix was just running ` sed -i.bak 's/\r$//' filename`  for UNIX endings.

silentninja:
@IyadElwy Thanks for confirming that the fix works.

As for the issue related to windows line endings, it could be fixed by cloning the repo inside a Linux subsystem. 
We have documentation about the fix in this wiki https://wiki.mathesar.org/en/community/setup_in_windows but it looks like it isn't mentioned in our `README.md` which is leading to new users missing out the instructions. Would like to send a PR to add it to our `README.md` ?

dmos62 (approved):
Looks good. Didn't test, since changes seem trivial (and @IyadElwy reported it working).

I agree with @kgodey that we need to track the planned improvements.

PR #1958, opened 2022-11-20
Added favicons to the platform by searching for the Font-Awesome icon…

### LINKED ISSUE #1520: Customize page favicons
## Current behavior

- All pages have the same favicon

## Desired behavior

- Our [Navigation Specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/navigation.md#navigation-via-page-url) require that we use distinct favicons for pages that represent different types of entities.

## Implementation

1. Review specs to see which entities require custom favicons.
1. Look in `src/icons.ts` to find the correct FontAwesome icons for those entities.
1. Generate custom favicons based on the FontAwesome icons. I like to follow [this guide](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs) for favicons, but we could probably get away with just using the SVG. Which browser's don't support SVG favicons these days? It might be good to look.
1. Use `<svelte:head>` to modify the HTML `<head>` element from within components like `TablePage.svelte`

## See also

- #3605 which is simpler (just one static favicon) and higher priority

IyadElwy:
…s, downloading, encoding and saving their svg, then saving the favicon element as svelte components and inserting them as needed

Fixes #1520

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
Nice work on this, @IyadElwy. The end result seems to behave well in Firefox and Chrome.

I'd like to clean up some of this code before merging though...

Your use of `data:image/svg+xml` is clever, and I think this approach should work well, even though it's not quite what I had mind. I was thinking we'd save plain old `.svg` files that the browser would request. If that were to be the implementation strategy, then placing those `.svg` files within `static-assets` would be appropriate. However, since we are using `.svelte` components instead, placing them in `static-assets` would be unconventional.

Here's what I'd like to do instead:

- Make a new `Favicon` component which lives at `src/components/Favicon.svelte`

    - Just like our `Icon` component, the `Favicon` component should accept a `data` prop as follows:

        ```ts
        export let data: IconProps['data'];
        ```

    - Then, the `Favicon` component should render that `<link>` tag as you've set it up in your favicon files.
    
    - It should use logic similar to `Icon` to build the paths in the svg. Because it needs to build it all into one string, you can use template literals.
    
    - Then you can either replace the angle brackets with their escaped equivalent, or you can encode the whole string as base64.

- Delete your new files in `static-assets`.

- Where you currently have code like

    ```svelte
    <FaviconTable/>
    ```

    We'd instead use

    ```svelte
    <Favicon {...iconTable} />
    ```

    This makes the favicon system more composable, allowing us to make use of the icons that we have already defined in `src/icons/index.ts`, some of which are now custom.

- So after all those changes, your PR would end up adding only one new file (`src/components/Favicon.svelte`), and modifying the `*Page.svelte` files you've already modified. Are you up for making these improvements?

- Also, make sure to run the [code formatting](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/README.md#formatting) when you're done with your changes.

IyadElwy:
Did you have something like this in mind ?

```svelte
<script lang="ts">
    import {IconProps} from "@mathesar-component-library/types";

    export let data: IconProps['data'];

    let path: String = '';
    $: [path] = data.icon;

</script>

<link rel="icon"
      href="{path}"
      type="image/svg+xml"/>


```

seancolsen:
@IyadElwy 

> Did you have something like this in mind ?

Yes, something like that. Building the path will require some more work, but yes, you're on the right track.

pavish:
@IyadElwy Are you still working on this PR?

There hasn't been any activity for over three months now and we might have to close this PR if we don't hear back from you. We'll wait for 3 more days (Monday, 2023-03-20 EOD Eastern Time) and close this PR if we don't receive further communication from you.

pavish:
Closed due to inactivity.

PR #1972, opened 2022-11-22
Change behavior of FK-Column Filter to be more similar to the filters for boolean columns.

### LINKED ISSUE #1523: Customize the filtering functions for FK columns
## Current behavior

- Filtering on an FK column presents the user with different filtering functions that don't quite make sense given our UX with the record selector. For FK columns, the value within the filter condition will be acquired through the (modal) Record Selector and afterwards displayed within the input using the Record Summary for the acquired record. The numerical primary key value used for the filtering condition won't be displayed to the user unless it happens to be contained within the Record Summary. Thus, for example, selecting all Publications with and author _greater than_ "Octavia Butler" doesn't quite make sense -- in the quantitative sense, at least 😜.

## Desired behavior

- Filtering functions for FK columns look more like the functions for boolean columns. We should only have "is empty" and "is equal to".

LFLATH:
The purpose of these changes is to make the filtering for FK columns behave more like the filtering for boolean columns. I don't believe that these changes are the best or most effective solution to the problem. However, they "work", so I thought I would submit a PR.

If this is not the correct way to submit a PR, I am really sorry!

Fixes #1523 

**Technical details**

The changes that I made are in filtering.ts and processedColumns.ts. In processedColumns.ts I added a new argument to the getFiltersForAbstractType() function. This argument is undefined when the column is not an FK column. I then use this argument in the filterDefinitionMap.get() call to check if the column is an FK column. If the column is an FK column, I use the type "Boolean" for the column. I chose this approach because it allows the FK column to access the boolean filters. If the argument passed is undefined, meaning the column is not an FK column, the filterDefinitionMap.get() uses the correct abstract type for the column. 
**Screenshots**

Changes in processedColumns.ts
Screenshot 2022-11-22 at 10 46 39 AM
Changes in filtering.ts
Screenshot 2022-11-22 at 10 47 24 AM

This image shows that Names_ID is an FK column.

This image shows that Names_ID has the correct filters available.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/abstract-types/operations/filtering.ts
```diff
@@ -218,13 +219,14 @@ export const filterDefinitionMap = getFilterDefinitionMap();
...
 ): Map<AbstractTypeFilterDefinition['id'], AbstractTypeFilterDefinition> {
   const allowedFiltersMap: Map<
     AbstractTypeFilterDefinition['id'],
     AbstractTypeFilterDefinition
   > = new Map();
 
-  filterDefinitionMap.get(categoryIdentifier)?.forEach((filter) => {
+  filterDefinitionMap.get((isFK !== undefined ? "boolean"  : categoryIdentifier))?.forEach((filter) => {
     allowedFiltersMap.set(filter.id, filter);
   });
```
pavish, line 231: @LFLATH 
As mentioned on [the comment in the issue](https://github.com/centerofci/mathesar/issues/1523#issuecomment-1322381116), these change should not be in `filtering.ts`.

It should be in a utility file as part of the filter-entry component, similar to the cell-fabric component.

mathesar_ui/src/stores/table-data/processedColumns.ts
```diff
@@ -103,7 +103,7 @@ export function processColumn({
       linkFk ? linkFk.referent_table : undefined,
       abstractType.cellInfo,
     ),
-    allowedFiltersMap: getFiltersForAbstractType(abstractType.identifier),
+    allowedFiltersMap: getFiltersForAbstractType(abstractType.identifier, linkFk),
```
pavish, line 106: The newly created utility function should be called here, instead of calling `getFiltersForAbstractType` directly.

mathesar_ui/src/stores/abstract-types/operations/filtering.ts
```diff
@@ -218,13 +219,14 @@ export const filterDefinitionMap = getFilterDefinitionMap();
...
   categoryIdentifier: AbstractTypeCategoryIdentifier,
+  isFK: FkConstraint
 ): Map<AbstractTypeFilterDefinition['id'], AbstractTypeFilterDefinition> {
   const allowedFiltersMap: Map<
     AbstractTypeFilterDefinition['id'],
     AbstractTypeFilterDefinition
   > = new Map();
 
-  filterDefinitionMap.get(categoryIdentifier)?.forEach((filter) => {
+  filterDefinitionMap.get((isFK !== undefined ? "boolean"  : categoryIdentifier))?.forEach((filter) => {
```
pavish, line 229: Is it better to have a separate method in this file for getting filters which are only for equality checks (to be used with PKs and FKs), which include the `equal` and `empty`. That method should then be called to get the filters, instead of getting the filters of boolean.

pavish (changes requested):
@LFLATH 

Good job navigating through the codebase and identifying the changes needed and making it work, as expected. However, I'd like a different approach to be followed here on how it is implemented.

You can keep the PR open, and make the changes on the same PR, and re-request review.

mathesar_ui/src/stores/abstract-types/operations/filtering.ts
LFLATH: Thank you so much for the guidance. I will start trying to implement these changes. I am still quite new to JS so it may take some time. Thank you again for the help, I really appreciate it.

LFLATH:
I made some changes and ran the code through Prettier and the Linter. I tried to make it more to the specifications. Let me know what other changes need to be made!

kgodey:
@LFLATH Please re-request review from @pavish using this button, it'll ensure he gets notified.
Screen Shot 2022-11-25 at 6 12 11 PM

LFLATH:
Thanks for the help, sorry about that!

pavish (approved):
@LFLATH The latest changes look good, however I had a few more suggestions. Since you're new to the stack, I thought it'd be easier to explain the difference if I added a commit myself and used that as an example. This might help you with future PRs.

Here's the commit: https://github.com/centerofci/mathesar/pull/1972/commits/a83e89bf23857697bdd9e567736cce16683c18dd

The difference:
1. There were a couple formatting errors and type errors from your last commit. I did notice that linting errors were all fixed.
   - The formatting errors can be identified by running `npm run check-format`, which calls prettier. Then, they can be fixed either manually or by using `npm run format` which will automatically make the fixes. **It is recommended that you fix formatting last, because any fixes you perform on linting and typecheck errors would require formatting.**
2. I removed the `isFk` argument from `getFiltersForAbstractType` and moved the FK related conditional logic to the `utils` file. The idea is that the files within `/abstract-types/` should only have to worry about type logic and not other column constraints and properties such as FKs.

pavish:
> If this is not the correct way to submit a PR, I am really sorry!
> Thanks for the help, sorry about that!

@LFLATH As an additional note, I wanted to ask you not to apologize. You took the time to explore Mathesar, understand what was required, did a great job with this PR, and we as a team thank you for helping us out! :)

LFLATH:
@pavish Thanks for the help! I have looked at the changes and they all make sense. I am super excited to continue working!

PR #1980, opened 2022-11-24
Add `remove` transform for hiding columns

### LINKED ISSUE #1867: Transformation for hiding columns
We want a DE transformation that selects a subset of columns.

dmos62:
Fixes #1867

The new transform's id is `remove` and it takes a list of aliases, in the same vein as the existing `select` transform.

Note, I moved a test function to a different file, and then added a test case to it. Sorry for the diff noise.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/transforms/base.py
+++ b/db/transforms/base.py
@@ -390,6 +390,49 @@ def _add_aliases_to_summarization_expr_field(
     return summarization
 
 
+class RemoveColumns(Transform):
+    """
+    We're implementing this transform in terms of selecting columns (since Postgres doesn't really
+    support "selecting everything except").
+    """
+    type = "remove"
+
+    def apply_to_relation(self, relation):
+        input_aliases = [
+            col.name
+            for col
+            in relation.c
+        ]
+        columns_to_select = self.get_columns_to_select(input_aliases)
+        select_transform = SelectSubsetOfColumns(columns_to_select)
+        relation = select_transform.apply_to_relation(relation)
+        breakpoint()
+        return relation
+
+    def get_unique_constraint_mappings(self, input_aliases):
+        columns_to_select = self.get_columns_to_select(input_aliases)
+        return [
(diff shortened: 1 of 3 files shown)


db/transforms/base.py
```diff
@@ -390,6 +390,48 @@ def _add_aliases_to_summarization_expr_field(
     return summarization
 
 
+class RemoveColumns(Transform):
+    """
+    We're implementing this transform in terms of selecting columns (since Postgres doesn't really
+    support "selecting everything except").
+    """
+    type = "remove"
```
kgodey, line 398: `remove` seems like a very generic name, can we call it `remove_columns` or something like that?

I'm not familiar with this area of the codebase so please ignore my comment if they are not helpful or conflict with the style.

dmos62: I considered that. As noted in the PR description, a close sibling of this transform is the `select` transform, which is just as generic, but happens to match an SQL term. That's the rationale behind calling this `remove`.

Standard SQL doesn't do "removing columns", though some variants have `SELECT * EXCEPT ...`, so another idea could be calling this transform `except`. We can also say `remove_columns`, but then let's rename `select` into `select_columns`.

kgodey: `hide` might be a good alternative (and also matches the UI). I think of `remove` as permanent somehow.

dmos62:
Renamed `remove` to `hide`.

Converted to draft, because the non-master branch this is based on is not yet merged.

mathemancer (approved):
I think this one looks good. I note that this won't throw any validation error if one or more of the columns passed for "hiding" aren't there, but I think that's okay given our use case.

PR #1987, opened 2022-11-26
Fix FK behaviour for Tables having round brackets

### LINKED ISSUE #1979: Creating a Table with round brackets results in API error
## Description
Creating a Table whose name contains round brackets, for example, a Table with the name "Cycle 6 (Release)" and then creating a foreign key referencing the Table causes most of the APIs to return an error. This happens due to a bug SQLAlchemy foreign key parsing regex.  

## Expected behavior
Since the bug is due to SQLAlchemy, it is not feasible to fix it. So we should prevent creating a Table whose name contains round brackets by validating the Table name on the backend.

## To Reproduce
1. Create a Table whose name contains round brackets, for example, a Table with the name "Cycle 6 (Release)"
2. Create a foreign key in a different that references any column of the Table created in Step (1).
3. Try refreshing the Tables page or send a request to `/tables/` API, the backend with throw an error

Anish9901:
Fixes #1979 

**Technical details**

This PR adds validation for names of the referent tables and filters them out if it contains round brackets.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja (changes requested):
We would also need to prevent the user from creating a Table with round brackets, can you add validation for that too? Other than that the PR looks good to me.

mathesar/api/serializers/links.py
```diff
@@ -82,6 +82,13 @@ def create(self, validated_data):
     }
     link_type = serializers.CharField(required=True)
 
+    def run_validation(self, data):
```
silentninja, line 85: The validation should not be a part of the Mapping Serializer rather it should be for the mapped Serializer.
The validation will fail when creating Many to Many links as the `ManyToManySerializer` does not have a `referent_table` field, you would need to check the `referents` field in that case. So you would need to write a different validation method for each of type of link

mathesar/api/serializers/links.py
```diff
@@ -82,6 +82,13 @@ def create(self, validated_data):
...
+        if referent_table := data.get('referent_table', None):
+            referent_table_name = Table.current_objects.get(id=referent_table).name
+            if referent_table_name.find('(') and referent_table_name.find(')') != -1:
+                raise InvalidReferentTableName(referent_table_name)
```
silentninja, line 89: It would be better to use the `field` parameter of the `InvalidTableName` Exception class to specify which field of the body is failing the validation instead of having a different Exception.

mathesar/api/serializers/constraints.py
```diff
@@ -147,6 +148,10 @@ def create(self, validated_data):
         return serializer.create(validated_data)
 
     def run_validation(self, data):
+        if referent_table := data.get('referent_table', None):
```
silentninja, line 151: It would be better to use the `field` parameter of the `InvalidTableName` Exception class to specify which field of the body is failing the validation instead of having a different Exception.

silentninja (changes requested):

silentninja (approved):
Thanks for the PR @Anish9901 It looks good to me.

Issue #1985, opened 2022-11-25
Users & Permissions design


kgodey:
# New pages

## New Admin section
We need a new section of Mathesar that's only accessible to administrators. For the purposes of this issue, we'll only be designing the "Users" sub-section but we'll be adding other sections as part of our installation and deployment work.

- [x] Main admin panel page
- [x] Users sub section
    - [x] See list of users
    - [x] See list of permissions for each user
    - [x] Add new user
      - [x] Handle errors when user cannot be created
    - [x] Delete user
      - [x] Handle errors when user cannot be deleted 
    - [x] Edit user details
      - [x] Handle errors when editing fails 
    - [x] Display permissions for a given user
    - [x] Reset user's password
      - [x] Handle errors when operation fails

## New User Settings page
We'll need a user settings page for the user to edit their own settings. Right now, that will only include a "Account" sub-section but we'll include more sections in the future for the user to configure their personal settings.

- [x] Main user settings page
- [x] Account sub-section
    - [x] See their details
    - [x] Edit their own details
      - [x] Handle errors when editing fails 
    - [x] Change their password
      - [x] Handle errors when operation fails
    - [x] Delete their account
      - [x] Handle errors when operation fails

## Styling for log in page
- [x] Show log in form
- [x] Styling for errors

## Change password page
This page will show up when an admin has reset the user's password and the user is logging in using the new password. The user will need to change their password before being allowed into the application.
- [x] Explain the page
- [x] Show form for changing the password
- [x] Styling for errors
- [x] Handle the case where someone opens the URL directly – should show an error message that tells the user how to change their password the normal way.

## Post log out page
Once users log out successfully, they should be directed to a page with a message that says "You've successfully logged out" or something like that.

- [x] Design the page 

# Changes to existing designs

## Navigation changes
The navigation should be updated to include the following
- [x] Link to the admin section (only visible to admins)
- [x] User settings page (visible to all users)
- [x] Log out action

## Database page changes
We need to have different actions / options available depending on the user's permissions:

- [x] Users with Manager permissions
  - [x] Show the list of people with access to the database
  - [x] Allow adding users to the DB
    - [x] Handle errors
  - [x] Allow removing users who have access to the DB
    - [x] Handle errors
- [x] Users with Editor & Viewer permissions
  - [x] Disable creating / deleting / editing schemas

## Schema page changes
We need to have different actions / options available depending on the user's permissions:

- [x] Users with Manager permissions
  - [x] Show the list of people with access to the database
[...]

PR #2000, opened 2022-11-30
Add Password reset View

silentninja:
Related to #1673 

PR adds a password reset view which the superuser would use to reset the password of any user. 

The API to call the Password Reset 
```
POST `/users/<user_id>/password_reset/`
{
  "password": str, //The new password of the user
}
```

The PR also adds a middleware to redirect users with default passwords to the password change page and the Django View for the password change page. 

Passwords are considered as `default`
- When a user's password is reset by the superuser

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey (changes requested):
This approach won't work for the plan in the [Users and Permissions design issue](https://github.com/centerofci/mathesar/issues/1985). We'll be building the admin panel in Svelte (I assume) and the reset password functionality will need to be part of that panel, not a separate page. We'll need a reset password API to power that.

@pavish please confirm.

pavish:
@silentninja @kgodey 

> This approach won't work for the plan in the https://github.com/centerofci/mathesar/issues/1985. We'll be building the admin panel in Svelte (I assume) and the reset password functionality will need to be part of that panel, not a separate page. We'll need a reset password API to power that.
> @pavish please confirm.

Yes, this is correct. We will be implementing all of the users and permissions frontend in svelte and will require APIs for them.

silentninja:
@pavish Is there any reason to implement login APIs using svelte? 

There are a few reasons I wouldn't use Svelte
- Login  APIs don't hold state across pages which makes a tool like svelte unnecessary. I am not sure where Svelte would be used especially after your [comment](https://groups.google.com/a/mathesar.org/g/mathesar-developers/c/QSprKWKb5JQ/m/R-r4s1VuBAAJ) where you have mentioned using Django templates for the login APIs.
- The Django forms already have built-in features like validation and reimplementing these looks redundant.

mathesar/api/ui/viewsets/users.py
```diff
@@ -16,6 +21,16 @@ class UserViewSet(AccessViewSetMixin, viewsets.ModelViewSet):
     pagination_class = DefaultLimitOffsetPagination
     access_policy = UserAccessPolicy
 
+    @action(methods=['post'], detail=False)
```
kgodey, line 24: Why not make this a detail view? You don't need the username then.

mathesar/api/ui/permissions/users.py
```diff
@@ -11,7 +11,7 @@ class UserAccessPolicy(AccessPolicy):
         },
         # Only superusers can create users
         {
-            'action': ['create'],
+            'action': ['create', 'password_reset'],
```
kgodey, line 14: Users should be able to reset their own passwords, it shouldn't just be superusers.

kgodey (changes requested):

mathesar/api/ui/viewsets/users.py
silentninja: That makes sense if the front end is listing users and knows about the `id`. I was thinking of it as a form for the superuser to enter the username of the user. I will change it to a detail view

mathesar/api/ui/permissions/users.py
kgodey: Okay yes, you're correct. We do need a reset password flow for users later (using email to verify), but that's not in scope right now.

silentninja: Created #2052 for the password change flow

kgodey (changes requested):
@silentninja Sorry for not flagging this earlier, but the [users and permissions spec]() says:

> Administrators can reset anyone’s password. We should follow security best practices so that the admin does not ever see the user’s permanent password.

It seems like with this implementation, the admin does see the user's permanent password, since they are setting it. Can we avoid this? Either by forcing users to change passwords on first login (not sure how we'd do this with Svelte) or the reset password flow should generate a link that the admin can send to the user to set their own password (also not sure how we'd do this with Svelte). Please discuss with @pavish if needed.

pavish:
@kgodey @silentninja 

I'm confused because the discussions here seem to be talking about different things. Here's what I have in mind regarding usage of Svelte/django templates:
1. The admin panel will be in Svelte. We need APIs for all operations here.
2. The login page will be in django templates. Only after successful login, we'll redirect to the app, which is in Svelte.
3. Password reset by users:
    - If the user wants to reset their own password because they can't login, it'll be part of the login page (django template). This is the [flow mentioned here](https://github.com/centerofci/mathesar/pull/2000#discussion_r1041446843).
4. Password change by users:
    - If the user wants to change their password, it'll be in Svelte. They can do this only after they're logged in already.
5. Password resets by admin:
    - This will be part of the admin panel, so everything will be in Svelte and be done using APIs.

Any UI after the user authentication is complete, will be in Svelte. Everything before that will be in Django templates.

## The password reset flow

### User account creation & first login
* I'm in favor of letting the admin type in the password, and then forcing the user to change their password at first login.
* The admin is the one creating the account, not the user. So the admin decides the username anyway, and it makes more sense for them to provide a password and force the user to change it.
* I don't see why we need a flow for a link for setting password here, since the admin has access to that link anyway, and that link is unsecure because anyone with that link can set the password for the target user.
* The implementation here would be in Django templates (no matter the flow), because the authentication process is not fully complete until the user changes their password and logs in again.
   - Secondly, we should not allow the user to access any other part of the app until they change their password.

### Password reset for existing user
* When the admin wants to reset a user's password, they should be allowed to type it in, and then force the user to change it at the next login, same as the previous case.
* Since the admin has full control of creating, deleting, and setting permissions for an account, this makes most sense.

The link for password reset seems like a good option only if the user themselves want to reset their password, because they've forgotten it, and we have a flow where the reset link is sent to their emails. I don't think it makes sense when the admin wants to reset the password.

kgodey:
> The flow for forcing the user to set password will also be in Django templates.

This is the part I wasn't sure whether it should be in Django or Svelte, thanks for clarifying @pavish.

@silentninja we also need to implement this part, either in this PR or another PR. I think it makes sense to implement in this PR since it's related to this feature.

silentninja:
> I don't see why we need a flow for a link for setting password here, since the admin has access to that link anyway

The same applies when an admin enters a password, he could log in with the same password and then change it. Our current method won't stop an admin if he is determined in seeing a users password.

kgodey:
> The same applies when an admin enters a password, he could log in with the same password and then change it. Our current method won't stop an admin if he is determined to see the user's password.

We're not trying to stop admins from getting access to user accounts by any means possible. We're just trying to increase the default level of security.

kgodey:
I've added a new "change password page" to the [Users & Permissions design checklist](https://github.com/centerofci/mathesar/issues/1985) based on the plan we've come up with here.

silentninja:
@kgodey You might have to review the PR again as the commits after your last review adds a lot more features to the PR. I have updated the description to reflect the changes

silentninja:
@kgodey @pavish When are we planning to update the styling for the Django templates?

silentninja:
We should be considering passwords set by the superuser during user creation as default password too and redirect the user to change the password when they log in initially but I haven't made any changes to our existing user flow in this PR as it is out of scope for the current PR.

I will create a separate issue to track it.

kgodey:
> When are we planning to update the styling for the Django templates?

Once @ghislaineguerin has completed #1985.

kgodey:
> We should be considering passwords set by the superuser during user creation as default password too and redirect the user to change the password when they log in initially

Good catch, I added another checklist item in #1673 for it

PR #2017, opened 2022-12-02
Stylize NULL values within record summaries

### LINKED ISSUE #2004: Record summaries of `NULL` values should not show "null" text.
## Reproducing
1. See Author in https://staging.mathesar.org/mathesar_tables/318/tables/19749/17. The record summary is configured to be `{first name} {last name}` and this author (who has no first name) is showing up as `null Abu'l-Fazl`
1. Also see Language in https://staging.mathesar.org/mathesar_tables/318/tables/19749/1, which is showing up as "null"

## What I'd expect
1. The record summary should be ` Abu'l-Fazl`
2. The record summary should be an actual `NULL` value

seancolsen:
Fixes #2004

### Before

image

### After

image

### Before

image

### After

image

## Notes

- This introduces usage of `{@html }` into our codebase, which (as somewhat of a purist) I was really hoping to avoid, due to the potential for XSS. It's a little hard for me to imagine how we would easily do this _without_ using `{@html }` though. I've taken care to escape the HTML entities from user input before inserting into the DOM (which should be evident from my sample data including HTML tags), but I still don't love it. If anyone has an idea for a better approach here I'd be open to re-thinking this one.

- As you can see in the last screenshot above, there's still an issue with dark text on dark background in the header. I notice that the issue is also present for other text in the header too, so I'm considering that issue to be unrelated to this work. We'll need to figure out a clever way of dealing with that, probably by using some CSS variables. I hope we can take that up separately from this PR.

- Also in the last screenshot you'll notice that the plain text "NULL" appears in the browser tab. We can't stylize text in the tab title, so I figured this is the best fallback. But I'm open to other approaches there.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey I've assigned you to review this, just to make sure the outcome is in-line with your expectations.

kgodey:
I think this works, thanks @seancolsen.

mathesar_ui/src/stores/table-data/record-summaries/recordSummaryUtils.ts
```diff
@@ -154,3 +163,14 @@ export function renderRecordSummaryForRow({
...
+const plainTextNull = 'NULL';
+const stylizedNull = '<span class="postgres-keyword">NULL</span>';
+
+export function displayRecordSummaryAsPlainText(str: string): string {
+  return str.replace(nullPlaceholderPattern, plainTextNull);
+}
+
+export function displayRecordSummaryAsHtml(str: string): string {
+  return escapeHtml(str).replace(nullPlaceholderPattern, stylizedNull);
+}
```
pavish, line 176: Is there any specific reason why we need this method to return a string?

Instead can we return a `(string | ComponentAndProps)[]`, and split the input string, and replace the null value with a component?

This will remove our need for using the `@html` tag.

pavish (changes requested):
@seancolsen I do not like the `@html` tag usage here, and I have a suggestion. I'm requesting changes. However, if it takes time or if you'd like to handle it later, let me know, and I'll approve the PR.

mathesar_ui/src/stores/table-data/record-summaries/recordSummaryUtils.ts
seancolsen: Ah, okay, this is brilliant. Using an _array_ of `string | ComponentAndProps` is the missing piece that I was not thinking of. Thanks.

I'd actually like to take this approach a bit further with some deeper refactoring though, because it will improve performance be avoiding the need to re-parse the record summary string. I've opened #2037 to track that work. Can we merge this PR for now, and then use the approach in that ticket later?

pavish (approved):

PR #2095, opened 2022-12-15
Creating more helpful error messages in the case where an email is invalud

ayao451:
Fixes #2053 

**Technical details**

Feel free to update the error message to whatever you want! We also included testing!

**Screenshots**

Screen Shot 2022-12-14 at 7 40 32 PM

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #2098, opened 2022-12-15
Creating More Helpful Error Messages in the Case Where an Email is Invalid

### LINKED ISSUE #2053: Show field-specific errors when a CHECK constraint fails on POST/PATCH records
## Steps to reproduce

1. Set up a table with an email column that has a unique constraint and contains a couple records with values for the column.

1. PATCH one of the records (e.g. at `/api/db/v0/tables/1270/records/5/`), attempting to set its email value to be a duplicate. You can do this via the front end on the record page.

1. Observe a response like the following

    ```json5
    [
      {
        "code": 4208,
        "message": "The requested update violates a uniqueness constraint",
        "field": null,
        "detail": {
          "constraint": 2316,
          "constraint_columns": [5264],
          "original_details": "Key (\"Email\")=(foo@bar.org) already exists."
        },
        "stacktrace": [
          // omitted for brevity
        ]
      }
    ]
    ```

    This is **good** because the front end is able to display an error message _on the specific field causing the problem_.

1. PATCH the same record, this time attempting to se the email value to a value like "nope" that is not a valid email address.

1. Expect to see an error response structured like the one above, with a `detail` property that the front end can _somehow_ use to determine the relevant field causing the error.

1. Instead, observe an error response like the following

    ```json5
    [
      {
        "code": 4999,
        "message": "(psycopg2.errors.CheckViolation) value for domain mathesar_types.email violates check constraint \"email_check\"\n\n[SQL: UPDATE \"Library Management\".\"Patrons\" SET id=%(id)s, \"First Name\"=%(First Name)s, \"Last Name\"=%(Last Name)s, \"Email\"=%(Email)s WHERE \"Library Management\".\"Patrons\".id = %(id_1)s]\n[parameters: {'id': 5, 'First Name': 'Laura', 'Last Name': 'Sotos', 'Email': 'this is not an email address', 'id_1': '5'}]\n(Background on this error at: http://sqlalche.me/e/14/gkpj)",
        "field": null,
        "detail": null,
        "stacktrace": [
          // omitted for brevity
        ]
      }
    ]
    ```

    This leads to poor UX on the record page as show in this unwieldy error message:

    image

ayao451:
Related to #2053

PR for gracefully handling CheckViolations caused by patching/posting invalid email record(s) in a column with  `mathesar_types.email` type.

Checklist
- [x] My commit messages follow [best practices](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ayao451:
Hi @kgodey I have tried for a little while, but still cant figure out what this issue is! I would appreciate the help :) Thank you so much!!!

Screen Shot 2022-12-15 at 6 32 57 AM

dmos62:
Hey @ayao451, thanks for the PR!

The issue you've dropped a screenshot for is a problem that randomly occurs in our testing suite (sigh). Normally it's fixed by rerunning the failed Github CI task. I'll do that now.

ayao451:
@dmos62  Oh no!! Let me know if there is anything else I can do to fix it!!!

kgodey:
@ayao451 A few things here:
- Please update the pull request title to describe the issue that's being fixed. The title you had in #2095 is fine. 
- Please close #2095 if it's no longer something that needs review
- Please fix formatting on the checklist above and check the items that you've completed.

ayao451:
@kgodey Fixed!! Is there anything else we should do?

mathesar/api/serializers/records.py
```diff
@@ -48,6 +48,10 @@ def update(self, instance, validated_data):
                     table=table,
                     status_code=status.HTTP_400_BAD_REQUEST,
                 )
+            elif "CheckViolation" in str(type(e.orig)):
```
silentninja, line 51: `CheckViolation` is a type of `IntegrityError` similar to `UniqueViolation` and is not related to the `TypeCast` Exception. So you would need to create a new Error class for it and add the logic for parsing the error message inside that class, similar to the UniqueViolationAPIException https://github.com/centerofci/mathesar/blob/632793a4df36c551ab042ebd5d14d12d3b2d0bdc/mathesar/api/exceptions/database_exceptions/exceptions.py#L20

Also, make sure to create a new Error code enum for invalid email exception in https://github.com/centerofci/mathesar/blob/632793a4df36c551ab042ebd5d14d12d3b2d0bdc/mathesar/api/exceptions/error_codes.py#L5 and use it as the error code for the email exception.

silentninja, line 51: It would be better to compare with the actual exception type rather than the string of the exception class. Something like `type(e.orig) == CheckViolation`

silentninja (changes requested):
Thanks for the PR @ayao451. I have also requested for few changes in specific comments.  

Also, Can you add a test case to verify if the exception is thrown correctly? You can use the below test case as a reference https://github.com/centerofci/mathesar/blob/632793a4df36c551ab042ebd5d14d12d3b2d0bdc/mathesar/tests/api/test_record_api.py#L1308

mathemancer:
Reassigning to @Anish9901 to get this merged. @Anish9901 If you think this is going to take any significant time to get into mergable state, please let me know and put it on the back burner. This is not high enough priority to focus on for more than a few hours right now.

mathesar/tests/api/test_record_api.py
```diff
@@ -1277,6 +1277,42 @@ def test_number_input_api_validation(empty_nasa_table, client):
...
+    email_col_id = table.get_column_name_id_bidirectional_map()[column_name]
+    data = {
+        email_col_id: 'foobar'
+    }
+    response = client.post(f'/api/db/v0/tables/{table.id}/records/', data=data)
+    response_data = response.json()
+    assert response.status_code == 400
+    assert response_data[0]['code'] == ErrorCodes.CheckViolation.value
+    assert response_data[0]['message'] == 'The requested insert violates a check constraint'
+    assert response_data[0]['detail'] == 'value for domain mathesar_types.email violates check constraint \"email_check\"\n'
```
silentninja, line 1293: The error can be rephrased as 
```suggestion
    assert response_data[0]['detail'] == 'Value for the column with type  mathesar_types.email violates check constraint \"email_check\"\n'
```

silentninja (approved):
Thanks for fixing the PR @Anish9901, nice work. 

@seancolsen Note that it is not possible to get the column causing the error as expected. So the front end would have to display this error as a row-wise error. I let you decide if that is okay for the UX

mathesar/tests/api/test_record_api.py
Anish9901: Actually, this is the original error message generated by postgres that is being passed on to the API, should it be altered?

seancolsen (changes requested):
- So here's the error response in this PR currently:

    ```json
    [
      {
        "code": 4215,
        "message": "The requested update violates a check constraint",
        "field": null,
        "detail": "value for domain mathesar_types.email violates check constraint \"email_check\"\n"
      }
    ]
    ```

- I don't know about this. It definitely does not fix the issue described in the ticket because it does not provide a _field-specific_ error.

- If we merge this PR as-is, then it actually makes the UX _worse_:

    image

    Before this PR, the error message was very ugly, but at least it mentioned "email", leading the user in the right direction.
    
    When the `detail` property of the error response is not a well-formed object, the front end relies on the `message` property. In theory we could alter that logic so that the `detail` property displays when it is a string. But I'm hesitant to do that because I'm not sure what other error messages potentially get worse with that change. We could concatenate the two, but it just feels a bit messy.

- I would be okay with merging this PR if we:

    1. Change the error response to:

        ```json
        [
          {
            "code": 4215,
            "message": "value for domain mathesar_types.email violates check constraint \"email_check\"\n",
            "field": null,
            "detail": null
          }
        ]
        ```

        This way, the user at least sees a bit more information in the error message to lead them in the right direction towards fixing their problem.

    1. Remove `Fixes #2053` from the PR description (leaving the ticket open). At some point in the future (low priority), we can continue discussion in the ticket about more creative ways to get a field-specific error message in this situation.
    
        For example, I'm wondering if it would be possible to alter all our `CHECK CONSTRAINT` definitions so that they raise [custom PostgreSQL errors](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) with strings that we could reliably parse into structured error codes within the service layer. That way we'd at least have structured information about _which check constraint_ failed. Then, in many cases, we could use that parsed error data with our knowledge of the column types to produce field-specific errors. It's a lot of work, so obviously not something to even continue researching until later. But I'd rather keep the ticket open because it does describe a meaningful UX problem which would still be unsolved after this PR.

seancolsen (approved):
This looks good now

seancolsen:
@Anish9901 I see in 39a2e62c5fd84aa1a6f34e56d5c07d75fc216392 you committed some changes to `mathesar_ui/package-lock.json`. I reverted them in 7abc3598711072e00dd029a043a61923b0c44343. You probably know this already, but we shouldn't commit changes to that file unless we're making changes to front end packages. I know it's a pain that the container sometimes modifies that file. Just try to remember to reset any changes made by the container so you don't accidentally commit them.

Anish9901:
@seancolsen Yeah my bad, I didn't mean to add `package-lock.json` in my commit.

Issue #2105, opened 2022-12-15
Pre-release visual consistency updates


kgodey:
We have noticed some inconsistencies between the Figma design guidelines and the implementation in our project. In particular, we need to fix the following issues:

- [x] Buttons
  - [x] Ensure use of primary, secondary and tertiary buttons is consistent
  - [x] Ensuring that all buttons have styles for all states
  - [x] Ensure consistent styling of button text (font size, font weight, text color)
- [x] Forms
  - [x] Form field labels are consistently styled
  - [ ] All form fields have consistent styles (includes styles for the default state, hover state, focus state, and error state)
- [x] Navigation
  - [ ] Consistently styling active and hover states for navigation menu items 
- [x] Headers
  - [x] Ensure consistent styling of all headers (font size, font weight, text color, background color)
  - [x] Apply consistent padding and margin to headers
- [x] Modals
  - [x] Verify that the background color for the modal is applied consistently throughout the application
  - [x] Confirm that the size of the modal is appropriate for the content it contains
  - [x] Check that all actions within the modal (such as buttons or links) are styled correctly and consistently
- [x] Alerts
  - [x] Ensure warnings, tips and system suggestions are used consistently
- [x] Tabs
  - [x] Ensure consistent styling of all tabs
  - [x] Ensure styles exist for all tab states (default, active, disabled, hover)
- [x] Toast Messages
  - [x] Colors match the design guidelines for each state (info, warning, error)
  - [x] Typography size and spacing is consistent
- [x] Typography
  - [x] Set variables for all font sizes
  - [x] Ensure that font sizes are consistently applied and that different font sizes are used appropriately
- [x] Layout
  - [x] Verify that the padding and widths of elements within the layout are consistent with the design
  - [x] Confirm that the layout looks good and functions correctly on different screen sizes and devices
- [x] Clean Up
  - [x] CSS variables - as per [Sean's suggestion](https://github.com/centerofci/mathesar/issues/2105#issuecomment-1358244320)
- [x] Data Import Preview
  - [x] Ensure that styles are applied consistently
  - [x] Ensure that layout is optimal for interacting with the table and form elements

seancolsen:
This list looks good to me.

One more idea:

- Look for CSS colors which are defined outside of CSS variables, and move them into CSS variables where possible.

    Here's a quick stab at a regex that finds offending CSS

    ```
[...]

PR #2107, opened 2022-12-15
Import styling

### LINKED ISSUE #1644: Update Data Import page based on new design
Blocked by #1643

pavish:
Fixes #1644 

**This PR:**
* Updates styling for import upload and preview pages

**Note:**
* The preview page styling does not match figma and contains improvisations which were discussed with @kgodey and @ghislaineguerin. The styling is not final and Ghislaine will be taking a look at pages with mixed layout content, which is out-of-scope for this PR.
* Toast alerts are not styled in this PR, as they seem to be tracked in https://github.com/centerofci/mathesar/issues/2105
* I **did not implement alerting the user while they try to close or move to another page**. This will require having to modify the router internals. I'll track this in a separate issue, and place it in post-release improvements. - Opened https://github.com/centerofci/mathesar/issues/2108 cc @kgodey 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
- Wow, this is a pretty huge improvement to the import process! Hats off!

- This is awesome! 👇

    image

- I'm happy you found `InsetPageLayout` and were able to utilize it.

- In f283e2c03edddf2e1102dc9d07c2319802da96f1 I fixed the way we're setting the color of the radio.

    My design of this component is such that the source of truth for the radio color is in HSL color space. Consuming components can customize the radio color via CSS variables like `--radio-hue`. The reason for this design is that I'm accounting for the radio being: (a) checked/unchecked, (b) hovered/not-hovered, and (c) enabled/disabled -- and I'm altering the radio color based on all those conditions. I have CSS like this which alters the color when the radio is disabled.

    ```scss
    &:disabled {
      --lightness: calc(var(--base-lightness) + 25%);
    }
    ```

    That sort of logic only works when the source of truth for the initial color is HSL.

    Your changes had overridden my HSL color definition with an RGB definition.

    I fixed this by taking that RGB value you supplied and converting it to HSL.

    My change might seem unimportant, given that the radio effectively only has hue when it's checked (and it's very unlikely to be checked and disabled at the same time) -- but note that this color logic is shared with the checkbox too, which does have hue when it's indeterminate and/or disabled. (Note the code comments atop the CSS, explaining the duplicated code.) I made my changes such that the color is kept consistent across radios and checkboxes.

- With a narrow viewport, we have some overflow here that would be good to fix:

    image

    You can either fix before merging or create a ticket.

- I'm disappointed with how much global CSS there is in `ImportUploadPage`. I see global CSS as an escape-hatch that should be used only as a last resort. In `ImportUploadPage` you're using it for typography and layout which should all be easy to do with scoped CSS.

- Somewhat related to the above point... In my styling work, I've been really trying to avoid putting lots of typography-related CSS into specific pages just to make the pages match Figma. I would rather we defer that work until we do some consistency styling across the entire app. The CSS in `ImportUploadPage` is doing things like setting the font size of the page title and setting the spacing of radio options, etc. I don't think we should setting those kinds of styles on a per-page basis. If we do set them on a per-page basis, then it will actually make the consistency work _harder_ because to achieve true consistency, Ghislaine will need to go track down all the page-specific CSS and remove it. I would rather we remove much of the CSS in `ImportUploadPage`, leaving the page looking more more the rest of the app and less like Figma. This step will also remove much of the global CSS, solving two problems at once.

- Also, I notice that some of the global CSS in `ImportUploadPage` is due to the following problem...

    You're setting `--spacing-x: 0.5em;`, presumably to target the `--spacing-x` value in `LabeledInput.scss`, increasing it from `0.3em`. But, it turns out that `--spacing-x` is (perhaps inadvertently) also targeting the value in `FieldsetGroup.scss`. So I'm guessing that's why you had to add more global CSS, inserting `margin-right: 0.5rem;` for each radio option.

    This problem is exactly why I recently proposed that we follow a pattern of using CSS variables like `--LabeledInput__spacing-x` and `--FieldsetGroup__spacing-x`.

    But just to be clear, in this case I'd rather solve the problem by setting the style at a lower-level anyway (so that it applies the same everywhere). In this case that would mean changing `--spacing-x-default` to  `0.5em` within `LabeledInput.scss`.

- `ImportUploadPage` and `ImportPreviewPage` are using `<h2>` for their page titles. Why is this not `<h1>`? To the points above, here I would expect to see an `<h1>` element, with no page-specific CSS applied.

- I opened the following tickets to track additional work that can be handled later:

    - #2110

Overall, I'm not sure whether it's best to make some of the above improvements in this PR or in future PRs. The styling simplifications I'm requesting _would_ help Ghislaine implement her consistency work, so I think it would be good to do soon, or at least make _her_ aware of the need to do it. I'll leave it to you to decide whether to merge this PR as-is or whether to do a little clean up before merging.

pavish:
@seancolsen 

> I fixed the way we're setting the color of the radio.

I have reservations of using hsl for radios & checkboxes. I agree with the reasoning, but I'd very much have a simpler radio & checkbox that is styled via theming options using colors directly, the same way we do for buttons and other components.

But it's a conversation to have later.

> With a narrow viewport, we have some overflow here that would be good to fix:

Thanks, I'll fix this.

> I'm disappointed with how much global CSS there is in ImportUploadPage. I see global CSS as an escape-hatch that should be used only as a last resort. In ImportUploadPage you're using it for typography and layout which should all be easy to do with scoped CSS.

I disagree with avoiding global styles. I understand scoping styles but I hate adding more elements to the DOM just to scope styles. Everywhere it required adding extra elements just for scoping, I've added global css instead. I think this requires more discussions.

> I've been really trying to avoid putting lots of typography-related CSS into specific pages just to make the pages match Figma. I would rather we defer that work until we do some consistency styling across the entire app

I agree with this point. The approach I've been following is to match figma, and defer this to the consistency work later.

> The CSS in ImportUploadPage is doing things like setting the font size of the page title and setting the spacing of radio options, etc. I don't think we should be setting those kinds of styles on a per-page basis. If we do set them on a per-page basis, then it will actually make the consistency work harder because to achieve true consistency, Ghislaine will need to go track down all the page-specific CSS and remove it. 

I added them here because this particular page is of a very different layout compared to all other pages. I think it'll be easier for Ghislaine to move the styles from here to a common place than styling them from scratch, which would make work faster not slower.

> I would rather we remove much of the CSS in ImportUploadPage, leaving the page looking more more the rest of the app and less like Figma. This step will also remove much of the global CSS, solving two problems at once.

I would like the page to look like Figma so that even if we do not get through consistency fixes on time, we still can release Mathesar without having to worry about a pending styling task.

> This problem is exactly why I recently proposed that we follow a pattern of using CSS variables like --LabeledInput__spacing-x and --FieldsetGroup__spacing-x.

I agree with this. We can make the changes here.

> ImportUploadPage and ImportPreviewPage are using <h2> for their page titles. Why is this not <h1>? To the points above, here I would expect to see an <h1> element, with no page-specific CSS applied.

I planned that we'd have global styling for h1 to h6 elements in either our App, or part of the component library, and h1 would be used in pages like 'Schema' or 'Database' where the font size of the header is bigger than the font size of headers in pages like Import, where we'd use h2.

seancolsen:
A few responses

- > I have reservations of using hsl for radios & checkboxes

    I added this to our [meeting agenda](https://hackmd.io/ctEvo2tzQTyFV_38noSSEQ?edit).

- > I disagree with avoiding global styles. 

    I added this to our [meeting agenda](https://hackmd.io/ctEvo2tzQTyFV_38noSSEQ?edit).

-  > h1 would be used in pages like 'Schema' or 'Database' where the font size of the header is bigger than the font size of headers in pages like Import, where we'd use h2.

    It sounds like you're saying you'd like to use h1 when you want a large font and use h2 when you want a smaller font. Is that what you're saying? If so, I'm very surprised by this opinion.

    The h1 and h2 elements have different _semantic meaning_. Changing the element just because you want a different font size does not make sense to me. Granted, the importance of adhering to these semantic rules is diminished for an app like ours. If we were building a content-focused site where SEO was important, then I'd be adamant about keeping _all_ our headings properly structured. With Mathesar I think we ought to follow best practices at least for the h1 element, giving every page one and only one h1 element.

pavish:
@seancolsen 

> The h1 and h2 elements have different semantic meaning. Changing the element just because you want a different font size does not make sense to me. Granted, the importance of adhering to these semantic rules is diminished for an app like ours. If we were building a content-focused site where SEO was important, then I'd be adamant about keeping all our headings properly structured. With Mathesar I think we ought to follow best practices at least for the h1 element, giving every page one and only one h1 element.

Okay, I agree. This completely slipped my mind, my comment does not make sense. I need to take some breaks I guess :)

pavish:
@seancolsen 

I've added https://github.com/centerofci/mathesar/pull/2107/commits/516b20826d2d523ad5f20eb60e4cd4ef42858299 which fixes the overflow issue. I've not made the component name scoping for spacing variables, I saw no need to use them here, so I removed them.

I've added https://github.com/centerofci/mathesar/pull/2107/commits/eb50b646c68a77ec76348be7a9a8a1d1a9345150 to fix the headers.

I'm going to go ahead and merge this PR since I already have your approval. Please feel free to review these commits post merge.

PR #2118, opened 2022-12-19
Click & Drag column headers to select multiple columns

### LINKED ISSUE #1622: Change behavior of cell selection via row/column headers
The cell selection functionality, when performed via the row headers or column headers doesn't match my expected behavior. Our [specs](https://github.com/centerofci/mathesar-wiki/blob/master/design/specs/table-inspector.md) aren't super clear on these details, so I'm using this ticket to get further into the weeds by describing my expected behavior.

I want the cell selection to work the way that LibreOffice works because that seems simple and effective. Some of the UX subtleties are difficult to describe, so if you're confused about what I'm asking for, play with LibreOffice to see how it works. I find it helpful to use LibreOffice as a working prototype.

Below I'll describe the changes I'd like in terms of columns and column headers -- but these changes should apply to rows and row headers too.

## Specific changes

- When all cells in the column are selected, clicking on the column header (or dragging from and to the same column header) resets the selection to zero cells. Instead, this should begin a new selection, selecting all the currently selected cells (effectively retaining the currently selected cells).

- When I drag to select multiple columns, currently no changes are made to the selection. Instead, this should begin a new selection, selecting the cells in all columns between (and including) the starting column and the ending column.

### LINKED ISSUE #1768: Cell selection should begin on mousedown
## Reproduce

1. Select a single table cell via click.
1. Within the Table Inspector, open the Column panel.
1. Observe UI to configure the column.
1. Press mouse down on the selected table cell.
1. Expect to continue seeing the same UI to configure the column.
1. Instead, observe the Column panel to contain a message indicating that no column is selected. This is because the set of selected cells was reset to be empty on mouse down, but it does not receive new entries until the user either releases the mouse or drags the selection into a new cell.

rajatvijay:
Fixes #1768 
Fixes #1622 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
This works pretty well. 

- I opened #2122 for a subtle edge-case that I think we can handle later.

- Also, I notice a bunch of new methods like `onColumnSelectionStart` in `SheetSelection`. Just a small nitpick... When I see a name like `onColumnSelectionStart`, I expect that the function will accept a callback. I think these functions would be more aptly named `handleColumnSelectionStart` (and similar). Or, better yet, name each function according to what it _does_ rather than how it's used, because how it's used may well change in the future, even if the behavior of the method does not change -- however, I recognize that sort of naming might be difficult in this case due to the complexity here. Not worth making any changes in this PR. Just a thought for the future.

seancolsen:
I found another regression that this PR seems to have introduced: #2130

PR #2165, opened 2022-12-22
Default display_options for inferred/created mathesar_money column

### LINKED ISSUE #2153: Money formatting should start off with correct decimal places
By default, money columns should have 2 decimal places in the display options.

Currently, when you import the money in the demo video, there are values like `$15.6` – by default, that should be `$15.60`.

Anish9901:
Fixes #2153

Set default display_options for newly inferred/created mathesar_money column.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
The PR works as intended but I am not sure if this is the right way to implement it as it could lead to unexpected behaviour. The reasons are
1. The display option field is set to `None` by default when a column is created. This means that the user has not created any display option and should default to whatever options the frontend considers as default. So this issue should be handled by the frontend just like how the `currency_symbol` is automatically considered as `$` when the `display_option` is `None`
2. We shouldn't be saving in the `to_representation` method, it is only for representing the value from the database, this is an unintended behaviour according to the framework and can lead to debugging nightmares. Instead, We should be setting up default display options object for a column on the backend instead of `None` if we don't want the frontend to handle defaults

@kgodey We can merge it for the demo video and look for an alternate solution later but I recommend considering reason 1 and have the frontend fix this

kgodey:
This is a nice to have for the demo video, so we don't need to merge it. I agree that saving in the `to_representation` field seems like an anti-pattern.

I don't think this should be the frontend's responsibility, though. I think we should have default display options in the backend when a money column is created (maybe in the model save method).

silentninja:
> I don't think this should be the frontend's responsibility, though.

The reason we have it as a frontend responsibility now is that the backend is not aware of the value of certain `display_options`. For example, the currency symbol can be known using the browser's locale by the frontend which the backend does not have access to. 
Moreover, this would mean we need to make all the display options optional and provide defaults as it would be idiomatic rather than having something ad-hoc for the Money field alone. 

But I don't have a strong opinion and can see advantages in both the options

kgodey:
> Moreover, this would mean we need to make all the display options optional and provide defaults as it would be idiomatic rather than having something ad-hoc for the Money field alone.

I think this makes sense.

> the backend is not aware of the value of certain display_options

The frontend should be able to know that the backend is picking default display options and override them if needed.

---

We can pick this back up after the holidays, there's too much architectural work involved to get this merged in by tomorrow.

Anish9901:
> We should be setting up default display options object for a column on the backend instead of `None`.

@silentninja Are you suggesting having a new default_display_options field for the column model or having default display options set for every column instead of `None`? 

Do you think its a good idea have a default_display_options map for each db_type and the frontend can just show a toggle that can be set to default_type_options or display_options defined by users?

silentninja:
> Are you suggesting having a new default_display_options field for the column model or having default display options set for every column instead of None?

- We should be "having default display options set for every column instead of None". 
- Alternatively, We could also have a map of default display_options and use the default value if the display_options value is `None`. This is very similar to the PR, except we don't save it instead replace the value during the representation. So we would always know that the user hasn't set any display options. 

My opinion would be to have the front end deal with setting the defaults as we frontend client would have better knowledge of the user preference(from browser defaults, locale etc). But I would like to hear others opinion on this.

> Do you think its a good idea have a default_display_options map for each db_type and the frontend can just show a toggle that can be set to default_type_options or display_options defined by users?

So the toggle will reset the display options to the default value? That would be a useful feature too. Also, how does this fit into the current discussion.

Anish9901:
> * We should be "having default display options set for every column instead of None".
> * Alternatively, We could also have a map of default display_options and use the default value if the display_options value is `None`. This is very similar to the PR, except we don't save it instead replace the value during the representation. So we would always know that the user hasn't set any display options.
> 
> My opinion would be to have the front end deal with setting the defaults as we frontend client would have better knowledge of the user preference(from browser defaults, locale etc). But I would like to hear others opinion on this.
> 
>>

Makes sense.

> So the toggle will reset the display options to the default value? That would be a useful feature too. Also, how does this fit into the current discussion.

Just an idea I thought would be useful.

kgodey:
What are the next steps for this PR?

silentninja:
@Anish9901 After assigning a default display_option in the `to_representation` please remove the statement to save it to the database. This would make sure the display_option is a default placeholder value and not the actual value saved by the user. 

This should be good enough for now. We can deal with a proper infrastructure for defaults in future when the need arises.

silentninja (approved):
Looks good, thanks @Anish9901!

PR #2172, opened 2022-12-23
Persist tab name in url for schema page

### LINKED ISSUE #1922: Persist schema page's tab name in the url

rajatvijay:
Fixes #1922

**Screenshots**
https://www.loom.com/share/e76803ee521c46d99436a32319609dbf

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/pages/schema/SchemaPage.svelte
```diff
@@ -87,7 +99,12 @@
     </slot>
   </AppSecondaryHeader>
 
-  <TabContainer bind:activeTab {tabs} uniformTabWidth={false}>
+  <TabContainer
+    {activeTab}
+    {tabs}
+    uniformTabWidth={false}
+    on:tabSelected={handleTabSelected}
+  >
```
pavish, line 107: The tabs here should not be buttons with an on click event navigating to a route. They should be anchors, so that users can also right click and open them from in a different browser window/tab.

pavish (changes requested):

mathesar_ui/src/component-library/tabs/TabContainer.svelte
```diff
@@ -10,23 +10,27 @@
...
   export let fillTabWidth = false;
   export let fillContainerHeight = false;
   export let uniformTabWidth = true;
 
   function selectActiveTab(e: Event, tab: Tab) {
-    activeTab = tab;
+    const hasLink = !!tab[linkKey];
+    if (!hasLink) {
+      activeTab = tab;
+    }
```
pavish, line 33: Why is active tab not set here for tabs with links?

pavish: This component is common and this property needs to be handled by the component. Binding to the property from the parent is optional, so it's not a concern if it gets changed internally in this scenario since we control the property on the parent.

pavish (changes requested):
@rajatvijay Looks good!

I do have a comment I'd like to be resolved before merging this in.

mathesar_ui/src/component-library/tabs/TabContainer.svelte
rajatvijay: @pavish 

This is done to avoid the state mess up created by the two data binding when the user does cmd+click on the tabs with links. 

Here is a video showcasing this: https://www.loom.com/share/e2be4607c15847beac3997c5c43df58c

pavish: @rajatvijay The solution made above is a workaround. The proper way to handle this is to select the tab only on the click event (not on the mouse down event), and to check if 'cmd/ctrl' is pressed down when click event is raised and if the tab is a link, and do nothing if it's a cmd+click and leave it to the browser to take over. I believe this is small enough to handle it in this PR itself.

pavish (changes requested):

pavish (approved):
@rajatvijay I went through the tab component and I'd very much like to deprecate this and write a new one altogether. I'd prefer the new component to be more headless like the ListBox. We can queue this for later, after our release.

For now, I'm gonna go ahead and merge this in, since a new component would lead to rewrite this logic anyway.

PR #2182, opened 2023-01-03
Have CI pytest task stop after first failed test

dmos62:
Right now, when Github CI is running our test suite and one of the tests fails, the CI will continue to run the remaining tests, which can take a long time, and we'll only know that a test failed after the CI task finishes.

This PR adds `--exitfirst` to our CI's pytest call, so that if a test fails, pytest stops immediately, thus shortening the CI task runtime and giving us feedback sooner.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -25,7 +25,7 @@ jobs:
         run: docker exec mathesar_service mkdir coverage_report
 
       - name: Run tests with pytest
-        run: docker exec mathesar_service pytest  --ignore=mathesar/tests/integration/ --cov-report=xml:coverage_report/coverage.xml
+        run: docker exec mathesar_service pytest --exitfirst --ignore=mathesar/tests/integration/ --cov-report=xml:coverage_report/coverage.xml
 
       - name: Save the coverage report
         uses: codecov/codecov-action@v2


silentninja (changes requested):
I don't know how useful this would be. There is also a scenario where multiple test cases could fail and we would have to run the test suite again to find the failing test. If the failing test is at the bottom of the list, this solution would make the feedback time longer than our current feedback time.

Is there a reason why you created this PR?

dmos62:
The specific situation that inspired this was a bug that was only reproducing itself in the CI, but not locally. I had to wait a long time for the CI to tell me if it failed (when it could have done so much faster, if `--exitfirst` was enabled), and I had to do it multiple times. I.e. iteration was very slow. Also, sometimes I don't run the full test suite on my machine before pushing, and am later surprised that the CI test run fails: I would like this feedback loop to be shorter.

I don't see a downside to having CI's test run fail early, because we don't really use the CI test run for debugging, but rather as a safeguard. If I have the CI failing, I might look at its log to tell me what kind of failure that is, but then I'll use my local environment for actual debugging, which is an order of magnitude faster than our CI runner.

@kgodey can you arbitrage?

kgodey:
It sounds like this solves a problem for @dmos62 but @silentninja, you're objecting to this because it _might_ cause other problems. Since this is a one line change, it seems like it's worth trying. It can be easily reverted if it causes more problems than it solves.

silentninja (approved):

silentninja:
> I don't see a downside to having CI's test run fail early, because we don't really use the CI test run for debugging, but rather as a safeguard.

Okay, this sounds reasonable. I am on the same page with this too. The only concern I have is that it would make the feedback longer in case of multiple test case failures, especially with external contributors. But for now, it shouldn't be a problem.

PR #2213, opened 2023-01-06
Reorder columns frontend

### LINKED ISSUE #1584: Use table-level display options to allow users to re-order columns
## Current behavior

- Columns are displayed within the Table Page and within the embedded Table Widget on the Record Page, and probably other places too.
- Columns are always ordered by their immutable Postgres attnum value (i.e. the order in which they were created). The user has no way to display newly-added columns prior to previously added columns.
- Some users may wish to modify the order in which columns are displayed.

## Desired behavior

- Within the Table Page, users can drag columns to reorder them.
- Within the table widgets (displayed on the Record Page), users can _not_ modify the column ordering (at least for the time being). The rationale here is that, because we intend to hide the linking PK column, it's possible the a reordering operation may be ambiguous without all columns present.

## UX

- The UX should be identical to Google Sheets -- you need to select the column before dragging it. When no cells are selected, initiating a drag event via the column header will begin selecting cells in multiple columns, allowing the user to select all cells in adjacent columns via dragging. When all cells within a particular column are selected, then that column's header becomes a target to begin a drag event which re-orders the column. The cursor will provide visual feedback as to the differences in behavior before dragging. This is a little hard to explain. Just try it on Google Sheets and you'll see how it works.

## Implementation

- The user-defined order is persisted via a table-level `display_options` property, `colum_order`, which is an array of `attnum` values.
- If the user has not customized the column order, then `colum_order` will be `null`.
- The order of columns (or cells) within API responses is undefined. Columns and cells can be returned in whatever order is convenient. The front end is responsible for sorting columns and cells into their appropriate order.
- Any time we display the columns within a table, we first display all the columns as listed in `column_order` (keeping in-mind that `column_order` may contain values for deleted columns which need to be filtered out), then we display any remaining columns, as sorted by their `attnum` values.
- When a user re-orders columns, the front end submits one PATCH request to update `colum_order`, providing the full array containing all `attnum` values.
- The above approach allows us to keep the columns metadata in Django thin and decoupled from Postgres. The computed order is deterministic and able to withstand DDL changes made outside of Mathesar without the need for reflection.
- The `colum_order` display option does not affect the selection of a column for the Record Summary. If a user wishes to change the column automatically selected for the Record Summary, they can do so by providing a custom summary template (not by re-ordering the columns).
- The `column_order` display option is not exposed directly via any UI, beyond the order in which the columns display.
- The `column_order` customization should apply identically for all users.

## Task scheduling

- This feature requires back end work and front end work. That work can be taken up in either order. If you want to implement the front end first, then begin by allowing the user to reorder columns without that order persisting via the API.

AlekLefebvre:
Fixes #1584

Allows the user to drag and drop headers of selected columns to re-order them.

**Technical details**
In addition to implementation detailed in the issue, this PR also modifies the SheetSelection to select columns based on IDs instead of indexes to account for the custom column order.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


rajatvijay:
I was trying to run this on my local just out of curiosity and found that the TablePage is broken due to the changes in this PR. I am attaching a SS of the error which is being displayed in the console. 

@AlekLefebvre you might want to look at this before the further review starts. 
Screenshot 2023-01-10 at 4 47 41 PM

AlekLefebvre:
Thanks @rajatvijay! It's fixed.

mathesar_ui/src/stores/table-data/tabularData.ts
```diff
@@ -102,8 +105,12 @@ export class TabularData {
         ),
     );
 
+    const table = get(currentTable) as TableEntry;
```
seancolsen, line 108: Relying on the `currentTable` store is problematic because it assumes that we'll never have an instance of `TabularData` for a table other than `currentTable`. One place where that assumption does not hold is the Record Page where we have one `TabularData` instance for the current table, plus any number of other instances for related tables too. The Record Selector also instantiates `TabularData` from the Schema Page when `$currentTable` is `undefined`.

To be fair, I would say that our current design of relying on this imported `currentTable` store (and others like it) is a deeper root cause of this problem because it leads people like you astray. We should aim for code that doesn't lead to these false assumptions. I'd rather we use imported stores only for things like "modal" and "toast" where we really _do_ expect a singleton (but that's a topic for another day).

For this PR, I think we should modify `TabularDataProps`, adding a new property like `table: TableEntry`. Then we would instantiate `TabularData` with a `TableEntry` everywhere. Searching our codebase for `new TabularData` seems to indicate that we're only instantiating `TabularData` four times, and in three of those cases we already have a `TableEntry` in scope. The exception is in `RecordSelectorWindow.svelte`, where it seems we could easily derive a `TableEntry` object from `tableId` (in scope) and the `tables` store (imported from `src/stores/tables.ts`).

@pavish How does the above paragraph sound to you. We would be making `TabularData` dependent on having a `TableEntry`. This would also allow us to persist the `TableEntry` within the `TabularData` instance, allowing us to see the table's name, for example. I have a vague memory that for some reason many months ago you didn't want to do this. If we don't want to require an entire `TableEntry` object, then we could pass in the minimum data needed here which would mean adding a property like `columnOrder: number[]` to `TabularDataProps`.

seancolsen (changes requested):
@AlekLefebvre 

Sorry it's taken me a while to review. We're in the midst of pushing pretty hard towards our first release.

Thanks for your work on this! I can tell you've put in some time, and I'm impressed with how far you've come and how well you're getting acclimated to our codebase.

Before we merge this PR, I'd like to polish this feature a bit more and address some regressions to current functionality. I don't want to strive for perfection here -- and I'm cautious about delaying merge for too long, lest we incur frustrating merge conflicts -- but I noticed a number of subtle UX shortcomings that I think would confuse users if left this feature as implemented here.

## Code issues which I fixed by pushing commits

- In 2a23aeefc022d5d8a06d4179892e681edb8b8f63 I made a small adjustment to a variable name for the sake fo our code standards.

## Code issues which I'd like you to fix

- (See one inline [comment](https://github.com/centerofci/mathesar/pull/2213#discussion_r1081799469))

## UX issues which I'd like you to fix

- I should be able to resize a selected column without dragging it.

- Ordering does not match my expectations when moving a column to the right. Say I begin with columns `A`, `B`, and `C`. Then I move `A` to the position between `B` and `C`. I expect to see columns ordered `B`, `A`, `C`. But instead I see columns ordered `B`, `C`, `A`. There appears to be an off-by-one error in here somewhere. Moving a column to the left does not have this problem.

- Column placement behavior should be changed when moving a column to the left.

    The behavior in this PR is:
    
    > While I'm dragging, hovering on a column header indicates that the dragged column will be placed _after_ the hovered column.

    I'd like to change this work as follows:

    > While I'm dragging to the right, hovering on a column header indicates that the dragged column will be placed _after_ the hovered column. While I'm dragging to the left (or over the dragged column's current position), hovering on a column header indicates that the dragged column will be placed _before_ the hovered column.

    That's how Google Sheets works and I think it's a little clearer.

- If I drag a column and drop it into its current position, then no changes should be made (i.e. no API request.)

- The drop target indicator...

    - ...needs to un-highlight after the drag event finishes. I'm observing that it consistently gets stuck in highlight mode.

    - ...should should match the styling of the column resize indicator (bigger and different color).

- When a column contains only one selected cell, I should be able to select all the cells in that column by clicking on the column header. Before this PR, that behavior worked. With the code in this PR, clicking on the column header of a column with only one selected cell does not appear to select any additional cells.

- The drag-start region...
    - ...needs to cover the entire column header cell. In this PR that area appears to only cover the text within the column header. I should be able to initiate a drag event from empty space to the right of the column name
    - ...should have `cursor: drag;` CSS applied to provide a cue the user that they can begin dragging.

- I shouldn't be able to move the column unless _all_ the cells in the column are selected. In this PR, I can drag columns which contain _some_ selected cells.


## Additional UX work that I'd like to handle later via tickets

If you're motivated, you can try tackling these items within this PR. But it's probably best to keep scope tight in this PR and push these off until later. I can create tickets for these items after merging this PR

- Improve visual feedback when dragging to re-order columns. I'd like it to look more like Google Sheets.

- Sort columns when they appear in other places too, e.g. Record Selector, Record Page, column "Select" and "MultiSelect" inputs (which occur in many places throughout the app), column selection pane of the Data Explorer. There are may be more places too.

- Allow dragging to sort columns within the Data Explorer

AlekLefebvre:
Thank you for the review! I'll get to work on these as soon as possible, it might take a week or two before I have time though.

mathesar_ui/src/stores/table-data/tabularData.ts
AlekLefebvre: I went ahead and implemented this solution. @pavish does it look ok to you? I can make changes to it if needed.

pavish: @AlekLefebvre It's going to take me a while to get my head into this. We're currently focused on our initial release which should be by the end of this month.

While there's chances of merge conflicts, erring on the side of caution w.r.t regressions, I'd like to wait until the initial release is done and tagged, before merging this PR in. I'll take a look at this sometime next week.

cc: @seancolsen

seancolsen (changes requested):
Thanks for following up with changes to address our review feedback, @AlekLefebvre. This is coming along nicely.

(I numbered my previous comments, for ease of reference.)

- It looks like you've addressed points 1, 2, 4, 5, 6, 7, 8, 10, and 11.

- I'm still observing problems **(3)** and **(9)**, exactly as I mentioned them in my previous review. Let me know if you need any clarification on the these problems.

Additionally, I found some more issues that seem to have stemmed from your follow-up changes:

- **(15)** The drag start region extends to the right of the resize handle

    There is a tiny bit of space (I think probably a 1px border) to the right of the resize handle which becomes a drag handle when the column is draggable. I think we should fix this so that it doesn't lead to unexpected drag events.

    Here's a video showing my cursor moving around at 500% zoom level to make this clear:

    https://user-images.githubusercontent.com/42411/221656821-141e99f1-7fad-458b-952d-b13e0a07a906.mp4

    I'd like to fix this so that there is no drag handle to the right of the resize handle. That border can be either be part of the resize handle or it can be dead space.


- **(16)** The cursor should not be `drag` when only some of the cells are selected. This is related to (11). You've fixed the column so that it can't be dragged. Now we need to fix the cursor so that it indicates that the column can't be dragged.

- **(17)** The column header should not move when showing the drop indicator. When dragging a column to the left, the column name of the drop target moves to the right a little bit when the drop target is hovered. It should not move. We can handle this via a subsequent ticket if necessary.

AlekLefebvre:
Hi @seancolsen,
I fixed #3 and 15-17. I haven't been able to replicate #9 this time, it's working for me and it also seems to be working in your video, the cursor changes to drag on the right side of the header. I'm able to initiate de drag from anywhere on the cell. Could you clarify what the issue is?
Thank you,

seancolsen (changes requested):
@AlekLefebvre (3), (15), (16), and (17) all look good!

- About **(9)**... After playing with this more, I'm seeing that I can only reproduce it in Firefox. I can't reproduce it in Chrome. Can you try reproducing it in Firefox? If you can identify a quick fix, feel free to incorporate it into this PR. Otherwise, I'll open a ticket for it after we merge.

I spotted a new issue that I'd like you to handle before we merge:

- **(18)**: When a column toggles between being selected and not being selected, its header label jumps around horizontally.

    https://user-images.githubusercontent.com/42411/223167396-f427200d-0c7f-428a-a09e-5600db99bf54.mp4

    In the above video, I see "First Name" and "Last Name" moving horizontally. Those labels should not move.

mathesar_ui/src/stores/table-data/tabularData.ts
pavish: I looked through the code and the approach looks good to me.

I agree that a store restructuring is definitely required, and perhaps also change the way we access them since `tables.data.get(someId)` can be `undefined` and a request to fetch it may not be in progress. For now, I think no changes in approach is required in this PR.

pavish (commented):
@AlekLefebvre I went through the code and it looks good to me. I added my comment regarding the approach of fetching tables in its respective comment thread: https://github.com/centerofci/mathesar/pull/2213#discussion_r1127383921. 

Great job on the effort you've put into this, I'm looking forward to getting this merged!

I've not tested it since @seancolsen's reviewing this primarily, I'll leave it to him to approve and merge after his comments are addressed.

AlekLefebvre:
Thank you @pavish and @seancolsen for your reviews.

(18) is fixed. (9) seems related to this Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=568313#c16 the only quick fix I can see is to replace the button with a div and keep the same classes. Please let me know if it's acceptable.

seancolsen (approved):
@AlekLefebvre This is looking good now. Very nice work on this PR, and thanks for your help in seeing a useful feature through from start to finish!

I've opened #2651 to track the remaining work with that weird FF-specific bug. Feel free to take it up if you like, but not pressure.

We'd love to keep you involved, so let us know if there are any other areas of Mathesar that you're interested in working on.

kgodey:
Really excited to see this merged. Nice work, @AlekLefebvre! Thanks for sticking through such a complex feature.

AlekLefebvre:
Yes, I'm interested in fixing those bugs. Let me know when master is merged back into develop and I'll get to work.

PR #2220, opened 2023-01-09
Separate load live demo data step

mathemancer:
Related to #1906 

This adds two endpoints:

```
POST /api/demo/v0/load_data/
GET /api/demo/v0/datasets_exist/
```

The first accepts an empty `POST` request, and loads demo data into the database. The second `GET`s a JSON of the following format:

```json
{
    "datasets_exist": <bool>
}
```
according to whether or not the expected demo schemas exist in the database. Note that it doesn't do anything except check for schemas of the expected names.

**Technical details**

- These endpoints only exist when Mathesar is running in live demo mode.
- These endpoints automatically use the DB associated with the user's session ID, so the UI doesn't need to submit any DB name info.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@kgodey @pavish I set up this PR so we can discuss what further steps are needed, and negotiate about that.

pavish:
@mathemancer
* I'd like 'datasets_exist' to be part of the 'common_data' provided to our templates. 
* I have a question (copied over from [this mail](https://groups.google.com/a/mathesar.org/g/mathesar-developers/c/TrisgdDuMuo/m/xWhjXFpAEgAJ)):
  * Scenario: `datasets_exist` is false, even though data is currently being loaded due to a previous request.
  * If the user clicks on 'Load data' (See scenario 2.2 in mail), how would the backend reset the data and reload again?
  * I assume we ignore the ongoing transaction, since we have no access to it. We create a new one, and associate it to the user session.
  * The stale data that gets created by the previous transaction gets removed by some other process.

demo/settings.py
```diff
@@ -6,6 +6,8 @@
     "demo.middleware.LiveDemoModeMiddleware",
 ]
 
+ROOT_URLCONF = "demo.urls"
```
kgodey, line 9: I'm glad you were able to avoid toggles in the main application. I think you might need to add a toggle for the `common_data` addition, though.

demo/install.py
```diff
@@ -14,6 +15,12 @@
 MOVIES_SQL_BZ2 = os.path.join(RESOURCES, "movie_collection.sql.bz2")
 
 
+def check_datasets(engine):
+    expect_schemas = [LIBRARY_MANAGEMENT, MOVIE_COLLECTION]
```
kgodey, line 19: I assume this will be augmented with new data sets later.

mathemancer: Exactly.

pavish:
@kgodey 

> How will users re-load data (e.g. if they've changed their schema and want to reset to the default sample data?)

The process would be similar to loading data the first time. Once they click 'Reload data', we'd set loading status to true and start re-loading.

The error that @mathemancer mentioned:
> Currently, a second request to the load_data endpoint for the same DB will return a 500, and the data will continue to load.

is not applicable when we have a dedicated loading status as mentioned in the [previous comment](https://github.com/centerofci/mathesar/pull/2220#issuecomment-1377649810).

demo/urls.py
```diff
@@ -0,0 +1,17 @@
...
+
+from config import urls as root_urls
+from demo import views
+
+
+urlpatterns = [
+    path('api/demo/v0/load_data/', views.load_data, name='load_data'),
+    path('api/demo/v0/datasets_exist/', views.data_exists, name='datasets_exist'),
+    path('<db_name>/', views.SchemasView.as_view(), name='schemas'),
+    re_path(
```
silentninja, line 11: This is a wildcard regex it would end up overriding all the URLs, so for any URL other than `api/demo`, `SchemaView` will be called. For example, calling `/auth/login/` will go to `SchemasView` instead of the Django login view. We would have to split the `SchemasView` regex in the mathesar URL into a separate path so that other apps can add extend the API paths without extending the `SchemasView` path

kgodey:
Since @mathemancer was able to get loading time down to sub-100ms, I don't think we need this functionality anymore. @mathemancer, do you just want to close this PR?

mathemancer:
Closing in favor of a different approach.

PR #2003, opened 2022-11-30
Allow the entire table card on the schema page to be clickable.

LFLATH:
I made a few changes in the HTML and CSS to allow the entire table card to be clickable.

Related to #1978 

These changes make the entire table card act as a link to the table.

Please let me know if anything I have done here it incorrect! I am still learning how this whole process works.

**Technical details**

I added a link tag around the header of the table card. This link goes to the same place as the Go to Table button. I then added some styles on hover. I also made a few changes to how the card is styled. I don't think these styles are correct. I added them more as a temporary fix to make this work.

**Screenshots**

image

## Checklist
 
- [ ] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [ ] My pull request targets the `master` branch of the repository
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

Issue #1042, opened 2022-02-03
Refactor CSS to avoid `!important` in component library


seancolsen:
For example, the `InputGroup` component introduces CSS which uses `!important`. I'm fine using `!important` in Mathesar CSS, but I think we should strictly avoid it within the component library since we are aiming for components which can be re-styled.

Abhishek140501:
Title
Refactor CSS to avoid !important in component library #1042

Description
This pull request aims to refactor the CSS code in the component library to remove the usage of !important declarations. The !important rule was causing specificity conflicts and making it difficult to override styles in certain scenarios. By refactoring the CSS, we can improve the maintainability and flexibility of the component library.

Changes Made
Reviewed the CSS codebase and identified all instances where !important was used.
Adjusted the CSS rules and selectors to remove the need for !important.
Tested the changes in different scenarios to ensure that the styles are still applied correctly without the use of !important.
Made necessary adjustments to maintain consistent styling across components.

Screenshot
![Screenshot (107)](https://github.com/centerofci/mathesar/assets/136616514/cb9bdf84-f12a-48e7-9bea-fbe91c81f84d)
![Screenshot (108)](https://github.com/centerofci/mathesar/assets/136616514/3716e74f-3f6d-4fa7-893c-03e4b4efb245)
![Screenshot (109)](https://github.com/centerofci/mathesar/assets/136616514/70e00626-9c9f-4fc7-9c3d-8bcef498d831)
![Screenshot (110)](https://github.com/centerofci/mathesar/assets/136616514/87990b64-ae1a-4035-af1e-11e1e49c9987)
![Screenshot (111)](https://github.com/centerofci/mathesar/assets/136616514/29d86e31-aac8-40f2-894d-1ea5501dc755)
![Screenshot (112)](https://github.com/centerofci/mathesar/assets/136616514/bf1a37ce-e065-4b55-b938-05194b0eace7)

Testing
I have tested the changes by incorporating them into an existing project that utilizes the component library. I verified the following:

Applied different styles to components and ensured that they were properly applied without the need for !important.
Checked for any unintended side effects or conflicts with existing styles.
Tested the components in different browsers and screen sizes to ensure cross-browser and responsive compatibility.
Related Issues
This addresses issue #1042.

Checklist
 The code follows the project's coding conventions and style guidelines.
 I have tested the changes and verified that they work as expected.
 The pull request has a descriptive title and includes all necessary information.
[...]

PR #2240, opened 2023-01-11
Make table cards clickable

### LINKED ISSUE #1978: On schema page, entire table card should be clickable
Currently, only the "Go to Table" area takes you to the table. But the entire card should be clickable and take you to the table (except for the "..." button and the "Find Record" button.

seancolsen:
Fixes #1978

## Before this PR

- Navigating to a table page required the user to click on the "Go to Table" link within the table card.

    image

    We want to be able to click on other areas of the card to access the hyperlink, like the table name.

- We had a PR #2003 from @LFLATH to fix this issue, but the PR had some problems, which stemmed from design shortcomings.

- @ghislaineguerin and I had a call where we discussed some design changes to the table card. She documented the [new design in Figma](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=4425%3A28850&t=piHbh9OqhGLLjRJ6-0).

- Using a horizontal ellipses for the menu trigger was a bit confusing when juxtaposed with the ellipses used for a truncated table name

    image

## Changes in this PR

- I created a new `HyperlinkCard` component.

    I'd like to eventually use it for the schema cards that we display on the Database Page too, but I didn't want to get too carried away with this PR.

    For consistency, I made the styling of the `HyperlinkCard` to mirror the styling used for the Schema Card.

- I modified the menu icon to use a vertical ellipses instead of a horizontal ellipses, and I made this change for the schema cards too.

- The entire table card can now be hovered

    image

- Then menu trigger extends to the full edges of the card, whereas previously the card had some padding which would have created a weird hyperlink click area between the menu trigger and the edge of the card.

    image

- I moved "Find a Record" into the menu.

    image

    @kgodey @rajatvijay and I had some discussion [in Matrix](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$1UtRqvDIJbgY_vgOQux8LGkZ3fKc1lWfyAY_-nh6fpg?via=matrix.mathesar.org&via=matrix.org) regarding discoverability of the Record Selector feature. The change in this PR _does_ make the Record Selector slightly less discoverable here, but I think that's okay.

- Truncated values are still visible on hover.

    image

    This was one of the problems with #2003 from @LFLATH. I had to take a bit of a fundamentally different approach here to fix that issue.

- I removed the `max-width` on the table cards. That was resulting in this minor anesthetic issue with spacing:

    image

    I think this spacing is better:

    image

## Issues

- @ghislaineguerin made this [Figma design](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=4425%3A28850&t=piHbh9OqhGLLjRJ6-0) which uses separate background colors for the card vs the page. 

    image

    I really like this design, but fully implementing it will require more far-reaching changes. For example, our `TabContainer` component requires that the tabs and the tab content both be placed in the same element, making it tricky to give the tabs the tan background shown above. I'd like to handle this later and leave it to @ghislaineguerin to prioritize/track.

- I don't love the styling when the dropdown trigger button is focused

    image

    This state results when you click on the menu trigger. With the menu open the focus indicator isn't super apparent. But if you click on the trigger again (closing the menu, while still keeping the trigger focused) then you enter the state in the screenshot above.

    This is tricky to change because of the usage of `!important` within our component library. We already have #1042 to remove usage of `!important`, so I'd rather leave this as-is for now. This is minor enough that I don't think it's worth tracking via a separate issue. I just wanted to mention it in case someone noticed it during review.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@kgodey this PR contains some design changes that you might like to look over.

mathesar_ui/src/components/HyperlinkCard.svelte
```diff
@@ -0,0 +1,106 @@
...
+  import { DropdownMenu } from '@mathesar-component-library';
+  import { iconMoreActions } from '@mathesar/icons';
+
+  export let href: string;
+  export let ariaLabel: string | undefined = undefined;
+
+  let isHoveringMenuTrigger = false;
+</script>
+
+<div class="hyperlink-card" class:hovering-menu-trigger={isHoveringMenuTrigger}>
```
rajatvijay, line 11: Since this is a common component, when used in other places might end up with invalid markup (button or links inside the wrapper link). This could happen majorly because its uses `slots` to render `top` and `bottom` contents and the consumer can pass a button or a link in those slots. 

Wouldn't it be better if we used something like this: https://svelte.dev/repl/40411730765d45c3a1b46c665562be70?version=3.55.1

Few minor things on the current implementation: 

1. I see a focus styling missing on the whole card but maybe this is something that we can handle when doing keyboard interactions. 
2. The hover styling of the vertical three dots menu doesn't seem to respect the border-radius of the card. This is something that you might want to handle in this PR itself.
Screenshot 2023-01-12 at 10 04 40 AM

rajatvijay (changes requested):

mathesar_ui/src/components/HyperlinkCard.svelte
seancolsen: > when used in other places might end up with invalid markup

- There are a number of things about your REPL code which confuse me. I think it might be best for us to hop on a call to hash this out. This table card has some nuances and tricky requirements. For example

    - (A) The whole card area needs to be clickable as a hyperlink, except the menu trigger.
    - (B) When hovering a truncated description, the user should see the popover from the truncation component.

    The combination of (A) and (B) lead me to the conclusion that the description needs to be a child of the hyperlink. I'm pretty sure that in your REPL we wouldn't be able to hover the table description to see its full value

    If this critique is something we could handle in a separate PR it might be nice, just so we can get this fix in before the release.

> focus styling missing on the whole card

- Yeah, I'd rather fix that separately from this PR. We have lots of other areas of the app that need similar improvements. I've opened #2248 to track that work.

> The hover styling of the vertical three dots menu doesn't seem to respect the border-radius of the card. 

- I've pushed ef85a568bbf40ea0f26c644201ce9d4713f5cb7b to fix that. It's weird to me that I wasn't seeing that hover problem when originally working on the PR (as is evident in my screenshots above). But I was indeed able to reproduce it now. Oh well! This fix ought to do the trick I think.

kgodey:
>  The change in this PR does make the Record Selector slightly less discoverable here, but I think that's okay.

I think more important that discoverability, this makes finding a record a 2-click operation instead of a 1-click operation. In our library use case, when a patron comes in, the librarian finds them from the schema page. This is a frequent operation, and the extra clicks add up over time. Also, the click target is much smaller, which shouldn't be the case for common oper

Additionally, here's some design issues I found:

1. Hovering over a long description obscures the name of the table. I think it would make more sense for the tooltip to be on the bottom rather than the top.
Screen Shot 2023-01-12 at 4 35 57 PM

2. We can end up with different size cards for tables without a description. I think this looks weird. I don't have a particular solution to suggest, but I did like that the previous design had a uniform size.
Screen Shot 2023-01-12 at 4 37 11 PM

---

After playing around it with it locally, I prefer our current (in `master`) table cards over the ones in this PR. I do like that the entire card is clickable, but otherwise it seems more inconsistent / harder to use.

kgodey:
Another design issue – the old design clearly shows when a table has been imported and the data needs to be confirmed. I haven't looked to see what we're doing in this PR, but it should be as clear.

Screen Shot 2023-01-12 at 5 02 49 PM

seancolsen:
@kgodey how would you like to proceed here? Being that you authored #1978 requesting that the table cards be clickable, would you prefer that we abandon that ticket and close this PR? Should we have another round of design? Ghislaine and I already spent some time together on a call to formulate the design that I implemented in this PR. Should we have a design session with the three of us together? I don't have an opinion. I'm fine with the design in `master` and fine with the design in this PR.

As I see it issues `1` and `2` are also present in `master`. E.g....

image

I can fix `1`. But `2` is tricky. I played with making the cards vertically shorter but that looked even worse. 

I agree that discoverability of the record selector could be a problem. I'm open to other suggestions here.

seancolsen:
Oh good point about the unconfirmed imports. This PR _does need additional work_ to handle that! I forgot about that.

kgodey:
@seancolsen Ideally, I'd like to update this PR to:
- Fix issue 1 with the tooltip positioning
- bring the "Find a Record" button back, perhaps as a single full width button similar to "Confirm Imported Data"
- Bring the "Confirm Imported Data" button back as well
- For issue 2, `master` doesn't have the of some cards being tall and some cards being short if they are description-less in different rows (see "Author" and "Paper-Author Map" in my screenshot). I think we should EITHER:
  - Make the cards the same height regardless of description being present and live with the empty extra space (current `master`'s solution). 
  - Have a masonry style layout for the cards (I'm guessing this will take a lot more work and we should do the first one).

pavish:
@seancolsen Looks like there are UX changes required here, so I'm going to hold off code review until they are compelete.

seancolsen:
@kgodey Ok how about this.

https://user-images.githubusercontent.com/42411/212226068-33292007-e1c4-456d-8395-9bfc535eb840.mp4

- In 74754ee65904b6625c31fb85ec466253fcc36b7c I fixed the issue with the tooltip positioning.

- In defb122f847a7e21a0c3f0859f92807d34d50d5f I combined TableCard and HyperlinkCard, which I hope will address @rajatvijay's concerns (because TableCard can not compose other component as HyperlinkCard could have). This card is getting sufficiently complex now that the HyperlinkCard abstraction doesn't feel worth it anymore anyway.

- In db38b14eabfd6e5d1ebbdb95514f76b31fa6ca04 I adjusted some styles, including some improvised changes...

    - I added a "Find a Record" button to the bottom of the card (and removed that entry from the dropdown menu)

    - I gave unconfirmed imports a distinct style. Notably, I didn't want the "Confirm Imported Data" to look quote the same as "Find a Record" because "Confirm Imported Data" is _the same link as the card_ whereas "Find a Record" is a button separate from the card's hyperlink.

    - I decreased the text size of "Find a Record". This is to address a concern of mine about the previous design where, when scanning the list of tables I would sometimes get lost within other text. With this change, the table name stands out more.

    - I also decreased the text size of the description for the same reason as above. This change also has the benefit of reducing the vertical size discrepancy between a card with no description and a card with a full description.

- I did _not_ make any changes to address this concern `2` of Kriti's.

    > master doesn't have the of some cards being tall and some cards being short if they are description-less in different rows

    @kgodey, as far as I can tell, the behavior you're critiquing is also present in `master`. Here's what I observe in `master`

    image

    Notice that the "Authors" card is short, the "Items" card is somewhat taller, and the "Publications" card is taller still. The difference in height is less apparent due to the additional constant factor of the bottom section of the card. I'm hoping that my changes above will bring this PR into close-enough parity with `master` to alleviate your concern.

    I agree that a masonry layout would be ideal here. Masonry layout is [coming](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Masonry_Layout) to browsers natively, but it'll be a while before it's viable. So you're correct that it wouldn't be quick to implement.

    I'd rather not make all cards as tall as the tallest possible card for the following reasons:

    - This is tricky to implement cleanly.
    - I predict that many people will have schemas with lots of tables that don't have descriptions. I think a whole grid of cards that are taller than they need to be will look awkward and decrease usability by requiring more scrolling to find a table.

    If we do still want to seek changes to address concern `2`, I'm hoping we can continue discussion/development in a separate PR.

kgodey:
I just watched the video, I did not play with it locally, but I think these changes satisfy all my concerns, nice work @seancolsen.  I'm unassigning myself from this PR. 

I'd suggest changing the "Confirm Imported Data" text to "Needs Confirmation" or something like that since it's no longer an action button.

> Notice that the "Authors" card is short, the "Items" card is somewhat taller, and the "Publications" card is taller still.

I didn't realize that. Maybe it's just that the original height difference in this PR was so different that it was jarring.

kgodey (approved):

rajatvijay:
@seancolsen looks good to me too. Since the `HyperlinkCard` does not exist anymore I will put my suggestion as an item for the next FE meeting as a low-priority item as we do not have a real use case for this now.

rajatvijay (approved):

seancolsen:
@kgodey I pushed 6bc0ed58ca31066580bc929507ccf074c00f8df7 changing the text to "Needs Import Confirmation". That seemed a bit clearer than "Needs Confirmation".

image

kgodey:
Looks good, thanks @seancolsen

pavish (approved):
Looks good

PR #2243, opened 2023-01-11
Improve "Delete Records" button text

### LINKED ISSUE #1679: Update table inspector message for deleting records.
If there’s only one record selected, the delete record action should just say “Delete record”, not “Delete 1 record”.

seancolsen:
Fixes #1679

| Before | After |
| -- | -- |
| image | image |

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/utils/languageUtils.ts
```diff
@@ -110,12 +114,53 @@ export function pluralize(
...
+    }
+    if ('word' in countWhenSingular) {
+      return countWhenSingular.word;
+    }
+    assertExhaustive(countWhenSingular);
+  }
+  return String(count);
+}
+
 export function labeledCount(
```
rajatvijay, line 149: This is the kind of function I would say should have some doc string on top of it to explain the meaning of `countWhenZero`, `countWhenSingular`, and their usages. It took me two rounds of reading to understand it completely(might be just me 😔 )

mathesar_ui/src/utils/languageUtils.ts
```diff
@@ -110,12 +114,53 @@ export function pluralize(
...
+function getCountText(
+  count: number,
+  countWhenZero: CountWhenZero,
+  countWhenSingular: CountWhenSingular,
+): string {
+  if (count === 0) {
+    if (countWhenZero === 'numeric') {
+      return String(count);
+    }
+    if ('word' in countWhenZero) {
```
rajatvijay, line 129: I generally prefer `hasOwnProperty`(or its latest incarnation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) instead of `in` as a convention to avoid looking into prototype chains un-intentionally.

rajatvijay (changes requested):

mathesar_ui/src/utils/languageUtils.ts
seancolsen: Interesting 🤔

- Thanks for pointing out `Object.hasOwn`. I had not seen that before. It looks nice, although I'd say [support](https://caniuse.com/mdn-javascript_builtins_object_hasown) is still not sufficient that we should use it in our codebase (without a polyfill).

- I've not seen `hasOwnProperty` used for type narrowing before.

    If I change the code to use what I _think_ you're suggesting, then:

    - I get this ESLint error.

        image

        We could silence that linting error if needed. But...

    - TypeScript does not seem to actually narrow the type based on that code

        image

    From the TypeScript code I've seen, usage of `in` seems to be idiomatic. Indeed, that's how the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) recommend narrowing types. That docs page has no occurrences of `hasOwnProperty`.

    Am I'm misunderstanding your recommendation here?

mathesar_ui/src/utils/languageUtils.ts
seancolsen: Thanks for this feedback. I've pushed cb577407d9add146b92cdcd5b2699065656e7148 to add some more documentation to the functions in that file. Does this help?

mathesar_ui/src/utils/languageUtils.ts
rajatvijay: @seancolsen its not about type narrowing - we can acheive this on `Object.hasOwnProperty` by overriding its type - rather about avoid looking into prototype chains. Anyways its a small thing hence approving the PR

rajatvijay (approved):

mathesar_ui/src/utils/languageUtils.ts
seancolsen: I'd be curious to see the code you'd prefer to use in this case.

Issue #2223, opened 2023-01-09
Live demo server often shows "502 bad gateway" error


kgodey:
Refreshing sometimes fixes it, but not always.

silentninja:
This can happen when a deployment is in progress. Are you trying to access when a deployment is in progress or when a deployment has taken place recently?

silentninja:
Lack of memory in the server can also be the reason. I will look into it when I have time.

kgodey:
> This can happen when a deployment is in progress. Are you trying to access when a deployment is in progress or when a deployment has taken place recently?

No, this is happening all the time.

> I will look into it when I have time.

@silentninja This is blocking the live demo / product from being usability tested (our top goal), so please make this your highest priority over later goals.

silentninja:
Update on the issue:
The error is happening due to server timeout. It takes more than 30 seconds for the `login` request to complete, so the gunicorn closes the connection as the timeout is 30 seconds. I suspect that loading the dataset synchronously is the problem.

silentninja:
#2230 should fix this error.

kgodey:
@mathemancer said on Matrix that he found a reflection-related bug that was causing this.

PR #2244, opened 2023-01-12
Improve demo loading time

mathemancer:
Related to #1906 

This changes to using a preloaded DB-level template when creating a new user demo database. This reduces the loading time to < 250ms for actually loading data (as per my profiling)

**Technical details**

This does _not_ fix the general over-reflection bug causing #2223 . That bug will make any request that involves reflection (including creating a new user DB) take at least 400ms times the number of databases. This PR only addresses the demo-data-specific performance.

With my testing, loading demo data seems to take 50-250ms plus the general reflection overhead on my machine.

I'm leaving the checklist in #1906 a bit vague w.r.t. any front end work until I get the follow up PR in and we see what the actual speed is like.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@pavish To check the performance, you should use a pristine DB

@silentninja I've put you as a reviewer to check that the code and methodology make sense (I'm not exactly an expert with the `AppConfig` subclassing).

pavish (approved):
The loading time has been reasonable enough locally for me so I'm not sure how to test this PR. I'm giving my approval and I believe the primary reason behind any delay/timeout is caused by reflection, which is covered in https://github.com/centerofci/mathesar/pull/2247.

mathemancer:
> The loading time has been reasonable enough locally for me so I'm not sure how to test this PR. I'm giving my approval and I believe the primary reason behind any delay/timeout is caused by reflection, which is covered in #2247.

Fair enough; I think we'll actually need to load up the demo server with multiple session DBs to really test this.

demo/apps.py
```diff
@@ -0,0 +1,41 @@
...
+    load_datasets(user_engine)
+    user_engine.dispose()
+
+
+class DemoConfig(AppConfig):
+    """Initialization manager."""
+
+    name = "demo"
+
+    def ready(self):
```
silentninja, line 39: This would be called any time we run a management command. So a template database will be dropped and a new one will be created each time we update the server, calling some Django commands like `migrate`. It would be better to create the template database when running the install script. 

But this is something we can fix later, please create an issue if you aren't planning to fix it later.

silentninja (changes requested):
Looks good to me, nice use of the templates! I just have one request though

silentninja (approved):

demo/apps.py
silentninja: After our conversation, I think we can fix it when a problem occurs.

PR #2252, opened 2023-01-13
Add autoincrement to existing integer primary key column of an import

### LINKED ISSUE #2144: ID sequence isn't being created for tables
## Description

I cannot add a new record to a new table because the ID does not have a default value. I set up this table by:
1. Importing [library_books.tsv.zip](https://github.com/centerofci/mathesar/files/10273401/library_books.tsv.zip) as "Items"
2. Pulling "Author First Name" and "Author First Name" into an "Authors" table
3. Moving "Author Website" to "Authors" table
4. Pulling "Publishers" into a "Publishers" table
5. Pulling "Title", "Publication Year", "ISBN", "Authors_id", and "Publishers_id" into a "Books" table

## Expected behavior

I should be able to add a new record!

## To Reproduce

~Try adding a new record on https://library.mathesar.org/mathesar_tables/4/tables/34/~

I needed to nuke this table since it was blocking my work. To reproduce, follow the steps under Description.

## Additional context

This seems related to
- #2116 
- #1356

silentninja:
Fixes #2144 

Removes disabling `autoincrement` when a dataset contains a default `id` column,  instead always update the default 
 pk column sequence to the max value of the column when exporting.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
This will break if the user's ID column non-integer, e.g., ISBN, UUID, Dewey, URI, etc.

I suggest a different approach:

If the user uploads a CSV with an `id` column, and it's not an integer, rename it to `id_old` (or something) and create our usual ID column. I see no reason to assume an `id` column of sequential integers will be the norm for user-provided data.

kgodey:
@silentninja @mathemancer what's the status of this PR?

silentninja:
I don't like the idea of changing the `id` field to `original_id`
1. It modifies the imported data without informing the user.
2. The user might want the existing `id` column from the imported data to be the actual primary key for various reason. For example, if the CSV files are relation, I would want to use the existing `id` column, so the references in other imported csv files are intact.

Alternate solutions that I can think of are
1. Revert back to calling the primary key column as `mathesar_id`. This prevents all the unnecessary problem
2. Have a toggle button that lets the user choose if we should use existing foreign key column or create a new one.

I considered other options but those are not feasible given our time constraints

silentninja:
> If the user uploads a CSV with an id column, and it's not an integer, rename it to id_old (or something) and create our usual ID column

@brent the problems mentioned by you can still happen with `INTEGER` column, for example, Dewey can be represented as INTEGER column.

kgodey:
@silentninja I don't think it's a problem to add a new `original_id` field. We are not deleting any user data. If the user wants to use their original column as the primary key, they can configure it that way after import.

Responses to your solution:
> 1. Revert back to calling the primary key column `mathesar_id`. This prevents all the unnecessary problem

This means that if they use their data outside of Mathesar, it will still have the "mathesar" name. That's worse.

> 2. Have a toggle button that lets the user choose if we should use the existing foreign key column or create a new one.

We'll need additional UX and frontend work for this, which will delay release further. This is not something critical enough to delay release on.

kgodey:
@silentninja The user can use the constraints modal to add/remove constraints. I'm not sure if we support adding/removing primary key constraints yet, but if not, we can add that in the future.

silentninja:
Okay, I will make the necessary changes.

silentninja:
@mathemancer On checking further, the behaviour you proposed seems to be the existing behaviour. Our current behaviour is 
1. We create an empty Table based on the csv dataset
2. If the created Table contains an `id` column Replace the `id` column with an `INTEGER` column
3. Try inserting the data
4. If it fails remove the Table and create a new Table with `id_original` column.

kgodey:
@silentninja If we already have this behavior, why do imported CSVs with the `id` column not have the sequence set correctly?

silentninja:
> "This will break if the user's ID column non-integer, e.g., ISBN, UUID, Dewey, URI, etc.

> I suggest a different approach:

> If the user uploads a CSV with an id column, and it's not an integer, rename it to id_old (or something) and create our usual ID column. I see no reason to assume an id column of sequential integers will be the norm for user-provided data."

@kgodey My reply was to the alternate behaviour mentioned above by @mathemancer. We already rename the existing `id` column to `original_id` if the column does not have a integer data

kgodey:
@silentninja Okay, thanks. This PR isn't ready for review yet, right?

silentninja:
> Okay, thanks. This PR isn't ready for review yet, right?

This PR does not need any changes so I changed the tags to review, sorry for not doing it earlier.

kgodey:
Okay thanks @silentninja

mathemancer:
> @mathemancer On checking further, the behaviour you proposed seems to be the existing behaviour. Our current behaviour is
> 
> 1. We create an empty Table based on the csv dataset
> 2. If the created Table contains an `id` column Replace the `id` column with an `INTEGER` column
> 3. Try inserting the data
> 4. If it fails remove the Table and create a new Table with existing `id` column to `id_original` column and create a new `id` column
> 
> Edit: The only change this PR adds is to set a sequence in case the existing `id` column has `Integer` data values.

Cool, thanks for the info.

mathemancer (approved):
@silentninja is right; this already works as I suggested. Sorry for the confusion.

PR #2253, opened 2023-01-13
Fix transforms not being able to define ordering

### LINKED ISSUE #2238: Order transformation does not sort results in queries
## Description
* A query request with order transformation does not provide the sorted output we would expect
* Example request:
```
{
   "base_table":4,
   "initial_columns":[
      {
         "id":18,
         "alias":"Patrons_Email"
      },
      {
         "id":16,
         "alias":"Patrons_First Name"
      },
      {
         "id":17,
         "alias":"Patrons_Last Name"
      }
   ],
   "transformations":[
      {
         "type":"order",
         "spec":[
            {
               "field":"Patrons_Email",
               "direction":"desc"
            }
         ]
      }
   ],
   "display_names":{
      "Patrons_Email":"Patrons_Email",
      "Patrons_Last Name":"Patrons_Last Name",
      "Patrons_First Name":"Patrons_First Name"
   },
   "parameters":{
      "limit":100,
      "offset":0
   }
}
```
* The result does not seem ordered in descending by email as expected

dmos62:
Fixes #2238

## Checklist
 
  ixdex.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
Unfortunately, we need to make this even a little more complicated.

The original issue that introduced this default ordering was that a partial defined order (e.g., the domain of an email field) would result in an indeterminate order for the result set (messing up, e.g., pagination).

This will fix the bug it's supposed to, but will reintroduce that old bug (unless I'm missing something).

I think we need to figure out how to merge the transform order with the primary key ordering to make a fully determined ordering in these cases.

dmos62:
@mathemancer Thanks for the review; I know how to fix this.

dmos62:
@mathemancer re-requesting review. A lot of hiccups, but seems to finally be working.

mathemancer (approved):
I think we should merge this, but create some issues to discuss and work on some details:

1. If there are multiple ordering steps (either `Order` transforms as part of the query, or `order_by` given as a query string parameter), the current setup will clobber any previous ordering with the next ordering step. The reason is that we create a deterministic ordering at each step, and we do this by adding the primary key (for example) as an ordering column. So, if step one orders by email domain, and step two orders by last name of a user, then there will be no trace of ordering by email domain in the final order.
2. At one point, you note in a comment that there's a bug whenever we have internal columns of JSON type, and no primary key in a table. This bug is still present for _any_ non-orderable column type if there is no primary key in the table (e.g., if the user attaches to a preexisting DB). It's easy to figure out whether or not a type is orderable in PostgreSQL. See the [docs page for ORDER BY](https://www.postgresql.org/docs/current/queries-order.html#id-1.5.6.9.5.10) and the [docs page for [Operator Classes](https://www.postgresql.org/docs/9.5/indexes-opclass.html).

mathemancer:
@dmos62 Leaving unmerged so as not to interfere with your workflow creating issues for (1) and (2) above.

Issue #2149, opened 2022-12-21
Breadcrumb out of sync after browser "back" button


seancolsen:
## Steps to reproduce

1. Use our Library Management schema.
1. Navigate to the Publications table page.
1. Group by Publisher.
1. Click on "Summarize Table", navigating to the Data Explorer
1. Observe the site navigation breadcrumb to be "Library Management / Data Explorer". Good.
1. Press the "Back" button on your browser.
1. Expect the navigation breadcrumb to contain "Library Management / Publications".
1. Instead, observe it to be "Library Management / Data Explorer" as before.

kgodey:
I ran into a stale breadcrumb issue on a call today, but I don't have reproduction steps handy. I know there was an Exploration Page breadcrumb in the nav even though I wasn't on an Exploration Page.

seancolsen:
Duplicate of #2030

PR #2254, opened 2023-01-13
Fix imperative breadcrumb logic

### LINKED ISSUE #2030: Breadcrumbs do not update when coming back from edit exploration to saved exploration
### Steps to reproduce

1. Go to any saved exploration and notice the breadcrumbs.
2. Click on the edit button on the top right and notice the breadcrumbs.
3. Click on the browser's back button. Notice that the page has changed(routing is successful) but the breadcrumbs do not change.

seancolsen:
Fixes #2030

## Reproducing the bug

- I was not able to reproduce the bug, given the steps listed in the ticket, but I was able to reproduce a bug with identical symptoms, given any of the following recipes:

- A:

    1. Use the steps in #2149 (duplicate issue)

- B:

    1. Load a page like http://localhost:8000/mathesar_tables/3/data-explorer/
    1. Click "Shortcuts" > "New Table from Data Import"

- C:

    1. Load a page like http://localhost:8000/mathesar_tables/3/explorations/1/
    1. Click "Shortcuts" > "New Table from Data Import"

## The fix

- My fix seems simple enough, but I'm not 100% confident that it will be free of regressions.

- I hammered at the app pretty hard, navigating from all sorts of pages to one another (and using the browser's "Back" button) -- and I couldn't spot any issues with the breadcrumbs.

## Notes for reviewers

- @rajatvijay I'd like to to review to verify that you can no longer reproduce the issue as you described it.

- @pavish I'd like you to review since this is peripherally related to the Data Explorer routing. When trying to reproduce this bug, I was only able to find issues when navigating away from a page within the Data Explorer system. Kriti also had reported on Matrix experiencing a bug like this with the Data Explorer, though she didn't have specific repro steps.

- I'd like both of you to try hammering at a bunch of pages via client-side navigation and browser "back" button.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


rajatvijay:
@seancolsen tried a bunch of cases, LGTM. 

I am also not able to reproduce the bug that I mentioned in the ticket, so we are GTG.

pavish (approved):
Looks good to me.

This might break if we have the same item object repeated in the breadcrumb, but our UX does not have such a scenario yet (and it probably wouldn't in the future), so we are good.

The reason this issue happened with the DataExplorer route is because of the way `EventfulRoute` and `MultiPathRoute` work. They are used in places where we want the same child without any re-render when urls get updated. Here the new route's load event may get dispatched before the old route's unload event and there's no way currently to force a standard behaviour, and I don't think it's worth focusing on at the moment.

PR #2453, opened 2023-02-10
Password change and reset throw 400

### LINKED ISSUE #2391: Password change and reset endpoints throw 500 instead of returning error messages
## Description
* Try updating a password with a common password such as `password`.
* The /password_reset and /password_change endpoints throw a 500
* They should return a 400 with formatted error messages

Related: 
* https://github.com/centerofci/mathesar/issues/1673

Hatilar420:
Fixes #2391

# Overview
1. The `/password_reset` and `/password_change` endpoints throw a 400 formatted response when the password is changed or reset with a common password

**Screenshots**

1. `/password_reset` formatted response.
image

2. `/password_change` formatted response.

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@Hatilar420 thanks for your contribution! this might take us a couple of weeks to get to since we're focused on launching our first release right now.

Hatilar420:
sure np @kgodey.

mathesar/api/exceptions/mixins.py
```diff
@@ -81,7 +81,7 @@ def _run_validator(self, validator, field, message):
                 args.append(field)
             validator(self.initial_data[field.field_name], *args)
         except (DjangoValidationError, RestValidationError) as err:
-            err_message = err.detail[0] if hasattr(err, 'detail') else err.message
+            err_message = err.detail[0] if hasattr(err, 'detail') else err.messages[0]
```
silentninja, line 84: - Django error objects sometimes contain only the `message` field. So `err.message` is sometimes valid too. So please revert its removal. In order to support various types of errors,  you would need to write the logic for each case.
- Also make sure to check if the `messages` array is not empty before fetching the first error object.

silentninja (changes requested):
Thanks for the PR @Hatilar420. The error is captured correctly but I have a few concerns which are commented on in detail below, these need to be addressed before the PR is merged.

mathesar/api/exceptions/mixins.py
Hatilar420: Valid points @silentninja, Will implement the suggested changes.

Hatilar420:
Hi @silentninja, I have implemented the changes you requested for. I have some doubts.
- What should value should we assign to the `err_message` variable if the `messages` array were to be empty (Which is really unlikely unless and until there is some change in `DjangoValidationError`, `RestValidationError` by the Django team)
- What should be the default value for `err_message`?

silentninja (approved):

silentninja:
> What should value should we assign to the err_message variable if the messages array were to be empty (Which is really unlikely unless and until there is some change in DjangoValidationError, RestValidationError by the Django team)
What should be the default value for err_message?

The above scenario should not likely occur, so we should leave it as is and get a 500 error thrown than hiding the problem

PR #2460, opened 2023-02-11
Fix un-authenticated behavior for all endpoints

### LINKED ISSUE #2193: Return 401 for unauthenticated requests to /api/ and redirect to login page for other urls
## Description
When user is not authenticated:
- Endpoints within /api/ should return a 401.
   - Currently some of them within /api/ return a 403, and some perform redirection to login page. They should all return a 401.
- All other endpoints which have associated views such as /, /<db_name>/, /<db_name>/<schema_id>/ etc., should redirect to login page.

## Frontend work
* Within api/requestUtils
  * Check if the response status is 401.
    * If yes, reload page.

Anish9901:
Fixes #2193

**Technical details**

This adds the DRF's `TokenAuthentication` authentication scheme that properly handles unauthenticated behavior by responding with `HTTP 401 Unauthorized` error when an unauthorized request is made to any `/api` endpoint.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


config/settings/common_settings.py
```diff
@@ -164,10 +164,10 @@ def pipe_delim(pipe_string):
 
 MEDIA_URL = "/media/"
 
-# Update Authentication classes, removed BasicAuthentication
 # Defaults: https://www.django-rest-framework.org/api-guide/settings/
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': [
+        'rest_framework.authentication.BasicAuthentication',
```
silentninja, line 170: `BasicAuthentication` has many [security vulnerabilities](https://www.whiteoaksecurity.com/blog/simple-security-fails-part-5-basic-authentication/), we should not be using it and I don't see why it is needed for to fix this issue

silentninja (changes requested):
I don't think this PR actually fixes the issue, instead replaces it with a completely different behaviour compared to the expected outcome mentioned in issue #2193. When a user is not authenticated, the outcomes listed below happen randomly,
  - A browser prompt asking for a username and password
   - Sometimes results in a redirect to the login page for `/api` endpoints
    - Sometimes results in 401 status code
I guess it is due to the authentication type changes due to how DRF handles authentication(tries all the authentication backends) and how authentication backends are chosen by Django, but in short this trouble in my opinion and I would prefer to avoid it.
    
~I think issue #2193 is happening due to the [`AccessViewSetMixin` ](https://github.com/rsinger86/drf-access-policy/blob/8c4630432aafd70ff3fecb70d0606dc30426dbf8/rest_access_policy/access_view_set_mixin.py) as it overrides the permission classes and uses its access policy permission classes before other DRF permission classes, so when an unauthenticated user tries to access an API, it results in 403 status code(thrown by the `AccessViewSetMixin` when any condition fails). So we should try to change the priority of the permission classes so that the DRF [`isAuthenticated`](https://github.com/centerofci/mathesar/blob/f33c7a63d2a0e7b2bc7ce0f9eb35b0783b2b6ae6/config/settings/common_settings.py#L174) permission class(added by default to all viewset) is the top priority as it is throws `401` instead of `403`~

Edit: Looks like throwing 403 for session authentication is the [expected behaviour](https://www.django-rest-framework.org/api-guide/permissions/#how-permissions-are-determined). There is a long discussion (https://github.com/encode/django-rest-framework/pull/1611) regarding this problem.

Anish9901:
> Sometimes results in a redirect to the login page for /api endpoints

@silentninja I didn't find any `/api` endpoint redirecting to the login page during local testing. Can you provide some endpoints that redirect to login page instead of 401s?

About the popups I thought that frontend could disable them as per the need. Correct me if I am wrong.

config/settings/common_settings.py
Anish9901: Replaced with `TokenAuthentication`.

I was unaware of the security vulnerabilities of `BasicAuthentication` as they weren't mentioned any where in the [docs](https://www.django-rest-framework.org/api-guide/authentication/#basicauthentication).

>  I don't see why it is needed for to fix this issue

`BasicAuthentication` did provide the appropriate `WWW-Authenticate` header which was necessary for returning `HTTP 401 Unauthorized` which was lacking in `SessionAuthentication` therefore I decided to go with it.

silentninja:
> > Sometimes results in a redirect to the login page for /api endpoints
> 
>  I didn't find any `/api` endpoint redirecting to the login page during local testing. Can you provide some endpoints that redirect to login page instead of 401s?

This behaviour does not happen with a particular API, instead, it is one of the randomly chosen actions when an unauthenticated user tries to access an API. I don't know of a reliable way to reproduce it. But lets not worry about it now. The commit doesn't seem to have that issue.

silentninja (changes requested):

mathesar/api/db/permissions/table.py
```diff
@@ -48,7 +48,7 @@ class TableAccessPolicy(AccessPolicy):
 
     @classmethod
     def _scope_queryset(cls, request, qs, allowed_roles):
-        if not request.user.is_superuser:
+        if not (request.user.is_superuser or request.user.is_anonymous):
```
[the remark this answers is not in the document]
Anish9901: I missed it, thanks for the catch!

silentninja (approved):

silentninja:
Thanks for the PR @Anish9901, it is working as expected. I will wait for @pavish to review it before merging the PR

pavish (approved):
@Anish9901 Looks good. I tested with a couple endpoints and it works as expected, I did not do a through test since @silentninja has already reviewed this.

I added a frontend commit https://github.com/centerofci/mathesar/pull/2460/commits/16918496879ebe40870e24ce306b136619ad75a2 to handle reload (which redirects to login page) when a 401 is encountered.

PR #1355, opened 2022-05-05
Upgrade text-numeric inferring

Jyuart:
Fixes #1107

It's an upgrade of the text (`TEXT`, `CHAR`, `VARCHAR`) --> numeric (`NUMERIC`) types inferring (and more comprehensive casting to numeric type accordingly since inferring is based on casting capabilities). Values like:
- 3.14
- 123,456.7
- 123.456,7
- 123 456,7
- 1,23,456.7
- 123'456.7
are now successfully recognized as of `NUMERIC` type

**Technical details**

It was inspired by this PR https://github.com/centerofci/mathesar/pull/1137 and is implemented in the same manner:
1. A separate function for getting a numeric array in the format: `['number', 'separator', 'floating-point']`. For example: `['331,209.05', ',', '.']`
2. Another function that removes locale-specific parts and extracts the actual number in the format with `.` as a floating-point and with no separators: `331209.05`
3. Supported formats are based on this comment: https://github.com/centerofci/mathesar/issues/1107#issuecomment-1123949497, but can be easily upgraded if necessary. Excerpt:
<html>
<body>

example locale | example format
-- | --
'en' | -123,456.7
'de' | -123.456,7
'fr' | -123 456,7
'hi' | -1,23,456.7
'de-CH' | -123'456.7

</body>
</html>

5. One test in the `test_table_api.py` file was updated

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

[...]

PR #2465, opened 2023-02-13
Handle human readable number casting in general

### LINKED ISSUE #1309: Handle human readable number casting in general
## Problem

Currently, casting from textual types (e.g., `TEXT`) to numbers relies on the default PostgreSQL number handling. This will be solved for the `NUMERIC` type once #1107 is solved, but we should also handle other number types.

## Proposed solution

We should write custom logic to allow handling casting more strings to number types.

- Casting to number types with decimals should use the same (or very similar) logic to casting to `NUMERIC`.
- Casting to number types without decimals (e.g., `INTEGER`) should use a subset of that logic that handles only grouping separators.

## Additional context

This was originally inspired by the custom casting work for money types. See #1137 . This issue should complete the work begun when solving #1107 .

Aditramesh:
Fixes #1309 

 Adds functionalities to handle casting of text to integer and decimal types which was implemented for numeric types in #1355

**Technical details**
Uses the exact same logic for decimal types but a subset of that for integer types  

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Aditramesh:
The pr is in very bad shape because this is my first pr,sorry for the trouble i havent added tests ,but wil add them soon once the work here is appoved

db/types/operations/cast.py
```diff
@@ -1027,6 +1036,128 @@ def _build_numeric_array_function():
...
+
+def _build_integer_array_function():
+    """
+    The main reason for this function to be separate is for testing. This
+    does have some performance impact; we should consider inlining later.
+    """
+    qualified_function_name = get_qualified_name(INTEGER_ARR_FUNC_NAME)
+
+    no_separator_big = r"[0-9]{4,}(?:([,.])[0-9]+)?"
+    no_separator_small = r"[0-9]{1,3}(?:([,.])[0-9]{1,2}|[0-9]{4,})?"
```
mathemancer, line 1050: These regexes match things which are not typically integers, e.g., `123456.1234` and `123.45` . They _don't_ match integers such as `12,345,678`. Please try to understand how the example regexes work, and use the appropriate pieces.

mathemancer (changes requested):
See my specific comments. Make sure you understand how the regexes you're using work, and that you understand the goal of the issue.

You should try to set up functions that can take typical 'string-like' integer input such as `12,345,678` and transform it into actual integers. For the 'decimal-like' part of the issue, you just need to add the decimal types other than `NUMERIC` into the same logic as the `NUMERIC`, and test that it works.

db/types/operations/cast.py
Aditramesh: Thank you @mathemancer ,working on both of these

Aditramesh: Hey @mathemancer i went through each of the regex and removed the decimal part from each of them so all the possibilities for integers are taken care in each case
    no_separator_big = r"[0-9]{4,}(?:([,])[0-9]+)?"
    no_separator_small = r"[0-9]{1,3}(?:([,])[0-9]{1,2}|[0-9]{4,})?"
    comma_separator_req_decimal = r"[0-9]{1,3}(,)[0-9]{3}+"
    **period_separator_req_decimal = r"[0-9]{1,3}(\.)[0-9]{3}"**
    comma_separator_opt_integer = r"[0-9]{1,3}(?:(,)[0-9]{3}){2,}"
    **period_separator_opt_decimal = r"[0-9]{1,3}(?:(\.)[0-9]{3}){2,}"**
    space_separator_opt_integer = r"[0-9]{1,3}(?:( )[0-9]{3})+"
    comma_separator_lakh_system = r"[0-9]{1,2}(?:(,)[0-9]{2})+,[0-9]{3}"
    single_quote_separator_opt_integer = r"[0-9]{1,3}(?:(\'')[0-9]{3})+"
however there are two cases for regex having period seperators,i have kept '.' and removed ',' as it acts as decimal point,is this right?

mathemancer: A few things:
- The `no_separator_...` ones should just be combined to a single regex that finds a string of numbers for integers.
- The `...req_decimal` ones don't make sense when looking for integers. The point of them is that given `12.345`, you can't tell whether `.` is a separator or decimal point unless you know the locale.

To answer your question, yes `,` acts as the decimal in those cases.

The goal is to match:
- 12345
- 1,234,567
- 1.234.567
And not to match:
- 1,2
- 1,23
- 1,234
- 1.2
- 1.23
- 1.234

It's more important, given where this function is used, to avoid matching ambiguous strings.

Aditramesh: That was very detailed,thanks for the help @mathemancer ,working on it .

db/types/operations/cast.py
```diff
@@ -1027,6 +1036,128 @@ def _build_numeric_array_function():
...
+        SELECT array_remove(ARRAY[{group_dividers_str}], null) INTO group_divider_arr;
+        SELECT actual_number_arr[1] INTO actual_number;
+        SELECT group_divider_arr[1] INTO group_divider;
+        RETURN ARRAY[actual_number, group_divider];
+      END;
+    $$ LANGUAGE plpgsql;
+    """
+
+
+def _build_decimal_array_function():
```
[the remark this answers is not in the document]
Aditramesh: The only place in the   _build_numeric_array_function(): where it being decimal or numeric will make a difference is in this line 
 qualified_function_name = get_qualified_name(NUMERIC_ARR_FUNC_NAME)
other than this it is independent ,so is this the only change required ? 

Also real and float types are a subset of numeric as given in the categories file ,so it should be taken care of in the numeric function itself 

if the same function is being used for both decimal and numeric types should the name of the function be changed ?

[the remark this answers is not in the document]
mathemancer: I don't see any reason to consider `REAL` or `FLOAT as different from `NUMERIC` for this logic. The only possible issue is that these target types are of finite precision rather than arbitrary.

You should remove the distinction between them and just use the same casting function for all these types.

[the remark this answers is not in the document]
Aditramesh: So casting everything to numeric types instead of decimal types in the cast function should take care of it ?

def create_decimal_number_casts(engine):
    decimal_array_create = _build_numeric_array_function(DECIMAL_ARR_FUNC_NAME)
    with engine.begin() as conn:
        conn.execute(text(decimal_array_create))
    type_body_map = _get_decimal_number_type_body_map()
    create_cast_functions(PostgresType.NUMERIC, type_body_map, engine)

dmos62:
@Aditramesh closing due to inactivity. @Aditramesh if you'd like to continue working on this, make sure that at least all tests are passing before asking for a re-review.

Aditramesh:
Okay @dmos62 ,i will make sure the tests pass before asking for a re-review ,thank you .

PR #2489, opened 2023-02-15
Table inspector Access controls

### LINKED ISSUE #2477: Access control in Table inspector

rajatvijay:
Fixes #2477 

**Technical details**
1. Added the ability to disable [FormBuilder](mathesar_ui/src/component-library/form-builder/FormBuilder.svelte)(the whole form in a single go) and the [FormElement](mathesar_ui/src/component-library/form-builder/FormElement.svelte)(single field at a time)
2. When the user does not have `canExecuteDDL` the whole `Properties` section is removed from the Table mode since that data is already available on the left of the `ActionsPane`

**When no canExecuteDDL**
Screenshot 2023-02-16 at 6 47 25 PM
Screenshot 2023-02-16 at 6 47 31 PM
Screenshot 2023-02-16 at 6 54 01 PM

**When no canEditTableRecords**
Screenshot 2023-02-16 at 6 55 29 PM

**When no canEditMetadata**
Screenshot 2023-02-16 at 7 00 08 PM
Screenshot 2023-02-16 at 6 58 47 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
This looks good except for one small issue:

- Editors and viewers should not see this save button.

    image

    There's a weird bug here. The save button appears when I change the cell selection from a cell within the "id" column to a cell within the "First Name" column. But then when I toggle between selecting a cell within the "First Name" column and a cell within the "Last Name" column, the Save button disappears and does not reappear.

rajatvijay:
> This looks good except for one small issue:
> 
>     * Editors and viewers should not see this save button.
>       image
>       There's a weird bug here. The save button appears when I change the cell selection from a cell within the "id" column to a cell within the "First Name" column. But then when I toggle between selecting a cell within the "First Name" column and a cell within the "Last Name" column, the Save button disappears and does not reappear.

ad54b0479

mathesar_ui/src/systems/table-view/table-inspector/column/SetDefaultValue.svelte
```diff
@@ -9,21 +9,23 @@
...
   export let column: ProcessedColumn;
+  export let canExecuteDDL: boolean;
 
   const tabularData = getTabularDataStoreFromContext();
   $: ({ columnsDataStore, recordsData } = $tabularData);
   $: ({ recordSummaries } = recordsData);
 
-  $: value = column.column.default?.value;
-  $: actionButtonsVisible = value !== column.column.default?.value;
+  $: initialValue = column.column.default?.value ?? '';
```
pavish, line 18: Do not assign an empty string here. The input is not always a text input, for eg., for boolean we have a dropdown. The value here is a generic type which is accepted by DynamicInput and inferred by the underlying component which is based on the column type.

pavish (changes requested):

mathesar_ui/src/systems/table-view/table-inspector/column/SetDefaultValue.svelte
rajatvijay: yes @pavish exactly that is why I did not re-assign it. I have pushed the final change now.

seancolsen (approved):
Looks good to me now. I'll wait for @pavish to re-review before merging.

mathesar_ui/src/systems/table-view/table-inspector/column/SetDefaultValue.svelte
```diff
@@ -9,21 +9,42 @@
...
+  $: sanitizedValue = value === '' ? undefined : value;
+
+  /**
+   * Not using strict equality here since
+   * numbers are being sent as string to the backend
+   * but the initialValue from BE is always of the
+   * correct data type.
+   */
+  // eslint-disable-next-line eqeqeq
+  $: actionButtonsVisible = sanitizedValue != initialValue;
```
pavish, line 38: There is a `initialInputValue` property in ProcessedColumn which should provide the value of the input when the default value is undefined. This eliminates the need for sanitizedValue.

I'm recommending a value & initialValue comparison with stringified values via JSON.stringify since this should provide the value as returned from the server. This is more versatile and works with types like JSON map, JSON list, and array types. The current check would not work for any of them.

```suggestion
  $: initialValue = column.column.default?.value ?? column.initialInputValue;
  $: value = initialValue;
  $: actionButtonsVisible = JSON.stringify(value) !== JSON.stringify(initialValue);
```

pavish (changes requested):
@rajatvijay @seancolsen I've not done a complete review since @seancolsen's already approved the PR.

I've only reviewed the updates based on the comment I had earlier.

mathesar_ui/src/systems/table-view/table-inspector/column/SetDefaultValue.svelte
pavish: https://github.com/centerofci/mathesar/pull/2489#discussion_r1111313262

mathesar_ui/src/systems/table-view/table-inspector/column/SetDefaultValue.svelte
rajatvijay: @pavish 

>There is an initialInputValue property in ProcessedColumn which should provide the value of the input when the default value is undefined. This eliminates the need for sanitized value.

Could not find this key inside the ProcessedColumn instead using: `getInitialInputValue(column.column);`

>I'm recommending a value & initialValue comparison with stringified values via JSON.stringify since this should provide the value as returned from the server. This is more versatile and works with types like JSON map, JSON list, and array types. The current check would not work for any of them.

There are two reasons for me not taking this approach

**First Reason:** 
When setting a default value for a column with data type number, let's say `value` will be `"1145"` and `initialValue` will be `1145`.
```typescript

$: value = "1145";
$: initialValue = 1145;
$: actionButtionsVisible = JSON.stringify(value) !== JSON.stringify(initialValue);

```

Here actionButtionsVisible will be false should be true since `JSON.stringify("1145") = "\"1145\""` and `JSON.stringify(1145) = "1145"`.

**Second Reason:**
We do not have any such column types as of now that will need JSON map, JSON list, or array types and the purpose of this PR is very different

pavish: @rajatvijay 

> Could not find this key inside the ProcessedColumn instead using: getInitialInputValue(column.column);

It's worth adding it to ProcessedColumn. You can refer data explorer which has it.

> We do not have any such column types as of now that will need JSON map, JSON list, or array types and the purpose of this PR is very different

We do have all of these column types implemented, please refer the cell-fabric component. I have disabled it while creating a column for the release. The user is still able to convert a string column to JSON list or map, and the database the user connects to can have these columns.

> Here actionButtionsVisible will be false should be true since JSON.stringify("1145") = "\"1145\"" and JSON.stringify(1145) = "1145".

I'd suggest a type check before this check, instead of `==`. For literals, we could do a `String(value) === String(initialValue)` and we can do a `JSON.stringify` check for objects.

rajatvijay: 385b210dd

pavish: Looks good

PR #2513, opened 2023-02-18
Add telemetry to frontend app

### LINKED ISSUE #2502: Set up analytics for live demo
We should set up analytics for our demo site. We can't add analytics to the main Mathesar codebase (since it's designed to be self-hosted), so we should add to the demo deployment via Ansible.

We use Simple Analytics. Here's the [docs](https://docs.simpleanalytics.com/). It would be great if we had events that helped us understand what actions users were taking and which data sets they were finding useful.

silentninja:
Fixes #2502 

Adds functions to log user interaction events. There is a slight change in the implemented infrastructure compared to agreed infrastructure in the [sync meeting](https://wiki.mathesar.org/en/meeting-notes/2023-02/2023-02-17-launch-check-in#adding-analytic-events-to-the-frontend). I have added the details in the `Technical Details` section below

**Technical details**
We dispatch the logged user interaction events as a window event, which is then picked up by an event listener that is available in the demo mode.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


silentninja:
@kgodey I would like you to review the logged event name and metadata attached to the events and make sure they are sensible.

mathesar_ui/src/pages/schema/SchemaPage.svelte
```diff
@@ -81,6 +83,13 @@
   }
 
   $: isDefault = schema.name === 'public';
+  onMount(async () => {
+    logEvent('opened_schema', {
+      database_name: database.name,
+      schema_name: schema.name,
+      page: 'schema',
```
kgodey, line 90: What does `page` refer to here? It's a little vague so I'd prefer a more verbose name.

kgodey (commented):
I have one comment @silentninja

mathesar_ui/src/pages/schema/SchemaPage.svelte
silentninja: @kgodey `page` refers to the screen/page where the event/action took place. Does the name `event_location`  or `source_page`  convey the intent better?


In case you want to understand the reason behind the `page` attribute. It is useful in case an action can be performed in more than one place. For example, we can create a Table from
- The schema overview page (/<db_name>/<schema_id>/) . The event will be `opened_create_table` and the associated metadata will be `page: schema`.  
- The tables page  (/<db_name>/<schema_id>/tables/). The event will be `opened_create_table` and the associated metadata will be `page: tables`.  
The `page` attribute is helpful if we want to analyse the preferred page to create a Table.

kgodey: Makes sense @silentninja. I would do something like "source": "schema_page" since not all starting points are pages (e.g. the record selector might be a source).

demo/templates/demo/analytics.html
```diff
@@ -1,4 +1,7 @@
 <script async defer src="https://sa.mathesar.org/latest.js"></script>
+<script>
+    window.addEventListener("event", (e) => sa_event(e.detail.eventName, e.detail.metadata));
+</script>
```
pavish, line 4: `sa_event` would be undefined until the async analytics script loads, so the function will throw.

Event listeners ignore any errors within them, so we don't have to worry about a DOM wide error. Also, we don't dispatch any event during that time so we probably wouldn't have any lost events.

I would still like some run time checks to ensure `sa_event` is defined, and `e.detail` is defined, and print a console error if not, to ensure we're aware of any future issues during development. Since we can't use typescript here, I feel that a runtime check would be good. This is a pattern I'd like to discuss @seancolsen @rajatvijay 

@silentninja No changes needed in this PR. If we do introduce some pattern around this, we'll take that up later.

mathesar_ui/src/pages/schema/SchemaPage.svelte
```diff
@@ -81,6 +83,13 @@
   }
 
   $: isDefault = schema.name === 'public';
+  onMount(async () => {
+    logEvent('opened_schema', {
+      database_name: database.name,
+      schema_name: schema.name,
```
pavish, line 89: This does not have to be placed within onMount.

`onMount` only gets called after the component is rendered to the DOM. The `logEvent` function has nothing to do with the DOM, so we can directly specify it on the component.

pavish (changes requested):
@silentninja This looks good. I have a couple minor comments.

Regarding the events, I know that simple analytics automatically tracks route changes. We have only one additional custom event here `opened_schema`.

I assume we'd add more events to provide us better insight. Is there an issue to track the events we want?

mathesar_ui/src/utils/telemetry.ts
```diff
@@ -0,0 +1,13 @@
+export function logEvent(
+  eventName: string,
+  metadata: Record<string, string | number | boolean | Date>,
+): void {
+  window.dispatchEvent(
+    new CustomEvent('event', {
```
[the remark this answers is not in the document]
silentninja: `userAction` makes sense.

pavish (approved):
Looks good!

PR #2514, opened 2023-02-18
In-app upgrades

### LINKED ISSUE #2396: UX and frontend implementation for upgrades
### Notes
- [Current Figma design](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=3804%3A28864&t=HyNupYmgZ9PqjEGr-0)
- [Issue with user flow](https://github.com/centerofci/mathesar/issues/227)

### Tasks
- Finalize the user flow
- Request Figma UX changes if needed
- Implement the frontend (create additional GitHub issues if needed)

### Feasibility 
Watchtower, our docker image upgrade backend, doesn't report progress. The only progress reporting available to the frontend will be periodically calling some HTTP endpoint on the service container to check whether it is online or not.

The final UX should take this into account.

seancolsen:
Fixes #2396

## How the upgrade system works

🔍 _click screenshots to enlarge_

1. The backend puts a "tag name" in the common-data blob to inform the front end of the version number of the current release. Right now that tag name is `'0.0.0'`, but late it will become `'0.0.1'` and so on.

1. When an admin user loads any page, the front end checks localStorage to see if it has cached data about Mathesar releases.
    - If the cache is _not_ usable, we make two requests to the GitHub REST API, asking for data about the release with the current tag name, and asking for data about the latest available release. These requests don't have any authentication, which means that GitHub's strictest rate limits apply (only 60 per hr per IP address). After fetching, we store the data in localStorage, along with a timestamp and an `inputHash` string set to the value of the tag name used for the fetch.
    - Subsequent cache checks will invalidate the cache if: (A) the timestamp is older than 24 hours, or (B) the `inputHash` value does not match the current tag name.
    - If the cache _is_ usable, then we set a `ReleaseDataStore` instance in context at the app root, providing information about the current release and the latest available release.

1. The `ReleaseDataStore` shows whether an upgrade is available, and we read it through context in several high-level places throughout the app to display a red dot indicator on menu icons.

    

1. The Software Update page shows info about releases, and when an upgrade is available, the latest release has an "Upgrade..." button.

    

1. Clicking on "Upgrade..." opens a modal with some additional information about the upgrade process.

    

1. Clicking "Continue" performs the upgrade. The front end makes a `POST` request to `/api/ui/v0/upgrade/`. The backend then tells watchtower to pull the latest image from DockerHub and rebuild.

    

1. When the upgrade completes, the page reloads showing the Software Update page again. At this point, the back end provides a newer tag name which will invalidate the cached release data, prompting the front end to re-fetch release info.

    

1. At this point, each browser for each admin user will continue checking for updates at most once per day. The user can forcibly re-check for updates by clicking "Check for Updates" from the Software Update page.

## Error scenarios

Generally speaking, this whole process has loads of failure modes. I've done my best to consider the most obvious and important scenarios, but there may be some that I'm missing.

- If the front end can't get any data from GitHub about the current release or the latest release, then the Software Update page will look like this:

    

    That's how the page looks now, actually! All the better-looking screenshots from above were using mock data.

- If the front end can load data from GitHub about the latest release, but not about the current release, then the Software Update page will look like this:

    

- In theory, the `/api/ui/v0/upgrade/` endpoint could respond with a non-2xx HTTP status. In practice we have determined this to be very unlikely. Unfortunately, failures are still likely to result in a 200 status. Nonetheless, I've still implemented an error message for this scenario that displays within the modal and provides a "Retry" button.

    

## Backend work still needed

- The `/api/ui/v0/upgrade/` endpoint currently accepts `GET` requests. It should be changed to accept `POST` requests instead (and should reject `GET` requests). I've implemented this PR using `POST` in preparation for the backend change.

### Showing release data

If you run this PR locally right now it won't appear to do much. There are several challenges which complicate manually QA-ing this work:

- The backend is currently handing off a tag name of `'0.0.0'` and GitHub doesn't have a release with that tag name.
- Because GitHub doesn't have _any_ releases, it doesn't have a "latest" release either.

In working on this PR, I've been tricking the front end into using the release data from the VSCode project. I still have the relevant code committed in this PR but commented-out. You can toggle the comments [here](https://github.com/centerofci/mathesar/blob/7a116ca48167e511f773306ca571747423083cd2/mathesar_ui/src/App.svelte#L20) and [here](https://github.com/centerofci/mathesar/blob/7a116ca48167e511f773306ca571747423083cd2/mathesar_ui/src/3rd-party-apis/github-releases.ts#L14) to put it into "mock" mode. That's how I generated the screenshots above.

### Performing an upgrade

Challenges:

- Testing the upgrade process can only be done in live mode. Switching a local installation from dev mode to live mode is a real pain. I'd really rather not do it again if I can avoid it 😬
- The backend needs to change from GET to POST as mentioned above.
- I'm not sure if our images on DockerHub are ready for testing this out yet

**I have NOT manually tested that the upgrade process actually works**. I'm not sure how I'd do that sort of testing without having live releases.

## Deviations from Figma

- Generally, I took the Figma designs somewhat loosely.

- For the sake of time, I'd rather not go through all the differences one-by-one, but you might like to compare my screenshots above to Figma and raise any concerns you have with my approach.

## Implementation Notes

- I changed the "General" page to "Software Update". Kriti [approved](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$1f_8WTwxbVZT9mcZRu2FBZRiYHucGl-bDIZmvSSlYrw?via=matrix.mathesar.org&via=matrix.org) this.

- For the red dot, I modified the `Icon` component. I considered many different approaches here, including an action and a wrapper component. I don't love commandeering `Icon` for this purpose, but it was the quickest way forward that I could find. There is non-trivial code here which applies as path mask to the main icon design in order to have a sort-of negative space border around the dot. I made sure to keep the default code path through the component as lean as possible, so that we don't run all that computation for every icon instance.

- Unrelated to this issue, I also added a custom page title in `UserListingPage.svelte`

- The localStorage caching functionality is decoupled from the specific use-case here and abstracted into a `CachedFetchStore` class. We might be able to adapt this for other uses in the future.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@seancolsen I think the UI looks fine. The only change I'd like is to use a different icon instead of the sparkly magic icon for "new upgrade available" (because we are using that icon with the purple background items to indicate automatic configurations decided on by Mathesar).

seancolsen:
@kgodey Would you care to suggest a replacement icon? Ideally from [FontAwesome](https://fontawesome.com/search).

kgodey:
@seancolsen It looks like Figma uses `circle-info` (or some similar looking icon), I'd just go with that.

dmos62:
Looks good to me.

seancolsen:
@kgodey I've pushed 6d6267d4ceef6614c0c49b19a89f22e574098b76 to improve the icon. You can see the results here:

image

Note that this still doesn't exactly match what Figma has:

image

I'm just going with the closest available FontAwesome free icon. FontAwesome does this [circle-info](https://fontawesome.com/icons/circle-info?s=regular&f=classic) _Pro_ icon that more closely matches what Figma has, but we can't use that. We _could_ make a custom icon, but I'd rather not hold up this feature for that work right now.

kgodey:
That's fine @seancolsen.

mathesar_ui/src/3rd-party-apis/github-releases.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+async function getOneRelease(
+  endpoint: string,
+): Promise<GitHubRelease | undefined> {
+  const response = await fetch(`${baseUrl}${endpoint}`, { mode: 'cors' });
+  if (response.status !== 200) {
+    return undefined;
+  }
+  return (await response.json()) as GitHubRelease;
+}
```
pavish, line 34: I'd like us to use the `getAPI` method from our `requestUtils` here, and modify the method to support headers. 

Is there a reason why we use fetch here directly? If there is, I'd rather have another function within `requestUtils` that composes fetch.

mathesar_ui/src/api/upgrade.ts
```diff
@@ -0,0 +1,11 @@
+export const upgradeApi = {
+  /** @throws Error for some very limited failure scenarios. */
+  async upgrade(): Promise<void> {
+    const response = await fetch('/api/ui/v0/upgrade/', {
+      method: 'POST',
+    });
+    if (!response.ok) {
+      throw new Error('Unable to upgrade.');
+    }
```
pavish, line 9: This request will not succeed because there is no CSRF token here.

1. I have the same comment as earlier. I'd like to use the existing methods, or add another common method to `requestUtils` to make requests.
1. It's easier in the long run to handles all requests from a single file. eg., It'll help if we want to adding common request headers, log requests, be able abort them with CancellablePromise, handle errors in a common manner etc.,

mathesar_ui/src/component-library/icon/Icon.svelte
```diff
@@ -57,6 +60,7 @@
 
   let faClasses: string;
   $: faClasses = concatClasses(classes, flip, rotate);
+  $: dot = hasNotificationDot ? getDot(viewBoxWidth, viewBoxHeight) : undefined;
```
pavish, line 63: I agree adding Dot to Icon is the quickest solution at the moment, so we can go ahead with this but, I'd eventually like to decouple it from the Icon component, and make it more common.

Please open an issue to keep track of this.

mathesar_ui/src/pages/admin-update/ReleaseBox.svelte
```diff
@@ -0,0 +1,142 @@
...
+    <div class="left">
+      <div class="logo"><Logo /></div>
+      <div class="name">Mathesar</div>
+      <div class="version">{release.tagName}</div>
+    </div>
+    <div class="right">
+      <div class="date">Released {dateString}</div>
+      <a href={release.notesUrl} class="notes">
+        Release Notes <Icon {...iconExternalHyperlink} />
+      </a>
```
pavish, line 59: I think it'd be better if we open this in a new browser tab. The icon also indicates that this opens in an external tab.

mathesar_ui/src/utils/cachedFetchStore.ts
```diff
@@ -0,0 +1,184 @@
...
+type TimestampedValue<T> = {
+  /** ms since UNIX epoch */
+  timestamp: number;
+  inputHash: string;
+  value: T;
+};
+
+export class CachedFetchStore<T>
+  implements Readable<TimestampedValue<T> | undefined>
+{
```
pavish, line 24: Nice 😎

pavish (changes requested):
@seancolsen Looks good!

I have a few minor comments, I'll leave it upto you to decide if you'd like to fix them in this PR or do it later.
I've not tested the actual upgrade process but I went through the code and the mocked requests, and it looks good.

I would like [this comment](https://github.com/centerofci/mathesar/pull/2514#discussion_r1113977548) to be handled in this PR itself.

mathesar_ui/src/3rd-party-apis/github-releases.ts
seancolsen: Fixed in ade9233c021bdb54b819eb76acbda052b01c14a5, using the approach we discussed on the call. I'm still calling `fetch` here, just to keep the changes as minimal as possible.

mathesar_ui/src/api/upgrade.ts
seancolsen: Fixed in b13a9e316cb5a3185de43e25a821b6538f22ded0.

mathesar_ui/src/component-library/icon/Icon.svelte
seancolsen: #2542

mathesar_ui/src/pages/admin-update/ReleaseBox.svelte
seancolsen: Fixed in a6fd82052517802baffb6f784da4e607323d57da.

seancolsen:
@pavish this is ready for re-review now

pavish (approved):

PR #2574, opened 2023-02-24
install.sh usability fixes

### LINKED ISSUE #2523: Installation usability test - high priority issues
## Problem

We did a round of usability testing on the `install.sh`; issues were found. This issue collects the feedback together. 

### Code
- [x] ~User should press ENTER at the `DOCKER SETUP` step so they have time to read.~ Not sure what this means.
- [x] Multiple people have reported not being able to log in or not being able to create a schema once logged in. Please verify this works. (Note from @kgodey: we have root SSH access + admin user access to this instance, if it will help debug).
- [x] Determine how `admin` user is being created in prod installation setup. This is causing a bug if the user also tries to create a superuser named `admin`.
- [x] Use proper user for `pg_isready` to avoid `FATAL:  role "root" does not exist` spam in logs.
- [x] Figure out warning when starting containers:
  `RequestsDependencyWarning: urllib3 (1.26.14) or chardet (5.1.0)/charset_normalizer (2.0.7) doesn't match a supported version!`

### Copy
- [x] ~Make some troubleshooting items into prerequisites (e.g., docker versions)~ Already fixed
- [x] Change "terminal window" -> "terminal" in docs - Fixed in #2549 
- [x] ~Format important words in documentation specially~ - Unclear what "important" means
- [x] ~Make separate section with description of files output by `install.sh` (environment file)~ - Already fixed
- [x] ~Add separate "Maintenance" section in docs for upgrades and so on.~ - New "Administration" section under "Installation" should fix this.

### Notes
- #2550 was pulled out from here.
- Styling related issues are tracked in #2457 
- Some issues pulled into #2570

silentninja:
Related to #2523 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


install.sh
```diff
@@ -312,7 +312,12 @@ printf "
 Service is ready and healthy!
 Adding admin user to Django webservice now.
 "
-sudo docker exec mathesar_service python manage.py createsuperuser --no-input --username "$superuser_username" --email "$superuser_email" 2> >(grep -vi warn)
+while !  sudo docker exec mathesar_service python manage.py createsuperuser --no-input --username "$superuser_username" --email "$superuser_email"
+do
+  read -r -p "Choose a different admin username [mathesar]: " superuser_username
```
kgodey, line 317: Please change the default username to `admin`

mathemancer: With this fix, that won't work. I.e., this fix essentially just prompts a user to create a different user if the username is already taken. The username in question, in this case, is `admin`. We would assume that's already taken, since we still don't know how it's being created.

I suggest we remove this change, and merge the other changes to get those fixes in.

mathemancer (changes requested):
I think we should remove the changes to the `install.sh` script, and keep the other changes.

silentninja:
> I think we should remove the changes to the install.sh script, and keep the other changes.
@mathemancer Even though we added the prompt to ask for different super username in case it conflicts with existing user to fix a bug which we cannot reproduce. We also want the install.sh to be idempotent.

mathemancer (approved):
After discussion with @silentninja and doing some of my own poking around, the `admin` username already being created bug appears to have disappeared on its own.

Given some other issues with silently passing over the `admin` name already existing, we're going to merge this without the fix on that.

mathemancer:
> @mathemancer Even though we added the prompt to ask for different super username in case it conflicts with existing user to fix a bug which we cannot reproduce, we also want the install.sh to be idempotent.

Let's handle that separately. We need some care when handling user input in that case.

silentninja:
>  the admin username already being created bug appears to have disappeared on its own.

The only scenario where I had `the admin username already being created` bug  was when I changed from development mode to production mode. This is warranted as we create a `admin` user automatically in development mode.

PR #2622, opened 2023-03-02
Esc key reverts edits

### LINKED ISSUE #1116: Esc key should cancel edits instead of saving
## Steps to reproduce

1. Begin editing a text cell.
1. Modify the contents.
1. Press the Esc key.
1. Expect the cell to transition to select mode, restoring the value before you began making your edits.
1. Observe the cell to transition to select mode, saving the value you entered.

## Expected behavior
1. For stepped input columns (all string based columns, number etc.,) which contain an edit mode, "Esc" key in edit mode should revert the value to what is present in the DB and cancel unsaved edits.
1. For all other columns, "Esc" key would essentially do nothing since value change is based on different interactions.

## Additional context

- The current behavior was established in #1080.
- "Save of Esc" is consistent with AirTable.
- "Cancel on Esc" is consistent with LibreOffice and Google Sheets

dauinh:
Fixes #1116

**Technical details**

- Modify `SteppedInputCell`
- Assign value of currently selected cell to `lastSavedInput`
If user hits "Esc" key, save `lastSavedInput`
If user hits "Enter", "Tab" or clicks mouse at a different cell, save current input

Please let me know if my approach is appropriate and if I need to make any change. Thank you! 
(I'm sorry for the messy commit history, I'm still learning how to navigate git...)

**Screenshots**

https://user-images.githubusercontent.com/62679553/222344471-3558dc22-b810-4061-aacb-9571edd92050.mov

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/components/cell-fabric/data-types/components/SteppedInputCell.svelte
+++ b/mathesar_ui/src/components/cell-fabric/data-types/components/SteppedInputCell.svelte
@@ -32,6 +32,8 @@
 
   let cellRef: HTMLElement;
   let isEditMode = false;
+  let isLastInputSaved = false;
+  let lastSavedInput: string;
 
   $: formattedValue = formatValue?.(value) ?? value;
   $: matchParts = (() => {
@@ -63,7 +65,20 @@
     isEditMode = false;
   }
 
+  function saveInput() {
+    lastSavedInput = value;
+    isLastInputSaved = true;
+  }
+
+  function revertEdit() {
+    value = lastSavedInput;
+    isLastInputSaved = false;
+  }
+
   function handleKeyDown(e: KeyboardEvent) {
+    if (!isLastInputSaved) {
+      saveInput();
+    }
     switch (e.key) {
       case 'Enter':
         if (isEditMode) {
@@ -75,9 +90,11 @@
         // not prevented, the textarea gets a new line break. Needs more digging
         // down.
         e.preventDefault();
+        saveInput();
         break;
       case 'Escape':
         resetEditMode();
+        revertEdit();
         break;
       case 'Tab':
         resetEditMode();
@@ -85,6 +102,7 @@
           originalEvent: e,
           key: e.key,
         });
+        saveInput();
         break;
       case 'ArrowLeft':
       case 'ArrowRight':
@@ -129,6 +147,7 @@
     if (!isActive) {
       dispatch('activate');
     }
+    saveInput();
   }
 </script>
 


seancolsen:
@dauinh This PR got us part of the way there, but it still had some problems. An unnecessary request was sent to save the value after canceling. After canceling, subsequent edits had an incorrect initial value for the input. And some of the code needed some cleanup and type fixes.

In working to formulate feedback for you to fix these issues, it ended up taking me less time to make the fix myself than to explain all the changes required. Thus, I'm closing this PR in favor of my changes in #2643.

dauinh:
@seancolsen I think I have a better understanding of component's lifecycle now. Thank you for your feedback. I hope I can be more helpful in the future

dauinh:
@seancolsen I think I have a better understanding after reading about component's lifecycle now! Thank you for your feedback, I will try to write cleaner code and be mindful about type 💪

One question: why do you need `bind:value` in the child component?

seancolsen:
@dauinh 

> why do you need bind:value in the child component?

Before my PR, `SteppedInputCell` only had one mechanism to modify `value`: calling `dispatch('update', {value})` which _saves_ the value. After my PR that component has a new mechanism to update `value`: calling `revertValue` which mutates its `valueProp`. Adding `bind:value` to usages of this component turn that prop into a two-way binding, allowing the `value` mutation to propagate up to the parent.

Let's look in `TextBoxCell.svelte` and follow a hypothetical scenario where the user cancels an edit:

1. First we enter edit mode. This exposes the `TextInput` component.
1. Then the user modifies the value via `TextInput`. The modified value propagates up from `TextInput` to `TextBoxCell` and then propagates back down into `SteppedInputCell`.
1. When `SteppedInputCell` gets the modified value, it saves the original value too, in case we want to revert.
1. The user cancels the edit. `SteppedInputCell` propagates the original value up to `TextBoxCell` (due to the two way binding you're asking about) and `TextBoxCell` propagates that original value back down into `TextInput`, thus **resetting the value within `TextInput`** so that subsequent user edits will begin with the original value, not the modified value.

Two-way binding can get a bit ugly sometimes, and this is arguably one of those times. If we were using a framework with one-way bindings (e.g. React, Solid), we'd probably end up using `SteppedInputCell` to intercept value modifications before they ever reached `TextBoxCell`. That would be a bit cleaner. And while we could refactor this Svelte code to behave in that manner, I didn't deem the extra refactoring to be worthwhile at the moment.

dauinh:
@seancolsen Thank you so much for the very detailed comment!

PR #2652, opened 2023-03-08
Disable all inputs while the form is processing #2132

### LINKED ISSUE #2132: Modify form system to disable all inputs while the form is processing
- (A) When adding a new column, we have a small form in a dropdown (`NewColumnCell.svelte`). This form disables its inputs while loading. This pattern is nice and we should employ it at a lower level so it is used throughout the app.

- (B) On the Record Page, we use the form system to save field values for the record. This form does _not_ disable its inputs while the form is loading. It should.

- We should make changes within `src/components/form` so that (B) has the same user experience as (A) (disabled inputs while loading). Then we should remove the code from (A) which manually sets the inputs to disabled -- thereby giving (A) the same developer experience as (B) (no need to think about disabling inputs while the form is loading).

- Notice that on the Record Page the `id` field is disabled even when the form is _not_ loading. We should ensure that this behavior remains.

DanyaalMahmood:
Fixes #2132 

This PR changes the behavior of the form in the record page. When the form is loading the inputs are disabled. 

https://user-images.githubusercontent.com/110989303/223845785-9e749313-7e23-4b1f-b073-b40d339f7a36.mp4

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


DanyaalMahmood:
Hey @seancolsen could you please rerun the tests.

seancolsen:
@DanyaalMahmood I've re-run them. You can also run them locally by following the instructions in our [README](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/README.md)

DanyaalMahmood:
@seancolsen the tests are failing due to sqlalchemy throwing errors that there are too many clients already. I don't think it's related to anything that this PR changes. What should I do?

seancolsen:
@DanyaalMahmood Don't worry about those. I'll have someone from the backend team take a look. Nothing else for you to do now other than to wait until I have time to review this. There are a number of other things ahead of this task in my queue. I'll probably get a chance to review it sometime next week.

DanyaalMahmood:
Alright. Thanks

seancolsen (changes requested):
Your approach of exporting a writable store here is not great.

```ts
export const disableInputs = writable(false);
```

This makes that property global, meaning that if one page has two forms, all inputs on both forms will be disabled when only one form is being submitted.

Instead, we should have a writable property as part of the form object.

DanyaalMahmood:
@seancolsen I've made the changes you requested. I have given the form object a writable property for disabling inputs. Now only the inputs for the form being submitted will be disabled.

seancolsen:
Thanks @DanyaalMahmood. I'll review this when I get a chance. Hopefully it will be this week.

seancolsen:
@DanyaalMahmood You've done a good job with this PR. I spent some time looking at it but I wanted to take this in a slightly different direction. I'm closing this PR in favor of #2762.

I know I told you:

> we should have a writable property as part of the form object.

But once I took the time to get my head deeper into this problem, I realized that wasn't _exactly_ the cleanest way to do this. Your PR requires that we manually grab the `disableInputs` property out of the form object and pass it around as needed. This is some extra work at a high level that I'd rather handle automatically at a lower level to avoid mistakes.

In my PR, I am using a `disabled` store on each field object, and reactively updating that store when the form's `requestStatus` changes. From the perspective of consumers of the form system, very little is different from before. Consumers need not think about any disabled properties, so long as they're using the `<Field>` component because each field store now comes with the `disabled` property. I did need to ensure that forms without error handling were also updating the `requestStatus`, but this is something we would have needed to fix in your PR too. There were also a few input types that didn't properly respond to being disabled, so I had to fix that as well.

Apologies that I took over this work. You can still count this as a contribution for GSoC purposes.

PR #2668, opened 2023-03-10
Don't re-fetch records after deleting rows

### LINKED ISSUE #2463: Bug when adding and deleting rows.
## Description

Weird bugs while adding new rows and deleting them in the table.
When we add more that 2 new rows in the table at once, and then try to fill it it works. But deleting any row after these operation causes the page to buffer. Though we find the row deleted when we refresh the page.

## Expected behavior

There should be no infinite buffering when deleting rows after adding many of them.

## To Reproduce

I am attaching a video:
[Screencast from 13-02-23 01:10:04 AM IST.webm](https://user-images.githubusercontent.com/120925871/218361008-d44a2c05-d26f-412a-a8dc-242699a2128e.webm)

seancolsen:
Fixes #2463

## Context

- We have a section at the bottom of the table page for newly-added records.
- When the user hits the "Refresh" button at the bottom of the sheet, we re-fetch all the records in the table, re-incorporating newly-added records into normal part of the table.

## Before this PR

- After deleting a record, we re-fetch all the records in the table.
- But we _don't_ delete records from the newly-added records section. In most cases this is fine.
- But if the newly-added records section contains any records (that weren't deleted), then the sheet ends up with **duplicate** records. The keyed for-each block that we're using causes Svelte to throw an error in this case. 

## Suggested fix by Pavish

@pavish [said](https://github.com/centerofci/mathesar/issues/2463#issuecomment-1460278505):

> The removal from newRecords needs to happen before fetching the latest data.

If I understand you correctly, it sounds like you're suggesting that we perform the same actions, but in a different sequence. I don't think that will work. I'd like to take this a different direction

## After this PR

- After deleting a record, **we don't re-fetch** the records in the table.

## Thoughts

- Skipping the re-fetch solves the problem in a different way, and I like this better because it's simpler.

- It also eliminates the need for the `retainExistingRows` param, which has always seemed fishy to me.

- If we really do want to re-fetch, it's worth considering where the newly-added records _should_ go. Should they stay in the newly-added section? If so, why? If we really want to keep them in the newly-added section, then we'll need to look through the newly _fetched_ records, removing the duplicates from _there_. It doesn't seem worth it to me.

- Consider this UX situation...

    1. I'm doing some data clean up. I want to look through a bunch of rows manually and make some adjustments, editing cell values as I go.
    1. To make my work resumable (so I can pause without losing my place), I sort the sheet by a sensible column.
    1. Along the way, I make some edits to cell values _within that column that I sorted on_.
    1. Then I delete a row.

    Let's step back and consider what most users would expect at this point. While I was making all those edits to the data within my sorting column, none of the rows were re-sorted. I could continue about my work incrementally. If I delete a row, I would not expect anything to re-sort. If we re-fetch, we re-sort. This is part of why I don't think we should re-fetch. I don't think users will expect us to re-fetch.

- If there is a valid reason for re-fetching after deleting, then I'm open to closing this PR and making this fix by either:

    - _Fully_ re-fetching (including wiping the newly-added section)

    - Or removing newly added rows from the set of re-fetched rows (but this would be my least favorite approach)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/records.ts
+++ b/mathesar_ui/src/stores/table-data/records.ts
@@ -344,38 +344,30 @@ export class RecordsData {
       });
   }
 
-  async fetch(
-    retainExistingRows = false,
-  ): Promise<TableRecordsData | undefined> {
+  async fetch(): Promise<TableRecordsData | undefined> {
     this.promise?.cancel();
     const { offset } = get(this.meta.pagination);
 
     this.savedRecordRowsWithGroupHeaders.update((existingData) => {
       let data = [...existingData];
       data.length = Math.min(data.length, get(this.meta.pagination).size);
-
       let index = -1;
-      data = data.map((entry) => {
+      data = data.map(() => {
         index += 1;
-        if (!retainExistingRows || !entry) {
-          return {
-            state: 'loading',
-            identifier: generateRowIdentifier('dummy', offset, index),
-            rowIndex: index,
-            record: {},
-          };
-        }
-        return entry;
+        return {
+          state: 'loading',
+          identifier: generateRowIdentifier('dummy', offset, index),
+          rowIndex: index,
+          record: {},
+        };
       });
-
       return data;
     });
+
     this.error.set(undefined);
-    if (!retainExistingRows) {
-      this.state.set(States.Loading);
-      this.newRecords.set([]);
-      this.meta.clearAllStatusesAndErrors();
-    }
+    this.state.set(States.Loading);
+    this.newRecords.set([]);
+    this.meta.clearAllStatusesAndErrors();
 
     try {
       const params = get(this.meta.recordsRequestParamsData);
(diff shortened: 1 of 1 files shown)


pavish:
@seancolsen 

If we do not refetch after delete, pagination would break, and users would not be able to rely on it. It would be more confusing from the user standpoint and a pain to maintain from a development standpoint.

Assume we have 120 rows, with 2 pages (100 rows in first page, second one has 20). If the user deletes 20 rows in the first page, we need to shift pagination accordingly. Without refetching we can only show 80 rows in first page even thought the page size is configured to be 100. Then if the user tries to go to page 2, it'll result in an error/empty set because the API will return no results. If we try to reset count and restrict to a single page, the user cannot see the remaining 20 rows, unless they do a manual refresh.

> Fully re-fetching (including wiping the newly-added section)

This needs a UX re-thought for letting the user know what happens after a delete operation, and I'd like us to figure this out later. For 0.1.1, I'd very much like the bug to be fixed and keep the existing UX as it is.

I'd like this PR to be closed.

seancolsen:
Closing in favor of #2694

PR #2687, opened 2023-03-15
feat: added truncation to cell

### LINKED ISSUE #2345: Use Truncate component in Record Selector table cells
## Current behavior

- We have a `Truncate` component that we use in some places when we might not have enough space to show some text. It shows the full text on hover. For example, here it is in action in the column header cell

    image

- On the Table Page, when the value in a data cell is too long, we truncate it, but we don't use the `Truncate` component. That means that users don't get the popover behavior which shows the full value.

    image

    Whether or not users _should_ see the popover on table cells is debatable. We have #2342 to show the entire cell content within the table inspector, so I don't think we need a popover here.

- _However_ in the Record Selector we're also rendering table cells, but our needs are a bit different. Here we don't have a table inspector. Also, we don't let the user resize columns. So we do have a compelling need to show some sort of popover for truncated text, otherwise the user may have no other means of distinguishing records that appear to be duplicates from the first portion of their cell content.

    Currently, we have a rather crude mechanism for displaying popover text -- we put it into the `title` attribute.

    image

    We have some [logic](https://github.com/centerofci/mathesar/blob/25b1f8496488ab30c4d0e0da747193ae99a3a3ca/mathesar_ui/src/systems/record-selector/RecordSelectorDataCell.svelte#L14L38) which shows the title tooltip only when the text is sufficiently long.

- Further, the table widget in the Record Page also renders table cells and doesn't have a table inspector. Here, I think it would be nice to at least experiment with showing the popover.

## Desired behavior

- We should use the `Truncate` component for table cells.

- On the Table Page, we should pass a prop down to all the `Truncate` components, setting `passthrough={true}`

- Within the Record Selector, and on the Record Page, we should keep `passthrough={false}` (as it is by default), producing table cells which show popover content when the values are truncated.

Ary0405:
Fixes #2345 

Added truncation behaviour to the cells.

**Screenshots**
Screenshot 2023-03-15 at 11 42 52 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/components/cell-fabric/data-types/components/SteppedInputCell.svelte
```diff
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { createEventDispatcher, onMount } from 'svelte';
-
+  import { Truncate } from '@mathesar-component-library';
```
pavish, line 3: Other components from '@mathesar-component-library' are already imported in this component. Truncate can be added along with them. Refer lines 4 to 8.

mathesar_ui/src/components/cell-fabric/data-types/components/SteppedInputCell.svelte
```diff
@@ -168,15 +169,17 @@
...
+      <span
+        class="content"
+        class:nowrap={!isActive && !isIndependentOfSheet}
+        class:truncate={isActive && multiLineTruncate && !isIndependentOfSheet}
+      >
+        <slot name="content" {value} {formatValue} {matchParts}>
+          <CellValue value={formattedValue} {matchParts} />
+        </slot>
+      </span>
+    </Truncate>
```
pavish, line 182: @Ary0405 Notice in line 176 that the component only truncates values if the conditions specified there are met. You're enclosing the entire block within the Truncate component and enabling it by default, which will break usage. 

1. When the component is used independently of the sheet, it should not get truncated. eg., in Data explorer, upon selecting a cell, you can notice the content in the Cell tab in the right pane, which is truncated. It shouldn't be truncated there. The prop tracking this is `isIndependentOfSheet `.
2. `multiLineTruncate` prop is used to check if the truncation should take multiple lines into account. That logic needs to be handled using the `lines` prop from Truncate component.

Please get rid of the existing truncate related classes i.e. `nowrap` and `truncate`, and handle all logic using the Truncate component. Here's a rudimentary suggestion:
```suggestion
    <div class="content">
      <Truncate
        lines={isActive && multiLineTruncate ? 2 : 1}
        passthrough={isIndependentOfSheet}
      >
        <slot name="content" {value} {formatValue} {matchParts}>
          <CellValue value={formattedValue} {matchParts} />
        </slot>
      </Truncate>
    </div>
```

pavish (changes requested):

pavish:
@Ary0405 This PR has been open for a while and there hasn't been further progress or communication. I'm going to close this. Please feel free to open another PR if you intend to continue working on it.

PR #2690, opened 2023-03-16
Removing SQLAlchemy 1

mathemancer:
Related to #2737 

This removes SQLAlchemy imports from:
- `db.tables.operations.alter` and
- `db.tables.operations.drop`.

**Technical details**

This also adds some scaffolding functions in `db.connection` that can be used to avoid wrangling around function signatures until later in the process. These help us run the new functions with old SQLAlchemy objects.

It also has a couple of prototype database functions for 
- Getting the joinable tables, and
- Setting up the Mathesar Views we'll use for DQL operations in the future.

The SQL scripts setting these up need to be run manually if you want to use them, since they're not yet integrated into the overall project.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/sql/prototype.sql
```diff
@@ -0,0 +1,491 @@
...
+Execute the given command, returning the command executed.
+
+Not useful for SELECTing from tables. Most useful when you're performing DDL.
+
+Args:
+  command: Raw string that will be executed as a command.
+*/
+BEGIN
+  EXECUTE command;
+  RETURN command;
```
dmos62, line 39: What's the motivation for returning the string representation of a command that was executed? Not criticizing, just wondering.

db/sql/prototype.sql
```diff
@@ -0,0 +1,491 @@
...
+END;
+$$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
+
+
+-- Alter Table: LEFT IN PYTHON (for now) -----------------------------------------------------------
+
+-- Update table primary key sequence to latest -----------------------------------------------------
+
+CREATE OR REPLACE FUNCTION
+__msar.update_pk_sequence_to_latest(table_name text, column_ text) RETURNS text AS $$/*
```
dmos62, line 254: Is this compatible with multi-column PKs?

db/sql/prototype.sql
```diff
@@ -0,0 +1,491 @@
...
+  SELECT string_agg(format('%s AS c%s', quote_ident(attname), attnum), ', ')
+    FROM pg_attribute
+    WHERE attrelid=table_id AND attnum>0 AND NOT attisdropped
+  INTO viewcols;
+  RETURN __msar.exec_ddl(
+    'CREATE OR REPLACE VIEW %s AS SELECT %s FROM %s',
+    viewname, viewcols, __msar.get_table_name(table_id)
+  );
+END;
+$$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
```
dmos62, line 339: Do I understand this correctly? A Mathesar view is a reflection of a user table. A Mathesar view is algorithmically named according to the user table's oid. Its column values are the reflected user table's column values, and its columns' names are the corresponding reflected table columns' attnums. Would be great to put this in a docstring somewhere around here.

dmos62 (commented):
Interesting read. Didn't manage to get through everything tonight, will finish next week.

db/sql/prototype.sql
mathemancer: I think this will make testing and debugging easier. No other motivation. We don't really need this string for anything. It might be safer to remove it (I.e., avoid inadvertently picking up something sensitive in a log somewhere)

db/sql/prototype.sql
```diff
@@ -0,0 +1,491 @@
...
+The template is given in the first argument, and all further arguments are used to fill in the
+template. Not useful for SELECTing from tables. Most useful when you're performing DDL.
+
+Args:
+  command_template: Raw string that will be executed as a command.
+  arguments: arguments that will be used to fill in the template.
+*/
+DECLARE formatted_command TEXT;
+BEGIN
+  formatted_command := format(command_template, VARIADIC arguments);
```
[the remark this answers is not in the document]
mathemancer: That's correct. This helps with sanitization since it lets us parameterize everything we could want to, meaning other pieces of the machine will be able to sanitize things. I.e., most machinery is set up to parameterize and sanitize values passed, but not names. So if we want to have, e.g. `psycopg`'s SQL object help us sanitize strings, we need to reframe names as values. We can't make the assumption that names are safe since they're user-defined, so setting up the function this way will make it easier to sanitize things. (Hopefully.)

db/sql/prototype.sql
mathemancer: No, this is a direct port of the current python function (which also doesn't handle multi-column primary keys).

My plan is to get the current logic ported over while making minimal changes. I'm hoping to use pretty much the same test suite.

However, reviewing all this is a great opportunity to write up bugs and issues about exactly this sort of thing.

I'm pretty sure (hoping) that this function is only called when we're already sure it's a single-column primary key, namely when we've set the default `id` column as `SERIAL PRIMARY KEY` after the splitting logic runs. If that's wrong, we definitely need a bug report issue for this.

db/sql/prototype.sql
mathemancer: You understand correctly. I'll figure out some reasonable way to document the purpose of these views, and how they work.

dmos62 (approved):
Approving. Also asked for a non-critical improvement.

db/install.py
```diff
@@ -30,6 +35,8 @@ def install_mathesar(
         if database_created:
             print(f"Installing Mathesar on PostgreSQL database {database_name} at host {hostname}...")
             install.install_mathesar_on_database(user_db_engine)
+            with open(PROTOTYPE_FILE) as f, engine.begin() as conn:
+                conn.execute(text(f.read()))
```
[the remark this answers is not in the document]
mathemancer: Naturally. This bit was a little too prototypey.

dmos62:
Looking good. Ping me if you think this needs a last review.

mathemancer:
@dmos62 Code-wise, this is ready. I'm just setting up some of the surrounding (meta)issues and project stuff so that I can note them in the PR description here for the convenience of future code archaeologists.

PR #2694, opened 2023-03-16
Clear new records array after deleting a record

### LINKED ISSUE #2463: Bug when adding and deleting rows.

seancolsen:
Fixes #2463

## Notes

@pavish this is the simplistic approached we discussed via a call wherein we clear all the new records after deleting a row. This has the UX downside of also clearing the unsaved records, but I think that's okay for now.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/records.ts
+++ b/mathesar_ui/src/stores/table-data/records.ts
@@ -457,16 +457,9 @@ export class RecordsData {
       identifiersOfUnsavedRows.forEach((identifier) =>
         successRowKeys.add(identifier),
       );
-
-      if (failures.size > 0) {
-        if (successRowKeys.size === 0) {
-          throw Error('Unable to delete row!');
-        } else {
-          throw Error('Unable to delete some rows!');
-        }
-      }
     }
-    let shouldReFetchRecords = false;
+
+    let shouldReFetchRecords = successRowKeys.size > 0;
     if (primaryKeysOfSavedRows.length > 0) {
       // TODO: Convert this to single request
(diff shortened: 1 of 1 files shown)


pavish (approved):
@seancolsen Thanks, looks good!

I added a commit https://github.com/centerofci/mathesar/pull/2694/commits/0c092b7bd3c63e49720ff6d5f4fda36af1e34e64 to refetch the rows when there are unsaved rows which are deleted.

This is because when an unsaved row is deleted, we still set the new rows to empty. If there are saved new rows, they disappear from the table. But they aren't refetched so the user cannot view them until they do a full refresh.

The commit also fixes throwing errors when delete fails. That block seems to have shifted to a place where it wasn't called at all. 

Please take a look and we can merge this in.

pavish:
This is approved but I'm waiting on @seancolsen to take a second look at my commit before merging it in, so that we don't introduce bugs right before a release.

Issue #2737, opened 2023-03-23
RSQLA1: Move DDL Operations to SQL Functions


mathemancer:
This is the first project in the removing SQLAlchemy metaproject.

This issue and issues downstream (i.e., linked) are the source of truth for any work included in this project until further notice.

This project should proceed in phases. These have some overlap, but they should proceed generally in order.

## Phase One: Create DDL functions in database

During this phase, the goal is to create the bulk of needed DDL functions in the database. Examples of such functions can be found in the file `db/sql/0_msar.sql`. Please use the docstring style there.

During this phase, you can either use the function in a python wrapper (see examples in `db/tables/operations/alter.py`) or you can put a **`# TODO`** where the function will be wrapped. The heuristic should be whether you'll have do do a bunch of extra work with tests and dependencies to wrap the function in Python. If so, skip it. If not, go ahead and do the wrapper.

### Issues

- [x] db.columns.operations.alter
- [x] db.columns.operations.create
- [x] db.columns.operations.drop
- [x] db.constraints.operations.create
- [x] db.constraints.operations.drop
- [x] db.links.operations.create
- [x] db.schemas.operations.alter
- [x] db.schemas.operations.create
- [x] db.schemas.operations.drop
- [x] db.tables.operations.alter
- [x] db.tables.operations.create
- [x] db.tables.operations.drop
- [x] db.tables.operations.split

## Phase Two: Replace remaining Python DDL functions with thin wrappers

During this phase, the goal is to finish up using the DB functions via thin python wrappers. Minimal logic should remain in Python, only to link up function signatures. Avoid major function signature changes in this phase. If any DB functions are missing, write them (of course). It's completely fine to create scaffolding functions at this point.

Checking off an item in this phase implies there are no remaining `sqlalchemy` imports in the module.

### Issues

**TODO**: make issues. The below modules should all be covered. Note that you may bleed into other modules. Use your judgement. Avoid sprawl.

- [x] db.columns.operations.alter
- [x] db.columns.operations.create
- [x] db.columns.operations.drop
- [x] db.constraints.operations.create
- [x] db.constraints.operations.drop
- [x] db.links.operations.create
- [x] db.schemas.operations.alter
- [x] db.schemas.operations.create
- [x] db.schemas.operations.drop
- [x] db.tables.operations.alter
- [x] db.tables.operations.create
- [x] db.tables.operations.drop
- [X] db.tables.operations.split

## Phase Three: Refactor and clean up

[...]

PR #2743, opened 2023-03-24
Split move merge sql

mathemancer:
Related to #2737 

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns/operations/alter.py
+++ b/db/columns/operations/alter.py
@@ -4,6 +4,7 @@ from sqlalchemy import DefaultClause, text, DDL, select
 from sqlalchemy.exc import DataError, InternalError, ProgrammingError
 from psycopg2.errors import InvalidTextRepresentation, InvalidParameterValue, StringDataRightTruncation, RaiseException, SyntaxError
 
+from db import connection as db_conn
 from db.columns.defaults import NAME, NULLABLE
 from db.columns.exceptions import InvalidDefaultError, InvalidTypeError, InvalidTypeOptionError
 from db.columns.operations.select import (
@@ -283,21 +284,29 @@ def _batch_alter_table_rename_columns(table_oid, column_data_list, connection, e
 
 
 def batch_alter_table_drop_columns(table_oid, column_data_list, connection, engine):
-    table = reflect_table_from_oid(
-        table_oid,
-        engine,
-        connection_to_use=connection,
-        # TODO reuse metadata
-        metadata=get_empty_metadata(),
-    )
-    ctx = MigrationContext.configure(connection)
-    op = Operations(ctx)
-    with op.batch_alter_table(table.name, schema=table.schema) as batch_op:
-        for column_data in column_data_list:
-            column_attnum = column_data.get('attnum')
-            if column_attnum is not None and column_data.get('delete') is not None:
-                name = get_column_name_from_attnum(table_oid, column_attnum, engine=engine, metadata=get_empty_metadata(), connection_to_use=connection)
-                batch_op.drop_column(name)
+    """
+    Drop the given columns from the given table.
+
+    Args:
+        table_oid: OID of the table whose columns we'll drop.
+        column_data_list: The attnums of the columns to drop.
+
+    Returns:
+        A string of the command that was executed.
+    """
+    columns_to_drop = [
+        int(col['attnum']) for col in column_data_list
+        if col.get('attnum') is not None and col.get('delete') is not None
+    ]
+
+    if connection is not None and columns_to_drop:
+        return db_conn.execute_msar_func_with_psycopg2_conn(
+            connection, 'drop_columns', int(table_oid), *columns_to_drop
+        )
+    elif columns_to_drop:
+        return db_conn.execute_msar_func_with_engine(
+            engine, 'drop_columns', int(table_oid), *columns_to_drop
+        )
 
 
 def batch_update_columns(table_oid, engine, column_data_list):
@@ -308,6 +317,8 @@ def batch_update_columns(table_oid, engine, column_data_list):
         metadata=get_empty_metadata(),
     )
     _validate_columns_for_batch_update(table, column_data_list)
+    print(type(table_oid))
+    print(column_data_list)
     with engine.begin() as conn:
         _batch_update_column_types(table_oid, column_data_list, conn, engine)
         _batch_alter_table_rename_columns(table_oid, column_data_list, conn, engine)
--- a/db/sql/0_msar.sql
+++ b/db/sql/0_msar.sql
@@ -130,18 +130,27 @@ $$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
 
 CREATE OR REPLACE FUNCTION
 msar.get_column_name(table_id oid, col_attnum integer) RETURNS text AS $$/*
-Return the name for a given table, qualified or quoted as appropriate.
-
-In cases where the table is already included in the search path, the returned name will not be
-fully-qualified.
+Return the name for a given column in a given table.
 
 Args:
-  table_id:   The OID of the table.
+  table_id:  The OID of the table.
+  column_attnum:   The attnum of the column in the table.
 */
 BEGIN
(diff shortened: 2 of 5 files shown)


db/sql/0_msar.sql
```diff
@@ -130,16 +130,48 @@ $$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
...
+Return a string giving the appropriate name of the casting function for the target_type.
 
 Args:
-  table_id:   The OID of the table.
+  target_type: This should be a type that exists.
 */
+DECLARE target_type_prepped text;
 BEGIN
-  RETURN attname::text FROM pg_attribute WHERE attrelid=table_id AND attnum=col_attnum;
+  WITH unqualifier AS (
```
dmos62, line 164: What's the rationale for unqualifying a type? Does this not introduce the possibility for a collission?

db/sql/0_msar.sql
```diff
@@ -130,16 +130,48 @@ $$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
...
+  ), unspacer AS(
+    SELECT replace(unqualified_type, ' ', '_') unspaced_type
+    FROM unqualifier
+  )
+  SELECT replace(unspaced_type, '"', '_double_quote_')
+  FROM unspacer
+  INTO target_type_prepped;
+  RETURN format('mathesar_types.cast_to_%s', target_type_prepped);
 END;
 $$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
```
dmos62, line 176: What do you think about including in the docstrings some commentary about the algorithm? Some of these functions are not easy to parse.

db/sql/0_msar.sql
```diff
@@ -308,9 +348,64 @@ END;
...
+  FROM pg_attribute
+  WHERE attrelid=table_id AND ARRAY[attnum::integer] <@ attnums
+  INTO columns;
+  RETURN __msar.drop_columns(__msar.get_table_name(table_id), variadic columns);
+END;
+$$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
+
+
+CREATE OR REPLACE FUNCTION
+msar.drop_columns(schema_ text, table_ text, columns variadic text[]) RETURNS text AS $$/*
```
dmos62, line 391: Any reason why here you use `x_` while in the above functions you use `x_name` or `x_id`? Also, in the first dispatch, you use `columns variadic text[]`, which, given the signature, does convey what identifiers these are, but not saying it in the name is inconsistent with `table_name` and `column_id`. I think the naming consistency could be improved. I like the `[qualifier]_[id type]` scheme, as in `table_name`.

db/columns/operations/alter.py
```diff
@@ -283,21 +284,29 @@ def _batch_alter_table_rename_columns(table_oid, column_data_list, connection, e
...
-            column_attnum = column_data.get('attnum')
-            if column_attnum is not None and column_data.get('delete') is not None:
-                name = get_column_name_from_attnum(table_oid, column_attnum, engine=engine, metadata=get_empty_metadata(), connection_to_use=connection)
-                batch_op.drop_column(name)
+    """
+    Drop the given columns from the given table.
+
+    Args:
+        table_oid: OID of the table whose columns we'll drop.
+        column_data_list: The attnums of the columns to drop.
```
dmos62, line 292: Doesn't seem accurate; sounds like it's an iterable of attnums, but it's actually an iterable of dict-likes.

db/tables/operations/alter.py
```diff
@@ -66,8 +66,14 @@ def update_pk_sequence_to_latest(engine, table, connection=None):
     name = table.name
     column = table.c[constants.ID].name
     if connection is not None:
+        # The quote wrangling here is temporary; due to SQLAlchemy's query
+        # builder.
```
dmos62, line 70: Appreciate the comment :+1:

db/types/operations/cast.py
```diff
@@ -240,6 +240,7 @@ def assemble_function_creation_sql(argument_type, target_type, function_body):
     """
 
 
+# TODO Replace with SQL version (msar.get_cast_function_name) when refactoring.
```
dmos62, line 243: :+1:

dmos62 (changes requested):

db/sql/0_msar.sql
dmos62: If this does introduce the possibility for a collission, but you're making the compromise so as to not have to refactor our casting functions, I think this warrants a TODO code comment.

db/columns/operations/alter.py
mathemancer: You're correct; Fixed now.

db/sql/0_msar.sql
mathemancer: We can do it where needed, but I prefer to minimize the number of places where we need to maintain the same algorithm. 

That said, I've expanded the docstring for this function, as it is a bit difficult.

db/sql/0_msar.sql
mathemancer: Yes, but I acknowledge I've made some compromises. Perhaps we can improve by putting our heads together. There are a few places where these names are used. 
- In the code itself: For this, more verbose might be okay (more later on that)
- In the `Argument data types` column of the output from `\df the_func_name` in `psql`.
- In the docstring of the function (which appears in the `\df+ the_func_name` output along with the code).

Long variable names are a major liability for the last two. I've (not completely arbitrarily) chosen to hard-limit line widths to 100 characters. The reason for that is that the output of descriptive functions and commands in `psql` is always _tabular_, and there isn't much point in describing functions or names at all if it makes the tabular output too wide (since it'll be quite illegible). 

Long function signatures really strain the output of `\df` and `\df+` in this respect. In fact, naming variables at all is not that common. I chose to use names, but try pretty hard to lean on types so the user could tell what they refer to. (I.e., if there are two IDs, I specify which is which; If there are two names, I give, e.g., `schema_, table_` so the next person can at a glance figure out which `TEXT` entry is which, etc.) 

You're right that my current setup is a bit inconsistent. The inconsistency crept in when creating functions that map between ID types (e.g., you don't want to use `table_` for the OID and name in the same function (and can't, in SQL)). I ended up going with adding the `_id` suffix for numeric IDs, but not `_name` for names. Again, horizontal space is generally at a premium wherever the signature is actually viewed by a user (as compared to the developer).

mathemancer: @dmos62 I've tried a different variable naming setup, explained in a comment at the top of the `0_msar.sql` file. Let's see what you think of it.

Anish9901:
Unassigning myself as @dmos62 has already reviewed the PR.

db/sql/0_msar.sql
```diff
@@ -156,6 +156,13 @@ CREATE OR REPLACE FUNCTION
 msar.get_cast_function_name(target_type regtype) RETURNS text AS $$/*
 Return a string giving the appropriate name of the casting function for the target_type.
 
+Currently set up to duplicate the logic in our python casting function builder. This will be
+changed. Given a qualified, potentially capitalized type name, we
+- Remove the namespace (schema),
+- Replace any white space in the type name with underscores,
+- Replace double quotes in the type name (e.g., the "char" type) with '_double_quote_'
+- Use the prepped type name in the name `mathesar_types.cast_to_%s`.
```
dmos62, line 196: Appreciated :+1:

db/sql/0_msar.sql
```diff
@@ -28,6 +28,27 @@ well as its name identifer(s).
 
 Note that these identification schemes apply to the public-facing functions in the `msar` namespace,
 not necessarily the internal `__msar` functions.
+
+NAMING CONVENTIONS
```
dmos62, line 32: Naming looks great!

PR #2751, opened 2023-03-27
Update README.md with troubleshooting instructions

dmos62:
Stemming from conversation in Matrix about instructing new contributors to use issue trackers for troubleshooting more.

Goal is to make Matrix the last go-to for troubleshooting, or in other words funnel more of the troubleshooting into searchable platforms like Github or StackOverflow, so as to reduce repetitive troubleshooting sessions in Matrix.

If this is merged, might make sense to link to this in our contributor and/or gsoc candidate guides.


dmos62:
@kgodey asking for your review since I'm not sure who's responsibility README falls under. @pavish asking for your review since we discussed something along these lines.

README.md
```diff
@@ -110,5 +111,13 @@ We want to make existing database functionality more accessible, for users of al
 ## Contributing
 We actively encourage contribution! Read through [our contributing guidelines](https://wiki.mathesar.org/community/contributing) to get started.
 
+## Bugs and troubleshooting
+
+If you run into problems, here's what you can try:
+1. Narrow down the problem you're having to a minimal search phrase
+2. Search on [our issue tracker](https://github.com/centerofci/mathesar/issues/)
+3. Search on [StackOverflow using our tag `[mathesar]`](https://stackoverflow.com/questions/tagged/mathesar)
+4. If you think this is a novel problem, ask on one of those platforms or [on Matrix](#join-our-community)
```
pavish, line 120: I think it might be better if the default goto action for novel problems is to open tickets instead of matrix chats.

pavish (commented):
@dmos62 I think this sounds good, I had a suggestion. I'll defer to @kgodey to review and approve the PR.

README.md
kgodey: I agree with @pavish, an issue is much preferable to a Matrix chat. 

Also, I think this belongs in our wiki contributing guide, not in the README. Our README serves a wide audience and the goals are to provide an introduction to the project and to link out to appropriate resources for various sub-groups (contributors, users, people evaluating us, etc.) If we put primary documentation here, it will end up becoming long and unwieldy.

kgodey (changes requested):
Please see in-line comment.

README.md
dmos62: > If you think this is a novel problem, ask on one of those platforms or on Matrix

This does instruct asking on the other platforms first and Matrix second. We can uncollapse "those platforms" so that it's something like:

```suggestion
4. If you think this is a novel problem, ask on [our issue tracker](https://github.com/centerofci/mathesar/issues/), [StackOverflow](https://stackoverflow.com/questions/tagged/mathesar), or [on Matrix](#join-our-community)
```

Another option could be to split asking on Matrix as a 5th step that says something like:
```suggestion
4. If you think this is a novel problem, ask on [our issue tracker](https://github.com/centerofci/mathesar/issues/) or [StackOverflow](https://stackoverflow.com/questions/tagged/mathesar)
5. In case you're having trouble formulating a report, reach out [on Matrix](#join-our-community)(#join-our-community)
```

@pavish @kgodey how do these sound?

@kgodey putting this on the wiki makes sense.

dmos62:
Closing to be redone on the wiki.

README.md
kgodey: I like the second option with the 5th step. 

Also, do we actually want to be monitoring Stack Overflow? I'd consider removing that in favor of just the issue tracker. I don't have a strong opinion, though.

dmos62: @kgodey I don't have a strong opinion either; I think SO has its uses, and I expect people to increasingly use it, but we probably shouldn't actively encourage people to post on it at this time :+1:.

pavish (approved):
I think this looks good.

dmos62:
When I submitted this a few months ago (during GSoC, I think), I was under the impression that our bug report funnels could be improved. Now that some time has passed, I'm not sure that is still true. Anyone that's been dealing with bug reports has input? This isn't exactly documentation, but might be within your domain of interest, @seancolsen.

seancolsen:
I'm neutral on this PR. Seems fine to me. And I'm not sure if we need it. I'd be okay with merging or closing.

kgodey (approved):

kgodey:
I'm merging this, I don't see how this hurts anything and it could help.

PR #2750, opened 2023-03-27
[#2355] Part 1- Fixes for Icon type in filter, sort and group 

hemant-hc:
Related to issue #2355 

This PR fixes the icon type issues for Filter, Sort and Group functionality.

In Part 2 - The following changes will be included:
1. Fix icon type for Append Column dropdown in record summary.
2. Fix icon type for New Foreign Key Constraint dropdown.

**Technical details**

I previously created a draft PR with the changes in only the filter functionality to get an approval on the approach.

I made the necessary change as suggested from the comment on the draft PR to the code: Changing `[ConstraintType]` (tuple) to `ConstraintType[]` (array). And implemented this approach to further solve the icon issues for sort and and partially for group functionality.

**Screenshots**

For Filter functionality:

image

For Sort functionality:

image

For Group functionality when a group is already selected:

image

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
[...]

PR #2768, opened 2023-03-29
[#2355] Part 2- Fixes for Icon type for Unique Constraint, Foreign Key Constraint and Record Summary

### LINKED ISSUE #2355: FK column names need link icon in some UI locations
## Good

These places in the UI show a link icon for an FK column. That's good.

image

image

image

## Bad

These places show that same column, but they have the wrong icon. They're showing the icon for the column's _type_, but they don't take the FK into account. We should change these places to show the correct icon.

image

image

image

image

image

image

image

## Implementation notes

- At first glace, this issue may appear rather trivial, but it's not. This is not a matter of swapping out one icon for another. Due to some complex requirements from the Data Explorer, column objects are somewhat polymorphic, which is part of what makes this icon work a little tricky and bug-prone.

- The `ColumnName.svelte` component receives a `DisplayColumn`, but several other types are able to satisfy `DisplayColumn`. The `ProcessedColumn` type is the one that will have sufficient info to render the correct icon. We need to go looking through usages of `ColumnName.svelte` and try to get the value supplied to its `column` prop to have a `constraintsType` property. If you can feed in a `ProcessedColumn` object then that will do the trick. But if that's tough then you might be able to compute a `constraintsType` property and tack that onto whatever is already being passed in.

- Notice how we have `SelectProcessedColumns.svelte`. That's good because it always accepts a `ProcessedColumn` object (which will have the correct icon). For single-select, we have `SelectColumn.svelte`. It might be a good idea to make a new component `SelectProcessedColumn.svelte` (which is a single-select for `ProcessedColumn` objects) and use that instead of `SelectColumn` where possible.

hemant-hc:
Fixes #2355 

This PR makes the following changes:
1. Fix icon type for AppendColumn dropdown under Record Summary, new Unique constraint and new Foreign Key constraint.
2. Refactoring: Move `getColumnConstraintTypeByColumnId` function to `columnUtils.ts` and modify its usages.

**Technical details**

-
**Screenshots**

For AppendColumn dropdown under Record Summary:

image

For new Foreign Key constraint:

image

For new Unique Key constraint:

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/systems/table-view/table-inspector/record-summary/AppendColumn.svelte
+++ b/mathesar_ui/src/systems/table-view/table-inspector/record-summary/AppendColumn.svelte
@@ -16,6 +16,7 @@
     if (linkFkType) {
       return true;
     }
+    return false;
   }
 
   function getIcon(column: Column) {
--- a/mathesar_ui/src/systems/table-view/table-inspector/record-summary/AppendColumn.svelte
+++ b/mathesar_ui/src/systems/table-view/table-inspector/record-summary/AppendColumn.svelte
@@ -2,8 +2,11 @@
   import { ButtonMenuItem, DropdownMenu } from '@mathesar-component-library';
   import { getTabularDataStoreFromContext } from '@mathesar/stores/table-data';
   import type { Column } from '@mathesar/api/types/tables/columns';
-  import { iconAddNew, iconTableLink } from '@mathesar/icons';
-  import { getColumnIconProps } from '@mathesar/utils/columnUtils';
+  import { iconAddNew, iconConstraint, iconTableLink } from '@mathesar/icons';
+  import {
+    getColumnIconProps,
+    getColumnConstraintTypeByColumnId,
+  } from '@mathesar/utils/columnUtils';
 
   const tabularData = getTabularDataStoreFromContext();
   $: ({ processedColumns } = $tabularData);
@@ -11,19 +14,18 @@
   export let columns: Column[];
   export let onSelect: (column: Column) => void;
 
-  function checkForForeignKey(column: Column) {
-    const linkFkType = $processedColumns.get(column.id)?.linkFk?.type;
-    if (linkFkType) {
-      return true;
-    }
-    return false;
-  }
-
   function getIcon(column: Column) {
-    const isForeignKey = checkForForeignKey(column);
-    if (isForeignKey) {
+    const constraintsType = getColumnConstraintTypeByColumnId(
+      column.id,
+      $processedColumns,
+    );
+    if (constraintsType?.includes('primary')) {
+      return iconConstraint;
+    }
+    if (constraintsType?.includes('foreignkey')) {
       return iconTableLink;
     }
+
     const icon = getColumnIconProps(column);
     return Array.isArray(icon) ? icon[0] : icon;
   }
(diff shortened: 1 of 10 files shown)


mathesar_ui/src/systems/table-view/table-inspector/record-summary/AppendColumn.svelte
```diff
@@ -1,13 +1,29 @@
...
+      return true;
+    }
+    return false;
+  }
+
   function getIcon(column: Column) {
+    const isForeignKey = checkForForeignKey(column);
+    if (isForeignKey) {
+      return iconTableLink;
+    }
```
hemant-hc, line 26: @seancolsen Looking at the implementation of `getColumnIconProps`, if there is a constraint type `primary` on the column then its icon takes precedence over the foreign key icon.

```
export function getColumnIconProps(
  _column: DisplayColumn,
): IconProps | IconProps[] {
  if (_column.constraintsType?.includes('primary')) {
    return iconConstraint;
  }

  if (_column.constraintsType?.includes('foreignkey')) {
    return iconTableLink;
  }

  ...
)
``` 
This behaviour is different from other dropdowns where foreign key icon takes precedence (as per my first [PR](https://github.com/centerofci/mathesar/pull/2750)).

**Do we want to give precedence to the primary icon?**

In that case we can modify `getColumnConstraintTypeByColumnId` in `columnUtils.ts` like this:
```
export function getColumnConstraintTypeByColumnId(columnId: number, processedColumns: Map<number, ProcessedColumn>) {
  const processedColumn = processedColumns.get(columnId);
  const constraintsType = Array.from(
    new Set(
      [
        ...(processedColumn?.exclusiveConstraints ?? []),
        ...(processedColumn?.sharedConstraints ?? []),
      ]
        .filter((constraintType) => !!constraintType)
        .map((constraintType) => constraintType.type),
    ),
  );
  return constraintsType;
}
```

seancolsen: @hemant-hc yes, I think it makes sense to give precedence to the primary icon. Feel free to make those changes in this PR

hemant-hc:
@seancolsen, i have made the required changes. Can you please review.

Screenshots:
image

image

seancolsen (approved):
All these changes are good, except for the changes in `SelectColumn.svelte` (I reverted those). You were getting `tabularData` from context which is not great (for reasons that are a little hard to explain quickly). It turns out that the `SelectColumn` component is only used within one other component -- `NewFkConstraint.svelte`. I made some comments in that component detailing the next steps to fix the problem. But this is super low priority. We expect that form to be rarely used. And even then, the bug here is super minimal. So I'm fine with closing the ticket associated with this PR and leaving the outstanding task as a simple TODO in the code.

I had actually begun making the refactoring changes myself within `NewFkConstraint.svelte` until I realized that I was going to run into a tricky git merge conflict with my work in #2794, so I backed off. I still have those changes and I'll probably open a PR myself to finish it up next week.
 
Thanks for your work on this, @hemant-hc!

Issue #2663, opened 2023-03-09
Speed up column reflection


dmos62:
Refer to this comment for profile at the time of writing: https://github.com/centerofci/mathesar/issues/2544#issuecomment-1450451207

Currently, `mathesar/state/django.py::reflect_columns_from_tables` (or more specifically `mathesar.models.base._set_default_preview_template` call inside) is taking 1.37/4.40s. See if that can be improved.

thesujai:
@dmos62 Thanks for assigning me this issue.
Here is a change i want to make which I think will optimize the function:

the `mathesar.models.base._set_default_preview_template` we can find `preview_settings` before if condition and simply do a if check  by `preview_settings.customized`. Also inside the if condition instead of of creating a new variable for template i.e 'preview_template' we can directly store template by `preview_settings.template=compute_default_preview_settings(table)`
and later save the settings
Like this:
    preview_settings = table.settings.preview_settings
    if not preview_settings.customized:
        preview_settings.template = compute_default_preview_template(table)
        preview_settings.save() 

This is better because we were using tables.settings twice now we are doing only once
 Let me know your views on this

dmos62:
@thesujai please continue working on the issue, but I'll unassign you so that other contributors can work on this in parallel. That's important since we have only so many issues available for GSoC candidates to contribute to on the backend.

Aritra8438:
Hi, @dmos62, while working on this issue, I have noticed that, ```compute_default_preview_template``` is taking **much longer** when it comes to **column reflection**.

Currently, ```compute_default_preview_template```  filters all the columns and checks which among them might be a preview column. **So, **if we have n tables with n columns**, it will take almost **O(n<sup>2</sup>)** time to filter out all the columns of a specific table and assign one as the preview column.** Moreover, this operation will be performed each time a **user creates a new table or creates a new column or deletes one.** 

One specific optimization we can do is to store the already calculated preview templates for each table and return them directly upon function call. We need to check if preview column still exists for a table as it might be deleted by the user. If it exists, we return it, else we calculate again to assign a new preview template. **This can be done using dictionary data-structure available in python and the time complexity will reduce to **O(log(n))**.**

Upon implementing this logic, there was a noticable change in **time-consumption of adding a column** (My local database contains around 18 tables with around 50 columns):

**Before implementation:**

[...]

PR #2770, opened 2023-03-29
Added Caching Logic of Preview Template to Speed Up Column Reflection

Aritra8438:
This PR optimizes #2663 by adding caching logic to speed up column reflection 

**Technical details**

Currently, ```compute_default_preview_template```  filters all the columns and checks which among them might be a preview column.
To optimize, ```CACHED_PREVIEW_FOR_TABLES = {}``` dictionary is used to memoiz the result of time-consuming ```compute_default_preview_template```function and returns it on function call.

**Further has been discussed in the issue thread, [here](https://github.com/centerofci/mathesar/issues/2663#issuecomment-1489217521).**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models/base.py
```diff
@@ -921,6 +922,11 @@ def _set_default_preview_template(table):
 
 
 def compute_default_preview_template(table):
+    if table.id in CACHED_PREVIEW_FOR_TABLES:
+        cached_preview_column = CACHED_PREVIEW_FOR_TABLES[table.id]
+        if Column.current_objects.filter(id=cached_preview_column):
+            preview_template = f"{{{cached_preview_column}}}"
+            return preview_template
```
dmos62, line 929: You've introduced a cache and you're performing two operations on it: setting and getting. Make those into discrete methods to improve maintainability.

mathesar/models/base.py
```diff
@@ -921,6 +922,11 @@ def _set_default_preview_template(table):
 
 
 def compute_default_preview_template(table):
+    if table.id in CACHED_PREVIEW_FOR_TABLES:
+        cached_preview_column = CACHED_PREVIEW_FOR_TABLES[table.id]
+        if Column.current_objects.filter(id=cached_preview_column):
```
dmos62, line 927: Use `.get` instead of `.filter`.

mathesar/models/base.py
```diff
@@ -57,6 +57,7 @@
 
 
 NAME_CACHE_INTERVAL = 60 * 5
+CACHED_PREVIEW_FOR_TABLES = {}
```
dmos62, line 60: Rename to something that expresses that you're not caching the actual preview, but just the id of the column from which preview will be built.

mathesar/models/base.py
```diff
@@ -938,6 +944,7 @@ def compute_default_preview_template(table):
         preview_column = primary_key_column
 
     if preview_column:
+        CACHED_PREVIEW_FOR_TABLES[table.id] = preview_column.id
```
dmos62, line 947: Why only cache when a preview column exists?

dmos62 (changes requested):
In one of the comments I pointed out that you're performing two operations on this cache you've introduced. There's a third one that's missing: clearing the cache. Look at the `mathesar.state` namespace. Define this cache there and make sure it's cleared with the rest of the state by `mathesar.state.reset_reflection`.

mathesar/models/base.py
Aritra8438: Sure, I will do that.

mathesar/models/base.py
```diff
@@ -921,6 +922,11 @@ def _set_default_preview_template(table):
 
 
 def compute_default_preview_template(table):
+    if table.id in CACHED_PREVIEW_FOR_TABLES:
+        cached_preview_column = CACHED_PREVIEW_FOR_TABLES[table.id]
```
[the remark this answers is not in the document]
Aritra8438: Okay.

Aritra8438:
> On the general concept: introducing state (a cache in this case) is risky,

Hi, @dmos62, can you explain a little bit? 
Is it general backend not to use state in backend? 
What is a good practice then?

mathesar/models/base.py
dmos62: Oh, use `.exists()` then. Good for readability.

dmos62:
@Aritra8438 generally speaking most problems we experience on the backend are caused by state. Specifically speaking, I'll walk back what I said somewhat. You have a column existance check when getting from the cache, so it seems like an outdated cache can't cause panic, since whatever column reference it returns, it will be for an existing column. I'll make a final judgment about this during a further review.

Aritra8438:
> @Aritra8438 generally speaking most problems we experience on the backend are caused by state. Specifically speaking, I'll walk back what I said somewhat. You have a column existance check when getting from the cache, so it seems like an outdated cache can't cause panic, since whatever column reference it returns, it will be for an existing column. I'll make a final judgment about this during a further review.

Hi, @dmos62, thanks for the explanation. I am planning on adding a few more optimizations. **Should I make it draft proposal**?

The optimizations I have planned:
- If we check first two columns of the filtered queryset, one of them can be definitely assigned as a preview column,
- If we delete the preview column, **maintaining two caches (or dictionaries), one for** ```table.id``` -> ```preview_column.id``` **and the other for**  ```preview_column.id``` -> ```table.id```, will make the algorithm even faster.

Aritra8438:
Hi, @dmos62, as you have said, **I have discretized the setting and getting method.**

Also I added **two more methods**, ```remove_preview_columnId_from_cache``` and ```remove_tableId_from_cache```.

 - **remove_preview_columnId_from_cache**: will remove the preview column from the cache, if user deletes the column.
 - **remove_tableId_from_cache**: will remove the table from the cache, if the table is deleted.

**Now, clearing the cache.** If we add ```clear_preview_cache``` inside ```reset_reflection```, it will have no effect, as every column addition or deletion calls ```reset_reflection``` and therefore will clear the cache. **Can we clear the cache on something more infrequent like login and logout?**

Whenver a column or table is added or deleted, the cache will be modified accordingly. **So, the cache should never grow beyond the number of tables in the database.** Based on that, where would it be a good option to place the function?

***Why 2nd cache (GET_TABLE_ID_FROM_PREVIEW_COLUMN_ID) ?***
   It will check if the preview column is deleted in O(1) time. Whenever the preview column is deleted, this cache will return the table associated with it and then the table will be deleted from the 1st cache.

Here is the performance before:

before_new

Here is the performance after:

after_new

If **state implementation** is risky, I will try to optimize the other part of the function after this.

mathesar/state/base.py
```diff
@@ -2,6 +2,40 @@
 from mathesar.state.metadata import reset_cached_metadata, get_cached_metadata
 from mathesar.state.cached_property import clear_cached_property_cache
 
+CACHED_PREVIEW_COLUMN_ID_FOR_TABLE_ID = {}
+GET_TABLE_ID_FROM_PREVIEW_COLUMN_ID = {}
+
+
+def get_preview_columnId_for_tableId(tableId):
```
dmos62, line 9: Use only lowercase in method names: `column_id` instead of `columnId`.

mathesar/api/db/viewsets/columns.py
```diff
@@ -207,6 +208,7 @@ def dependents(self, request, pk=None, table_pk=None):
 
     def destroy(self, request, pk=None, table_pk=None):
         column_instance = self.get_object()
+        remove_preview_columnId_from_cache(columnId=column_instance.id)
```
dmos62, line 211: You don't want to do selective state updates. Only full resets via `reset_reflection`. Well, not completely full. `reset_reflection` takes a database as argument.

mathesar/state/base.py
```diff
@@ -2,6 +2,40 @@
...
+    if tableId in CACHED_PREVIEW_COLUMN_ID_FOR_TABLE_ID:
+        columnId = CACHED_PREVIEW_COLUMN_ID_FOR_TABLE_ID[tableId]
+        GET_TABLE_ID_FROM_PREVIEW_COLUMN_ID.pop(columnId, None)
+        CACHED_PREVIEW_COLUMN_ID_FOR_TABLE_ID.pop(tableId, None)
+
+
+def clear_preview_cache():
+    CACHED_PREVIEW_COLUMN_ID_FOR_TABLE_ID.clear()
+    GET_TABLE_ID_FROM_PREVIEW_COLUMN_ID.clear()
+
```
dmos62, line 38: This file is not the appropriate place for this logic. E.g. routines related to `clear_dj_cache`, and it itself, aren't defined in this file.

mathesar/state/base.py
```diff
@@ -20,6 +54,7 @@ def reset_reflection(db_name=None):
 
     Note, this causes immediate calls to Postgres.
     """
+    # clear_preview_cache()
```
dmos62, line 57: This is probably exactly where you want to clear the "preview cache".

dmos62: Btw, consider a more descriptive name for the cache.

dmos62 (changes requested):

dmos62:
> If we add clear_preview_cache inside reset_reflection, it will have no effect, as every column addition or deletion calls reset_reflection and therefore will clear the cache.

Column addition and deletion aren't the only places that trigger reflection. But, you might be right. Our approach may be fundamentally flawed. If we need a cache during reflection, but its validity is dependent on reflection having been performed, we cannot use it. And, we are not willing to introduce a separately updated cache (as you have done here), for technical debt management reasons.

There might not be a way to fix this approach (though I encourage you to think on that). We'll take this PR into consideration when reviewing your GSoC proposal, so don't worry about not being eligible.

dmos62:
> Can we index the database based on the model index reference `table`?

Not sure, you'd have to check Django docs.

Aritra8438:
Closing this.
- `SQLAlchemy` will evetually be removed.
- This PR had many problems (check the comments).

PR #2640, opened 2023-03-06
Add client side validations for email and username

IamEzio:
Fixes #2555 

I went through the official [documentation](https://docs.djangoproject.com/en/4.1/ref/contrib/auth/) on how Django validates the `username` and `password` field. 

 - Username should be `150 characters or fewer. Usernames may contain alphanumeric, _, @, +, . and - characters.`

 - Passwords can  `be arbitrarily long and can contain any character.` Additionally, we can validate if the inputs for `New Password` and `Confirm Password` fields are same wherever applicable.
 
 - Email inputs should pass the regex ``/^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/``. Took help from [this](https://stackoverflow.com/a/46181) thread.

**Screenshots**

https://user-images.githubusercontent.com/89375125/223125439-95f1e2d9-0336-4e33-9cfd-a4e33e8ecc60.mp4

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
[...]

PR #1932, opened 2022-11-14
Add new abstraction to help with form validation, and use it within the "Extract Columns" modal

seancolsen:
## Why I opened this PR

So I got a little carried away on a somewhat tangential front end project during the past few work days, and now I want to come clean and introduce it. I know we ought avoid tangents like this right now, being laser focused on the live demo and release, so I'm definitely guilty of following my curiosity too far down the rabbit hole here. Sorry :grimacing:  But I'm hoping it will pay off...

Last week I was working on building the UI for the record summary template configuration. It's a relatively simple form, but I was frustrated with having done the same thing over and over again to implement client-side form validation. And many times I simply _haven't_ implemented validation because doing so is kind of tedious.

I had an idea for a small helper utility, and when I started chasing that idea it blossomed into something that I think is really cool!

## What's in this PR

- 6d7bb94bc94ca9d15f4f2508b25af2b8cd235a7c adds the the form validation system, and makes a few tweaks to our component library so the form system can better utilize it.
- 15cc7a01deb507a42a07eae6feb0734da8715ac4 refactors the "Move Columns" modal to use the new system, improving the client-side validation quite a bit.

## Why I didn't choose an existing form validation library

I actually spent quite a bit of time looking at existing solutions and took some [notes](https://hackmd.io/YPvOIu7rSHKx6eOB9UnIYA) on my research. I can review them with anyone who is interested. I spent some time playing with some of these libraries and didn't like any of them.

## An example form

Let's look at the form in my tests...

```ts
type Beverage = 'juice' | 'water' | 'beer';

const name = requiredField('', [uniqueWith(['Jim'], 'Name already taken.')]);
const age = requiredField<number | undefined>(undefined, [
  min(16, 'You must be at least 16 to register.'),
]);
const beverage = requiredField<undefined | Beverage>(undefined);

const form = makeForm({ name, age, beverage }, [
  comboInvalidIf(
    [age, beverage],
    ([a, b]) => b === 'beer' && a < 21,
    'Too young for beer.',
  ),
]);
```

```html
<Field label="Name" field={name} />
<Field label="Age" field={age} input={NumberInput} />
<Field label="Beverage" field={beverage} />
<FormSubmit {form} onCancel={form.reset} onProceed={save} initiallyHidden />
```

image

image

image

image

image

## How it works

1. You make a new field:

    ```ts
    const name = requiredField('', [uniqueWith(['Jim'], 'Name already taken.')]);
    ```

1. `name` is now a `Field`, which mostly looks like this:

[...]

PR #2794, opened 2023-04-07
Clean up some form validation code

seancolsen:
## Notes

I happened to come across some code in `components/form/validators.ts` that I thought could use some clean up, so this is a quick PR I made on a whim. Some of the mess stemmed from #2640, but I think that the committers in that PR were also following patterns that I established when writing the `min` and `max` functions [here](https://github.com/centerofci/mathesar/pull/1932/commits/6d7bb94bc94ca9d15f4f2508b25af2b8cd235a7c#diff-a36aaf0f5b3845b86ae5c3682d46af136e1cf46c574608f9ec49f8ee9ce60aabR75-R93).

After writing those `min`/`max` validators, I later ended up making [improvements](https://github.com/centerofci/mathesar/pull/2080/commits/9fc9ef790139807c05e8905637135644ca58ba07#diff-a36aaf0f5b3845b86ae5c3682d46af136e1cf46c574608f9ec49f8ee9ce60aabR87-R112) to the validation system so that validators on required fields would not need to bother dealing with `null` and `undefined`. Validators on required fields are assured to run after the "required" validation has already run. That commit also has some nice TS to ensure that validation functions passed to `requiredField` are typed different from validation functions passed to `optionalField`. I think I should have cleaned up the `min`/`max` validators when making these improvements. I did that cleanup in this PR now.

In general, I'm trying to make sure that, as we continue to build out more validators in `validators.ts`, we don't need to put a bunch of boilerplate in each of them.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/systems/users-and-permissions/UserDetailsForm.svelte
```diff
@@ -34,16 +34,13 @@
...
-    ),
-    matchRegexAllowNullUndefined(
+    maxLength(150, 'Username cannot be longer than 150 characters.'),
+    matchRegex(
       /^[A-Za-z0-9_@.+-]*$/,
       'Username can only contain alphanumeric characters, _, @, +, ., and -.',
     ),
   ]);
-  $: email = optionalField(user?.email ?? '', [validateEmailAllowEmpty()]);
+  $: email = optionalField(user?.email ?? '', [isEmail()]);
```
pavish, line 43: @seancolsen The case here is that email should allow empty values.

Try opening an user with an existing value for the email field, and clear it. This will result in a validation error.

pavish (changes requested):

mathesar_ui/src/systems/users-and-permissions/UserDetailsForm.svelte
seancolsen: Thanks for pointing this out!

I added 24a2a67af4e86bb15dfeb47f25705b1ab449eb9a to fix this at the form level. Now the validation functions for `requiredField` _and_ `optionalField` use **filled** values. The validation logic returns early any time the value is considered empty. For required fields, empty values give a validation error. For optional fields, empty values bypass all subsequent validation.

pavish (approved):

PR #2813, opened 2023-04-13
Make sure multi-db tests are passing

### LINKED ISSUE #2007: Make sure backend supports using multiple user databases at the same time
Related: #2006 

With the exception of a few tests, backend has been using only a single user database. Now we'll need to support switching between user databases and possibly using multiple user databases at the same time. This issue is about verifying that backend supports that.

dmos62:
Related #2007

Some maintenance on multi-db tests, in preparation for db-switching functionality on frontend. Biggest change is reenabling tests that were marked for skipping for some reason.

Also did basic smoke checking locally.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
These tests assume that all DBs will be in the settings dict, and that we'll be able to modify the settings dict at runtime.

You can't modify the settings dict at runtime and expect things to work. They were flaky because of that (there's some kind of thread-dependence issue), and so I disabled them. I _don't_ think they should be reenabled. The reason I didn't delete them outright was that I assume we'll want to make valid versions of these at some point.

https://docs.djangoproject.com/en/4.2/topics/settings/#altering-settings-at-runtime

dmos62:
@mathemancer Do you have ideas about how to fix this? I can only think of redoing how we define user databases (which would be nice), but I'm not sure we have time for that right now.

I think it's important to test multi-db support, since we'll now have user-facing functionality dependent on that. Do you agree?

If we don't know how to fix this without making a project out of it, and the problem is only that these tests *may* have random false failures, maybe let's enable them and see what happens? If false failures do happen, and they mess with our workflow, we can then use that as justification to invest more time into this.

mathemancer:
@dmos62 I agree that now is not the time for modifying how we define DBs (though I also agree that we should fix that ASAP).

Couldn't we come up with a way to make the tests accurate to our current setup? I.e., we're not supporting modifying the settings dict at runtime (in memory) anyway, so why should we test that?

Wouldn't it be possible to just use a different settings file during testing, or even set up separate tests for this that use a multi-db settings file? What happens if they actually have multiple DBs defined in their settings file? I'm not actually sure what will happen to our tests, or these tests in that case.

dmos62:
@mathemancer we've been setting up and tearing down the Django `settings` dict-like for every single test call though: https://github.com/centerofci/mathesar/blob/develop/mathesar/tests/conftest.py#L115

Are you sure that fixing this now is warranted?

kgodey:
Django docs for overriding settings in tests: https://docs.djangoproject.com/en/4.2/topics/testing/tools/#overriding-settings, just in case it helps.

mathemancer:
> @mathemancer we've been setting up and tearing down the Django `settings` dict-like for every single test call though: https://github.com/centerofci/mathesar/blob/develop/mathesar/tests/conftest.py#L115. That doesn't make this pattern more supported by our stack, but it is pretty widespread in our testing suite.
> 
> Are you sure that fixing this now is warranted? Setting up a separate testing suite with its own Django configuration sounds like overkill (that's what would be required, if I'm not mistaken). I've tested running Django with multiple user dbs hardcoded in the config and it worked fine, for what it's worth.

I think turning these tests back off would be far preferable to setting a trap in our testing suite to forget about. We'd inevitably end up building more on top of the same misguided setup, making the problem worse and worse.

> Django docs for overriding settings in tests: https://docs.djangoproject.com/en/4.2/topics/testing/tools/#overriding-settings, just in case it helps.

Unless I'm misunderstanding, that page indicates that using those methods on `DATABASES` will not have the expected effect.

dmos62:
@mathemancer

> a trap in our testing suite to forget about

[We have an issue to track this](https://github.com/centerofci/mathesar/issues/2816), and when we refactor how we define user dbs, all of this will have to be changed, so these few tests, and whatever tests we derive from them, aren't something we can forget, unless we forget to fix user db definition.

As for it being a trap, I'm still not sure what the problem is here, but you mention it being about multi-threading, so how about we restrict these tests to running in the same `pytest-xdist` runner?

Or, better yet, since we don't use xdist at the moment, how about we create an issue for reenabling `pytest-xdist` and make it blocked on #2816?

kgodey:
> when we refactor how we define user dbs

Why are we refactoring how we define user DBs?

dmos62:
Me and @mathemancer resolved this in sync:

1. Add code comments in the tests, explaining the context and possibility for false failures;
2. Create an issue for reenabling `pytest-xdist` and make it blocked on https://github.com/centerofci/mathesar/issues/2816;
3. Merge this.

dmos62:
@kgodey does the issue description in https://github.com/centerofci/mathesar/issues/2816 provide enough context? We can continue discussing there, if needed.

kgodey:
@dmos62 I looked at #2816 but there's not enough context there. I don't understand the underlyng problem, and so I can't evaluate whether redoing the way we define user DBs the only/best solution to it. It might be better to keep the GitHub issue focused on explaining the problem rather than presuming the solution.

dmos62:
@kgodey I'll respond in #2816.

PR #2836, opened 2023-04-24
Fix caret out of view when using Input on Chrome

### LINKED ISSUE #2758: Text caret sometimes scrolled out of view when typing into FormattedInput on Chrome
## Description
* Edit a number cell, type in a large value
* Notice that the value gets updated and once the length exceeds what's visible on the screen, the cursor is not visible and further updates are not shown.
* The entered value does get updated but it's not visible until the user exits edit mode and enters edit mode again.

https://user-images.githubusercontent.com/11850603/228086282-d1b716e8-ff6f-434d-b27d-69a13bc52851.mov

## Expected behavior
* The cursor should always be visible and newly entered values should be shown.

I remember this was working as expected when the feature was implemented. This is a possible regression which was introduced later, but before we released.

dev-cj:
Fixes #2758 

**Technical details**
Fixed the issue by scrolling left on element in `handleChildBeforeInput`

**Screenshots**

https://user-images.githubusercontent.com/47112778/234040531-67d16be9-99fd-4a3e-b641-d9a3592ea54f.mp4

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
@dev-cj Thanks for this PR. Unfortunately this is not enough to fully fix this issue.

Did you read my [comment](https://github.com/centerofci/mathesar/issues/2758#issuecomment-1487587271) on the ticket? I said:

> - Putting some code like this...
> 
>     ```js
>     if (element) {
>       element.scrollLeft = 1000;
>     }
>     ```
> 
>     ...after the `setSelectionRange` _kind of_ works. It works if you're adding text to the end of the contents. But if you have a long string and you put the caret at the _start_ of the string and hit backwards-delete, then we have the reverse of the current problem (albeit a more rare bug, but a bug nonetheless).

The fix is this PR is effectively identical to that example code that I already explained will not fully fix this issue. If my comment is confusing, I'll be happy to clarify or answer any specific questions you have. I think we need a more involved fix or a more clever solution to this problem.

dev-cj:
> @dev-cj Thanks for this PR. Unfortunately this is not enough to fully fix this issue.
> 
> Did you read my [comment](https://github.com/centerofci/mathesar/issues/2758#issuecomment-1487587271) on the ticket? I said:
> 
> > * Putting some code like this...
> >   ```js
> >   if (element) {
> >     element.scrollLeft = 1000;
> >   }
> >   ```
> >   
> >   
> >       
> >         
> >       
> >   
> >         
> >       
> >   
> >       
> >     
> >   ...after the `setSelectionRange` _kind of_ works. It works if you're adding text to the end of the contents. But if you have a long string and you put the caret at the _start_ of the string and hit backwards-delete, then we have the reverse of the current problem (albeit a more rare bug, but a bug nonetheless).
> 
> The fix is this PR is effectively identical to that example code that I already explained will not fully fix this issue. If my comment is confusing, I'll be happy to clarify or answer any specific questions you have. I think we need a more involved fix or a more clever solution to this problem.


@seancolsen 
I missed your comment there. Thanks for your work and explanation 🙇🏻 
The getSelection should move the caret into view of the input. But it's not working on Chrome only, even the Svelte example similar to this approach is broken on chrome but works fine in firefox. [example](https://svelte.dev/repl/f45d38ec017749ddbcb86310d1102f82?version=3.46.4) Change the text limit to 50. 

```
I tried to fix this by using the caret offset in pixels and scrollLeft approach.
get caret offset from left (in pixel) => only scroll left when offset is more than the width of the element (scroll only the difference)
```
 

### This is how far I was able to solve it.
1. Keeps caret in view 
2. Backwards delete also works fine when the caret is in start

## Caveats
1. When editing a long string at far-right position but before last characters. Rest of string from right gets out of view. Because caret offset is still more than the element width.



https://user-images.githubusercontent.com/47112778/235203149-b2f84b25-1a16-43d1-a519-5534b4720a25.mp4

seancolsen:
@dev-cj I'm confused by your last comment. What do you see as the next step for this PR?

I just looked it over again, and I'm still observing the problem I already described. For example In Chrome backwards delete from the start of the text scrolls to the end of the text.

[Screencast_20230428_143420.webm](https://user-images.githubusercontent.com/42411/235226806-49baa46d-8126-4cdb-bce3-5a0dfda80fb8.webm)

dev-cj:
@seancolsen Pushed the changes which will reflect the changes shown in my last video [comment](https://github.com/centerofci/mathesar/pull/2836#issuecomment-1527814738)

I think to get rid of the issue, changing the logic in FormattedInput is necessary, For e.g, formatting, parsing, and throwing error in onchange instead of before input.

seancolsen:
@dev-cj I spent a long time playing with your changes and working on this today.

Some reactions to your changes:

- Good idea, looking for a third party library. I looked around too, and couldn't find any that seemed better than [textarea-caret-position](https://github.com/component/textarea-caret-position) which you picked.

- If we're going to include third party code without modification (as you seem to have proposed), then it would be best to import the package through NPM instead of copy-pasting the code (as you did).

- However, there's some code in that NPM package that I don't really like, especially for our use case here. I ended up doing some work to re-implement this somewhat from scratch, but drawing heavily on the strategy used by that NPM package.

- For our use case, we only care about `input` &mdash; we don't care about `textarea`. And we only care about width (no height). Those assumptions allow us to eliminate a significant amount of the complexity inherent in that code from NPM.

- Some of the behavior of that code from NPM felt a bit buggy to me. For example, the cloned element would line wrap it the text content contained consecutive dashes (`--`) which caused very strange scroll jumps.

- Also, the vast majority of the time the user is typing into the input, there is no need to create a DOM element and measure it. That sort of overhead makes me a bit nervous from a performance standpoint. The code I wrote bypasses this DOM manipulation for several "happy paths" that should be more common than the "worst case scenario". This logic is documented in the new `README.md` file which I added to explain how this code works.

Since I've added a lot of new code here, I'd like to get another core team member to review it. @pavish I'm requesting you for this (in part because you're the issue author).

seancolsen:
Since I've added a lot of new code here, I'd like to get another core team member to review it. @pavish I'm requesting you for this (in part because you're the issue author).

The easiest way to try out these changes is to type into the record summary template editor here:

image

Just put `{id}` somewhere into your text so that the input doesn't complain at you.

There are several branches of logic to consider when testing this, and I've described them in the new `README.md` that I added as part of this PR. The most complex branch is (D). Make sure to hit that in your test. Some suggestions:

- Fill the input with text. Use arrow keys to put your caret at the start. Then use arrow keys to put your caret at the position within the text nearest to the right edge of the input. Then type a wide character such as W. Before this PR, the `W` would end up off screen. After this PR, the text content is automatically scrolled into view.

- Try similar things at the other side of the input with backspace.

- Play with backspace, forwards delete, and entering text. Play with putting your caret somewhere and then using your mouse scroll or touch to scroll the content without changing the caret position &mdash; then modify the content.

I've hammered on this pretty hard myself and I can't get it to do anything weird.

dev-cj:
@seancolsen  
Thank you for your valuable comments https://github.com/centerofci/mathesar/pull/2836#issuecomment-1558312319


- As I couldn't solve the issue for all cases with direct use of the code as an NPM package, I pasted the code such that it will be helpful to solve it with tweaks and added references to the original code.
- Thank you for the redo of the implementation, thorough documentation, and helpful images. Going through your changes and seeing your approach has been a valuable learning experience for me. I will incorporate these into my skill set.

pavish (approved):
Looks good to me!

PR #2891, opened 2023-05-29
Prevent default credentials being used for removed DBs

### LINKED ISSUE #2877: Database removed from .env file does not get removed
## Description
* Add a new database to .env file, start Mathesar. It gets added.
* Remove the database and restart Mathesar. Notice that it does not get removed.
* I added two databases `mt1` and `mt2`. I removed `mt2`. This was the response from the server within `common_data.databases`:
   ```
    [
      {
        "id": 1,
        "name": "mt2",
        "deleted": false,
        "supported_types_url": "http://localhost:8000/api/ui/v0/databases/1/types/"
      },
      {
        "id": 2,
        "name": "mt1",
        "deleted": false,
        "supported_types_url": "http://localhost:8000/api/ui/v0/databases/2/types/"
      }
    ]
   ```
* Note that 'mt2' has id 1 (I'm not sure if this was the id it had before), and that the value of `deleted` is `false`.

pavish:
Fixes #2877 

The root cause behind the issue:
* If a user DB is configured once, the name of the DB is always stored in our `mathesar_database` table. We have a bool column `deleted` to mark if the DB is removed.
* In our case, the user adds a DB and then removes it. So, it's present in the `mathesar_database` table, but not in the `DATABASES` dict.
* When attempting to connect to a DB which isn't present in the `DATABASES` dict, the credentials of the `default` db was being used. This was implemented to support dynamically created DBs for the demo, however, this also affects non-demo environments.
* If the credentials for the `default` DB matches with the DB which is removed, the connection succeeds. This is the case when all of them run on the same server with the same user, which is the default setup of our docker-compose installation methods.
* Since the connection succeeds, the `deleted` flag isn't modified, and we are able to access DBs which the user expects to have been removed. These DBs also show up on the UI.

The changes in this PR:
* Only uses the credentials of the `default` DB if Mathesar is running in demo mode.
  * We have separated all our demo logic with a middleware and separate settings file. My commit checks the `MATHESAR_LIVE_DEMO` setting in our main application code. I don't see an easier alternative to move this check within the demo module. 
* Passes down only the DBs that have `deleted:false` to the frontend via `common_data`. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/database/base.py
+++ b/mathesar/database/base.py
@@ -13,7 +13,10 @@ def create_mathesar_engine(db_name):
     try:
         credentials = _get_credentials_for_db_name_in_settings(db_name)
     except KeyError:
-        credentials = _get_credentials_for_db_name_not_in_settings(db_name)
+        if hasattr(settings, 'MATHESAR_LIVE_DEMO') and settings.MATHESAR_LIVE_DEMO:
+            credentials = _get_credentials_for_db_name_not_in_settings(db_name)
+        else:
+            raise
     return engine.create_future_engine_with_custom_types(**credentials)
 
 
--- a/mathesar/views.py
+++ b/mathesar/views.py
@@ -33,11 +33,11 @@ def get_schema_list(request, database):
 
 
 def _get_permissible_db_queryset(request):
-    qs = Database.objects.all()
+    qs = Database.objects.filter(deleted=False)
     permission_restricted_qs = DatabaseAccessPolicy.scope_queryset(request, qs)
     schema_qs = Schema.objects.all()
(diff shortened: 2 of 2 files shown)


mathemancer (approved):
Looks good, tests are passing. Thanks!

mathemancer (approved):
Looks good, tests are passing. Thanks!

PR #2932, opened 2023-06-09
Add `median` aggregation function

### LINKED ISSUE #2931: Support for `median` aggregation function while summarizing columns

Aritra8438:
Fixes #2931 

**This pull request adds support for `Median` aggregation as a transformation step in the data explorer.** 

User will be able to:
- Get `median` of any column based on `Group By` clause.
---
**Technical details**

I have made the following changes:
- Add `Median(DBFunction)` class to support median aggregation
- Add `median` suffix to the output alias of the aggregated column.
- Add `Median` to the list of transformation steps in the data explorer.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
+++ b/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
@@ -15,7 +15,7 @@ function mapAllInputTypesToOneOutputType(
   );
 }
 
-function mapInpuTypesToCorrespondingOutputTypes(): AbstractTypeSummarizationFunctionsResponseValue['inputOutputTypeMap'] {
+function mapInpuTypesToTheSameOutputTypes(): AbstractTypeSummarizationFunctionsResponseValue['inputOutputTypeMap'] {
   return Object.fromEntries(
     Object.values(abstractTypeCategory).map((t) => [t, t]),
   );
@@ -44,7 +44,7 @@ const functionsResponse: AbstractTypeSummarizationFunctionsResponse = {
   },
   median: {
     label: 'Median',
-    inputOutputTypeMap: mapInpuTypesToCorrespondingOutputTypes(),
+    inputOutputTypeMap: mapInpuTypesToTheSameOutputTypes(),
   },
 };
 
--- a/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
+++ b/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
@@ -15,7 +15,7 @@ function mapAllInputTypesToOneOutputType(
   );
 }
 
-function mapInpuTypesToTheSameOutputTypes(): AbstractTypeSummarizationFunctionsResponseValue['inputOutputTypeMap'] {
+function mapInputTypesToTheSameOutputType(): AbstractTypeSummarizationFunctionsResponseValue['inputOutputTypeMap'] {
   return Object.fromEntries(
     Object.values(abstractTypeCategory).map((t) => [t, t]),
   );
@@ -44,7 +44,7 @@ const functionsResponse: AbstractTypeSummarizationFunctionsResponse = {
   },
   median: {
     label: 'Median',
-    inputOutputTypeMap: mapInpuTypesToTheSameOutputTypes(),
+    inputOutputTypeMap: mapInputTypesToTheSameOutputType(),
   },
 };
 
--- a/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
+++ b/mathesar_ui/src/stores/abstract-types/operations/summarization.ts
@@ -15,6 +15,12 @@ function mapAllInputTypesToOneOutputType(
   );
 }
 
+function mapInpuTypesToCorrespondingOutputTypes(): AbstractTypeSummarizationFunctionsResponseValue['inputOutputTypeMap'] {
+  return Object.fromEntries(
+    Object.values(abstractTypeCategory).map((t) => [t, t]),
+  );
+}
+
 const functionsResponse: AbstractTypeSummarizationFunctionsResponse = {
   distinct_aggregate_to_array: {
     label: 'List',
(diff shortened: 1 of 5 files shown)


mathesar_ui/src/stores/abstract-types/operations/summarization.ts
```diff
@@ -36,6 +42,10 @@ const functionsResponse: AbstractTypeSummarizationFunctionsResponse = {
       [abstractTypeCategory.Duration]: abstractTypeCategory.Duration,
     },
   },
+  median: {
+    label: 'Median',
+    inputOutputTypeMap: mapInpuTypesToCorrespondingOutputTypes(),
+  },
```
seancolsen, line 48: @Aritra8438 I would expect `median` to look the same as `sum`. Why is it different?

Aritra8438: @seancolsen, median aggregation will be available for all column types.

If we have an array of `Text` i.e. ```["John","Alice", "Paul"]```, 
```"John"``` will be the median.

seancolsen: Ok, interesting. So any type that can be sorted can have a median? I didn't know that. Sounds reasonable though. In that case we can keep this structure. Can you change the name of that function from `mapInpuTypesToCorrespondingOutputTypes` to `mapInputTypesToTheSameOutputType`?

Aritra8438: Sure @seancolsen.

I would like to explain a bit if this helps you.

There are two related PostgreSQL functions, `PERCENTILE_DISC` and `PERCENTILE_CONT`, and this [SO blog](https://stackoverflow.com/questions/23585667/percentile-disc-vs-percentile-cont) explains how they are different.

If we assume two numeric arrays,

| array  | `PERCENTILE_DISC` | `PERCENTILE_CONT`|
| ------------- | ------------- | --------------|
| `[1,2,3,4]`  | 2  | 2.5 |
| `[1,2,3]`  | 2  | 2 |

`PERCENTILE_CONT` interpolates when there are multiple central elements, whereas `PERCENTILE_DISC` returns the first one.

We have used `PERCENTILE_DISC` and so will be able to calculate the median for all column types (as there won't be any scenario to interpolate two texts)

seancolsen: I see. Hmmm. Did @mathemancer weigh in on this choice? You highlight an interesting problem. `PERCENTILE_DISC` and `PERCENTILE_CONT` can give different results. This might confuse some users. Personally, I don't see the utility of computing the median of a non-numeric type. Why would someone want this? If we don't actually need that, then I'd lean towards `PERCENTILE_CONT` because I think it will align better with expectations. What do you think @mathemancer? Of course we could have two separate median functions, but that may be overkill.

Aritra8438: Yes, @seancolsen. @mathemancer suggested that we should go for `discrete` one because,
- `Median` then will be available for all column types.
- It will be more interpretable initially.

> because I think it will align better with expectations.

We may selectively use `PERCENTILE_CONT` for numeric columns in the backend. @mathemancer, should I do this?

seancolsen: > selectively use PERCENTILE_CONT for numeric columns in the backend

Oh, good idea!

Aritra8438: @seancolsen, I mistakenly changed the name to `mapInputTypesToTheSameOutputType-s`. 
When I was fixing the error, I noticed the function name might describe that we are mapping to the same output type. (e.g. every type to `number` or `text`)

seancolsen (approved):
Looks good!

mathesar_ui/src/stores/abstract-types/operations/summarization.ts
mathemancer: > I see. Hmmm. Did @mathemancer weigh in on this choice? You highlight an interesting problem. `PERCENTILE_DISC` and `PERCENTILE_CONT` can give different results. This might confuse some users. Personally, I don't see the utility of computing the median of a non-numeric type. Why would someone want this? If we don't actually need that, then I'd lean towards `PERCENTILE_CONT` because I think it will align better with expectations. What do you think @mathemancer? Of course we could have two separate median functions, but that may be overkill.

I think we should use the discrete definition. 

There are multiple acceptable definitions of a median, and the most useful ones in the field (i.e., when dealing with real data from a sample) generally return a value from the set from which the sample is pulled. For example, for a median of integers, you want an integer. For a median of a set of monetary values, you don't want half cents. If you're pulling from a sample of birthdates reported as timestamps, you want a timestamp. `PERCENTILE_DISC` enables all of these, but `PERCENTILE_CONT` doesn't. In fact, the latter can only return intervals or double precision floats.

To be fair, there are also some use cases where the interpolation is useful, but I think those are rarer outside of middle school textbooks, where I acknowledge 'median' is usually defined using the interpolated version. Mainly, if you have a double precision sample, and so you lose nothing by using `PERCENTILE_CONT`.

I suppose the eventual solution could be to build a custom aggregation function that interpolates, but still keeps the result in the sample set.

Regarding non-numeric types, I agree that strings are less useful (though you can still use this to divide your data into halves, e.g., to assign lists of names to two sales people). However, there are some types that _are_ useful:
- time/date types (this is a big one). It's not uncommon at all to want an "average" birthdate or the like. Median and mode are the only ones that suffice for that, and the median (in the postgres implementation) has to be discrete.
- arrays of numeric types. Less common, but certainly has use cases.

If all of this is unconvincing, we could discuss more tomorrow during our call.

seancolsen: I'm convinced. Thanks @mathemancer

mathemancer (approved):
Okay, I think this looks good as well.

PR #2907, opened 2023-06-02
Improve readability of docs

kgodey:
TBD

## Checklist
 
- [ ] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [ ] My pull request targets the `develop` branch of the repository
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #2937, opened 2023-06-09
Update documentation styles for active and hover

ghislaineguerin:
This PR addresses @seancolsen comments on [#documentation](https://github.com/centerofci/mathesar/pull/2907) regarding on hover and active states for navigation and table of contents.


kgodey:
I only have one thing to add to @seancolsen's review, which is that on dark mode, the "current page" styling is the brightest thing on the page and draws my attention too much. It looks like the primary action item on the page, whereas it should just be a subtle indication of "you are here".

Screen Shot 2023-06-12 at 5 22 41 PM

kgodey (changes requested):
Please see main comment in thread.

seancolsen (approved):
Looks good! Thanks.

kgodey (approved):
Looks great, @ghislaineguerin!

Issue #2202, opened 2023-01-05
Allow deleting a record with CASCADE


seancolsen:
## Problem

- You can't delete a record which is referenced by other records

## Desired outcome

- We should give users a way to delete such records

## Design challenges

- At what point does the user make the decision to use CASCADE or not use CASCADE?
- How does the user communicate that choice to Mathesar?
- How does Mathesar communicate the behavior of CASCADE to the user?

seancolsen:
I just want to say that I bump into this somewhat regularly and it's really frustrating. I would like to elevate the priority of this issue, so I'm moving it to the "High Priority" milestone in hopes that we can address it at some point in the coming months.

mathemancer:
My suggestion is to start with helping users with setting the `ON DELETE` behavior. Options are:

- `NO ACTION` (default; throws error when attempting to delete, current behavior)
- `RESTRICT` (disallows deleting referenced record. Throws more precise error)
- `CASCADE` (delete referencing row (not just the fkey value)
- `SET NULL` (just clear the referencing value if there's no conflicting constraint, leave the rest of the row)
- `SET DEFAULT` (does what it says on the tin. If there's no default set on the referencing column, then this is equivalent to `SET NULL`).

I think we could consider also having a piecemeal "fix this one value" flow in case a user wants one `ON DELETE` action generally, but wants to use a different action for some value. But, I'd defer that until after we have the `ON DELETE` behavior sorted out.

seancolsen:
@mathemancer made some additional [comments](https://github.com/mathesar-foundation/mathesar/issues/4178#issuecomment-2618062274) in a duplicate issue:

> To reiterate what I said in [#4164](https://github.com/mathesar-foundation/mathesar/issues/4164):
> 
> - Letting them do this from the deletion operation itself is not a feature, it's a footgun.
> - We should instead help them set up their data model with the appropriate `ON DELETE` setting (apparently `CASCADE`) for their usecase.
> - We could, maybe, offer help for that operation when dependent rows exist along the lines of "consider modifying the relevant foreign constraint" with a link or some helpful UI indicator.
> 
> To put a fine point on it: If I were a DBA who installed software that let my users access the database, and that software made it "convenient" for users to delete data in violation of the data model, I'd be apoplectic.
> 
> If a user has a data model with `ON DELETE RESTRICT`, it is not our place to make it convenient to ignore that.

seancolsen:
@mathemancer 

[...]

PR #2985, opened 2023-07-01
Updated frontend to send a single bulk delete request instead of one request for each record

### LINKED ISSUE #2701: Update frontend to send a single bulk delete request instead of one request for each row
* When rows are deleted from a table, we currently send a delete request for each row from the frontend.
* As of https://github.com/centerofci/mathesar/issues/2237, there is a new backend API to delete records in bulk.
* We should update the frontend to send a single request to delete the records.

thesujai:
Fixes #2701

**Technical details**

- This PR is a improvement of #2746 based on the suggestion there by @pavish 
- Modified `mathesar_ui/src/api/utils/requestUtils/DeleteAPI` with optional argument for handling bulk delete request.
- Implemented the modified function in `mathesar_ui/src/stores/table-data/records/deleteSelected` for sending single bulk delete request instead of one request for each record.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/api/utils/requestUtils.ts
+++ b/mathesar_ui/src/api/utils/requestUtils.ts
@@ -162,8 +162,8 @@ export function putAPI<T>(url: string, data: unknown): CancellablePromise<T> {
   return sendXHRRequest('PUT', url, data);
 }
 
-export function deleteAPI<T>(url: string): CancellablePromise<T> {
-  return sendXHRRequest('DELETE', url);
+export function deleteAPI<T>(url: string, data?: unknown): CancellablePromise<T> {
+  return sendXHRRequest('DELETE', url, data);
 }
 
 export function uploadFile<T>(
--- a/mathesar_ui/src/api/utils/requestUtils.ts
+++ b/mathesar_ui/src/api/utils/requestUtils.ts
@@ -162,7 +162,10 @@ export function putAPI<T>(url: string, data: unknown): CancellablePromise<T> {
   return sendXHRRequest('PUT', url, data);
 }
 
-export function deleteAPI<T>(url: string, data?: unknown): CancellablePromise<T> {
+export function deleteAPI<T>(
+  url: string,
+  data?: unknown,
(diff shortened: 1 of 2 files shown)


seancolsen (changes requested):
Thanks for this pr, @thesujai. It mostly looks great. There's one more thing I'd like to handle though...

It's possible that attempting to delete records will fail at the DB layer. For example, you might be trying to delete a record which is referenced by other records. We don't [yet](https://github.com/centerofci/mathesar/issues/2202) support deleting via CASCADE, so the related records will need to be deleted first.

- Before this PR:

    Failure to delete such records had decent error behavior, allowing the user to understand what happened and understand the steps needed to fix it.

    image

- After this PR:

    When a single record is being deleted, the UI handles errors nicely. Good!

    image
    
    But when multiple records are being deleted, the UI does not handle errors very well.

    image


Changes I'd like to see to improve the problematic error scenario noted above

- All loading spinners should clear when the request resolves.
- The error message from the API should display in the toast message. (No need to display it in the row header cells, because the API doesn't give us per-row error messages.)

thesujai:
Hey @seancolsen, Thank you for the review
I made the suggested changes, here are some UI which will looks different:

- When deleting multiple records containing rows with foreign references:
Screenshot from 2023-07-11 13-26-26

- When deleting a single row with foreign reference:
Screenshot from 2023-07-11 13-26-08

Earlier when multiple rows were selected containing ones with foreign references, some were deleted leaving the ones with foreign reference untouched. But with the new changes if the selection contain any foreign reference row, it will not delete any of the rows(Which i think would be more safe)

seancolsen (approved):
Looks great, @thesujai. Thanks!

I also added 00a06e9f97c40859fe4186d42bfbd31aee908172 to make use of our `pluralize` function for the error message.

Issue #2988, opened 2023-07-03
URL cell loses focus when clicking on its contained hyperlink


seancolsen:
## Steps to reproduce

1. Prerequisites:
    1. On the table page click on a cell to focus it.
    1. When the cell is focused, you can use the <kbd>Up</kbd> and <kbd>Down</kbd> arrow keys to move the active cell. This is good.
    1. Click on the "Filter" button above the table to open the filtering dropdown. Click it again to close it. Now that button is focused instead of a cell, and you won't be able to use the arrow keys to move the active cell. This is okay, but important to understand.

1. Set up a column with a URL type.
1. Into a cell in that column, enter a relatively short URL like `http://a.b`
1. Click on different cell to make your URL cell inactive.
1. On your inactive URL cell, click _the link text_.
1. Observe that the cell is selected and active (has a thick border). Good!
1. Expect the cell to also be focused, but instead observe that it is not focused (because the link inside the cell is focused instead). Bad.
1. Click on some empty space in the cell but outside the link. Now observe the cell to be focused. Good.

manthan-jsharma:
this looks like a valid issue to me

@pavish , @rajatvijay , can i pick this up ?

Issue #2987, opened 2023-07-03
Cell loses focus when clicking on its outline


seancolsen:
## Steps to reproduce

1. Prerequisites:
    1. On the table page click on a cell to focus it.
    1. When the cell is focused, you can use the <kbd>Up</kbd> and <kbd>Down</kbd> arrow keys to move the active cell. This is good.
    1. Click on the "Filter" button above the table to open the filtering dropdown. Click it again to close it. Now that button is focused instead of a cell, and you won't be able to use the arrow keys to move the active cell. This is okay, but important to understand.

1. Increase your browser zoom to something like 300% to make this easier.
1. Click on a cell to focus it.
1. Click on the blue cell border.
1. Expect the cell to remain focused.
1. Observe that the cell loses focus.

manthan-jsharma:
@pavish , @rajatvijay , is somebody looking into this issue ? 

I'm interested in picking this up

sahil-adit:
I am interested to solve this issue,Can you please provide some more information?

seancolsen:
@sahil-adit you're welcome to work on this. Please submit a PR or ask specific questions. If you have questions about this issue, you can ask them here. Or if you have questions about getting set up with Mathesar development, please ask them in Matrix.

abhirupPann:
Is this issue still open?

seancolsen:
Yes, @abhirupPann this is still open. You're welcome to submit a PR for it.

Mahas1234:
Are there any more fixes needed. I'm looking for first contribution.

seancolsen:
This issue is still open @Mahas1234, feel free to submit a PR.

smit9924:
Is this issue still open? I would like to fix this up

Anish9901:
Yes, this is still open @smit9924. We are preparing for our beta release currently, so some pieces of the app might be broken in the develop branch. Please feel free to reach out on the issue or on one of our [Matrix](https://wiki.mathesar.org/community/matrix/) channels if you need help regarding this issue.

prakhargupta-jan:
Is it ok to work on an issue without it being assigned to you I hope I'll not be stepping on anyone's toes. new to Open Source

smit9924:
@Anish9901 I debug this issue and would like to have some suggestions:
(1) On click the filter button it is common that cell loses the focus and it might be expected as user click the other element out of the table. I don't feel there isn't anything unexpected over there. According to me we should not lock focus on table cell.
[...]

PR #2989, opened 2023-07-03
Improve cell focus behavior

### LINKED ISSUE #1892: Re-focus table cell after opening record selector on that cell
## Steps to reproduce

1. Go to the Table Page for a table that has a foreign key column.
1. Open the Record Selector from an FK cell.
1. Close the Record Selector, either via Esc, the close button, or by submitting a value.
1. Expect the cell within the table to remain selected and active.
1. Instead observe that after closing the Record Selector, the cell is no longer active.

## Related

- #1917

### LINKED ISSUE #2380: Synchronize cell "active" and "focused" states
## Current behavior

- Every cell in the sheet has the following state:

    - **active** or **inactive**
    
        No more than one cell can be active at a time, and we do our best (imperatively) to ensure that we always have one active cell.

    - **selected** or **unselected**
    
        Multiple cells can be selected within the sheet. We do our best (imperatively) to ensure that the active cell is also selected.

    - **focused** or **unfocused**

        No more than one cell can be focused at a time. But it's quite common to have _zero_ cells focused -- and necessary so that the browser can focus other UI on the page. We have rather airtight logic to ensure that a _focused_ cell is always _active_.

- The user can interact with a table cell via the keyboard in several ways (e.g. `Enter` to enter edit mode, arrow keys to move the active cell, etc) -- but **only when the cell is _focused_**. If you click on some other UI like empty space above the table or a button somewhere else, the cell loses focus and those keyboard actions no longer have any effect even though the cell may still be _active_, but not the other way around.

- We indicate that a cell is _active_ by displaying a thick blue border if the cell allows editing (or displaying a grey border if the cell does not allow editing).

- We have no indicator for the cell's _focused_ state.

## Problems with the current behavior

- I don't care whether a cell is _active_ -- I care whether a cell is _focused_. I can't identify any special behavior that an active-unfocused cell has over an inactive-unfocused cell. I can't do anything with it if it's not focused. Maybe I'm overlooking something here though.

- The current behavior has the effect of making me unsure whether I can use the keyboard to interact with the cell, so I subconsciously end up using the mouse more.

## Comparisons with other products

- **Google Sheets** and **LibreOffice Calc** both behave very similarly to our current behavior. While that may be an argument for keeping it, I still don't like it.

- **AirTable** almost entirely synchronizes their active cell with their focused cell, but they also discard the entire selection (and active cell) when the user focuses on other UI. I like the simplicity of this behavior, but it wouldn't be compatible with our table inspector design because we need to maintain a selection of cells while user is focusing on UI within the table inspector.

## Desired behavior

- Cell selection would work the same as it currently does, but we would eliminate the active/inactive state, keeping only focused/unfocused. When a cell is focused it gets a thick blue border (or grey if disabled). When a cell is unfocused it doesn't have the border. That's actually the behavior I built within the Record Selector, and I think it works well.

@kgodey @pavish @rajatvijay @ghislaineguerin I'm curious what you think about this idea. It seems like a change that would probably be easy to make.

seancolsen:
Fixes #1892
Fixes #2380

## Notes

- I started working on #1892 first but ended up chasing down #2380 in the process. That's why this PR addresses them both.
- #1892 is pretty straightforward, and I think it should be easy to see that this PR addresses it. Launch the record selector from an FK cell. Close the selector either by submitting a value or by closing the model. Before this PR, the FK cell would no longer be focused (although it was hard to tell). After this PR, the cell regains focus, allowing you to use the keyboard to continue navigating from cell to cell.
- #2380 is the more nebulous issue, where we had some discussion in the ticket about various implementation approaches and only seemed to reach a clear agreement on the main objective. So this PR is a rough stab at addressing that main objective. Basically: the cell still has an 'active' state, kept separate from its focused state; but now the cell is _styled_ differently depending on whether it is focused. We had previously styled the cell outline gray for disabled cells. I repurposed that gray styling to indicate the state when the cell is active but no focused.
- I agree with @pavish that we need to retain state to track when a cell is active but not focused. And I think we're both in agreement that we should try to prevent the user from inadvertently causing the cell to enter this state. For example, after the record selector closes, we should re-focus the cell to avoid the "active but not focused" state.
- What's nice about these changes is it makes it easier for us devs to spot that pesky "active but not focused" state. In playing with the app after these changes, I noticed and reported two new bugs: #2987 #2988

## Before

- Cell outline was blue when cell was active and not disabled.
- Cell outline was gray when cell was active and disabled.
- The focus state did not affect the cell outline.
- Opening and closing the record selector caused the cell to lose focus

## After

- Cell outline is blue when the cell is active and focused.
- Cell outline is gray when the cell is active and not focused.
- Cell has no outline otherwise
- The cell is re-focused after the record selector closes

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish (approved):
Looks great! Nice work!

PR #2991, opened 2023-07-04
Removed dependency on lazydict and support Python 3.10

### LINKED ISSUE #2960: Remove the lazydict dependency
## Problem
This project has a depencdency on lazydict (https://github.com/janrain/lazydict) which is no longer supported (latest version is 7 years ago)  and does not work with Python 3.10.

## Proposed solution
The lazydict is used only on one place in the code  https://github.com/centerofci/mathesar/blob/76f9d8517dd28d5255852b49f9ab864b5721fd8e/mathesar/api/display_options.py#L4
to avoid re-evaluating the `_money_display_options_schema`... There are a lot of ways that this can be resolved, for example instead of the lazydict we could use a normal dict and wrap `_money_display_options_schema` with `functools.cache` (https://docs.python.org/3/library/functools.html#functools.cache) to evaluate it only once.

## Additional context
I'd be happy to provide a PR for resolving this.

Abhishek140501:
Fixes #2960

Updated code to remove the dependency on lazydict and ensure compatibility with Python 3.10. Instead of using lazydict, the _money_display_options_schema function is now wrapped with functools.cache to evaluate it only once.

Technical details:
- Replaced lazydict dependency with functools.cache
- Updated _money_display_options_schema function to use cache decorator
- Verified compatibility with Python 3.10

Screenshots: 
Screenshot (106)

Checklist:
- [x] My pull request has a descriptive title.
- [x] My pull request targets the `develop` branch of the repository.
- [x] My commit messages follow best practices.

Developer Certificate of Origin: I certify that my contribution complies with the Developer Certificate of Origin.

### DIFF
--- a/mathesar_ui/src/App.svelte
+++ b/mathesar_ui/src/App.svelte
@@ -1,10 +1,7 @@
 <script lang="ts">
-  import { get } from 'svelte/store';
-
   import { Confirmation, ToastPresenter } from '@mathesar-component-library';
   import { confirmationController } from '@mathesar/stores/confirmation';
   import { toast } from '@mathesar/stores/toast';
-  import { setUserProfileStoreInContext } from '@mathesar/stores/userProfile';
   import {
     RecordSelectorController,
     setRecordSelectorControllerInContext,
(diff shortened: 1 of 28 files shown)


kgodey:
@Abhishek140501 Thanks for the contribution! Please revert the changes in the `.md` files, it looks like they're unrelated.

PR #2993, opened 2023-07-05
Remove lazydict dependency

### LINKED ISSUE #2960: Remove the lazydict dependency

spapas:
Fixes #2960

This PR removes the lazydict dependency that is not compatible with python > 3.9

**Technical details**

Please notice that this PR will evaluate the function `_money_display_options_schema` only once when the application is starting, while with the previous version that was using the lazydict would evaluate it once when it was called. 

We could implement a *lazy* solution instead however it would need more changes to the source code and I really don't think it's worth it, I actually prefer evaluating the function when the project is starting.
`

## Checklist
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/api/display_options.py
+++ b/mathesar/api/display_options.py
@@ -91,4 +91,3 @@ DISPLAY_OPTIONS_BY_UI_TYPE = {
     # NOTE: below callable will be evaluated only once
     UIType.MONEY: _money_display_options_schema(),
 }
-
--- a/mathesar/api/display_options.py
+++ b/mathesar/api/display_options.py
@@ -88,6 +88,5 @@ DISPLAY_OPTIONS_BY_UI_TYPE = {
             {"name": "show_units", "type": "boolean"},
         ]
     },
-    # NOTE: below callable will be evaluated only once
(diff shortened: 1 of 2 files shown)


dmos62 (approved):
Looks good, thanks!

PR #3004, opened 2023-07-06
Add `Peak Day of Week` aggregation function.

### LINKED ISSUE #3003: Support for `Peak day of week` aggregation function while summarizing columns
## Problem
We need to create an aggregate function for `Day of Week` which imitates `average`, the only difference will be it will take modular mathematics into consieration.

## Proposed solution
It would be nice to support such an aggregate as a transformation step while summarizing Date and DateTime columns.

## To Discuss
- `Peak day of week` aggregation function should only be available for columns with **Date** and **DateTime** data type.

## Additional context
This feature request is a part of GSoC '23 project, [Add More Summarization Functions](https://summerofcode.withgoogle.com/programs/2023/projects/D9XvkQCN).

Aritra8438:
Fixes #3003 

This PR adds `Peak Day of Week` aggregation function to the Mathesar UI.

**Technical details**

I have done the following things:
- Add custom PostgreSQL function for `Peak Day of Week` aggregation.
- Add `Peak Day of Week` to the Mathesar UI.
- Add tests for the same.

**Screenshots**

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/sql/3_msar_custom_aggregates.sql
```diff
@@ -37,3 +37,61 @@ CREATE OR REPLACE AGGREGATE peak_time (TIME)
     finalfunc = final_func_peak_time,
     initcond = '{0,0}'
 );
+
+
+CREATE OR REPLACE FUNCTION dow_to_degrees(_date DATE)
```
mathemancer, line 42: better to use `date_` here for the variable name.

db/sql/3_msar_custom_aggregates.sql
```diff
@@ -37,3 +37,61 @@ CREATE OR REPLACE AGGREGATE peak_time (TIME)
...
+
+
+CREATE OR REPLACE FUNCTION degrees_to_dow(degrees DOUBLE PRECISION)
+	returns INT AS $$
+    SELECT (ROUND(degrees * 7 / 360)::int) % 7;
+$$ LANGUAGE SQL;
+
+
+
+CREATE OR REPLACE FUNCTION dow_to_string(_dow INT)
```
mathemancer, line 55: Is `dow` a keyword? If not, use that for the variable name instead. If so, use `dow_`.

db/sql/3_msar_custom_aggregates.sql
```diff
@@ -37,3 +37,61 @@ CREATE OR REPLACE AGGREGATE peak_time (TIME)
...
+$$ LANGUAGE SQL;
+
+
+CREATE OR REPLACE FUNCTION degrees_to_dow(degrees DOUBLE PRECISION)
+	returns INT AS $$
+    SELECT (ROUND(degrees * 7 / 360)::int) % 7;
+$$ LANGUAGE SQL;
+
+
+
```
mathemancer, line 54: Just two lines here
```suggestion


```

db/sql/3_msar_custom_aggregates.sql
```diff
@@ -37,3 +37,61 @@ CREATE OR REPLACE AGGREGATE peak_time (TIME)
...
+$$ LANGUAGE SQL;
+
+
+CREATE OR REPLACE AGGREGATE peak_day_of_week (DATE)
+(
+    sfunc = accum_dow,
+    stype = DOUBLE PRECISION[],
+    finalfunc = final_func_peak_dow,
+    initcond = '{0,0}'
+);
```
mathemancer, line 97: Please end files with a newline character

mathemancer (changes requested):
Overall, I think this is pretty good, but I think we should make it less discrete. By that, I mean that we shouldn't box into dates from timestamps and so on, but go the other way. The better measurement than a discrete DOW is the number of seconds since the beginning of the week. I.e., peak _time_ in the week. In so far as the user wants to know a peak day, this is derivable from peak time in the week. But, it's not the same the other way.

Let me know if that doesn't make sense.

Many (most) comments from the previous PR carry over here. But I want to emphasize:
- Please make sure to add comments to each function explaining them (like the `0_msar.sql` file)
- Please namespace the functions under the `msar` schema (crucial)

seancolsen (commented):
Hmmm.

The front end code changes look good, insofar as they follow the pattern we've established.

Buuut, there's an underlying problem here that I'm really scratching my head about how best to resolve...

**The weekday names will be hard to localize.**

Here are some more thoughts:

- For Mathesar's **Date** type (as an example), the user can choose between these formatting options

    

    I was not deeply involved in the design decisions that led to that interface for configuring the date formatting, but I know that we briefly considered giving the user more control over customizing the formatting. I think it's something we may still take up in the future. But I bring mention all of this to demonstrate that currently Mathesar has no way for a user to show the name of a weekday, so we don't really have a precedent to follow.

- @rajatvijay is working on an [internationalization project](https://github.com/centerofci/mathesar-wiki/blob/master/projects/internationalization.md), and we've spent quite a while hammering down some detailed [specs](https://github.com/centerofci/mathesar-wiki/blob/master/engineering/specs/internationalization.md) for it. The specs discuss [_Translating server-side rendered pages & API error messages_](https://github.com/centerofci/mathesar-wiki/blob/master/engineering/specs/internationalization.md#translating-server-side-rendered-pages--api-error-messages), but they do not discuss anything like _"translating localized cell values"_ because our approach thus far has been to handle that sort of problem with display options. If we wanted to continue using the display options approach here, it seems to me that we'd need an entirely new UI type for something like "Day of week", but I really grimace at the thought of that because it just seems like too much additional complexity. To do this sort of localization in a clean way, we should be pass a normalized value over the API and then allow the front end to format the value. For a weekday, that normalized value would probably be an integer.

- I really don't have a good solution to propose. I can think of many approaches which I _don't_ like.

I'm torn on how to deal with this problem.

I'm sorry to be a stick in the mud, but I do have some broader concerns that I'd like to relay while on this subject. These concerns apply to other summarization functions in addition to "Peak day of week", but I'll stick them here for lack of a better place.

- I predict that many users won't understand the meaning of "Peak day of week". I would expect many users to incorrectly assume that it means the most common day of week, but it's not. I notice that ("Monday", "Wednesday") dates give a peak weekday of "Tuesday". Other users may be unclear on the meaning entirely. We could potentially link out to some user docs which explain this "peak" concept in more detail. Some people might want (or even naively expect, if they've not thought deeply) the week to "wrap" on weekdays. For example, data with ("Monday", "Friday") gives a peak weekday of "Saturday". But some users might expect it to give "Wednesday", and other users might expect it to give "Friday" or "Monday".

- At our last meeting about List Data Types, I expressed some concern about filtering as applied to lists. I have a similar concern with the direction I see these summarization functions heading. Perhaps this ship has already sailed, with the plans that we've laid out. My concern with filtering lists was that we'd need to present the filters to the user in a "pre-composed" manner. I have that same concern here. Instead of giving users a whole bunch of aggregate functions, I'd rather give users _formulas_ which would allow them to compose their own aggregate behavior as they see fit. We could have a scalar function to extract an integer weekday from a date. Then we could have an aggregate function to compute a "peak" value, given any number along with the bounds of its wrapping window. Then we could have a scalar function (or user-defined case expression) to format an integer weekday index as a localized day of week. The user could then compose those functions to build all kinds of cool stuff. If we give them this list of pre-composed aggregate functions, the resulting functionality is an awkward mix of ambiguity and limitations.

Don't get me wrong &mdash; I think these "peak" aggregation functions are cool! I don't want to pump the brakes too hard on this work. Perhaps these are just thoughts to keep in mind for the future when we make larger improvements to the Data Explorer.

Aritra8438:
Thanks, @seancolsen.

> * I predict that many users won't understand the meaning of "Peak day of week".

This is one of our concerns, Brent and I were planning on achieving a more accurate version of this. Nevertheless, users might not be able to understand what peak day of week represents.

> * We could potentially link out to some user docs which explain this "peak" concept in more detail.

Yesss.

> * I would expect many users to incorrectly assume that it means the most common day of week, but it's not.

This aggregation can be done using `Mode`. But, many users might not be aware of that. Should we consider changing the aggregate name to `most frequent`? 


> * For example, data with ("Monday", "Friday") gives a peak weekday of "Saturday". But some users might expect it to give "Wednesday", and other users might expect it to give "Friday" or "Monday".

The peak day of week name is indeed ambigous, as what is done with the column is not clear from the aggregate name. We are actually doing modular average, so `mean` won't fit either (as, people would then expect ("Monday", "Friday") to give "Wednesday").


**But why may "Saturday" be a better choice than "Wednesday"?**

Because, it minimizes the absolute distance from those days. So, it is a better central measure as it minimizes the variance. That said, when the user wants to wrap on weekdays, "Wednesday" will be the only choice.

Aritra8438:
Hi, @mathemancer.

- I have made the functions less discreet.
- given them proper naming.

Also, if the `peak_time_of_week` is "Wednesday 9:00 PM", the function will return "Wednesday" only (not "Thursday")

mathemancer:
Regarding internationalization, I agree that @seancolsen 's point is quite valid. 

Regarding the `peak_time` or `peak_dow` name, I agree that it doesn't quite capture things. We're basically calculating a mean (average) of cyclical data. I guess we could just use `average_time` or `average_dow` ?

That said, we already have `mode` available (or do we? If not we should definitely add that to the TODO list for summarizations, @Aritra8438 ). And even though bi- or multi-modal data occurs in non-cyclical situations, the mean is still one of the most common and useful stats you can calculate on a dataset.

Regarding leaving composition to the user: I mostly agree with @seancolsen 's points, with some caveats. Having composable pieces allows users to create exponentially more functions (since a composition of functions is a function) than we actually implement ourselves. This is a seriously powerful concept. That said, we don't have any clue right now (@seancolsen 's recent email notwithstanding) for how to implement composition in this context at the UI layer. We can't even portray the `point` datatype in a useful way in the UI as far as I know, and that would be the input-output type for some of the functions being composed. Also, our method of matching up the input-output types, while serviceable, may not be clear enough once things get complicated. Given that, I really want to avoid keeping useful features from users while we ponder more powerful features. I also think that even if we have more powerful composition in the future, it doesn't mean that well-named functions that encapsulate difficult concepts are then obsolete.

Regarding whether these 'average on a cycle' functions should be precomposed or not under any circumstance: Such functions may not occur to many users, even given the necessary pieces to build them. These functions essentially calculate a center of mass of a bunch of vectors, where the vectors come from interpreting cycles in data as revolutions around a unit circle. You then reinterpret the resulting center of mass vector as a member of the original data type. The fact that such a calculation lets you answer questions like 'when should I have extra staff available at my 24 hour call center?' is something I think we need to show to the user by giving them precomposed functions with understandable names. 

With that said: I think we will quickly experience diminishing returns on precomposing these functions. So, We're not planning to continue in this vein after the two we already have queued. I'll take another look and try to figure out whether we can break these part into constituent atoms that make sense in the course of this internship, or whether we should hold off. Our current lack of support for the `point` type makes me skeptical, though.

seancolsen:
@mathemancer @Aritra8438 I thought we decided to close this PR instead of merge it.

Aritra8438:
Yes @seancolsen. Sorry, I forgpt to close it.
How did it get merged?

seancolsen:
Ok that makes sense @Aritra8438. I don't think we need to do anything else.

Issue #2709, opened 2023-03-18
No tables visible when viewing an existing schema


hydrandt:
## Description
When I connect to an existing schema, which was built on the [Graphile Starter project](https://github.com/graphile/starter), there are no tables.

## Expected behavior
Tables should be visible:) They are correctly counted in the list of schemas, but as soon as I enter the schema, I just get "You've created a new schema, now add tables to it."

## To Reproduce
1. install using official install script; connect to existing database of the Graphile Starter project
2. open the app_public schema
3. no tables are visible

## Environment
 - OS: Debian Sid
 - Browser: Firefox
 - Browser Version: 110.0.1

## Additional context

There are following warnings about unrecognized column types:

```
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'citext' of column 'email'                                                                                                
mathesar_service                |   return Table(                                                                                                                                                                                      
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'point' of column 'coordinates'                                                                                           
mathesar_service                |   return Table(                                                                                                                                                                                      
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'citext' of column 'username'                                                                                             
mathesar_service                |   return Table(                                                                                                                                                                                      
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'point' of column 'location_coordinates'                                                                                  
mathesar_service                |   return Table(                                                                                                                                                                                      
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'citext' of column 'email'                                                                                                
mathesar_service                |   return Table(                                                                                                                                                                                      
mathesar_service                | /code/db/tables/operations/select.py:20: SAWarning: Did not recognize type 'point' of column 'coordinates'                                                                                           
mathesar_service                |   return Table(  
```

pavish:
This is a user reported issue, we need someone from the backend team to prioritize this and pick it up.
cc: @dmos62, @silentninja, @mathemancer, @Anish9901 

 I'm moving this out of 0.1.2 since there are several unknowns with the issue.

notramo:
`citext` is an often used type, mostly used for e-mail addresses, as they are case-insensitive. It can be found in nearly every webapp which uses e-mail addresses. Could you prioritise this issue?

notramo:
[...]

Issue #3007, opened 2023-07-07
Add support for `point` column type


dmos62:
Reported here: https://github.com/centerofci/mathesar/issues/2709

dmos62:
[This comment in related thread applies](https://github.com/centerofci/mathesar/issues/2959#issuecomment-1649843939).

Issue #2959, opened 2023-06-21
Add support for `citext` column type


silentninja:
## Problem
Mathesar currently does not support [`citext`](https://www.postgresql.org/docs/current/citext.html) and this makes it impossible to use Mathesar to manage the database of a [Lucky](https://luckyframework.org/) project, as the default migration creates a `citext` column for the e-mail address. Probably several other back-end frameworks are also affected by this.

## Proposed solution

SQLAlchemy does not support `citext` so a custom SQLAlchemy type for `citext` type should be added to Mathesar.

## Additional context
- Related to #2709 which reports the missing `citext` type along with few other types
- A [comment on #2709](https://github.com/centerofci/mathesar/issues/2709#issuecomment-1601565406) adds another request for the `citext`.

Comments by [notramo](https://github.com/notramo)

> I took a look at SQLAlchemy versions, and it seems that [v2.0.7](https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_2_0_7) added support for `citext` type. Mathesar uses the 1.x.x version of SQLAlchemy currently.

silentninja:
@notramo Thanks for reporting! We are in the process of migrating away from SQLAlchemy so it won't make sense to upgrade to SQLAlchemy 2.0 but we should be able to fix this issue by having a custom type for `citext` type. 

I am moving this issue to the upcoming milestone 0.1.3, so you can expect this issue to be fixed when we release Mathesar version 0.1.3.

notramo:
@silentninja, `citext` is basically `text`, the only difference is that the `=` comparison operator is case-insensitive. What about treating it as `text`, until a proper type is added? The comparison happens inside the Postgres engine anyway. AFAIK, every other aspects are the same as `text`.

mathemancer:
As I noted in [this comment](https://github.com/centerofci/mathesar/pull/3040#issuecomment-1651023079), we could give hints to the front end based on Postgres's own categorizations (someday). I checked by installing the `citext` extension. It's properly detected as a 'String type' in the `pg_type` table, which we could use to give some useful info to the front end.

PR #3040, opened 2023-07-13
Support unknown types (backend)

### LINKED ISSUE #3024: Support unknown Postgres types
Mathesar currently (erroniously) presumes that each Postgres data type it sees will be known via the relevant enums ([see comment](https://github.com/centerofci/mathesar/issues/3024#issuecomment-1632812393)) and, when an unknown type is encountered, Mathesar panics ([see bug report](https://github.com/centerofci/mathesar/issues/2709)). We want Mathesar to handle unknown types gracefully.

dmos62:
Fixes #3024

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/types/operations/convert.py
```diff
@@ -18,12 +19,11 @@ def get_db_type_enum_from_id(db_type_id):
         try:
             return MathesarCustomType(db_type_id)
         except ValueError:
+            if db_type_id == UnknownType().id:
+                return UnknownType()
```
dmos62, line 23: This change especially, but also the other change in this file might seem funny. The motivation was to get `get_db_type_enum_from_id` and `get_db_type_enum_from_class` to work with unknown types, and it turns out that most of the rest of the code fell into place.

kgodey:
Does this also fix related issues #2709, #2959, and #3007? Or do those need additional work?

silentninja:
@kgodey It fixes partially fixes #2709 by not throwing up an error when an unknown type is encountered. In order to fix #2709, #2959 and #3007 we need to add support for those types(citext, point) instead of considering them as "Unknown Type"

dmos62:
@kgodey I'll expand on @silentninja's answer by saying that this fixes "unknown types" (types we've not enumerated) breaking the column and record APIs (and maybe reflection as well, I don't recall). There are two complimentary directions to go from here:

1. have backend report the actual type identifiers of unknown types: this PR groups all of them under the umbrella identifier `__unknown__`; I didn't do this in this PR, because it's potentially-complex and it's unnecessary in order to downgrade this problem from a highest-priority blocker to a medium-priority UX concern;
2. have backend know things about some of these types; for example, even if Mathesar doesn't crash when `citext` is used, it still doesn't know that it's a text type, so it's not able to provide the right subset of filters or the right type casting choices (actually we disallow casting from unknown types).

kgodey:
Thanks @dmos62, sounds good. I do agree not crashing is the most immediate thing to fix, and 1 and 2 are secondary (but helpful to have).

silentninja (approved):
Looks good!

On a side note, it took some time for me to understand what was going on, mostly because there is an intermediary SQLAlchemy type involved in the mapping process. We definitely need to re-visit the type mapping and see if this can be made simple.

mathemancer:
@dmos62 In [the docs](https://www.postgresql.org/docs/13/catalog-pg-type.html), they give the following table:

Code | Category
-- | --
A | Array types
B | Boolean types
C | Composite types
D | Date/time types
E | Enum types
G | Geometric types
I | Network address types
N | Numeric types
P | Pseudo-types
R | Range types
S | String types
T | Timespan types
U | User-defined types
V | Bit-string types
X | unknown type

These codes are associated with all types in the DB, and you can query the `pg_type` table to get this info. I was thinking we could use these categories to simplify some of our logic, and also to do a better job of types which are unknown to Mathesar, but _are_ known to the DB.

As an example, suppose the `mathesar_types.email` type was unknown to Mathesar. This type is implemented as a domain over `TEXT`. This means that the `pg_type` table puts it into the `S` (for String) category, which we could at least use to give the front end a clue about how to handle input for the type and so on, even without knowing anything about it in the back end.

PR #3088, opened 2023-07-24
Superuser create

### LINKED ISSUE #3052: Add superuser creation page
## Problem
A Mathesar superuser is required before a user can start using Mathesar.  The process to create a Mathesar superuser is done using the command line which is slightly complicated and varies based on the installation types. Moreover having to fiddle around with the command line to create a superuser makes it hard to provide a one-click install experience. 
 
## Proposed solution
There should be a page for creating a superuser. This page should be accessible only if a super does not exist. 

## Proposed implementation
We will be creating a super user creation page using Django templates. The login page will redirect to this superuser creation page if the super-user does not exist.

silentninja:
Fixes #3052 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/users/decorators.py
+++ b/mathesar/users/decorators.py
@@ -1,0 +1,30 @@
+import functools
+
+from django.contrib.auth import get_user_model
+from django.shortcuts import redirect
+from django.urls import reverse
+
+
+def superuser_must_not_exists(view_function=None):
+    """
+    Decorator for views that redirects to the home page if at least one superuser exists
+    """
+    @functools.wraps(view_function)
+    def wrapper(request, *args, **kwargs):
+        if get_user_model().objects.filter(is_superuser=True).exists():
+            return redirect(reverse('home'))
+        return view_function(request, *args, **kwargs)
+    return wrapper
+
+
+def least_one_superuser_exists(view_function=None):
+    """
+    Decorator for views that checks if at least one superuser exists
+    and redirects to the superuser create screen if none exists
+    """
+    @functools.wraps(view_function)
+    def wrapper(request, *args, **kwargs):
+        if get_user_model().objects.filter(is_superuser=True).exists():
+            return view_function(request, *args, **kwargs)
+        return redirect(reverse('superuser_create'))
+    return wrapper
--- a/mathesar/templates/registration/superuser_create.html
+++ b/mathesar/templates/registration/superuser_create.html
@@ -95,7 +95,7 @@
     {% endif %}
     <div class="footer">
       <button class="btn btn-primary submit-button" type="submit">
-        Log In
+        Create Superuser
       </button>
     </div>
   </form>
--- a/mathesar/templates/registration/superuser_create.html
+++ b/mathesar/templates/registration/superuser_create.html
@@ -1,0 +1,102 @@
+{% extends 'mathesar/login_base.html' %}
+
+{% block h1 %}Welcome! Lets create a superuser{% endblock %}
+
+{% block box_content %}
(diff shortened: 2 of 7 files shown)


kgodey:
Please do not merge this until it fully works (i.e. it's styled, it's documented, and it wouldn't block the release if we cut it tomorrow).

silentninja:
> Please do not merge this until it fully works (i.e. it's styled, it's documented, and it wouldn't block the release if we cut it tomorrow).

Won't this lead to a huge PR which will be hard to review?  I thought we decided in case a feature is not complete before the cycle ends, we would be reverting or disabling it using techniques like a feature flag during the 2 weeks meant for release.
 So in this case to disable the feature we need to remove the URL for the superuser to create a page and remove the logic which redirects to the superuser-create page.

kgodey:
> As I recall, we previously agreed that if a feature isn't fully completed before the development cycle ends, we should revert or disable it using techniques like a feature flag during the two-week release phase.

My recollection is that we would disable incomplete features using techniques like a feature flag _before_ the PR was merged. I'm fine with doing that for this PR.

> Won't this lead to a huge PR 

You could also stack other PRs on top of this one so they can all be reviewed individually but merged together.

mathesar/users/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    @functools.wraps(view_function)
+    def wrapper(request, *args, **kwargs):
+        if get_user_model().objects.filter(is_superuser=True).exists():
+            return redirect(reverse('home'))
+        return view_function(request, *args, **kwargs)
+    return wrapper
+
+
+def least_one_superuser_exists(view_function=None):
```
mathemancer, line 20: I think just `superuser_exists` would be simpler here.

mathesar/users/decorators.py
```diff
@@ -0,0 +1,30 @@
+import functools
+
+from django.contrib.auth import get_user_model
+from django.shortcuts import redirect
+from django.urls import reverse
+
+
+def superuser_must_not_exists(view_function=None):
```
mathemancer, line 8: Wouldn't returning some error be more robust here? Something about using redirects as a security measure in this way makes my skin itchy.

mathemancer (changes requested):
Okay, this seems to work as intended. Maybe it's my admitted inexperience with web security and so on, but something about the redirect-as-security seems sketchy here. Is it completely impossible to spoof the post request submitted by the create button? If there's any gap in the security on the create superuser page, it's pretty bad since it's not authenticated at all, and allows , well, creating a super user.

Also, I agree with @kgodey that we should figure out how to turn this off before merging.

mathesar/users/decorators.py
silentninja: We are **returning** a `redirect` without going to the next step, so even if it does not get redirected (which shouldn't happen) the user will never be able to create a superuser. Also, we are using `redirects` as the strategy for our redirecting unauthenticated users to the login page, so I don't see much of a difference.

mathemancer, line 8: Okay, thank you for the explanation. My only request now is to make the following grammar change:

```suggestion
def superuser_must_not_exist(view_function=None):
```

(and also at the call)

mathemancer (approved):
Okay, I'm satisfied on the security front. I made a small grammar suggestion.

Issue #3036, opened 2023-07-12
Implement frontend flow: User viewing shared entity


pavish:
I've not added a description, since I'll be implementing this.

Refer [Product spec](https://wiki.mathesar.org/en/product/specs/publicly-shareable-links) for detailed info.

PR #3093, opened 2023-07-24
Shareable links frontend - shared table consumer view

pavish:
Related to #3036 

**This PR:**
* Allows viewing shared tables on the frontend

**Testing instructions:**
* Generate a shared link for a table via API:
  - A post request to `/api/ui/v0/tables/1/shares` with `{ enabled: true }` should share the table and generate a slug. The response will be as follows:
    ```
    {
      "id": 1,
      "slug": "874cfd2b-d580-4ea1-9ded-c226d46f9443",
      "enabled": true
    }
    ```
* Visit `http://localhost:8000/shares/tables/874cfd2b-d580-4ea1-9ded-c226d46f9443/` and you should be able to see a readonly view of the table.
* The following cases should be taken into account, when the link is valid:
  - a. The link should work when you're not logged into Mathesar.
  - b. The link should work when the user is logged in but does not have access to the table's schema or DB.
  - c. The link should work when the user is logged in and has access to the table.

**Note for reviewers:**
* Case 'b' mentioned above was the trickiest to implement, please pay extra attention to it.
* Inorder to make 'b' more stable, a complete store refactor is required on the frontend. In this PR, a workaround has been implemented for the time being. Some examples on why a store refactor is required:
   - For case 'b', code like this `$tables.get(id)` will not work because the user does not have access to the table, and therefore `tables` store will not contain the requested table.
    - Imports like `$currentDbAbstractTypes` will return empty values because the user does not have access to the database.
* The workaround implemented here is to separate the routing contexts of shared routes and normal authenticated routes. This effectively makes the routing of shared routes server-side-only.
* This is more of a band-aid solution and we should get into refactoring the stores as soon as possible in upcoming cycles.

**Cases not handled in this PR:**
* _Filtering FK columns_, and _context menu links_ do not work because the user does not have access to these entities. They are currently visible on the UI. We'll have to come up with a UX and have some discussions around how to handle these cases. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
This is looking good @pavish! I have some changes to request, but I'm approving so that you can decide whether to address them in this PR or a subsequent one.

- I see we're now using `RouteObserver.svelte` instead of `Observer.svelte`. Do we still need `Observer.svelte`? It doesn't seem to be used anywhere. Maybe delete it?

- The name "share consumer" is a bit weird. This name is used in many places throughout this PR, either as the type `ShareConsumer` or the variable `shareConsumer`.

    First of all, I like to avoid words where the part of speech is ambiguous. "Share" is not great because it's commonly used as a verb, but we're using it as a noun in this case. It's not a huge problem, but I'd say it reduces readability slightly.

    "Consumer" is the bit I have a harder time with. I don't see how the meaning of "consume" is relevant in this case.

    Suggested alternatives:

    - "sharing entry"
    - "public share"

- In `TableDescription.svelte` and `TableName.svelte`, you've added `{...$$restProps}`. Can you please also define a `$$Props` interface so as to conform to our [code standard](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/STANDARDS.md#when-using-restprops-define-a-props-type)?

- I see you've noted some remaining work to do after this PR. I'd like to make sure you address these things eventually:

    - Disable re-reordering of columns
    - Don't show the icon hyperlink to the record page within the PK cell
    - Remove the following entries in the cell context menu:
        - "Set to NULL"
        - "Go to Record Page"
        - "Go to Linked Record" _(shown only for FK columns)_
    - Remove the "Go to Record Page" entry from the row header context menu

pavish:
@seancolsen

> I see we're now using RouteObserver.svelte instead of Observer.svelte. Do we still need Observer.svelte? It doesn't seem to be used anywhere. Maybe delete it?

Thank you for pointing this out. I initially had a [large PR](https://github.com/centerofci/mathesar/pull/3061/files#diff-c0aa7bcaff0c17c6b69b16a4b663d84f0d1e223b5fe758742958e4fdfcd077ea) which I split into frontend and backend PRs. I missed removing Observer.svelte when I pulled in the frontend related changesets. 

I've removed it in https://github.com/centerofci/mathesar/pull/3093/commits/f87afe9a035ca988473f5acefc18265a7e7e3307

> The name "share consumer" is a bit weird. This name is used in many places throughout this PR, either as the type ShareConsumer or the variable shareConsumer.

I do agree the name is a bit weird. I tried to come up with a number of different names and it took me a lot of time figuring out a good name than to figure out the implementation, so I went ahead with `shareConsumer`.

Here's the reasoning behind it.
* We use the term `shares` in the backend to denote models, and APIs related to sharing, and I wanted to keep it the same.
* We have two contexts in which the routes, pages, and components might refer to 'Shares'
   - a: Creating & modifying a share. This is performed by a manager who has access to a table/query.
   - b: Viewing a shared entity. This is performed by anyone who has a link.
* I had to differentiate cases a and b, and I referred to case `b` as `ShareConsumer` in the code, as it 'consumes' a shared entity.

> Suggested alternatives: "sharing entry" "public share"

The alternatives here are too ambiguous to differentiate the two cases.

I'm open to more suggestions. Meanwhile, I'll also be thinking of some. I think we can handle this in future PRs.

> In TableDescription.svelte and TableName.svelte, you've added {...$$restProps}. Can you please also define a $$Props interface so as to conform to our [code standard](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/STANDARDS.md#when-using-restprops-define-a-props-type)?

I've added a commit: https://github.com/centerofci/mathesar/pull/3093/commits/812012ed3c86e32a3a494f4077331a52f99df3ca to explicitly declare the prop I needed. It seemed the fastest way to resolve this for now.

> I see you've noted some remaining work to do after this PR. I'd like to make sure you address these things eventually

Yes, I had considered all of them in mind and I now notice that I have not explicitly mentioned them in the description. Thanks for noting all the cases down.

PR #3097, opened 2023-07-25
Add DDL functions for altering columns

mathemancer:
Related to #2737 

This PR removes SQLAlchemy usage from the `db.columns.operations.alter` module by adding SQL functions to handle those operations.

**Technical details**

#### Python layer
- Main modified python function is `db.columns.operations.alter.alter_column`.
- Removed from the `db.columns.operations.alter` module.
- Left a couple convenience functions in the `db.columns.operations.alter` to avoid massive refactor.
  - These should probably be removed in a future PR, i.e., wrapped into the main `alter_columns` function.
- Modifies the `alter_column_type` convenience function signature to use `column_attnum` rather than `column_name`.
- I left a bunch of Python tests in python that should probably be moved to SQL. Namely, the casting tests. The prerequisite to moving those tests to SQL in a decent way is to move the logic that writes the casting functions down to the SQL layer. Otherwise, we need the webservice to write those functions so we can run DB tests, which I want to avoid.
- Because most python layer functions are in place, it wasn't necessary to mock out the DB functions in testing and make sure the python layer is wrangling input to output properly.

#### SQL layer
- Main added SQL function is `msar.alter_columns`.
- Adds an `msar.is_default_possibly_dynamic` function in SQL; not yet wired up to python.
- This PR uses a number of technical details about how PostgreSQL converts strings to and from internal representations of types with options. I've attempted to thoroughly document these details in the relevant functions.

### Dynamic defaults
- This PR technically adds support for altering the type of columns with dynamic defaults, which I don't think we allowed before.
- We still don't allow creating dynamic defaults from the API.
- We still don't allow dropping or otherwise modifying dynamic defaults from the API.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/sql/0_msar.sql
```diff
@@ -902,29 +1032,49 @@ All fields are optional, and a null value as input returns 'text'
...
-    ''
-  ) || CASE WHEN topts.array_ THEN '[]' ELSE '' END
+  __msar.get_formatted_base_type(
+    COALESCE(
+      msar.get_fully_qualified_object_name(typ_jsonb ->> 'schema', typ_jsonb ->> 'name'),
+      typ_jsonb ->> 'name',
+      'text'  -- We fall back to 'text' when input is null or empty.
+    ),
+    typ_jsonb -> 'options'
+  ) || CASE WHEN (typ_jsonb -> 'options' ->> 'array')::boolean THEN '[]' ELSE '' END
```
dmos62, line 1044: This function seems shorter, which is nice, but it's become more horizontally intense. Could you split some of the lines? Examples of horizontally intense expressions: `) || CASE WHEN ... THEN ...`, `format_type((typ_jsonb ->> 'id')::integer, (typ_jsonb ->> 'modifier')::integer)`.

db/sql/0_msar.sql
```diff
@@ -902,29 +1032,49 @@ All fields are optional, and a null value as input returns 'text'
...
+{
+  "name": <str> (optional),
+  "options": <obj> (optional)
+}
+
+*/
+SELECT msar.build_type_text(
+  jsonb_strip_nulls(
+    jsonb_build_object(
+      'name', COALESCE(typ_jsonb ->> 'name', old_type),
```
dmos62, line 1069: I'm surprised that the `typ_jsonb` struct doesn't consider the type name a required field. When does that offer advantages? I see how that makes handling it more fiddly, however. You mention the use case where you want to only change the options of a type, but in such a case the type name/id would be easily accessible, right? What's the win here?

By the way, loving the documentation.

db/sql/0_msar.sql
```diff
@@ -1590,6 +1740,318 @@ END;
...
+However, we do throw an error in cases where the schema exists, but the type casting function
+doesn't. This is assumed to be an error the user should know about.
+
+Args:
+  val: This is quite general, and isn't sanitized in any way. It can be either a literal or a column
+       identifier, since we want to be able to produce a casting expression in either case.
+  type_: This type name string must cast properly to a regtype.
+*/
+SELECT CASE
+  WHEN EXISTS (SELECT 1 FROM pg_namespace WHERE nspname='mathesar_types') THEN
```
dmos62, line 1809: I recognize the `SELECT 1` trick from having to do the same, but the intent wouldn't be immediately obvious to someone else, possibly. Maybe move checking whether a schema exists to a function?

db/sql/0_msar.sql
```diff
@@ -1590,6 +1740,318 @@ END;
...
+    msar.get_cast_function_name(type_::regtype) || '(' || val || ')'
+  ELSE
+    val || '::' || type_::regtype::text
```
dmos62, line 1812: Why is this recovery necessary?

Multi-functional or overloaded (not sure what's the best term) functions are quite common here. I wonder why.

dmos62 (changes requested):
This is a half-finished review. I'll pick it back up tomorrow.

db/sql/0_msar.sql
```diff
@@ -252,6 +252,32 @@ END;
...
+msar.is_default_possibly_dynamic(tab_id oid, col_id integer) RETURNS boolean AS $$/*
+Determine whether the default value for the given column is an expression or constant.
+
+If the column default is an expression, then we return 'True', since that could be dynamic. If the
+column default is a simple constant, we return 'False'. The check is not very sophisticated, and
+errs on the side of returning 'True'. We simply pull apart the pg_node_tree representation of the
+expression, and check whether the root node is a known function call type. Note that we do *not*
+search any deeper in the tree than the root node. This means we won't notice that some expressions
+are actually constant (or at least static), if they have a function call or operator as their root
+node.
```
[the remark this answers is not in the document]
mathemancer: I did quite a bit of research about this. It's not possible to actually access a navigable tree form of the `pg_node_tree` type in SQL (using PG core functionality). All the functions for doing that are C functions which are inaccessible from the DB.

The only way to improve, therefore, would involve learning enough about this poorly-documented internal format to actually construct a tree, and the advantage would be that we could notice (for example) if a pure function had only constant inputs, and call that output therefore constant.

It might make an interesting research project someday...

mathesar/tests/api/test_column_api.py
```diff
@@ -494,7 +494,7 @@ def test_column_update_type_dynamic_default(column_test_table, client):
     response = client.patch(
         f"/api/db/v0/tables/{column_test_table.id}/columns/{column.id}/", data=data
     )
-    assert response.status_code == 400
+    assert response.status_code == 200
```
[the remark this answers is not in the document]
mathemancer: We didn't allow this before since we didn't have a way to construct `cast_to_new_type(adefault_generating_func())` in python. This is because SQLAlchemy receives _all_ default values as simple strings, and it's rather tricky to then reconstruct the underlying function call in the case of a dynamic default value.

This issue is negated when doing the work on the DB.

db/sql/0_msar.sql
mathemancer: Okay, I split a couple of lines up a bit. I did leave one of the function calls on a single line, because splitting it up made it harder (for me) to see the nesting in that case.

db/sql/0_msar.sql
mathemancer: It's accessible in the database layer, but this JSON was constructed in the Python layer. If we have to look up the type of a column during the call where it's generated, it's a round trip to the DB. Plus, I find these lookups to be cleaner and generally _less_ fiddly here in the DB than they were in python.

Another clean solution would be for the front end to add the type name when it originally calls the API, I suppose. There may be other problems with that; not sure.

db/sql/0_msar.sql
mathemancer: Good point; done.

db/sql/0_msar.sql
mathemancer: If I'm understanding your question, the `ELSE` is needed since we want to avoid requiring the `mathesar_types` schema being installed for the functions here to work. You could simply install the `msar` schema and get that functionality by itself if desired. Eventually, I think we want to split this even further.

Temporarily, it's also useful for testing, since the `mathesar_types` schema can only (at this juncture) be installed via the webservice, and our DB tests don't actually have the Mathesar service available in the CI environment. I don't consider this reason to be permanent, since we should eventually move the construction of the `mathesar_types` schema and its installation into SQL.

mathesar/tests/api/test_column_api.py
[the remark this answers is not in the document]
dmos62: Ah, so we're now, when casting, also redefinining the dynamic default to cast itself to the new type, thus not actually losing the dynamic default? That would set up an interesting "entropy" collection scheme, where if you repeatedly cast the dynamic default between a few types you lose information, not that that's a problem.

dmos62 (approved):

mathemancer:
I'm going to go ahead and merge this. @Anish9901 Feel free to do a retroactive review.

db/sql/0_msar.sql
mathemancer: I think you're right. There is a smell here:
- It's a bit inconsistent when we consider type options to be attached to the 'whole' type.
- E.g., if we change from `CHARACTER(256)` to `CHARACTER VARYING(256)`, the frontend has to submit the type option length=256 again.

PR #3113, opened 2023-07-30
Shared queries - Auth handling for query requests, frontend consumer view, basic API tests

### LINKED ISSUE #3034: Handle authentication & authorization bypass for shared entities
I've not added a description, since I'll be implementing this.

Refer [Product spec](https://wiki.mathesar.org/en/product/specs/publicly-shareable-links) for detailed info.

### LINKED ISSUE #3036: Implement frontend flow: User viewing shared entity

pavish:
Fixes #3034 
Fixes #3036 

*The diff appears large due to the tests added, the actual code is of reasonable size.*
*This is mainly frontend code, the backend changes are minimal.*

**This PR**:
* Bypasses authentication for query requests for valid share uuids
* Adds custom common_data for shared tables required by frontend
* Allows viewing shared queries on the frontend
* Adds basic share crud API tests

**Flows built in this PR:**
* Requests to the following endpoints should succeed with a valid `share-link-uuid`:
   - `/api/db/v0/queries/<query_pk/?shared-link-uuid=<uuid>`
   - `/api/db/v0/queries/<query_pk/results/?shared-link-uuid=<uuid>`
* The following user flows should work for the above endpoints:
   - Unauthenticated requests with valid `shared-link-uuid`
   - Authenticated requests made by users who do not have explicit access the query, but have valid `shared-link-uuid`
   - Authenticated requests made by users who have access to the query, with or without valid `shared-link-uuid`

**Testing instructions:**
* Generate a shared link for a query via API:
  - A post request to `/api/ui/v0/queries/1/shares` with `{ enabled: true }` should share the query and generate a slug. The response will be like follows:
    ```
    {
      "id": 1,
      "slug": "584cfd0p-890j-95ec-9ded-c226d46f9542",
      "enabled": true
    }
    ```
* Visit `http://localhost:8000/shares/explorations/584cfd0p-890j-95ec-9ded-c226d46f9542/` and you should be able to see a readonly view of the query.
* The following cases should be taken into account, when the link is valid:
  - a. The link should work when you're not logged into Mathesar.
  - b. The link should work when the user is logged in but does not have access to the query's schema or DB.
  - c. The link should work when the user is logged in and has access to the table.

**Frontend notes**:
* The FE initially used only the `/run` endpoint for running queries to keep code common between exploration viewing route, and data explorer routes.
* However, permission checks made it clear that the FE had to use the `/results` endpoint for the exploration page, which is also the proper endpoint to use for it. We didn't do this earlier due to a time crunch. This change is implemented in this PR.

**Backend notes**:
* The tests added are only for CRUD operations on /shares endpoints of tables and queries.
* The auth bypass tests for the /tables/* and queries/* endpoints are not implemented in this PR.
* They will be covered in upcoming PRs. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/api/db/viewsets/queries.py
+++ b/mathesar/api/db/viewsets/queries.py
@@ -87,7 +87,6 @@ class QueryViewSet(
                 "records": paginated_records.data,
                 "output_columns": columns,
                 "column_metadata": column_metadata,
-                "parameters": {k: json.loads(request.GET[k]) for k in request.GET},
             }
         )
 
--- a/mathesar/api/db/permissions/query.py
+++ b/mathesar/api/db/permissions/query.py
@@ -1,6 +1,8 @@
 from django.db.models import Q
 from rest_access_policy import AccessPolicy
 
+from mathesar.api.utils import SHARED_LINK_UUID_QUERY_PARAM
+from mathesar.api.permission_utils import QueryAccessInspector
 from mathesar.models.users import Role
 
 
@@ -8,18 +10,25 @@ class QueryAccessPolicy(AccessPolicy):
 
     statements = [
         {
+            'action': [
+                'retrieve',
+                'results',
+            ],
+            'principal': '*',
+            'effect': 'allow',
+            'condition_expression': 'is_atleast_query_viewer'
+        },
+        {
             # Restrictions for the create method is done by the Serializers when creating the query,
             # As the permissions depend on the base_table object and not on the query object.
             'action': [
                 'list',
-                'retrieve',
                 'destroy',
                 'create',
                 'run',
                 'update',
                 'partial_update',
                 'columns',
-                'results',
                 'records',
 
             ],
@@ -43,3 +52,11 @@ class QueryAccessPolicy(AccessPolicy):
             qs = qs.filter(permissible_database_role_filter | permissible_schema_roles_filter)
 
         return qs
+
+    def is_atleast_query_viewer(self, request, view, action):
+        query = view.get_object()
+        return QueryAccessInspector(
+            request.user,
+            query,
+            token=request.query_params.get(SHARED_LINK_UUID_QUERY_PARAM)
+        ).is_atleast_viewer()
--- a/mathesar/api/ui/viewsets/shares.py
+++ b/mathesar/api/ui/viewsets/shares.py
@@ -19,7 +19,7 @@ class SharedTableViewSet(AccessViewSetMixin, viewsets.ModelViewSet):
         serializer.save(table_id=self.kwargs['table_pk'])
 
 
-class SharedQueryViewSet(viewsets.ModelViewSet):
+class SharedQueryViewSet(AccessViewSetMixin, viewsets.ModelViewSet):
     pagination_class = DefaultLimitOffsetPagination
     serializer_class = SharedQuerySerializer
     access_policy = SharedQueryAccessPolicy
--- a/mathesar/tests/api/query/conftest.py
+++ b/mathesar/tests/api/query/conftest.py
@@ -12,32 +12,42 @@ def academics_ma_tables(db_table_to_dj_table, academics_db_tables):
 
 
 @pytest.fixture
-def minimal_patents_query(create_patents_table, get_uid):
-    base_table = create_patents_table(table_name=get_uid())
-    initial_columns = [
-        {
-            'id': base_table.get_column_by_name('Center').id,
-            'alias': 'col1',
-        },
-        {
-            'id': base_table.get_column_by_name('Case Number').id,
-            'alias': 'col2',
-        },
-    ]
-    display_names = {
-        'col1': 'Column 1',
-        'col2': 'Column 2',
-        'Checkout Month': 'Month',
-        'Count': 'Number of Checkouts',
-    }
-    display_options = {
-        'col1': dict(a=1),
-        'col2': dict(b=2),
-    }
-    ui_query = UIQuery.objects.create(
-        base_table=base_table,
-        initial_columns=initial_columns,
-        display_options=display_options,
-        display_names=display_names,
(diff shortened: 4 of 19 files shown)


pavish:
Python linting seems to be failing probably due to an issue with GH: https://github.com/centerofci/mathesar/actions/runs/5709234709/job/15467788737?pr=3113.

mathesar/api/db/viewsets/queries.py
```diff
@@ -30,10 +31,22 @@ class QueryViewSet(
...
+        if self.action == "results" or self.action == "retrieve" or self.action is None:
+            # Permission for both these actions are handled at QueryAccessPolicy
+            # Scoping is automatically handled by DRF since these are nested urls
+            # which contain the pk of the query
+            # It is essential to include None here, because it's called again by
+            # `filter_queryset` with action `None` after the query is formed
+            # TODO: Move these actions to separate viewsets to make this cleaner
+            queryset = UIQuery.objects.all().order_by('-created_at')
+        else:
+            queryset = self.access_policy.scope_queryset(self.request, UIQuery.objects.all())
```
dmos62, line 48: I'm having trouble understanding what's happening here. If permissions for these actions are handled in `QueryAccessPolicy`, why not use it here? Is scoping handled automatically, because the url contains the pk, which implies the necessary level of permission?

I trust that this code behaves correctly, but it's pretty hard to understand. You added a TODO comment for making this clearer. I think it would make sense to do that for this PR.

dmos62 (changes requested):
Looks good, appreciate the tidy tests; `QueryViewSet` needs some clarification, see comment.

mathesar/api/db/viewsets/queries.py
pavish: @dmos62 

For both the actions that are listed here `results` and `retrieve`, we do not require scoping, since they request one particular query, which is identified since it's part of the url. `QueryAccessPolicy` takes care of checking if the user has access to that particular query.

> You added a TODO comment for making this clearer. I think it would make sense to do that for this PR.

The work required will need refactoring of this file and bunch of related files, which takes focus away from my current project. I'd like to take it up during the cooldown period or later. I'll create followup tickets for it. Does this sound good?

dmos62: Does the complexity stem from reusing the same viewset and the same access policy for both shared and regular queries? Are there disadvantages to using separate classes for both?

Unfortunately, I'm still having trouble reviewing this. For merging, in my opinion, there needs to be a refactor, or we can hop on a call and figure out how to make the code and documentation clearer without a refactor.

mathesar/api/db/viewsets/queries.py
```diff
@@ -30,15 +31,37 @@ class QueryViewSet(
...
+        """
+        should_queryset_be_scoped = \
+            QueryAccessPolicy.get_should_queryset_be_scoped(self.action)
+        if should_queryset_be_scoped:
+            queryset = self.access_policy.scope_queryset(
+                self.request,
+                UIQuery.objects.all()
+            )
+        else:
+            queryset = UIQuery.objects.all().order_by('-created_at')
```
dmos62, line 61: I noticed that the `order_by` transformations is applied here (which is not useful, I think), but not on the scoped queryset. Is this an oversight?

dmos62 (approved):

mathesar/imports/base.py
```diff
@@ -11,7 +11,7 @@
 
 
 def create_table_from_data_file(data_file, name, schema, comment=None):
-    engine = create_mathesar_engine(schema.database.name)
+    engine = schema._sa_engine
```
dmos62, line 14: @mathemancer quick look? @pavish told me this PR worsened the Postgres connection bottleneck; after a quick look around I found that our `mathesar` layer tests rely on this routine a lot, which doesn't use the engine cache. I'm hopeful that this will improve resource usage in tests outside this PR as well.

seancolsen (approved):
I've reviewed this and it looks good to me.

It's not clear to me whether the back-end review is complete yet, so I'm holding off on merging though. Please feel free to merge when ready.

mathesar/api/db/viewsets/queries.py
pavish: @dmos62 

> but not on the scoped queryset. Is this an oversight?

The `get_queryset` method applies a `order_by` transformation when returning the queryset. So, this is handled for both cases.
```python
    def get_queryset(self):
        queryset = self._get_scoped_queryset()
        schema_id = self.request.query_params.get('schema')
        if schema_id:
            queryset = queryset.filter(base_table__schema=schema_id)
        return queryset.order_by('-created_at')
```

I added a commit to remove the redundant `order_by` transformation above: https://github.com/centerofci/mathesar/pull/3113/commits/bb41400ec9609ba0108dd51da243883d7b896737

mathesar/api/db/viewsets/queries.py
pavish: The discussion was resolved over a call, and @dmos62 added a few commits to make this portion more readable.

mathesar/imports/base.py
mathemancer: I think it may have had something to do with sanitizing the engine. However, it may just be a general oversight compounded by pattern-following. I have no problem with making this change.

pavish: Thanks @dmos62 and @mathemancer. I'm going to go ahead and merge the PR in.

PR #3114, opened 2023-07-31
Dont show "Go to Record Page" for error rows

### LINKED ISSUE #2234: Only show "Go to Record Page" context menu entry for rows that have saved records
## Steps to reproduce

1. Go to the table page

1. Right-click on the row header for the placeholder row at the bottom of the table.

1. Expect that the context menu does not contain a link to the record page.

1. Instead, observe a link, which has `NaN` in its URL

    image

## Implementation notes

- We already have a utility function `rowHasSavedRecord` which should be used here. But that utility function doesn't work as expected when a new row is being created with errors (because the row may in fact have a record in it).

- Note that we do allow "Delete Record" for placeholder rows with unsaved records -- it just removes the record from the front end (without making any requests).

Pulkit0729:
Fixes #2234 

**Technical details**

I have added a functio in records.ts for checking if the given row contains error or not. 
**Screenshots**

[screen-capture (2).webm](https://github.com/centerofci/mathesar/assets/82268257/b1169918-929f-4aee-b77e-674105015049)

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/records.ts
+++ b/mathesar_ui/src/stores/table-data/records.ts
@@ -724,6 +724,15 @@ export class RecordsData {
     return [...get(this.savedRecords), ...get(this.newRecords)];
   }
 
+  rowHasError(row: Row) {
+    const pkColumn = get(this.columnsDataStore.pkColumn);
+    const rowKey = getRowKey(row, pkColumn?.id);
+    if (get(this.meta.rowsWithClientSideErrors).has(rowKey)) {
+      return true;
+    }
+    return false;
+  }
+
   destroy(): void {
     this.promise?.cancel();
     this.promise = undefined;
--- a/mathesar_ui/src/systems/table-view/row/RowContextOptions.svelte
+++ b/mathesar_ui/src/systems/table-view/row/RowContextOptions.svelte
@@ -48,12 +48,15 @@
   }
 </script>
 
-<LinkMenuItem
-  href={$storeToGetRecordPageUrl({ recordId: recordPk }) || ''}
-  icon={iconExternalLink}
->
-  Go to Record Page
-</LinkMenuItem>
+{#if !recordsData.rowHasError(row)}
+  <LinkMenuItem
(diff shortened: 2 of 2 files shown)


seancolsen:
@Pulkit0729 I'll review this, but it will probably be at least a week before I can get to it. I have a lot of other PRs in my review queue currently.

seancolsen (changes requested):
@Pulkit0729 I took a look at this today. Sorry that it's accumulated merge conflicts now. Can you fix these? I'd like you to make some other changes too.

- I don't think it's a good idea to use `rowsWithClientSideErrors`. We need to differentiate between rows that have already been saved and rows that have not yet been saved. A row can be saved but still have client side errors too &mdash; for example, if the user tries to edit the value of a cell within a saved row but the server responds with a validation error due to the violation of a unique constraint. Instead, I think it would be better to check if the record within the row has a value set for the primary key column. If the PK value is null, then we can deduce that the row has not yet been saved.

- Also, the code you added to `records.ts` has a subtle reactivity problem.

    I imagine you'll probably discard this code to address the above point, but I want to give some critique on it anyway because I don't want you to repeat this pattern in whatever new solution you come up with.

    Within the component the value of the expression `recordsData.rowHasError(row)` will only change when `row` changes. But that expression is also dependent on `recordsData.columnsDataStore.pkColumn` and `recordsData.meta.rowsWithClientSideErrors` and changes to those stores will not reactively update the value returned from your `rowHasError` method because you've used the `get` function. One way to fix this would be to move the logic from your `rowHasError` method into a series of reactive statements inside `RowContextOptions.svelte`.

seancolsen:
@Pulkit0729 are you still planning to work on this? If we don't hear from you in the next couple weeks we may need to close this PR.

PR #3117, opened 2023-07-31
Remove db superuser requirement

### LINKED ISSUE #2990: The requirement of superuser postgresql access is problematic
## Problem
Mathesar needs a Postgres superuser to function correctly, from the docs at https://docs.mathesar.org/installation/build-from-source/ 

## Proposed solution
The mathesar user should not require superuser access. 

## Additional context
The superuser is a global permission meaning that a user that has superuser permission will be able to access (and do stuff) not on the mathesar but *all* the databases of the RDBMS. Considering that many production systems have a single RDBMS hosting many application this is a major problem since the mathsar user won't have any access boundaries on the same RDBMS. The mathesar user access can be unlimited but *must* be bounded without the mathesar database.

silentninja:
Fixes #2990 

The PR now creates a non-existent database during installation only if the database user has the privilege to create a database and gracefully prints a warning message instead of breaking the install script.

I have also updated the documentation to remove the need for superuser privilege and replaced it with ownership privilege of the database

**Additional context**
I couldn't change our test case to use just the ownership privilege of the database as our test suite needs some more additional privileges like `CREATEDB` which made the refactor really complicated.
But I tested it by commenting out the `drop_database` and `create_database` functions and the test cases that don't depend on those functions pass as expected. I think we should revisit this when we rework our test cases.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


docs/docs/installation/docker/index.md
```diff
@@ -24,8 +24,8 @@ You'll need to install **[Docker](https://docs.docker.com/desktop/)** v23+
 #### Database for Mathesar's internal usage
 You'll need to:
 
-- Create a PostgreSQL database for Mathesar's internal usage.
-- Create a database user for Mathesar to use. The user should be a `SUPERUSER`, [see PostgreSQL docs for more information](https://www.postgresql.org/docs/13/sql-createrole.html).
+- Create a database user for Mathesar to use.
+- Create a PostgreSQL database for Mathesar's internal usage owned by that database user, [see PostgreSQL docs for more information](https://www.postgresql.org/docs/9.0/ddl-priv.html)
```
mathemancer, line 28: Use the v13 docs, please.

docs/docs/installation/docker/index.md
```diff
@@ -40,11 +40,11 @@ Have the following information for all databases you'd like to connect to Mathes
 - Database hostname
 - Database port
 - Database name
-- Database username (should be a `SUPERUSER`, see above)
+- Database username (should be owner of the database, see above)
```
mathemancer, line 43: Do we actually need owner? Doesn't `USAGE` on the DB suffice?

docs/docs/installation/docker/index.md
```diff
@@ -40,11 +40,11 @@ Have the following information for all databases you'd like to connect to Mathes
...
 - Database password
 
 !!! warning "Database creation"
-    Whenever the Docker container is started, we will attempt to create any databases in this list that don't already exist. So you don't need to ensure that they are created before installation.
+    Whenever the Docker container is started, we will attempt to create any databases in this list that don't already exist if the user has [`CREATEDB` privilege](https://www.postgresql.org/docs/current/sql-createrole.html). So you don't need to ensure that they are created before installation.
```
mathemancer, line 47: Use the v13 docs, please.

docs/docs/snippets/docker-compose-prerequisites.md
```diff
@@ -24,7 +24,7 @@ You can create a new PostgreSQL database while setting up Mathesar or use our UI
 To connect Mathesar to an existing database:
 
 - The external database should be able to accept network connections from your Mathesar server.
-- You'll need to set up a database user for Mathesar to use. The user should be a `SUPERUSER`, [see PostgreSQL docs for more information](https://www.postgresql.org/docs/13/sql-createrole.html).
+- You'll need to set up a database user for Mathesar to use. The user should own that database, [see PostgreSQL docs for more information](https://www.postgresql.org/docs/9.0/ddl-priv.html)
```
mathemancer, line 27: v13 docs, please!

mathemancer, line 27: I think we can get away with the user having `USAGE` on the database.

mathemancer (changes requested):
Shouldn't we go for `USAGE` on databases? I think the owner of the database is unique, and I'd assume that you specifically _wouldn't_ want to let mathesar use that user if you're worried about security.

kgodey:
I'd like to review this from a documentation perspective once other issues are addressed, please tag me before merging.

silentninja:
I don't think databases have `USAGE` privilege.  Postgres docs mention only `CREATE | CONNECT | TEMPORARY | TEMP` as the available permissions. Moreover, even if it did have `USAGE` privilege, we still need to `DROP` and `CREATE` schemas and other objects which a `USAGE` privilege does not allow. I couldn't even use `GRANT ALL ON DATABASE` because it doesn't allow for deleting a schema.


[Postgres docs for reference](https://www.postgresql.org/docs/current/sql-grant.html)
```
GRANT { { CREATE | CONNECT | TEMPORARY | TEMP } [, ...] | ALL [ PRIVILEGES ] }
    ON DATABASE database_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
    [ GRANTED BY role_specification ]
```

mathemancer (approved):
Okay, you're correct about `USAGE`. I did some more reading, and it seems we're in a bit of a quandary, and the actual review we need here is from a kind of product level. My biggest problem with this change is that the `OWNER` of a database is unique, and I would imagine a DBA would be similarly reluctant to change the unique owner of a database to a `mathesar` user, or to grant Mathesar access to the pre-existing, unique, owner of a database's user credentials.

I suppose the plus side is that a DBA could set up a special user who owns nothing else, and make that user the owner of a database they want to use Mathesar with. However, that would _remove_ ownership from whoever owned it previously. And if the previous owner was not a superuser, that reduces the privileges of that other user's account.

After some reading and thinking, the only actual permissions we need on a database are `CREATE` and `CONNECT`, and we'd probably need `USAGE` and `CREATE` on the `public` schema, since I assume parts of our codebase assume access to `public`. That would let us connect to the database, and set up Mathesar schemas, which is all that's needed for installation. However, you wouldn't really be able to see any objects in any schemas except the `public` one, or ones created by the `mathesar` user. 

@kgodey Do you have any product considerations here we should think about? Is "Mathesar needs unique ownership over DBs it connects to" a reasonable choice, or a  better one than "Mathesar needs a superuser on your DB cluster"? I'm not sure. As the originator of the issue pointed out, a superuser has access to _all_ DBs on a cluster, whereas an owner could be restricted to just the one. The alternative is to sink more development time into coming up with a more precise privilege stance and testing that it works.

mathemancer:
Also, @silentninja I'm approving so I'm not blocking anything once @kgodey weighs in, but please wait for her input before merging.

kgodey (dismissed):
I'm in agreement with @mathemancer's concerns.

>  Is "Mathesar needs unique ownership over DBs it connects to" a reasonable choice, or a better one than "Mathesar needs a superuser on your DB cluster"?

I think I'd rather have "Mathesar needs a superuser on your DB cluster" than "Mathesar needs unique ownership over DBs it connects to". At least you can have multiple superusers, you can only have one owner of a DB, and I don't think most people would be comfortable setting that to a machine user (or using their own credentials). 

> The alternative is to sink more development time into coming up with a more precise privilege stance and testing that it works.

I thought this was what we were doing for this issue.

---

To get this merged, I would be okay with updating the docs to say you _either_ need the user to be a superuser or be the database owner, but I don't think it makes sense to replace the superuser instructions with DB owner instructions. However, I don't think this fixes the larger issue, and we still need to do work on coming up with a more precise privilege set and using that instead.

silentninja:
> The alternative is to sink more development time into coming up with a more precise privilege stance and testing that it works.

> I thought this was what we were doing for this issue.


As I mentioned in my [previous comment](https://github.com/centerofci/mathesar/pull/3117#issuecomment-1663232128), Postgres won't let you delete a `Schema` unless you are the owner of the database or a superuser. We cannot have precise permissions because Postgres does not allow for a broad delete privilege to all the schemas Something like `GRANT DELETE ON DATABASE db` does not **exists**.

kgodey:
@silentninja Sorry, I didn't quite process your earlier comment since I was relying on @mathemancer's summary of the situation.

We should update the docs to say you _either_ need the user to be a superuser or be the database owner, and _explain why_. 

The fundamental issue we are trying to solve is that potential users are hesitant to accept the permissions we need. This PR does not fix that issue – it just replaces one set of permissions with another, and there are reasonable objections to both. Explaining why we need the permissions in a clear and succinct way would at least address the fundamental issue.

I also think we need to have a larger product discussion around how we can work around this in the future – perhaps we can disable the ability to delete a schema through the UI (if that's the only thing we need superuser/owner permissions for), or perhaps we can have some sort of "sudo mode" like GitHub has for operations that require superuser/owner credentials. @silentninja and @mathemancer I'll leave it up to you to create an issue to track this future work.

silentninja:
> The fundamental issue we are trying to solve is that potential users are hesitant to accept the permissions we need. This PR does not fix that issue

This is correct. I was trying to fix #2990 which is aimed at having privileges specific to a database(based on the description) but this PR title made it looks like it was solving a different issue. 

I will create two new issues to unblock this PR as it is preventing a user from using Mathesar 
- Documenting the requirement for superuser or owner privileges.
- A draft issue for tracking product-level discussion on avoiding the need for powerful privileges for superuser/owner

As for this PR, I will update the docs to say you either need a superuser or a database owner

rajatvijay:
@seancolsen assigning this to you since this needs documentation review.

seancolsen (approved):
I've added some minor docs language changes in a766eacd1137ad4d2765c1600926f4d3aec47474, and I've approved this PR from a documentation perspective only. I'm re-assigning it to you in @silentninja, just in case there is more work here that needs to be done. I haven't read through the entire diff or comments here.

PR #3121, opened 2023-08-02
Add Postgres to Mathesar docker image

### LINKED ISSUE #3072: Add a Docker image containing the Mathesar server along with a Postgres server
## Problem
The current docker image requires the user to already have a Postgres server running on their machine. This is a hassle for someone looking to try out Matheasar quickly.

## Proposed solution
Provide a docker image containing a Postgres server along with the Mathesar server

silentninja:
Fixes #3072
The PR adds a Postgres server to the docker image. The in-built database server starts only if the environment variable `DJANGO_DATABASE_URL` is not set, this is done to reduce unnecessary resources. 

The following behavior takes place when `DJANGO_DATABASE_URL` env variable is not set
- The inbuilt Postgres server is started
- Default DB user is created with 
  -  username  as `mathesar` 
  -  password  as `mathesar` 
- Default database is created with
  -  database name as `mathesar_django`
- The  `DJANGO_DATABASE_URL` env variable is set to point to the default database running inside the docker image
   - The connection string value for the env variable will be `postgres://mathesar:mathesar@localhost:5432/mathesar_django`
- The  value of the  `MATHESAR_DATABASES` env variable is set to `(mathesar_tables|postgresql://mathesar:mathesar@localhost:5432/mathesar)`

**Technical details**
Most of the information is the same as the [official Postgres Dockerfile](https://github.com/docker-library/postgres/blob/43d17d5ced92f230fa8c196e746f2e2aa288e5e8/13/bullseye/Dockerfile). I made the following changes
- Added the Postgres repo for installing Postgres 13. Debian buster(used by our docker image) comes with Postgres 11 by default
- Set the following default credentials for the database that is created initially
  -  username  as `mathesar` 
  -  password  as `mathesar` 
  -  database as `mathesar_django`

  

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Dockerfile
```diff
@@ -3,7 +3,112 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+	gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4; \
+	gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu; \
+	gpgconf --kill all; \
+	rm -rf "$GNUPGHOME" /usr/local/bin/gosu.asc; \
+	apt-mark auto '.*' > /dev/null; \
+	[ -z "$savedAptMark" ] || apt-mark manual $savedAptMark > /dev/null; \
+	apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false; \
+	chmod +x /usr/local/bin/gosu; \
+	gosu --version; \
+	gosu nobody true
```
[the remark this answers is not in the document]
mathemancer: Also the certificates and wget are already installed, I believe. Alsoalso, I don't think we need `wget`. Just use curl.

Dockerfile
```diff
@@ -3,7 +3,112 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+
+# make the "en_US.UTF-8" locale so postgres will be utf-8 enabled by default
+RUN set -eux; \
+	if [ -f /etc/dpkg/dpkg.cfg.d/docker ]; then \
+# if this file exists, we're likely in "debian:xxx-slim", and locales are thus being excluded so we need to remove that exclusion (since we need locales)
```
mathemancer, line 43: We're not, right? We're starting from the python buster image, which isn't on slim.

Dockerfile
```diff
@@ -3,7 +3,112 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+
+RUN mkdir /docker-entrypoint-initdb.d
+
+ENV PG_MAJOR 13
+
+RUN set -ex; \
+    curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - ; \
+    echo "deb http://apt.postgresql.org/pub/repos/apt/ buster-pgdg main" > /etc/apt/sources.list.d/pgdg.list; \
+    mkdir -p /etc/postgresql-common/createcluster.d/; \
+    echo "create_main_cluster = false" > /etc/postgresql-common/createcluster.d/no-main-cluster.conf; \
```
mathemancer, line 60: Why specify this? What's wrong with the defaults?

Dockerfile
```diff
@@ -3,7 +3,112 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+ENV PG_VERSION 13.11-1.pgdg110+1
+
+# make the sample config easier to munge (and "correct by default")
+RUN set -eux; \
+	dpkg-divert --add --rename --divert "/usr/share/postgresql/postgresql.conf.sample.dpkg" "/usr/share/postgresql/$PG_MAJOR/postgresql.conf.sample"; \
+	cp -v /usr/share/postgresql/postgresql.conf.sample.dpkg /usr/share/postgresql/postgresql.conf.sample; \
+	ln -sv ../postgresql.conf.sample "/usr/share/postgresql/$PG_MAJOR/"; \
+	sed -ri "s!^#?(listen_addresses)\s*=\s*\S+.*!\1 = '*'!" /usr/share/postgresql/postgresql.conf.sample; \
+	grep -F "listen_addresses = '*'" /usr/share/postgresql/postgresql.conf.sample
+
```
mathemancer, line 80: Why does this part exist? The comment doesn't help me understand.

Dockerfile
```diff
@@ -3,7 +3,112 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+RUN mkdir -p /var/run/postgresql && chown -R postgres:postgres /var/run/postgresql && chmod 2777 /var/run/postgresql
```
mathemancer, line 81: Why?? All these things are set up by installing postgres, right?

mathemancer (changes requested):
My main overall comment is that this is _waay_ more complicated than it warrants. I can't see any advantages we're getting over just 
- adding the postgres repo
- Installing postgres
- Running it with `pg_ctl`  (and enabling it so it starts whenever you start the container)

We don't need to be able to do 90% of the stuff enabled by the run script, IMO, and using the above would avoid all manner of weird `chmod`ding and `chroot`ing.

I can already tell by looking at this that no one's going to want to maintain it, and we're going to end up with weird cruft, things you cant touch without breaking, etc. And out-of-date versions of packages.

Dockerfile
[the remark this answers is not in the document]
silentninja: I got a package missing error when I tried to install it from the Debian repo and I ended up copying the instructions from the gosu repo. But it was an oversight on my end, I replaced it with installing from debian repo.

silentninja:
Thanks for the review @mathemancer. I removed most of the irrelevant steps. I didn't remove the steps that change the permissions during runtime because we need it for users running docker without root privileges.

kgodey (approved):
Made some minor updates to docs. @silentninja you forgot to fix the nav and the main page, I updated those too.

I did not review anything except the docs.

@mathemancer please also review the docs.

Dockerfile
```diff
@@ -3,7 +3,49 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
+# server stops when all sessions are terminated.
+#
+# See https://www.postgresql.org/docs/12/server-shutdown.html for more details
+# about available PostgreSQL server shutdown signals.
+#
+# See also https://www.postgresql.org/docs/12/server-start.html for further
+# justification of this as the default value, namely that the example (and
+# shipped) systemd service files use the "Fast Shutdown mode" for service
+# termination.
+#
```
mathemancer, line 44: I think you can remove this part of the comment.

docs/docs/installation/docker/index.md
```diff
@@ -19,46 +19,18 @@ You should have permission to run Docker containers on the system.
...
-!!! warning "Database creation"
-    Whenever the Docker container is started, we will attempt to create any databases in this list that don't already exist. So you don't need to ensure that they are created before installation.
-
 ## Installation Steps
 
 1. Run the Mathesar Docker Image
 
     ```bash
     docker run \
       --detach
```
mathemancer, line 28: missing backslash

db-run.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
```
mathemancer, line 2: I think you should `set -e` here. There's no point in continuing if this breaks (is there?).

db-run.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Loads various settings that are used elsewhere in the script
+# This should be called before any other functions
+docker_setup_env() {
+	declare -g DATABASE_ALREADY_EXISTS
+	# look specifically for PG_VERSION, as it is expected in the DB dir
+	if [ -s "$PGDATA/PG_VERSION" ]; then
```
mathemancer, line 9: This test is defective. 
- Missing a `$`
- `$PG_VERSION` is not set.

Wouldn't the major version suffice anyway?

mathemancer (changes requested):
This is a major improvement over the last iteration. Thank you for that!

However, while it works in the docker compose setup, it doesn't work when run as a single image. The problem is in the `db-run.sh` script. It has a couple of typos that make the 'is there already a db?' test defective.

### Docs Review
- I suspect we can lower the required docker version by alot. Maybe not worth worrying about right now
- The uninstalling instructions are out-of-sync for the docker install. If they follow the instructions for installing, they won't have the separate DB to worry about.
- There's a missing backslash. See my specific comment.
- We need to add instructions for starting/stopping the service.

silentninja:
@mathemancer I ran across a lot of bugs  or unintented effects trying to use `pg_createcluster`. These are the options I tried

#### pg_createcluster (Works with some inconvenience to the user)

When creating a cluster with `pg_createcluster`, it ends up creating the conf directory in `/etc/postgresql/<version>/<cluster-name>/` instead of the data directory which is in `/var/lib/postgresql/<cluster-name>`. The config files are needed for starting the cluster.  So now the user has to mount `/etc/postgresql/<pg-version>/<cluster-name>/` in addition to the data directory, So in case of a postgres version update the user needs to make changes to the mounted volume path.

#### initdb and pg_ctl (Fails due to permission errors)

I tried looking into `initdb` and `pg_ctl` as a alternative. Both of these tools cannot be run as root and docker initializes a volume folder with root privileges(pg_createcluster changes the ownership during the runtime as it is ran as root process). So `initdb` does cannot change the privilege of the data directory(mounted as a volume) and we need to change the privileges before using `initdb` as a postgres user which is what the code in the initial commit of the PR was actually doing.
 
 
 #### Mix of both pg_createcluster and pg_ctl(Fails on restart)
 
I thought I finally got something working with a good UX in this commit but unfortunately it has its own set of problems too. when the postgres server it always starts in recovery mode. @mathemancer Is there anything I am missing here

kgodey:
> I suspect we can lower the required docker version by alot. Maybe not worth worrying about right now

Please create an issue to track this.

db-run.sh
silentninja: No, the test is correct. `PG_VERSION` is a file inside the data directory which is created when the cluster created.

Dockerfile
```diff
@@ -3,24 +3,24 @@ ARG PYTHON_REQUIREMENTS=requirements.txt
...
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*; \
     :
 
 ENV PATH $PATH:/usr/lib/postgresql/$PG_MAJOR/bin
 
-ENV PGDATA /var/lib/postgresql/data
-
-VOLUME /var/lib/postgresql/data
+ENV PGDATA /var/lib/postgresql/mathesar
+VOLUME /etc/postgres/15/mathesar
```
mathemancer, line 22:
```suggestion
VOLUME /etc/postgresql/15/mathesar
```
Also, why not mount `/etc/postgresql/` instead? I thought you were going to try that so we can avoid fiddling around with mounting different version numbers if we want to upgrade and so on.

mathemancer (changes requested):
I think this simplifies things nicely, but:
- there are a number of unaddressed comments from my previous review; some are important.
- I think we need to block this on compatibility with PostgreSQL 15. There's (at least one) issue that keeps the tests from passing on PG15.
- See my specific comment about the mount point

silentninja:
@mathemancer I made changes to Github action to use the docker image and pass the credentials as an environment variable when calling in docker-exec (to avoid breaking any workflow using docker-compose). So we are using the inbuilt postgres server to test Postgres version 13,14 and 15

silentninja:
@mathemancer I addressed all the review feedback, and added the docker image to our test suite. The docs don't need to be updated as there aren't any breaking changes to our test runner. 
I would like to get a final review to make sure your concerns are addressed.

mathemancer (commented):
Overall, I think this is getting much better! However:
- Running the container twice results in an error:
```
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
Starting inbuilt database
Error: cluster configuration already exists
```
See my specific  comment about the documented `run` command.

docs/docs/installation/docker/index.md
```diff
@@ -19,49 +19,17 @@ You should have permission to run Docker containers on the system.
...
-      --detach
-      -e DJANGO_DATABASE_URL='<replace with a postgres connection string>' \
-      -e MATHESAR_DATABASES='(<unique_db_key>|<replace with a postgres connection array>)' \
-      -e SECRET_KEY='<replace with a 50 character string>' \
-      -e ALLOWED_HOSTS='.localhost, 127.0.0.1, [::1]' \
+      --detach \
       -v static:/code/static \
       -v media:/code/media \
+      -v postgresql_config:/etc/postgresql/ \
+      -v postgresql_data:/var/lib/postgresql/data \
```
mathemancer, line 32: Is this correct? Am I missing something? I think this is misaligned with the volume specified in the dockerfile.

mathemancer (changes requested):
Sorry, mis-clicked on my previous review. It should have been 'request changes' and included the specific comment about the volume mountpoint.

mathemancer (changes requested):
I think the config volume still doesn't match.

mathemancer (approved):
Okay, I think this is working, and I can find no other problems.

This is a real improvement to the installation for a user trying out Mathesar, nice work!

PR #3168, opened 2023-08-24
Add constraint copying to column extration logic

### LINKED ISSUE #3160: Critical regression: Extraction removes fkey links
## Description

When extracting columns, if any extracted column is a foreign key, it loses its reference.

## Expected behavior

We should not drop those references.

## To Reproduce

Follow the steps of the Library demo video until you extract the columns to create the Books table. Note that in the books table, the Publisher and Author columns are no longer fkeys.

## Additional context

This almost certainly occurred when moving this logic to SQL. Sadly, it seems we didn't have a test case covering this eventuality.

mathemancer:
Fixes #3160 

This adds logic to correctly copy foreign key constraints when extracting columns to a new table.

**Technical details**

This logic was lost when originally porting the column extraction to SQL. We missed it due to the fact that our new constraint logic operates almost entirely on the 'table constraint' level, whereas the previous python logic sometimes operated on the 'column constraint' level.

I think the new paradigm of 'everything is a table constraint' will make things more consistent moving forward, avoiding this kind of problem in the future.

Sadly, I had to comment out the same tests here that I did in my PR #3156 . This will probably cause merge conflicts when we merge this back into develop.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/sql/0_msar.sql
```diff
@@ -1878,8 +1913,8 @@ DECLARE
   constraint_defs __msar.con_def[];
 BEGIN
   fq_table_name := format('%s.%s', __msar.get_schema_name(sch_oid), quote_ident(tab_name));
-  column_defs := msar.process_col_def_jsonb(null, col_defs, false, true);
-  constraint_defs := msar.process_con_def_jsonb(null, con_defs);
+  column_defs := msar.process_col_def_jsonb(0, col_defs, false, true);
+  constraint_defs := msar.process_con_def_jsonb(0, con_defs);
```
silentninja, line 1917: I really got confused as the function name is very similar to `msar.process_col_def_jsonb`. I would prefer to call it as `msar.process_column_def`. Just wanted to put my thoughts out, no need to take any action now.

db/sql/0_msar.sql
```diff
@@ -2333,7 +2369,7 @@ BEGIN
     msar.get_relation_namespace_oid(tab_id),
     new_tab_name,
     extracted_col_defs,
```
silentninja, line 2371: Now I see why I missed this bug when reviewing the column moving logic PR. I was assuming `extracted_col_defs` contained constraints too. 

`msar.process_col_def_jsonb`  actually handles `NOT NULL` constraint because it cannot be added as a table constraint. Handling constraints in two different places is pesky. We should add comments to the `msar.process_col_def_jsonb` function to clear up the confusion

silentninja (approved):
Looks good! I made some comments related to readability and should not block this PR as the requests are related to broader concerns. Please feel free to merge this PR.

db/sql/0_msar.sql
mathemancer: I recognize that in this case the abbreviations are a little confusing. I'd rather make more distinguishable abbreviations in this case rather than lengthening the function names. As described elsewhere, horizontal space is at a premium for function signatures (including names). I'll try to think of something eaiser.

db/sql/0_msar.sql
mathemancer: As I described above, the problem is actually that we're dealing with constraints in both ways in Python, and I wanted to unify it to table-only constraints. Irritatingly, `NOT NULL` isn't really implemented as a constraint at all in PostgreSQL. It's a column property (like its type or default). I.e., the nullability of a column is stored as a boolean in the `pg_attribute` table, not as a row in the `pg_constraint` table. So the fact that you can't store a null value in a given column is treated the same as the fact that you can't store a value of `'abc'` in a `numeric` column. However, SQLAlchemy, numerous online resources, and even the PostgreSQL docs often call `NOT NULL` a constraint, leading to this seeming inconsistency, and lots of confusion all around. Quite irritating.

I'll try to add some comments to that effect, though.

db/sql/0_msar.sql
```diff
@@ -2324,6 +2359,7 @@ The extraction takes a set of columns from the table, and creates a new table fr
 */
 DECLARE
   extracted_col_defs CONSTANT jsonb := msar.get_extracted_col_def_jsonb(tab_id, col_ids);
+  extracted_con_defs CONSTANT jsonb := msar.get_extracted_con_def_jsonb(tab_id, col_ids);
   fkey_name CONSTANT text := msar.build_unique_fkey_column_name(tab_id, fk_col_name, new_tab_name);
```
Anish9901, line 2363: Just a minor nitpick this name got me confused as to if we were creating a unique foreign key i.e a one-to-one link instead of a many-to-one link between the extracted tables.

PR #3170, opened 2023-08-24
API for database config 

### LINKED ISSUE #3157: API for adding/modifying user database credentials in Mathesar
This issue is for tracking the backend work needed for implementing the design in #3159

## Proposed solution
- Create API to add the user database credentials in the internal Django database
  - The following information will be sent from the client
    - Unique name for the database
    - Database name
    - Database username
    - Database password
    - Database host
    - Database port
  - The database credentials other than the unique name should be encrypted using the `SECRET_KEY` when stored in the database and decrypted back using the same `SECRET_KEY` when decrypting back
  - Mathesar schema (containing Mathesar types and schema) should be automatically installed into the user database when it is added to Mathesar
  - Can be added only by the superuser
 
- Create API to edit the user database credentials in the internal Django database
  - The following information can be edited
    - Database name
    - Database username
    - Database password
    - Database host
    - Database port
  - Can be edited only by the superuser
  
- Create API to remove the user database credentials
  - Optionally, accept boolean to remove Mathesar Schema when the user database credential is removed from Mathesar
  - If the boolean to remove Mathesar schema is `True`, we should remove Mathesar schema installed on the database first before removing the credentials from Mathesar
  - Can be removed only by the superuser

### LINKED ISSUE #3158: Update the database objects created using the env to be consistent with the database objects created using the API
Follow-up of #3157  Database objects added using the environment variable `MATHESAR_DATABASES` should also include `username`, `password`, `hostname` and `port` to be consistent with the database objects created using the database API. These databases objects added using the environment variable should also be read-only and should not be editable using the API.

## Additional context

- [This is the line](https://github.com/centerofci/mathesar/blob/8018a4404301e896b95ffe857acb55669456a434/mathesar/apps.py#L13) which adds the environment variables

Anish9901:
Fixes #3157
Fixes #3158

**Technical details**

API Signature:
- Create DB credentials:
```python
POST /api/db/v0/databases/
```
```sql
{
    "name": "mathesar", (unique)
    "db_username": "mathesar",
    "db_password": "mathesar",
    "db_host": "mathesar_dev_db",
    "db_port": 5432
}
```
- Alter DB credentials:
```python
PATCH /api/db/v0/databases/<id>
```
```sql
{
    "name": "mathesar", (unique)
    "db_username": "mathesar",
    "db_password": "mathesar",
    "db_host": "mathesar_dev_db",
    "db_port": 5432
}
```
- Delete DB credentials:
```python
DELETE /api/db/v0/databases/<id>
```
- Delete DB credentials & msar SQL & types schemas:
```python
DELETE /api/db/v0/databases/<id>/?del_msar_schemas=True
```
Changed `common_data` dict:
- If someone edits their database credentials outside of mathesar the django model for that connection becomes stale.
- To address this the `databases` key in the `common_dict` will now provide a dict something like this for the corresponding stale database object:
```sql
{
    'name': 'mt_tables',
    'editable': True,
    'error': 'Error connecting to the database'
}
``` 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
There is a issue with the install script that I'd like to point out, whenever there is a change in the function signature of our SQL functions the install script executes without termination even though psycopg exceptions can be noticed in the logs.

I don't think it's a good idea to reinstall those schema without the consent of the users as there might be db objects dependent on those schemas.

mathesar/api/db/viewsets/databases.py
```diff
@@ -32,6 +34,35 @@ def get_queryset(self):
             Database.objects.all().order_by('-created_at')
         )
 
+    def create(self, request):
+        serializer = DatabaseSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        credentials = serializer.validated_data
+        if is_valid_pg_creds(credentials):
```
silentninja, line 41: It would be better to perform this validation inside the serializer.

mathesar/api/serializers/databases.py
```diff
@@ -7,12 +7,17 @@
 
 
 class DatabaseSerializer(MathesarErrorMessageMixin, serializers.ModelSerializer):
+    name = serializers.CharField()
+    db_username = serializers.CharField()
```
silentninja, line 11: Is there a reason to explicitly specify the field types here, Doesn't drf infer from the model? It would be good to add comments on the need for overriding the field type

silentninja (changes requested):
This is a good start! However, there are a few issues

- The Database model does not contain the unique name field. We cannot use the database name as a reference as it can be changed too even though the data inside the database is the same. Please add it to the `Database` model.

- I have made specific comments to highlight other issues

silentninja:
> whenever there is a change in the function signature of our SQL functions the install script executes without termination even though psycopg exceptions can be noticed in the logs.

I thought we decided not to have a breaking function signature between release. If you are noticing this in the `develop` branch I don't think it is a problem. If not please raise an issue for it

> I don't think it's a good idea to reinstall those schema without the consent of the users as there might be db objects dependent on those schemas.

For the sake of easier upgrades, we decided to not have any breaking changes. So I don't think it will affect any dependent objects

Anish9901:
> The Database model does not contain the unique name field. We cannot use the database name as a reference as it can be changed too even though the data inside the database is the same. Please add it to the Database model.

`unique` is set to `True` in the Database model for the `name` field, am I missing something?

mathesar/exception_handlers.py
```diff
@@ -61,6 +62,8 @@ def mathesar_exception_handler(exc, context):
         response = exception_handler(api_exception, context)
 
     if response is not None:
+        if isinstance(exc, MathesarValidationException):
+            return response
```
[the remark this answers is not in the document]
Anish9901: [This](https://github.com/centerofci/mathesar/blob/00204a7198cd21351c121e739fcd76014790942b/mathesar/api/exceptions/validation_exceptions/base_exceptions.py#L21) not being a list was causing `is_pretty()` to fail, fixed now!

silentninja:
> unique is set to True in the Database model for the name field, am I missing something?

If the database object is added using a .env file, the only way to track the database object across database name change is by using a unique database key which is not tied to the property of the database. So we need a unique database key on top of database name

silentninja (approved):
Thanks @Anish9901, it looks good to me! I would like to wait until the frontend work is done to merge this PR as it will be easier to confirm if something is broken.

Issue #3069, opened 2023-07-19
Support for column descriptions/comments


domdomegg:
## Problem

Managing large Postgres instances can become difficult if they're not well documented. As such, it's often desirable to have a way to add explanatory comments on tables or columns. This can be helpful for keeping things organised and understandable.

This also would align with features from other accessible data management platforms, like Airtable: https://support.airtable.com/docs/adding-descriptions-in-airtable

## Proposed solution

Postgres allows storing comments alongside tables and columns. They can be read back with the comment retrieval functions: https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-COMMENT-TABLE

It would be neat if Mathesar could store and retrieve database comments using these functions.

## Additional context

I think we could do comment retrieval without being able to write them, at least as an initial version.

kgodey:
We already support reading and writing schema and table comments (they're surfaced as "description" in the UI to make it more user-friendly for non-technical users). We don't support column comments yet, but it should be quick to add.

We may also want to consider some sort of popover or help text that explains that descriptions are actual database comments.

seancolsen:
I updated the title to reflect that the remaining work relates only to _columns_.

@domdomegg if you're experiencing any difficulty using comments on tables or schemas, please let us know. It should already be working bidirectionally. That said, if you update a table or schema comment outside Mathesar, you'll need to [manually sync those changes](https://docs.mathesar.org/user-guide/syncing-db/) into Mathesar before you can see the comments in Mathesar.

kgodey:
@seancolsen I'd consider also making a separate issue for this potential action item:

> We may also want to consider some sort of popover or help text that explains that descriptions are actual database comments.

seancolsen:
Good idea, @kgodey. I've splintered the design work off into #3071. As such, I've un-tagged this issue as "design" so that we can keep it focused on implementing comments for columns (which I think is rather straightforward and should not require any design work). I've also un-tagged this as draft because I think we have enough info to get started on this.

domdomegg:
Thanks both! I think the main thing I guess I wanted to raise was the column descriptions, which I think this issue tracks accurately (I didn't see the table descriptions, and while I care about it being stored in the Postgres database for interoperability with other tools, I don't think it's critical that this is made clear).

kgodey:
Thanks for clarifying @domdomegg!

seancolsen:
[...]

PR #3186, opened 2023-08-28
Comments for columns, backend impl.

dmos62:
Backend aspect of #3069

The API body field that's been added is `description`. See this test for details: https://github.com/centerofci/mathesar/pull/3186/files#diff-bbad274a71f6001d9791919ecf60e3eb7428896050b8fa7055b7ae4bf03c9f55R828

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
Sorry this took so long to review!

Overall, this is working, but with some limitations. I'm not seeing a way to _remove_ the comment from a column, or set it to `NULL` when desired. I think you should emulate the way that we're handling dropping the default value setting of a column (or use some other method).

As far as the style, I would prefer to avoid looping over arrays since the syntax is inconsistent and hard to remember (compared with looping over results). I prefer the `unnest` style used in the rest of the file. I've made a specific code suggestion to this effect.

Line length: I'm trying to be pretty aggressive about limiting line lengths to 100 chars, since horizontal space is at a premium where a user may want to display these functions to see their docstrings. Same for the signatures. I'm using
```sh
grep '.\{101,\}' db/sql/0_msar.sql
```
to find lines of 101 or more characters. You have a few long lines.

My biggest change to request is to avoid mixing processed and unprocessed arguments to functions. As an example, you have a function that expects unsafe, unprocessed schema and table names, but also a safe, quoted, processed comment. This is quite dangerous, and mixing these will lead to bugs and security issues if the caller isn't extremely careful.

I suggest using the following functions:

- `__msar.comment_on_column(tab_name text, col_name text, comment_ text)`:  
  This function should assume safety for all values, same as other `__msar`-namespaced functions, and thus perform no further processing. You've already written this function.
- `msar.comment_on_column(tab_id oid, col_id integer, comment_ text)`:
  This function should assume the `comment_` is unprocessed, and do its own quoting.

If you _want_, you could add another:

- `msar.comment_on_column(sch_name text, tab_name text, col_name text, comment_)`:
  This function should assume all arguments are unprocessed, and do its own quoting accordingly. Note that this function isn't actually needed for the PR, but it would be nice to have for users who want to use `msar` functions to comment on columns on the DB without looking up the relevant oid and attnum.

The call in `msar.add_columns` would be modified to use the processed table name. You could just reuse the result from the `__msar.add_columns(__msar.get_relation_name(tab_id)` in that function.

dmos62:
> avoid mixing processed and unprocessed arguments to functions. As an example, you have a function that expects unsafe, unprocessed schema and table names, but also a safe, quoted, processed comment. This is quite dangerous, and mixing these will lead to bugs and security issues if the caller isn't extremely careful.

Hm, I'm not sure we're seeing my code the same way.

I used below comment at the top of the file as a guide for which functions should sanitize input. As I understand, private functions should not quote and sanitize their inputs: that should only be done by public functions. Note that all of my `__msar.comment_on_column` functions are private.

```
These are the schemas where the new functions will generally live:

      __msar:  These functions aren't designed to be used except by other Mathesar functions.
               Generally need preformatted strings as input, won't do quoting, etc.
        msar:  These functions are designed to be used more easily. They'll format strings, quote
               identifiers, and so on.

```

Further, you say that these private functions sanitized some of the input, but not all of it. Are you referring to one of them calling the public function `msar.get_fully_qualified_object_name` that takes string schema and table names and outputs a concatenated string name? I did not notice that it also expects the inputs to be unquoted, I'll fix that. Is that enough to resolve this conflict?

db/sql/0_msar.sql
```diff
@@ -2192,6 +2233,18 @@ BEGIN
       msar.process_col_alter_jsonb(tab_id, col_alters)
     );
   END IF;
+
+  -- Here, we perform all description-changing alterations.
+  PERFORM __msar.comment_on_column(
+    tab_id := tab_id,
+    col_id := (col_alter->>'attnum')::integer,
```
dmos62, line 2240: I like the `AS x(attnum integer, description text)` syntax, but I couldn't figure out how to use it and to be able to do `jsonb_key_exists(col_alter, 'description');` lower down as well. Am I missing something?

db/columns/operations/select.py
```diff
@@ -8,6 +9,14 @@
 from db.utils import execute_statement, get_pg_catalog_table
 
 
+def get_column_description(oid, attnum, engine):
+    conn_str = str(engine.url)
+    with psycopg.connect(conn_str) as conn:
+        row = conn.execute(f"SELECT col_description({oid}, {attnum})").fetchone()
+        description = row[0]
+    return description
```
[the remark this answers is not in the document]
dmos62: I'll see about doing that.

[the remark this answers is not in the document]
dmos62: > At first glance, this seems like it would be worth making a DB layer wrapper for this so it fits with the rest of our machinery. What do you think?

Could you expand on this a bit more? I understand the "general 'run a function and return the cursor' python wrapper" idea, but I'm not sure about your first suggestion.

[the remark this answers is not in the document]
mathemancer: The first suggestion would be to create something like `msar.col_description(tab_id oid, col_id integer)` on the DB that's just a thin wrapper around the PostgreSQL `col_description` . The value of that is that it'd be callable from Python using the same `execute_msar_func_with_engine` _Python_ wrapper that we use elsewhere. I'm trying to keep all new calls to the database funneled through a pretty controlled spot right now (the `db.connection` module), since we'll need to refactor that eventually and I don't want to hunt through every `conn.execute` in the codebase at that point.

db/sql/0_msar.sql
```diff
@@ -112,6 +113,15 @@ $$ LANGUAGE sql RETURNS NULL ON NULL INPUT;
 ----------------------------------------------------------------------------------------------------
 ----------------------------------------------------------------------------------------------------
 
+CREATE OR REPLACE FUNCTION __msar.jsonb_key_exists(data jsonb, key text) RETURNS boolean AS $$/*
+Wraps the `?` jsonb operator for improved readability.
+*/
```
mathemancer, line 118: Great idea.

db/sql/0_msar.sql
```diff
@@ -126,6 +136,20 @@ END;
...
+Return the fully-qualified name for a given database object (e.g., table).
+
+Args:
+  sch_name: The schema of the object, quoted.
+  obj_name: The name of the object, unqualified and quoted.
+*/
+BEGIN
+  RETURN format('%s.%s', sch_name, obj_name);
+END;
+$$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
```
mathemancer, line 150: Is this function still necessary, given that it's only called in the function immediately below? I think it would be okay to combine the two.

db/sql/0_msar.sql
```diff
@@ -956,7 +981,8 @@ SELECT array_agg(
     -- set the default value for the duplicate column if specified
     CASE WHEN copy_defaults THEN pg_get_expr(adbin, tab_id) END,
     -- We don't set a duplicate column as a primary key, since that would cause an error.
-    false
+    false,
+    col_description(tab_id, pg_columns.attnum)
```
mathemancer, line 985: We should consider eventually making this configurable (it might be nice to have a comment like "Copy of Column X" by default here. However, let's leave it like you've got it for this PR.

db/sql/0_msar.sql
```diff
@@ -1301,7 +1328,9 @@ WITH attnum_cte AS (
         format('%L', col_def_obj ->> 'default')
     END,
     -- We don't allow setting the primary key column manually
-    false
+    false,
+    -- Set the description for the column
+    COALESCE(quote_literal(col_def_obj ->> 'description'), NULL)
```
mathemancer, line 1333: This is redundant, isn't it? `COALESCE` will only fall back to the next argument if the first turns out to be `NULL`. I think you can remove the `COALESCE` call entirely.

db/sql/0_msar.sql
```diff
@@ -1310,8 +1339,9 @@ SELECT array_cat(
   CASE
     WHEN create_id THEN
       -- The below tuple defines a default 'id' column for Mathesar.  It has name id, type integer,
-      -- it's not null, and it uses the 'identity' functionality to generate default values.
-      ARRAY[('id', 'integer', true, null, true)]::__msar.col_def[]
+      -- it's not null, it uses the 'identity' functionality to generate default values, doesn't
+      -- have a comment.
+      ARRAY[('id', 'integer', true, null, true, NULL)]::__msar.col_def[]
```
mathemancer, line 1344: After rethinking this, I think we _should_ put a default comment. 

@ghislaineguerin Do you have some feedback on this? The question is whether the default ID column we create should have a default comment. I suggest:

> Mathesar default ID column

db/sql/0_msar.sql
```diff
@@ -1353,6 +1383,16 @@ DECLARE
...
   PERFORM __msar.add_columns(__msar.get_relation_name(tab_id), variadic col_create_defs);
+
+  PERFORM
+  __msar.comment_on_column(
+    tab_id,
+    col_create_def.name_,
+    col_create_def.description
+  )
+  FROM unnest(col_create_defs) AS col_create_def
+  WHERE col_create_def.description IS NOT NULL;
```
mathemancer, line 1394:
```suggestion
BEGIN
  col_create_defs := msar.process_col_def_jsonb(tab_id, col_defs, raw_default);
  PERFORM __msar.add_columns(fq_table_name, variadic col_create_defs);

  PERFORM
  __msar.comment_on_column(
    fq_table_name,
    col_create_def.name_,
    col_create_def.description
  )
  FROM unnest(col_create_defs) AS col_create_def
  WHERE col_create_def.description IS NOT NULL;
```
Then you'd also have 
```
  fq_table_name text := __msar.get_relation_name(tab_id);
```
in the `DECLARE` block.

This lets you remove one of the private functions, and also makes this function call more similar to the one just above it.

db/sql/0_msar.sql
```diff
@@ -2227,6 +2269,22 @@ BEGIN
...
+      quote_literal(col_alter->>'description') AS comment_
+    FROM jsonb_array_elements(col_alters) AS col_alter
+    WHERE __msar.jsonb_key_exists(col_alter, 'description')
+  LOOP
+    PERFORM __msar.comment_on_column(
+      tab_id := tab_id,
+      col_id := description_alter.col_id,
+      comment_ := description_alter.comment_
+    );
+  END LOOP;
```
mathemancer, line 2286:
```suggestion
  -- Here, we perform all description-changing alterations.
  FOR description_alter IN 
    SELECT
      (col_alter->>'attnum')::integer AS col_id,
      col_alter->>'description' AS comment_
    FROM jsonb_array_elements(col_alters) AS col_alter
    WHERE __msar.jsonb_key_exists(col_alter, 'description')
  LOOP
    PERFORM msar.comment_on_column(
      tab_id := tab_id,
      col_id := description_alter.col_id,
      comment_ := description_alter.comment_
    );
  END LOOP;
```

This change lets you remove a function, simplifying the whole thing a bit.

db/sql/0_msar.sql
```diff
@@ -2227,6 +2269,22 @@ BEGIN
...
     );
   END IF;
+
+  -- Here, we perform all description-changing alterations.
+  FOR description_alter IN 
+    SELECT
+      (col_alter->>'attnum')::integer AS col_id,
+      quote_literal(col_alter->>'description') AS comment_
+    FROM jsonb_array_elements(col_alters) AS col_alter
+    WHERE __msar.jsonb_key_exists(col_alter, 'description')
```
mathemancer, line 2279: Nice.

db/sql/0_msar.sql
```diff
@@ -2237,6 +2295,122 @@ END;
...
+  tab_id: The OID of the table containg the column whose comment we will change.
+  col_name: The name of the column whose comment we'll change
+  comment_: The new comment. Any quotes or special characters must be escaped.
+*/
+SELECT __msar.comment_on_column(
+  __msar.get_relation_name(tab_id),
+  col_name,
+  comment_
+);
+$$ LANGUAGE SQL;
```
mathemancer, line 2369: This function could be removed, in combination with my other comment.

db/sql/0_msar.sql
```diff
@@ -2237,6 +2295,122 @@ END;
...
+  tab_id: The OID of the table containg the column whose comment we will change.
+  col_id: The ATTNUM of the column whose comment we will change.
+  comment_: The new comment.
+*/
+SELECT __msar.comment_on_column(
+  __msar.get_relation_name(tab_id),
+  msar.get_column_name(tab_id, col_id),
+  comment_
+);
+$$ LANGUAGE SQL;
```
mathemancer, line 2390: This function could be removed, in combination with the other comment.

mathemancer (changes requested):
For testing, I think we should add the following:
- At least one test to make sure comments are added when adding columns
- At least one test to make sure the comment isn't deleted when altering columns without passing the `"description"` key
- At least one test where the comment has a `;` in it for both adding and altering columns.

I also found a couple of functions that could be removed, with only minor places to the spots they're called.

Also, I think we should consider adding a default comment to our ID column.

db/sql/0_msar.sql
ghislaineguerin: We're adding these column comments to enhance the user/dev experience, correct? As of now, we don't show column comments in the UI, so I can't weigh in on that. But considering the context in #3069, it seems like this is a good solution. The text looks good to me.

db/sql/0_msar.sql
dmos62: There might be conflict in parameter quoting. Let me see.

db/sql/0_msar.sql
dmos62: It's not necessary, but doesn't it make sense to have this function be available in the private and public contexts? Considering that the parameter quoting requirements are different?

db/sql/0_msar.sql
dmos62: I left this one around. I made the analogue public function be based on this one instead.

db/sql/0_msar.sql
dmos62: Done in https://github.com/centerofci/mathesar/pull/3186/commits/90ed05cd0fe71ff6dd68acd254742158fca8edf6

db/sql/0_msar.sql
dmos62: Done in https://github.com/centerofci/mathesar/pull/3186/commits/90ed05cd0fe71ff6dd68acd254742158fca8edf6

db/sql/0_msar.sql
dmos62: Relevant commits https://github.com/centerofci/mathesar/pull/3186/commits/4bdc9b8c43a42e83dc5868686a29ff60cbfbd5c1 and https://github.com/centerofci/mathesar/pull/3186/commits/625b868f51376b3908f5a0df7438ccc4d2119008

dmos62: Relevant commits https://github.com/centerofci/mathesar/pull/3186/commits/4bdc9b8c43a42e83dc5868686a29ff60cbfbd5c1 and https://github.com/centerofci/mathesar/pull/3186/commits/625b868f51376b3908f5a0df7438ccc4d2119008

db/sql/0_msar.sql
dmos62: Relevant commits https://github.com/centerofci/mathesar/pull/3186/commits/4bdc9b8c43a42e83dc5868686a29ff60cbfbd5c1 and https://github.com/centerofci/mathesar/pull/3186/commits/625b868f51376b3908f5a0df7438ccc4d2119008

db/sql/test_0_msar.sql
```diff
@@ -1600,10 +1608,47 @@ BEGIN
   RETURN NEXT col_type_is('col_alters', 'col_opts', 'numeric(5,3)');
   RETURN NEXT col_not_null('col_alters', 'col_opts');
   RETURN NEXT col_not_null('col_alters', 'timecol');
+  RETURN NEXT is(col_description('col_alters'::regclass::oid, 2), 'This is a comment!');
+  RETURN NEXT is(col_description('col_alters'::regclass::oid, 3), NULL);
 END;
 $f$ LANGUAGE plpgsql;
 
 
+CREATE OR REPLACE FUNCTION test_comment_on_column() RETURNS SETOF TEXT AS $$
```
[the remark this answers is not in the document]
dmos62: Done in https://github.com/centerofci/mathesar/pull/3186/commits/39701c7dc82340d69405fe887a1952a9592c7b5a (and later commits)

mathemancer (approved):
Overall, this is looking really great. Thanks for taking the time to go back-and-forth on this, @dmos62 .

PR #3189, opened 2023-08-29
Add script for creating debian package

### LINKED ISSUE #2765: Create debian or flatpak images for Mathesar
## Problem
Installing on Debian OS without using Docker is quite complicated and there are too many steps involved which makes the installation process really brittle. This can be avoided by providing a `.deb` package.
 
Proposed solution
- We should create a `.deb` file which should be aimed at 
     -  "debian:bullseye"
     -  "debian:bookworm"
     -  "debian:sid"
     - "ubuntu:focal"
     - "ubuntu:jammy"
     - "ubuntu:kinetic"
     -  "ubuntu:lunar"
    
- We should target Python >= 3.7
- Write a script to generate `.deb` package using [dh-virtualenv](https://github.com/spotify/dh-virtualenv)
- Upload the `.deb` file to mathesar PPA

References:
- [Synapse script for generating deb file](https://github.com/matrix-org/synapse/blob/develop/scripts-dev/build_debian_packages.py)

- From a [reddit comment](https://www.reddit.com/r/programming/comments/11mw9u1/comment/jbm7hg8/?utm_source=share&utm_medium=web2x&context=3))
> Looks fantastic, will def give this a go when a .deb or flatpak happens.

silentninja:
Fixes #2765

This PR adds the [release-scripts/build-debian.sh](https://github.com/centerofci/mathesar/pull/3189/files#diff-37969797ae50c3fab7884fa1afdfd2b05ce23c130f339dde44b4b639ba64c0ca) for generating a Debian package of Mathesar. 
The generated Debian package is specific to the OS version it was generated on. So if it was generated on Debian bullseye, it can be installed only on machines running Debian bullseye and for example, cannot be installed on Debian buster. 
We would be running this script on a build service like [Open Build Service](https://openbuildservice.org/), it is provided here only for generating and testing manually.

The generated Debian package contains a virtualenv directory of all the dependencies of Mathesar and Mathesar Python modules. When the debian package is installed, the system python is symlinked to this virtual env. The systemd file then uses this virtual environment to start Mathesar.

The following changes are needed for the Debian package to function correctly

- Removal of Python typings - This is needed for supporting various Python versions. We could use backports but I decided to remove it as we are not using typings in a lot of places.
- Some non-python resources are referenced as Python modules instead of being referenced using the filesystem location - This is needed because Debian packages the files in the mathesar repository as Python modules.

## Notes for the reviewer
- The aim of this PR is to generate the debian package manually by running the  [release-scripts/build-debian.sh](https://github.com/centerofci/mathesar/pull/3189/files#diff-37969797ae50c3fab7884fa1afdfd2b05ce23c130f339dde44b4b639ba64c0ca) script and test the generated package on a Debian machine to see if the .deb file works as intended. 
In practice, the script will run a build service like [Open Build Service](https://openbuildservice.org/) which will download the release source code along with static assets, and use it for generating the Debian file.

- Instructions for building the Debian image are in the [Developer Guide](https://github.com/centerofci/mathesar/blob/ef22367d64dc1b07a4c6a9b10f815a00acc4644e/DEVELOPER_GUIDE.md)

- Before installing Mathesar, you need to install a Postgres server as we don't support SQLite as an internal database as of now

```
sudo apt install postgresql
sudo -u postgres psql
CREATE USER mathesar WITH SUPERUSER ENCRYPTED PASSWORD 'mathesar';
CREATE DATABASE mathesar_django;
```

- Installing the Debian package can be done by 

```
sudo apt install mathesar_0.1.3_amd64.deb
```

- Check the status of mathesar by

```
systemctl status mathesar.service
```

- If the mathesar service is installed on a remote server. You need to do the following

  - Add the IP address of the server to allowed hosts
    
     ```
     sudo nano /etc/mathesar/.env
     ```
     and add the IP address to the `ALLOWED_HOSTS` and restart the service
     

   - Since the mathesar server runs on port 8000 which is not exposed to the public. You would need to set up a reverse proxy to listen on port 80 or redirect the port 80 traffic to port 8000.

   ```
  sudo socat tcp-listen:80,reuseaddr,fork tcp:localhost:8000
  ```
  
- Updating the debian package can be tested by making changes to the version number in `pyproject.yaml` and running the build script again. The generated debian package can be installed to update the existing Mathesar running on the OS

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (approved):
I was able to get Mathesar running on a Debian (bullseye) machine as a `.deb` package! Nice work, @silentninja .

For future reference, the correct command to install the package is

```
sudo apt install ./mathesar_0.1.3_amd64.deb
```
I'm not sure why the `./` prefix is necessary, but it didn't work (on my test machine) without that. I assume it flags the installation package as existing on the local filesystem.

Moving forward, I think we should consider reorganizing the docs a bit, however I was able to use the ones added in this PR to build the package, so they suffice.

I think we should hold off on merging until #3206 is merged (via `develop`) into this branch, to make sure the python changes don't affect PG version compatibility.

Issue #3027, opened 2023-07-11
Support for importing Excel files


dmos62:
Part of [GSoC 2023 project for supporting the import of JSON and Excel into Mathesar](https://wiki.mathesar.org/community/gsoc/project-ideas/import-excel-json-files).

Expected outcomes once this issue is completed: Users should be able to upload Excel files in the current import UI seamlessly. If the file is structured so that it cannot be imported, they should see a clear error message explaining why.

```[tasklist]
- [ ] #3059
- [ ] #3083
- [x] Merge Excel normalization code #3132 
- [x] Merge remaining tests #3132 
- [x] Add `sheet_number` param in datafile API to accept the sheet number to use in the Excel file #3195 
- [ ] Create Documentation for types of allowed Excel #3204 
- [ ] Improve UI for better user understanding of the new feature #3211 
```

seancolsen:
Closing as the remaining work is tacked in #3031

Issue #3030, opened 2023-07-12
Updating the way we handle datafiles in the backend


IamEzio:
## Problem

Currently we store our data files as objects of the class `DataFile` that looks something like this:
```
class DataFile(BaseModel):
    created_from_choices = models.TextChoices("created_from", "FILE PASTE URL")
    file_type_choices = models.TextChoices("type", "CSV TSV JSON")

    file = models.FileField(upload_to=model_utils.user_directory_path)
    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True, on_delete=models.CASCADE)
    created_from = models.CharField(max_length=128, choices=created_from_choices.choices)
    type = models.CharField(max_length=128, choices=file_type_choices.choices)
    table_imported_to = models.ForeignKey(Table, related_name="data_files", blank=True,
                                          null=True, on_delete=models.SET_NULL)

    base_name = models.CharField(max_length=100)
    header = models.BooleanField(default=True)
    delimiter = models.CharField(max_length=1, default=',', blank=True)
    escapechar = models.CharField(max_length=1, blank=True)
    quotechar = models.CharField(max_length=1, default='"', blank=True)
```

This was working perfectly till now. But now that we are expanding our import feature to JSON and Excel files, this class definition needs to be updated.

Problems:
1. The attributes like `header`, `delimiter`, `escapechar` and `quotechar` are not required while dealing with JSON Files.
2. The attributes like `delimiter`, `escapechar` and `quotechar` are not required while dealing with Excel Files.
3. The attribute `max_level` (will be added as an API param for importing JSON) is not required while dealing with CSV and Excel Files.
4. The attribute `sheet_number` (will be added as API param for importing Excel -- to provide info about sheet number in excel file) is not required while dealing with CSV and JSON files.

## Proposed solution

We should add more models for handling each of these datafiles, each extending the original datafile model that has core attributes used by all datafiles.

```
class DataFile(BaseModel):
    created_from_choices = models.TextChoices("created_from", "FILE PASTE URL")

    file = models.FileField(upload_to=model_utils.user_directory_path)
    user = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True, on_delete=models.CASCADE)
    created_from = models.CharField(max_length=128, choices=created_from_choices.choices)
    table_imported_to = models.ForeignKey(Table, related_name="data_files", blank=True,
                                          null=True, on_delete=models.SET_NULL)
[...]

PR #3195, opened 2023-09-01
Add `sheet_index` param in datafile API to handle excel files with multiple sheets

IamEzio:
Fixes part of #3027 

This PR does the following:
- Adds code to accept `sheet_index` parameter which is the sheet number in an excel file.
- Updates excel importing code to handle cases when `header` is made false (i.e, use default column names).

**Screenshots**

https://github.com/centerofci/mathesar/assets/89375125/615a0e29-7a9d-4229-901e-bf531a5bca0d

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


IamEzio:
@dmos62 PTAL. Thanks!

mathesar/imports/excel.py
```diff
@@ -50,9 +50,16 @@ def remove_empty_rows_and_columns_from_dataframe(df):
...
+    header_row = 0 if data_file.header else None
+    try:
+        dataframe = remove_empty_rows_and_columns_from_dataframe(
+            pandas.read_excel(data_file.file.path, data_file.sheet_name, header=header_row)
+        )
+    except ValueError:
+        # If sheet number has been passed as a parameter instead of sheet name.
+        dataframe = remove_empty_rows_and_columns_from_dataframe(
+            pandas.read_excel(data_file.file.path, int(data_file.sheet_name), header=header_row)
+        )
```
Anish9901, line 62: I am a bit skeptical about this, I think we should _only_ use the index of the file so that our django model could stay consistent, this could be achieved by searching the name of the excel sheet in the list of all the sheet_names and then finding its index. However, finding the list of sheet_names could cause a lot of overhead using the `ExcelFile.sheet_names` or `df.keys()` as they require to load the whole excel file in memory instead of just the specified sheet as done here. I did however find a [method](https://stackoverflow.com/a/56320206) to efficiently extract sheet names but idk if its worth it. I'd like to know your thoughts on this @dmos62.

dmos62: I notice there's some confusion between index and name here. Sometimes the `sheet_name` is cast to an `int`, which I presume implies that it acts as an index. `sheet_index` and `sheet_name` should be distinct and if we'd want to use some kind of polymorphic variable that's sometimes a name and sometimes an index, we'd probably name it `sheet_id` (since both a name and an index are identifiers), but I don't like the idea of a variable sometimes being one thing and sometimes something else.

For what it's worth, I think an index identifier has a better simplicity/usefulness relation than a name identifier. We could support both, but keeping it minimal is desirable right now, I think.

The SO answer @Anish9901 linked to mentioned that loading excel files gets exponentially slower with size (in Pandas). Supposing that's true, we'd do well to fix that at some point. Maybe now. If not, we should definitely put a `TODO` comment with a link to that SO thread explaining what the problem is and what the solution could be (where the relevant code is found).

mathesar/utils/datafiles.py
```diff
@@ -114,13 +114,18 @@ def create_datafile(data):
...
+        if sheet_name.isdigit():
+            sheet_name = str(int(sheet_name) - 1)
         datafile = DataFile(
             file=raw_file,
             base_name=base_name,
             type=type,
             created_from=created_from,
             header=header,
-            max_level=max_level
+            max_level=max_level,
```
dmos62, line 127: Argh, this is a parameter for JSON parsing, right? I do hope we get to fix all of this logic being lumped together during your project @IamEzio .

dmos62 (changes requested):

mathesar/utils/datafiles.py
IamEzio: Yes @dmos62, this is one of our postponed task: #3030. We'll create different models for different types of datafile each extending the current DataFile class with only the essential attributes.

mathesar/tests/api/test_table_api.py
```diff
@@ -242,6 +242,7 @@ def _create_table(client, data_files, table_name, schema, import_target_table, d
 
     response = client.post('/api/db/v0/tables/', body)
     response_table = response.json()
+    print('\n\n', response_table, '\n\n')
```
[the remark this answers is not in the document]
IamEzio: Thanks @Anish9901, removed it now

IamEzio:
@dmos62 @Anish9901 I have updated the PR. PTAL. Thanks!

mathesar/utils/datafiles.py
```diff
@@ -114,18 +114,17 @@ def create_datafile(data):
         )
     else:
         max_level = data.get('max_level', 0)
-        sheet_name = data.get('sheet_name', '1')
+        sheet_index = data.get('sheet_index', 1)
```
dmos62, line 117: Why have default be 1 if the index is 0-indexed? This also conflicts with the analogous field in the `DataFile` django model. Is this an oversight?

dmos62 (changes requested):

mathesar/utils/datafiles.py
IamEzio: I was reducing the `sheet_index` value received from the API just a line below, so I thought that if the POST request doesn't have this value, we can assign it default value 1 and then reduce it.

On second thought, I agree that the above method is confusing. I have two solutions:
1. Keep the value 0-indexed everywhere and inform users to give sheet index based on 0-indexing.
2. Keep the value 1-indexed everywhere -- when asking users and when assigning default value (which would be 1). We'll just reduce it when we call `pandas.read_excel()`.

@dmos62 what are your thoughts?

dmos62: I think 0-indexing is standard and we should go with that.

IamEzio: @dmos62 I have updated it now. PTAL. Thanks!

dmos62 (approved):

PR #3203, opened 2023-09-11
Migrate django_db to sqlite

### LINKED ISSUE #3055: Add support for SQLite as a additional datasource for the internal database
## Problem.
Our current installation process requires the user to set up a Postgres server, and a database in order to start Mathesar which adds more complexity to our installation process. While we provide a default Postgres server along with our docker-compose, other installation packages don't come with a default Postgres server.

## Proposed solution
We should provide using `SQLite` as a datasource option for our internal database. This makes it easier for someone with an existing external Postgres database to try out Mathesar

## Additional context
Postgres-specific field is used in only one place
https://github.com/centerofci/mathesar/blob/1d49c44d26ebcf7777b7ba29572f678930ba5942/mathesar/models/base.py#L890

Migration should be written to change the `ArrayField` to a `JsonField` with `list` as the default initializer. The migration should also take into consideration to move the existing data from the `ArrayField` into `JsonField`

Anish9901:
Fixes #3055

**Technical details**

This PR removes the necessity to use postgres for internal django models db, whenever env `DJANGO_DATABASE_URL` is not set the app will now create a fresh sqlite db, however if `DJANGO_DATABASE_URL` is set the app will respect the value of the env variable and create a db for the given db connection.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
Note to reviewer: This PR requires to reset our django migration so it'd be better if #3170 gets merged before this PR as the previous also has migrations.

The reason for this reset is because our tests will use postgres and it doesn't allow altering an `integer[]` field to `jsonb`.

config/settings/common_settings.py
```diff
@@ -94,11 +94,19 @@ def pipe_delim(pipe_string):
     db_key: db_url(url_string)
     for db_key, url_string in decouple_config('MATHESAR_DATABASES', cast=Csv(pipe_delim))
 }
-DATABASES[decouple_config('DJANGO_DATABASE_KEY', default="default")] = decouple_config('DJANGO_DATABASE_URL', cast=db_url)
+
+try:
+    DATABASES[decouple_config('DJANGO_DATABASE_KEY', default="default")] = decouple_config('DJANGO_DATABASE_URL', cast=db_url)
```
silentninja, line 99: Wouldn't it be better to set the default SQLite dictionary as a default value? The default dictionary can be stored in a variable and used as a default value 
```suggestion
DEFAULT_DATABASE = {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
    }
DATABASES[decouple_config('DJANGO_DATABASE_KEY', default="default")] = decouple_config('DJANGO_DATABASE_URL', cast=db_url, default=DEFAULT_DATABASE
```

mathesar/api/serializers/table_settings.py
```diff
@@ -39,3 +39,9 @@ def update(self, instance, validated_data):
             instance.column_order = column_order_data
             instance.save()
         return instance
+
+    def to_internal_value(self, data):
+        column_order = data.get('column_order', None)
+        if isinstance(column_order, list):
+            data['column_order'] = list(map(int, column_order))
```
silentninja, line 46: Won't setting this validation as a [ListField](https://www.django-rest-framework.org/api-guide/fields/#listfield) work?

silentninja (changes requested):
The PR is missing migration for moving existing data from an `ArrayField` to JSONField. Have you tested with an existing database created using Mathesar 0.1.3?

I also have a few more concerns which I added as specific comments

Anish9901:
> The PR is missing migration for moving existing data from an ArrayField to JSONField.

Postgres disallows casting a col from `integer[]` to `jsonb` and thus breaks our tests if a migration is added to alter it. I'll add the migration along with the already altered migration(`0001_initial.py`).

mathesar/api/serializers/table_settings.py
Anish9901: 😲 Didn't knew setting a child would also typecast, I thought it was just type checking. Anyway, fixed now.

config/settings/common_settings.py
Anish9901: Wow, really clean, but this doesn't work because of `cast`. I've added the connection string as the `default` now and it works flawlessly.

silentninja (approved):

PR #3222, opened 2023-09-22
Paste API

### LINKED ISSUE #2844: API to create or update records in bulk
## Limitations of the current API

- Creating and modifying records requires one request per record

## Use cases

- [Pasting data into cells](https://github.com/centerofci/mathesar/issues/2812) requires modifying existing records and also (according to the [spec](https://wiki.mathesar.org/design/specs/clipboard-interactions/)) creating new records too (when the pasted data extends beyond the viewable records).
- We may eventually want to implement other features like setting multiple cell values to null, or doing something akin to the "fill" feature that spreadsheets offer. This would require modifying records in bulk.

## Similar work

- #2641 added a new endpoint to delete records in bulk
- #1442 added a a new endpoint to create records in bulk

    In Matrix @silentninja [said](https://matrix.to/#/!UZILDSNKobkelUYwBp:matrix.mathesar.org/$xmthisi3X5a1oUgDE1FZJkQeFClC8vee4Vu9scsctww?via=matrix.mathesar.org&via=matrix.org):

    > I don't think you can use the [existing_import API](https://github.com/centerofci/mathesar/pull/1442) for pasting the data into cells because the `existing_import` API creates a temporary Table from a datafile and then copies the data from the temporary Table to the Table the data has to be imported into, meanwhile the case of pasting data into cells is a much simpler task. So creating a new temporary Table and copying data from that temporary table might be an overkill for the copy-paste feature

## A "Paste" example

1. User copies these eight cells:

    

1. User pastes the copied data into the sheet at the following starting position:

    

1. The front end uses the API to:

    - **Update** records 249 and 250, modifying `First Name` and `Last Name`.
    - **Create** two new records by supplying values for `First Name` and `Last Name`.
    
1. The API responds with the full records for all the records modified and created, allowing the front end to display the new data.

## Thoughts on API design 

- I want to leave this ticket somewhat open-ended to start, in case the backend team would like to propose an API design.

- It would be nice to perform these changes within a single transaction. From a UX perspective, if something goes wrong, no changes should be made. That would mean creating one API to handle both updating records and creating records.

- How does the API behave when we tell it to update a record which no longer exists?

    Pretend we're following the "Paste" example above but in between steps `1.` and `2.` another user deletes record 249.

    - One possibility would be to _create_ a new record with id 249.
    - Another possibility would be to _fail_. I think I prefer this behavior.

- For the copy-paste use case at least, we can rely on our our data being rectangular. That is, we don't need to update one field in record 249 and a different field in record 250 &mdash; we need to update all the same fields in all records. I think it's fine to bake this assumption into the API, especially since I predict it may be necessary for a performant implementation. For example, we'd would want to avoid a naive implementation which makes one `UPDATE` query per row.

    Assuming we do bake this "rectangular" assumption into the API behavior, it may be worth considering the ways in which the API design convey that behavior. For example, we might be inclined to design the request body like this:

    ```json
    [
      { "1": 249, "2": null,     "3": "Lloyd"     },
      { "1": 250, "2": "Shelby", "3": "Shepherd"  },
      {           "2": "Robert", "3": "Mccormick" },
      {           "2": "Dawn",   "3": null        }
    ]
    ```

    But with that design, an API consumer could submit a rather strange request to update two records like this:

    ```json
    [
      { "1": 249,               "3": "Lloyd" },
      { "1": 250, "2": "Shelby"              }
    ]
    ```

    Because of the potential for requests like the the above, we'd need to clearly communicate the behavior of the API under these circumstances. For record 249, column 2, would the API leave the value untouched? Or would it set the value to `NULL`? Or maybe the request would fail?

    Just as food for brainstorming, a request schema like this might go further towards communicating this "rectangular" requirement to API consumers. I don't know.

    ```json
    {
      "columns": [1, 2, 3],
      "data": [
        [ 249, null,     "Lloyd"      ],
        [ 250, "Shelby", "Shepherd"   ],
        [ null, "Robert", "Mccormick" ],
        [ null, "Dawn",   null        ]
      ]
    }
    ```

- Unlike our proper `records` endpoint, the front end will not have any good way to display cell-level errors for this paste feature, or for the other features I'm imagining we'd build with this API. For that reason, the requirements around the error response are more loose. When the request fails due to some data problem (e.g. constraint violation, etc) the front end will display a toast message. The API should provide enough info in the error message to help the user fix their problem, but it certainly doesn't need to list _all_ the error, or even be super specific about which cell caused the error.

dmos62:
Fixes #2844

Implements paste API.

Initially we envisioned a generic bulk upsert API that could be used by the frontend paste functionality, hence the title of #2844. I explored that idea extensively, but opted for a non-generic API specifically for pasting.

In my estimate, the bulk upsert as expected by a user using paste in a spreadsheet is wildly different than the bulk upsert expected by database administrator doing `INSERT ON CONFLICT DO UPDATE`. Trying to accomodate both seemed like a good way to waste a lot of time and end up with bad UX for both.

Instead of summarizing my concerns, I'll summarize the features of the proposed paste API:

```python
updates = [
	dict(
		changes=dict(
			Status='0',
		),
		conditionals=dict(
			id=record_0_id,
		),
	),
	dict(
		changes=dict(
			Status='1',
			Center='2'
		),
		conditionals=dict(
			id=record_1_id,
		),
	),
]
inserts = [
	dict(
		Status='3',
		Center='4',
	),
	dict(
		Status='5',
		Center='6',
	),
]
data = dict(
	updates=updates,
	inserts=inserts,
)
client.post(f'/api/ui/v0/tables/{table.id}/records/paste/', data=data)
```

Above is the basic outline of the API. The endpoint accepts (via `POST`) a dict, that has two fields: updates and inserts.

It's assumed that frontend will always know whether a given row being pasted should result in an update or an insert. Making that explicit prevents surprises in case of relevant changes having been made that were not synced with the calling frontend.

Updates are a sequence of dicts: each dict has a `changes` key and a `conditionals` key: `changes` define the field-value pairs to change in the row that is satisfied by the equality comparisons defined by field-value pairs in `conditionals`. Think how `UPDATE` works in SQL. An update with `changes` being `{'field1':'val1'}` and `conditionals` being `{'id':'val2'}` requests that for the single row whose `id` field equals `val2`, its `field1` field's value is set to `val1`.

If more than one row would be updated by a given update, that's a fatal exception, because when pasting a row it should never result in more than one updated row. The paste API attempts to minimize surprises to the end user.

Inserts are a sequence of dicts: each dict defines the field-value pairs that should be set on the new resulting row.

An insert will always result in a new row or will panic if that's impossible for whatever reason.

Any panic encountered will rollback all changes made. Paste is meant to be an all-or-nothing operation.

Also, this API panics if any user-defined changes to `SERIAL` or `IDENTITY` columns are requested. These column types need special handling when mutating their values, because there are sequences involved, and that seems hard to safely generalize (i.e. prevent surprises).

All updates are performed before any inserts. All updates are performed in order requested; same for inserts.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/connection.py
+++ b/db/connection.py
@@ -1,8 +1,9 @@
 from sqlalchemy import text
 import psycopg
+import json
 
 
-def execute_msar_func_with_engine(engine, func_name, *args):
+def execute_msar_func_with_engine(engine, func_name, *parameters):
     """
     Execute an msar function using an SQLAlchemy engine.
 
@@ -14,15 +15,15 @@ def execute_msar_func_with_engine(engine, func_name, *args):
         *args: The list of parameters to pass
     """
     conn_str = str(engine.url)
+    statement, parameters = _get_parametrized_statement_and_parameters(
+        func_name, parameters
+    )
     with psycopg.connect(conn_str) as conn:
         # Returns a cursor
-        return conn.execute(
-            f"SELECT msar.{func_name}({','.join(['%s']*len(args))})",
-            args
-        )
+        return conn.execute(statement, parameters)
 
 
-def execute_msar_func_with_psycopg2_conn(conn, func_name, *args):
+def execute_msar_func_with_psycopg2_conn(conn, func_name, *parameters):
     """
     Execute an msar function using an SQLAlchemy engine.
 
@@ -33,11 +34,34 @@ def execute_msar_func_with_psycopg2_conn(conn, func_name, *args):
         func_name: The unqualified msar function name (danger; not sanitized)
         *args: The list of parameters to pass
     """
-    args_str = ", ".join([str(arg) for arg in args])
-    args_str = f"{args_str}"
-    stmt = text(f"SELECT msar.{func_name}({args_str})")
+    statement, parameters = _get_parametrized_statement_and_parameters(
+        func_name, parameters
+    )
     # Returns a cursor
-    return conn.execute(stmt)
+    return conn.execute(statement, parameters)
+
+
+#TODO document
+def _get_parametrized_statement_and_parameters(func_name, parameters):
+    parameter_placeholders = ','.join(
+        ['%s'] * len(parameters)
+    )
+    statement = f"SELECT msar.{func_name}({parameter_placeholders})"
+    adapted_parameters = _adapt_parameters(parameters)
+    return statement, adapted_parameters
+
+
+def _adapt_parameters(args):
+    """
+    Processes parameters, serializing dicts and lists into JSON strings.
+    """
+    return [
+        json.dumps(arg)
+        if type(arg) is dict or type(arg) is list
+        else arg
+        for arg
+        in args
+    ]
 
 
 def load_file_with_engine(engine, file_handle):
--- a/db/records/operations/paste.py
+++ b/db/records/operations/paste.py
@@ -1,0 +1,7 @@
+from db import connection as db_conn
+
+
+def bulk_paste_records(table_oid, updates, inserts, engine):
+    db_conn.execute_msar_func_with_engine(
+        engine, 'bulk_paste_records', table_oid, updates, inserts
+    )
--- a/db/sql/0_msar.sql
+++ b/db/sql/0_msar.sql
@@ -2574,3 +2574,126 @@ BEGIN
   RETURN jsonb_build_array(extracted_table_id, fkey_attnum);
 END;
 $f$ LANGUAGE plpgsql;
+
+
+----------------------------------------------------------------------------------------------------
+----------------------------------------------------------------------------------------------------
+-- RECORD UPDATING FUNCTIONS
+----------------------------------------------------------------------------------------------------
+----------------------------------------------------------------------------------------------------
+
+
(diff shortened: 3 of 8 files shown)


seancolsen:
@dmos62 I assume this PR is to fix #2844, is that right? If so, can you update the PR description?

seancolsen:
@dmos62 can you give an update one where you're at with this PR? How close to completion is it? What work is remaining?

dmos62:
I'll update the top post with what the goal/spec of the PR is.

As to the progress, here's the summary, but keep in mind that I might not recall some details. The PR seems 80% complete. The SQL logic and tests seem to be there. The API tests are there for the update aspect of pasting. The API tests for the insert aspect of pasting are missing.

mathemancer:
Closed for now, but we may grab some pieces of this later.

Issue #3230, opened 2023-10-03
Schemas are not returned in the APIs right after editing the database connection details


rajatvijay:
## To Reproduce

1. Add a new database connection from the mathesar UI.
3. Rename the database from the postgres server. 
4. Reload the database page of this particular database on mathesar.
5. Expect connection error on the page.
6. Click on the "Edit Connection" button. 
7. Change the name & click on the "Update the connection" button. 
8. Get redirected to the database page of this particular database. 
9. Expect to see the schemas from this database. Instead see no schemas being listed. 

## Additional information

1. Frontend calls the `POST: /api/ui/v0/reflect/` API right the editing of the database connection succeeds. 
2. To reproduce this please use the branch: `db-connection-pages`

dmos62:
Still can't reproduce. Could be some kind of race condition, but the `/api/ui/v0/reflect/` seems to be blocking.

I'll start out by having backend automatically reflect when you edit a connection string. And, to only reflect the affected db (the `/reflect` endpoint reflects everything).

Anish9901:
> I'll start out by having backend automatically reflect when you edit a connection string. And, to only reflect the affected db (the /reflect endpoint reflects everything).

@dmos62 I already tried to do that, but it results in a bunch of failed test cases.

dmos62:
@Anish9901 yeah, I'm running into some tech debt. You'd want to reflect the database whenever you make changes to the connection string (because it might now be a totally different database), but during reflection we actually make changes to the database models too (to mark them as deleted or not deleted), which causes infinite recursion. I think it's fairly important to get reflection right; I'm seeing what I can do.

Anish9901:
One possible solution I see for this problem is to detect which fields are altered in the model and triggering reflection only when `db_name`, `name`, `host` or `port` are changed, but that might not be a quick fix given the urgent priority of this issue.

dmos62:
> One possible solution I see for this problem is to detect which fields are altered in the model and triggering reflection only when `db_name`, `name`, `host` or `port` are changed, but that might not be a quick fix given the urgent priority of this issue.

I had similar ideas. Though I opted to just get rid of the `deleted` field altogether. Seemed like a confused concept, and it wasn't used by frontend. And, there was a whole bunch of other buggy weirdness associated with it. Now, we just do a connectability check whenever listing databases in the API, and we don't persist the result of that. If it turns out that we want to persist it for some reason (right now we don't need to), we can come up with something better than `deleted`.

[...]

PR #3245, opened 2023-10-13
Fix various bugs in database reflection, connection caching

dmos62:
Related to #3230

Noteworthy changes:

- Introduce `db.credentials.DbCredentials`; a way to uniquely identify a connection to a database; improves on how we previously did that by passing around multiple arguments (like hostname, db_name, username etc);
- Use of `DbCredentials` to key the engine cache; before we were incorrectly using the db_name to identify which database a given engine is for: that would break down when you had two databases with the same name (but different host or port);
- Not a change, but discovered that our `reset_reflection` and `make_sure_initial_reflection_happened` constructs still key database by their name (that's a significant bug);
- Have reflection be triggered whenever a database model is created saved or updated: our reflection policy has been that whenever something changes that might affect Postgres state, we reflect: however, this has not been observed when it came to Database model creations/updates: we were able to create a Database, or even change what actual Postgres database it's refering to without triggering a reflection: now that's imposible;
- Cleaned up engine management in our testing suite and reflection code: unfortunately, this was necessary, because above change to database reflection (effectively reflecting earlier), 1) caused bugs in testing fixtures due to now incorrect assumptions about when the first reflection would be triggered, and 2) increased simultaneous connection use beyond the 100 Postgres limit, which made the testing suite inoperable; 2) was due to incorrect connection management, which was made worse by the new reflection changes, because they caused more reflections to happen in fixtures (because they make a lot of changes to dbs);
- Worth noting that above changes to testing suite brought simultaneous connection use down to single digits per testing thread, thus we should be able to run tests via ~9 concurrent threads (again), though haven't thoroughly tested this yet.

## Checklist
 
  ixdex.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dmos62:
Not sure what's happening with the CI. Tests pass fine on my 4 GB dev server, but exit with 137 (mem usage over limit, I presume) over here. Should not block review @mathemancer.

dmos62:
Found that the CI was failing during tests because there was a typo in code that was being called by `db/install.py::install_mathesar`, but no test called it, so tests passed locally, but during setup the CI would break. Added a basic test for this.

dmos62:
Above comment was misled by a false-positive check suite by the CI. Now, the tests truly pass on the CI. PR ready for merging, once reviewed.

mathemancer:
We've completely reorganized the way we're going connections and credentials for databases, and this PR isn't really relevant in that context. I'm going to close it, though we may grab some code from it if needed.

PR #3316, opened 2023-11-20
Added drag functionality to update precendence for sort condition

### LINKED ISSUE #2349: Allow dragging to modify precedence of sorting conditions
## Context

On the Table Page, we have this UI to add/edit sorting conditions

image

## Current behavior

- Modifying the precedence of sorting conditions requires the user to remove the conditions and re-add them.

- The column selection dropdown is only enabled within the last entry.

## Desired behavior

- Users can sort entries via dragging to modify their precedence. _(More details below.)_

- Column selection dropdowns in all entries are enabled. The array of _options_ within each column selection dropdown is reactively derived so as to include only the table columns which are not specified by any sorting conditions. (These changes are necessary to make the logic for each sorting entry independent of its index within all sorting entries.)

### Dragging behavior specifications

- There is a [grip](https://fontawesome.com/icons/grip?s=solid&f=classic) icon to the left of the column selection dropdown. Hovering the grip icon shows a `grab` cursor.

- Users can grab the grip via mouse or touch. Use [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events).

- When grabbed:

    - Within the bounds of the sorting entries container, the sort entry moves vertically with the pointer. It does not move horizontally.

    - Other entries within the list slide as they are re-ordered. When sliding, their movement is animated and their z-index is such that they appear _under_ the entry being moved.

    - The sorting precedence is _not_ modified while dragging.

    - The pointer can move anywhere on the page while dragging. Use [pointer capture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) on the `body` for this.

    - The cursor shows as `dragging`.

- When released:

    - The precedence of sorting entries is modified.

    - A new request is sent to reload the table with the updated sorting. (This should happen automatically when modifying the sorting store)

- Ideally we would implement all of the above without any extra JS dependencies. If you are implementing this and can make a strong case for using a 3rd party library, then we'd be open to considering it before you begin implementation.

@kgodey @ghislaineguerin and @pavish I'd like your opinion on the UX changes I'm proposing here. I will un-draft this issue and apply a "help wanted" label if you all sign off.

@pavish If we implement this from scratch, I'd like to use [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) even though [browser support](https://caniuse.com/pointer) is still only at ~95%. Is that okay with you? It seems okay to me for this feature because it's a convenience feature, not a make-or-break feature. Users can still work around it in the rare case that their browser does not support pointer events.

dikwickley:
Fixes #2349 

**Technical details**

I have implemented the drag functionality from scratch without using any third party libraries.
It basically works by creating a ghost element that is draggable. While it is being dragged, it's position is used to calculate what index it is on and that updates the underlying list simultaneously.
When the pointer is released that list is updated to the original sort store.
The slide animation is achieved by https://svelte.dev/docs/svelte-animate#flip

@seancolsen please review.

**Screenshots**

https://github.com/mathesar-foundation/mathesar/assets/31622972/26ffb861-3399-471e-9030-e74e46b0c8e8

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (commented):
I've given this a very quick, preliminary review. This is really impressive work, @dikwickley!

I've tested this locally. It works perfectly.

The code needs some clean up:

- Please fix the linting errors.
- Fix the icons so that they're cleaner. See the code in that same file that looks like `<Icon {...iconAddNew} />`. Make your icon code work like that one. We shouldn't see `<svg>` inside a component unless it's a very low-level component.
- 100 extra lines in this one component just to handle the dragging isn't great. I'm not opposed to merging it like this. But ideally, we'd abstract that code out into its own component so that the concerns are better separated and components can be more easily reused. See that comment you have, _"Functionality for making sort entries draggable"_... A comment like that is a good sign that the code needs to be pulled out into a new abstraction. The comment is useful, but this is the sort of code that can get real messy over time. The comment applies to the next 100 or so lines, but comments have no way to communicate that sort of scope. Other abstractions like functions, modules, components, and actions are better able to wrap up big chunks of code like this into nice tidy boxes so that they have descriptive names around all the code at once and clear interfaces with the world outside that code. If you want to take a crack at pulling this code out into some other abstraction, I think that would be cool. But it's not strictly necessary to merge this PR. I also wouldn't be opposed to using a 3rd party library for this, especially if there is one that plays well with Svelte. You might even be able to look at some of the drag and drop libraries to see how they structure their abstractions.

I'd like to review the code more thoroughly, but I'm heading off for a 1 week vacation soon so I'm short on time. I'll get back to this PR when I return. We'll definitely merge this. We just need to get it cleaned up a bit. Thanks for your hard work thus far!

seancolsen:
Oh wait: you won't be able to merge it yourself. You can submit a review or just post a comment, and I'll merge.

dikwickley:
Thanks @seancolsen, i'll take a look.

dikwickley:
@seancolsen this is really amazing. You have abstracted it out in a much functional way. I did try to abstract it out myself but failed as the code was quite coupled.
I have tested it out for all the use cases and it works flawlessly. I think it's good to go 🚀

seancolsen:
Thanks!

PR #3340, opened 2023-12-08
Translate /src/systems, support passing arguments to RichText slots

pavish:
**This PR:**
* Translates components within `/src/systems`.
* Updates RichText component to support passing arguments to slots, enabling translation string syntax like:
  ```json
  "click_for_info": "Click [linkComponent](here) for more information."
  ```
* Removes a few slate components that are no longer used.

**Notes for reviewers:**
* I've not translated errors that are never meant to occur in prod. These are exhaustive checks that are in place for making development easier.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (commented):
@pavish I'm reviewing the RichText component changes only.

I like the spirit of the changes to RichText. But I have some concerns with the syntax (which I recall expressing when you introduced this idea to me on a call).

The syntax `[linkComponent](here)` looks so much like a markdown hyperlink that I'm worried editors will mix up the order of `linkComponent` and `here`. I can see why you chose that order though — it's in keeping with putting the component name in the square brackets. I think on our call I had proposed reversing them with a syntax like `[here](linkComponent)`, but I can see how that would also be confusing because the square brackets would be inconsistent with our existing RichText syntax — so I rescind that proposal.

What do you think about using one of these alternate syntaxes:

- `Click [linkComponent(here)] for more information.`
- `Click [linkComponent:here] for more information.`
- `Click [linkComponent here] for more information.`

pavish:
@seancolsen I thought a bit about this syntax.

The major concern I wanted to address is to ensure that translators know that anything within square brackets `[linkComponent]` **should not** be translated, however, everything within normal brackets `(here)` **should** be translated.

The alternative syntaxes you've provided would make it confusing to translators.

> looks so much like a markdown hyperlink that I'm worried editors will mix up the order of linkComponent and here

I'm not concerned about this as much as these strings are only mentioned in a json file. I'm not sure if this is a concern we should address.

Having said that, I'm open to brainstorming a better syntax. I'd like to do that separately from this PR though.

seancolsen (approved):
@pavish and I chatted on a call about this.

**Our plan:** We are going to merge this in as-is and then be sure to re-visit the syntax questions before we hire people to perform the translation work. We may decide to re-think the syntax within RichText component more broadly once we have a clearer idea of all the cases it needs to handle. At that point, if we make changes to the syntax, then we'll implement those syntax changes (modifying all the strings) before training any translators on it.

PR #3343, opened 2023-12-09
[Bugfix:API] Fix DatetimeFieldOverflow error handling

### LINKED ISSUE #3331: Internal Server Error when searching with partial dates in Find a Record
## Description

Sometimes when partially entering the date in the Record Search Field, it randomly throws internal server errors.
Sometimes it throws the intended `InvalidDateFormatAPIException` but other times it `UnboundLocalError`.

Here is a recording of the behaviour:

https://github.com/mathesar-foundation/mathesar/assets/31622972/6e7885c2-9d19-473a-86e0-8c63c5e68c0f

Also, it would be good UI/UX to show the user the date is invalid. In case when the user has just put an incorrect date time format and the server returns a 400 BAD REQUEST InvalidDateFormatAPIException. The user won't know as the table still shows the loading state.
image

## Expected behavior

- Server should not throw 500 internal server error and always throw InvalidDateFormatAPIException in case of wrong date format.
- User should be conveyed if their input date is incorrect.

## Environment
 - OS: macOS
 - Browser: Chrome
 - Other info:

sahilsuman933:
Fixes #3331

This change incorporates an additional error from psycopg2 within the try-catch exception block in DatetimeFieldOverflow to handle out-of-range date values. Due to this, we encountered an UnboundLocalError, which shouldn't have been the case; instead, it should have been an error in the format of InvalidDateFormatAPIException.
 
**Screenshots**

Error Message:
image

After implementing this error handling case:
image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
Hey, @sahilsuman933 please add a test case for this.

mathesar/api/db/viewsets/records.py
```diff
@@ -103,7 +103,7 @@ def list(self, request, table_pk=None):
                 status_code=status.HTTP_400_BAD_REQUEST
             )
         except DataError as e:
-            if isinstance(e.orig, InvalidDatetimeFormat):
+            if isinstance(e.orig, (InvalidDatetimeFormat, DatetimeFieldOverflow)):
```
Anish9901, line 106: Add an else to raise `e`, for the case in which we're unable to catch the specific exception.

Anish9901 (changes requested):

sahilsuman933:
There is no need for an additional test case because there is already a test case that should cover this issue. You can verify this by checking the test case `99/99/9999`. Previously, it resulted in a 500 Internal Server Error, but after my fix, it no longer throws that HTTP error.

Test Case: 
image

SRC: /mathesar/tests/api/test_record_api.py 

Before my fix:
image


After my fix:
image


// @Anish9901

mathesar/api/db/viewsets/records.py
sahilsuman933: Do you mean that I should add an `except Exception as e` to handle all errors, or do you want me to create a separate `elif` statement for `DatetimeFieldOverflow`?

sahilsuman933: Is this what you want?
image

Anish9901: Your current Implementation is fine just add an else block to throw any _other_ `DataError`, I want to eliminate the senarios where we are catching the exception correctly(`DataError` in this case) but not throwing it.

Anish9901:
> There is no need for an additional test case because there is already a test case that should cover this issue. You can verify this by checking the test case 99/99/9999. Previously, it resulted in a 500 Internal Server Error, but after my fix, it no longer throws that HTTP error.

Was this the case while running the pytest as well? As from the video in the issue description it seems like the exception was getting thown the 2nd time someone tried to add an invalid date.

sahilsuman933:
> Was this the case while running the pytest as well? 

No, I didn't encounter any issues while running the pytest. When I manually tested it with the table data used by pytest, it ran without any problems. Surprisingly, it didn't even raise an error for the value `99/99/9999`, which should have triggered an "invalid_date_format" issue.

image

Instead of returning a count of 0, it should throw an error for an invalid date. However, this was not the case with the test database used by pytest.

I am using patents table data:
image

Now, when I tried using it with another test data, specifically the "invalid_reference_base_table.csv" table, it started throwing an "InvalidDateFormat" error.

> As from the video in the issue description it seems like the exception was getting thown the 2nd time someone tried to add an invalid date.

If you carefully observe the video, you'll notice that the date format should be in MM-DD-YYYY, and the person was attempting to enter a value greater than 12, which caused the "DatetimeFieldOverflow" error. This is because there are only 12 months in a year.

mathesar/api/db/viewsets/records.py
sahilsuman933: Added these changes.

mathesar/api/db/viewsets/records.py
```diff
@@ -103,11 +103,16 @@ def list(self, request, table_pk=None):
...
```
Anish9901, line 106: Please split these to raise `InvalidDateFormatAPIException` and `InvalidDateAPIException` respectively.

mathesar/api/db/viewsets/records.py
```diff
@@ -103,11 +103,16 @@ def list(self, request, table_pk=None):
...
                 raise database_api_exceptions.InvalidDateFormatAPIException(
                     e,
                     status_code=status.HTTP_400_BAD_REQUEST,
                 )
+            else:
+                raise database_api_exceptions.RaiseExceptionAPIException(
```
Anish9901, line 112: `MathesarAPIException` would be more appropriate here, check the following for reference.
https://github.com/mathesar-foundation/mathesar/blob/d375575199e81205ea9cd4a7a3b928398f43ec65/mathesar/api/serializers/records.py#L62-L63

Anish9901 (changes requested):
I noticed a couple more things since my last review:
- We have a separate exception class for `DatetimeFieldOverlflow` namely `InvalidDateAPIException`.
-  I was able to reproduce the issue locally using our pytest test suite, and you should be adding a test case addressing this issue, the test case mentioned in your screenshot refer to record update instead of record search you can see the difference by noticing the patch and get requests.
Look at `record_test_search` for reference.

mathesar/api/db/viewsets/records.py
sahilsuman933: I have addressed all these changes that you mentioned in my latest commit.

sahilsuman933:
@Anish9901 The test case is failing and it is because the `Patent Expiration Date` is being treated as text rather than DATE as we discussed earlier. So, should I address this particular issue in the same pull request or should I create a new pull request for it.

sahilsuman933:
@Anish9901 Now the tests related to invalid dates are passing. Thank you so much for your guidance :)

Anish9901 (approved):
Alright, looks good now @sahilsuman933. Thanks for your contribution!

PR #3345, opened 2023-12-10
Parsing date entries on frontend.

### LINKED ISSUE #2327: Parse user entry of dates and times on the front end
## Context

In the following places, we accept user-entry of date/datetime/time values:

- Cells, when saving values
- Filter conditions, when filtering the table page, or an exploration
- Inputs when searching via the record selector
- Inputs when saving a record on the Table Page
- Inputs, when setting a default value for a column

## Current behavior

- In all the above contexts, the front end attempts to parse the user input into a structured value. If it succeeds, it sends an ISO-formatted string for that value. If it fails, it sends the raw user input. Sending the raw user input allows us to handle special values like "today" and "now" that postgres is able to understand. We don't want to parse those values into dates on the front end because we want to allow the user to set filter conditions that will change based on the current date (e.g. setting the default value for a date column to "today").

- This behavior is problematic because it produces this bug described in #1890.

    image

## Desired behavior

- The front end maintains a list of reserved "special" date values, so far only `today`, `now`, `yesterday`. If the user input matches a special value, then the front end sends that special value. Otherwise...

- The front end attempts to parse the user input into a structured value. If it succeeds, it sends an ISO-formatted string for that value. If it fails, then it displays a client-side error, allowing the user to fix the problem before submitting any API requests.

## Additional considerations

- I noticed that for some formats (e.g. `1/1/2020`) the Postgres parsing succeeds but our current parsing logic on the front end does not. When implementing this ticket, we'll need to improve the parsing logic on the front end to handle more cases like this and avoid UX regressions.

## Previous discussion

- This started in #1890
- Then we [chatted on Matrix](https://matrix.to/#/!vAyoAQoixqNrvBuIcH:matrix.mathesar.org/$LNuRmMfzsZ_8eV2Q-Z4tZ5kwfYUt_KsPVWQiRzSxYbU?via=matrix.mathesar.org&via=matrix.org)
- We had a [meeting](https://github.com/centerofci/mathesar-wiki/blob/master/meeting-notes/2023-01/2023-01-20-partial-user-input-of-dates.md) where we formulated the plan above

## See also

- #3953

dikwickley:
Fixes #2327

**Technical details**

This PR adds an `allowPartialParse` flag to the `FormattedInput`. This flag prevents strings that are not formatted as they are incomplete. Incomplete date inputs are enabled with a `has-error`class that signifies that the input is not ready to be submitted and is in incorrect format.
Only parsed dates are sent to the backend now rather than sending raw user input like before.
Also added few more formatting strings to format special cases.

**Screenshots**

https://github.com/mathesar-foundation/mathesar/assets/31622972/9c35bc7c-9907-425c-a3c8-8fcaa1de460c

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish I had initially assigned this to me to review since I opened the ticket. But now I see that it's targeting code that you wrote, so I think it would be better if you could review it. I've assigned you.

Issue #3445, opened 2024-02-13
Install from scratch documentation feedback


seancolsen:
@alvorithm emailed me the following feedback about our [Install from scratch](https://docs.mathesar.org/installation/build-from-source/) documentation (paraphrased slightly):

Potential changes:

- (A) When installing under `$HOME`, one needs to change the gunicorn service to run as `$USER`. Otherwise the permission error takes some time to work out (and I guess one doesn't want to give gunicorn.gunicorn control over the entire mathesar software folder?).

- (B) Once installed, we should go to `localhost:8000` (currently no port is indicated)

- (C) The default caddy.service file from the standard .deb stable package comes with other AmbientCapabilities, maybe a line comment there could help the rushed installer notice the difference.

- (D) Within the installation steps, I created a `<my-db>` database with owner mathesar. Is it correct to say that this is already a 'user' database, meant for user data?

    The confusion is due to having to create an admin user TTW. This user, with name  `<admin-user>` != 'mathesar' does not allow connecting to what (I believe) is the user database.

    What should one do here? Create a new DB via the psql prompt with user `<admin-user>` and connect to it, or re-use the 'mathesar' credentials and connect to `<my-db>`? I notice that even though I use the mathesar credentials, `<admin-user>` is listed as the sole user with access.

    It is almost obvious that I am missing some really basic distinction about how Postgres manages connection credentials vs. admin users (?), but given how smooth the rest of the process was, you might want to add a hint at the end of the flow so that the user successfully makes their first connection. 

Overall feedback:

- It is a lengthy procedure but overall well documented. I now have mathesar running on my laptop.
- For a network host I'd much prefer getting it to run under guix (similar to nix, https://guix.gnu.org/), but I don't know yet how to create services in Guix (they use their own shepherd, instead of systemd).

seancolsen:
I tried to make some progress on making changes to address this feedback, but I got hung up because I didn't understand the inner-workings of the installation process well enough.

@Anish9901 do you think you'd be able to make some docs changes to address this feedback?

kgodey:
Added the "user reported" tag since this is feedback from a user.

Anish9901:
I'll give it a shot @seancolsen.

vr-varad:
@Anish9901 are u working on it or i can give it  a try??

seancolsen:
Sorry @vr-varad but this issue is restricted to maintainers.

PR #3478, opened 2024-03-07
Extend mathesar to support python 3.10 and above

### LINKED ISSUE #3485: Add support for Python versions 3.10 and above
## Problem

We currently publish supporting only python 3.9 in our [build from source documentation](https://docs.mathesar.org/installation/build-from-source/#software) and all of our docker images are published using the same. This is a problem since modern Debian versions now ship with Python 3.10+ natively installed.

## Proposed solution

We should support Python versions from 3.9 all the way till 3.12(The latest Python version supported by Django)

Anish9901:
Fixes #3485 

**Technical details**

This PR extends mathesar to support Python versions 3.10, 3.11, and 3.12! This also adds the same in the testing matrix and the python tests are set up to be run with the cross-product of each supported python version with each supported postgres version. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
@Anish9901 FYI there's a package called `drf-friendly-errors-django-4` that should be a drop in replacement for `drf-friendly-errors`, I found it when testing 0.1.4 installation with 3.11, [here's my notes](https://hackmd.io/9MZdwS6VQEGXvEA3caQZcw?view).

https://pypi.org/project/drf-friendly-errors-django-4/

Anish9901:
I did try that package @kgodey, but it has some additional changes that conflicts with our current error codes and hence breaks some of our tests.

Anish9901:
I am thinking of handling doc changes regarding python version in #3445.

kgodey:
@Anish9901 Please handle the doc changes in this PR, otherwise the docs won't be in sync with the code. Also #3445 doesn't appear to be in this release, should it be?

mathesar/tests/api/test_record_api.py
```diff
@@ -1527,6 +1527,7 @@ def test_record_post_exclusion_violation(reservations_table, engine, client):
         ("room_number" WITH =, TSRANGE("check_in_date", "check_out_date", '[]') WITH &&);"""
     )
     with engine.begin() as conn:
+        conn.execute(text("""SET search_path="Reservations";"""))
```
mathemancer, line 1530: @Anish9901 Why was this change necessary? Will this affect app behavior? I.e., do we now need to check that we're manually setting the search path when we previously weren't?

mathemancer (changes requested):
I have a quick question. See the specific comment.

mathemancer:
@kgodey Regarding drf-friendly-errors, this is a temporary measure until we remove that dependency entirely.

mathesar/tests/api/test_record_api.py
Anish9901: The problem is that these tests would fail when running our entire test suite but would pass when run individually or with the `--lf` flag. I think the issue has to do with the way pytest handles its fixtures. This change was irrelevant to this PR however, the aforementioned issue existed since these tests were initially added so I decided to fix them in this PR. I have removed the `SET search_path` statement and fixed the underlying fixture that was causing the issue in the following commit:  https://github.com/mathesar-foundation/mathesar/pull/3478/commits/c3120967fb50933c3d20a58383c898ccee23fb2e.

mathemancer (approved):
Thanks for the explanation, and for your work on this PR. It looks great!

PR #3494, opened 2024-03-18
Release 0.1.6

Anish9901:


Anish9901:
Assigned @seancolsen & @mathemancer to check for red flags from Frontend and Backend perspectives respectively.

mathemancer (approved):
Looks okay to me. The main thing I'd like to flag is that this version adds a dependency on GNU `gettext` for setting up Mathesar from scratch. While less onerous than the thankfully removed Node dependencies, I still think we should consider whether there's a way to avoid that in future releases.

mathemancer:
> Looks okay to me. The main thing I'd like to flag is that this version adds a dependency on GNU `gettext` for setting up Mathesar from scratch. While less onerous than the thankfully removed Node dependencies, I still think we should consider whether there's a way to avoid that in future releases.

Issue about the dependencies: #3514

seancolsen (approved):

PR #3499, opened 2024-03-20
Fix CSRF failures when app is behind a reverse proxy

### LINKED ISSUE #3497: CSRF Failed Error
## Description
The error message 
"CSRF Failed: Origin checking failed - https://internal.mathesar.org/ does not match any trusted origins."
is appearing when performing certain actions in Mathesar like creating a new internal db or saving a record. 

## Additional context
This happened while testing internal.mathesar.org

Anish9901:
Fixes #3497

**Technical details**

Django 4 introduced some additional security measures w.r.t. CSRF protection. `CsrfViewMiddleware` specifically the [`_origin_verified()`](https://github.com/django/django/blob/stable/4.2.x/django/middleware/csrf.py#L276) method which relies on `is_secure()` to verify if the protocol used to make request by the client matches that to the request actually made to the django server.

Client(makes request via **https**)-->Caddy(makes request via **http**)-->Django

There is a mismatch between the protocol used by the client and caddy to make requests, this mismatch causes `is_secure()` to be False. Caddy in addition to make requests via HTTP, sends a header named [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) which tells django that the original request was made using HTTPS in our case. Adding `SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')` to our settings tells django to trust the X-Forwarded-Proto header that comes from our proxy, and any time its value is 'https', then the request is guaranteed to be secure.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -3,6 +3,14 @@ from config.settings.common_settings import * # noqa
 # Override default settings
 DEBUG = False
 MATHESAR_MODE = 'PRODUCTION'
+
+'''
+This tells Django to trust the X-Forwarded-Proto header that comes from our proxy,
+and any time its value is 'https', then the request is guaranteed to be secure
+(i.e., it originally came in via HTTPS).
+'''
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
 # Use a local.py module for settings that shouldn't be version tracked
 try:
     from .local import * # noqa 


mathemancer (approved):
I checked locally and this seems to sort it, but we need to double-check on internal.mathesar.org once this is merged.

PR #3526, opened 2024-04-05
 Navigate to linked table via cell context menu 

### LINKED ISSUE #2350: Navigate to linked table via cell context menu
## Current behavior

- We have this context menu on all cells.

    image

## Desired behavior

- When opened via a foreign key cell (as I've done in the above screenshot)...

    - (A) The "Go To Linked Record" functions the same, but is instead labeled as:

        > Open `Patricia Anderson`

        The icon is `iconRecord`

    - (B) The menu contains an additional entry labeled like this:

        > Open **Patrons** Table

        The icon is `iconTable`

- The (B) menu entry is also present when the context menu is opened via a foreign key column's header cell.

@kgodey Are these UX changes okay with you? If so, we can un-draft this ticket and mark it as "help wanted" and "good first issue".

hitenvidhani:
Fixes #2350 

**Screenshots**

On clicking a record which is inked
image

On clicking a record which is **not** linked
image

On clicking column name which is linked
image

On clicking column name **not** linked
image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
Thanks @hitenvidhani this is a good start. I have some changes to request:

- In `ColumnHeaderContextMenu.svelte`

    - Don't use `router.goto`. Instead, use `LinkMenuItem` like we are using in `RowCell.svelte`.

    - You have:
    
        ```svelte
        {$_('open')}
        <Identifier>{column.name}</Identifier>
        {$_('table')}
        ```

        This structure might be problematic when translated into other languages and thus does not follow best-practices for i18n. Apologies that we don't yet have any documented code standards which explain this. (We only recently enabled i18n.) The problem here is that some languages might follow different grammatical structures to translate this same text and the words might become rearranged or split depending on context. For this reason, translation needs to take place at a less granular level. We need to translate pieces of text at a higher level than this.

        For this particular string you'll need to make use of the `RichText.svelte` component. Look at the documentation within the component to understand how it works, and check other usages of it too. You'll need to add a new string to `src/i18n/languages/en/dict.json` like this:

        ```json5
        {
          // ...
          "open_named_table": "Open [tableName] table",
          // ...
        }
        ```

        Then pass that string into `RichText.svelte` with the other props and child components as necessary.

    - In that string above, don't use the _column_ name. Instead use the the _table_ name.


- In `RowCell.svelte`, you've added code which uses the following expression:

    ```js
    $recordSummaries.get(String(column.id))?.get(String(value))
    ```

    Because this expression is also present higher up in the component, it would be nice to extract up into a common `recordSummary` variable that gets used in both places. This would improve code cleanliness.

mathesar_ui/src/systems/table-view/row/RowCell.svelte
```diff
@@ -167,9 +171,7 @@
       {value}
       {isProcessing}
       {canViewLinkedEntities}
-      recordSummary={$recordSummaries
-        .get(String(column.id))
-        ?.get(String(value))}
+      recordSummary={recordName}
```
hitenvidhani, line 174: > extract up into a common recordSummary variable that gets used in both places.

Throws lint error hence making this change 
`  175:36  error  'recordSummary' is already declared in the upper scope on line 105 column 6  @typescript-eslint/no-shadow`

hitenvidhani: same reason to keep the variable as `recordName` and not `recordSummary`

seancolsen (approved):
Thanks @hitenvidhani.

I added a few other commits to clean things up. Please look over them and make sure you understand them. Ask questions if anything I did seems concerning or doesn't make sense.

mathesar_ui/src/systems/table-view/row/RowCell.svelte
seancolsen: I pushed 9609d37cf81a27f510888e5c66a60b346efb7b8b which fixes this problem at the lower scope instead of the higher. This way we get the preferred name across more lines of the file. And we use the `rs` name in a very small scope.

seancolsen:
Good questions, @hitenvidhani 

- **how to sort imports**
    
    I use the "Organize Imports" action in VS Code. I have a keyboard shortcut set to do this. At some point I'd really like to get some tooling set up in our repo which does this for us. Ideally it would be something that integrates with Prettier so that we when run Prettier the imports are automatically sorted. I you are interested in working on that it could be a nice little self-contained project. I can help clarify the requirements if this is confusing.

- **demo site**

    We update it manually after each release. The `develop` branch is not as stable as our releases. Like, we might have mid-cycle bugs that crop up during feature development. We make an effort to fix bugs like that during our QA phase before each release. So we don't want to run the demo site off the develop branch because such regressions could give people a bad impression of Mathesar.

hitenvidhani:
Sure @seancolsen 
I did a bit of research and it looks like we'd have to use the plugin `prettier-plugin-svelte` and add specific rules for the files in `.prettierrc`. It would be great if you could clarify the requirements further.

seancolsen:
@hitenvidhani I opened #3542 with some clearer requirements. We can continue to discuss there.

PR #3538, opened 2024-04-15
Remove demo code

### LINKED ISSUE #3531: Remove demo code
For the beta release, the demo logic and associated code will no longer be in the Mathesar service. It would be good to remove that code _before_ doing major work on permissions, etc. since it would be more difficult to maintain the demo while also overhauling permissions.

mathemancer:
Fixes #3531 

This PR removes all Demo code which is otherwise unused from Mathesar.

**Technical details**

Some demo code was used for example datasets in other context, and so was moved to a different location rather than removed entirely.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


kgodey:
I assume we want to wait until the April release branch is cut (cc @ghislaineguerin) before we merge this.

mathemancer:
> I assume we want to wait until the April release branch is cut (cc @ghislaineguerin) before we merge this.

Definitely.

@pavish Any opinions on whether we should remove the demo-mode-related translations? I know we paid for them, but we can always recover them through git or other means when/if we want them back.

pavish:
@mathemancer 

> Any opinions on whether we should remove the demo-mode-related translations? I know we paid for them, but we can always recover them through git or other means when/if we want them back.

Yes, we should remove any key we're no longer using.

kgodey:
Agreed that paying for the translations isn't a reason to keep them around if we no longer need them.

mathemancer:
Okay, this is now ready for review. From @pavish , I'm just requesting a quick check to make sure there aren't any obvious problems where the front end might break because it expects something (e.g., a flag or variable) to be available from the back end.

Dockerfile
```diff
@@ -106,18 +106,6 @@ RUN rm -rf ./mathesar/tests ./db/tests
...
-FROM common AS demo
-
-# Install prod requirements
-RUN pip install --no-cache-dir -r requirements-demo.txt
-
-EXPOSE 8000
-
-ENTRYPOINT ["./run.sh"]
-
-
```
pavish, line 120: Since we no longer have the `demo` stage, I think we should remove the `common` stage as well.

mathesar/templates/mathesar/login_base.html
```diff
@@ -112,41 +112,6 @@
...
-            <li>{% translate "Username" %}: <strong>{{live_demo_username}}</strong></li>
-            <li>{% translate "Password" %}: <strong>{{live_demo_password}}</strong></li>
-          </ul>
-        {% endif %}
-
-        {% translate "Keep in mind that the data in the live demo is reset regularly." %}
-      </div>
-    </div>
-  {% endif %}
-
```
pavish, line 149: We would have to figure out a different way to mention that the demo data is reset regularly to users accessing the demo site. Do we have an approach figured out on how to do that?

pavish (changes requested):
@mathemancer The code changes look good to me. I have a couple comments.

mathesar/templates/mathesar/login_base.html
mathemancer: I don't have an approach figured out yet. I figure let's get everything out for now, and then only add things back into the service if absolutely needed later. Couldn't we have a custom front page served by the demo microservice? It could pull the needed assets from the "real service". Not sure, though.

Dockerfile
mathemancer: Fair enough.

mathemancer:
@pavish I made the requested change.

Anish9901 (approved):
Looks good to me!

Anish9901:
Waiting on your re-review @pavish, please merge this if you don't have any more concerns.

mathesar/templates/mathesar/login_base.html
pavish: I think having a custom front page should be fine.

However, we currently show a banner on the app while someone is viewing the demo. This PR removes that banner as well. I'm cool with removing it and adding it later based on requirements. I do think this needs to be tracked in a separate issue.

pavish (approved):
@mathemancer Thanks, looks great!

We will have to create an issue to track this: https://github.com/mathesar-foundation/mathesar/pull/3538#discussion_r1568677772

PR #3523, opened 2024-03-29
Add demo target to Dockerfile

mathemancer:
Fixes #3522 

This adds a build stage to the Dockerfile for version 0.1.6, and adds a setting to fix the issue #3497 , but in the demo setting.

Edit: Now the default build is `production`, other targets work as expected.

## Checklist
 
- [X] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [X] My pull request targets the `develop` branch of the repository
- [X] My commit messages follow [best practices][best_practices].
- [X] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [X] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3554, opened 2024-04-22
Release notes 0.1.7

ghislaineguerin:
Release notes for 0.1.7

### DIFF
--- a/docs/docs/releases/0.1.7.md
+++ b/docs/docs/releases/0.1.7.md
@@ -16,6 +16,8 @@ See our guide on [upgrading Mathesar to 0.1.7](../administration/upgrade/0.1.7.m
 
 Users can now navigate to linked tables from the cell context menu, providing a more seamless experience when working with linked data.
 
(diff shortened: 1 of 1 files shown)


seancolsen (changes requested):
- For #3526 (Linked Table Navigation) it would be nice to add a screenshot. Keep in mind that we might also end up wanting to use this screenshot for the release blog post.

- To me, #3523 (Demo Target) should go in Maintenance, not Groundwork. I consider Groundwork to be for unfinished _features_. Perhaps I should have been more clear about that when I wrote the template. I think of the Groundwork section like: _"hey, there's something cool we're working on and you're going to like it when we're finished"_.

- Likewise I think #3524 (RPC Endpoint) should go in Maintenance, not Improvements because this change doesn't actually add any value to the user for this release. Similarly, I also suggest removing _"introduces a new RPC endpoint implementation for superuser functions"_ from the short description at the top of the document.

- Fix markdown formatting around `_#3357`

seancolsen (approved):

PR #3597, opened 2024-05-16
Refactor list of RPC methods to test

seancolsen:
This PR cleans up the code in test_endpoints.py so that it's easier to add new RPC methods to that file.

This is its own PR because:

1. It would be nice to get this merged soon so that we can all work on top of it.
1. I ran into some weird test failures that I want help troubleshooting in isolation.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3598, opened 2024-05-16
Implement `schemas.list` RPC method

### LINKED ISSUE #3596: Implement `schemas.list` RPC method

seancolsen:
Fixes #3596

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/0_msar.sql
+++ b/db/sql/0_msar.sql
@@ -684,12 +684,9 @@ FROM (
   SELECT 
     s.oid AS oid,
     s.nspname AS name,
-    d.description AS description,
+    obj_description(s.oid) AS description,
     COALESCE(count(c.oid), 0) AS table_count
   FROM pg_catalog.pg_namespace s
-  LEFT JOIN pg_catalog.pg_description d ON
-    d.objoid = s.oid AND
-    d.objsubid = 0
   LEFT JOIN pg_catalog.pg_class c ON
     c.relnamespace = s.oid AND
     c.relkind = 'r'
@@ -698,8 +695,7 @@ FROM (
     s.nspname NOT LIKE 'pg_%'
   GROUP BY
     s.oid,
-    s.nspname,
-    d.description
+    s.nspname
 ) AS schema_data;
 $$ LANGUAGE sql;
 
--- a/db/sql/0_msar.sql
+++ b/db/sql/0_msar.sql
@@ -689,6 +689,8 @@ FROM (
   FROM pg_catalog.pg_namespace s
   LEFT JOIN pg_catalog.pg_class c ON
     c.relnamespace = s.oid AND
+    -- Filter on relkind so that we only count tables. This must be done in the ON clause so that
+    -- we still get a row for schemas with no tables.
     c.relkind = 'r'
   WHERE
     s.nspname <> 'information_schema' AND
--- a/db/sql/0_msar.sql
+++ b/db/sql/0_msar.sql
@@ -662,6 +662,45 @@ SELECT EXISTS (SELECT 1 FROM pg_attribute WHERE attrelid=tab_id AND attname=col_
 $$ LANGUAGE SQL RETURNS NULL ON NULL INPUT;
 
 
+CREATE OR REPLACE FUNCTION msar.get_schemas() RETURNS jsonb AS $$/*
+Return a json array of objects describing the user-defined schemas in the database.
+
+PostgreSQL system schemas are ignored.
+
+Internal Mathesar-specifc schemas are INCLUDED. These should be filtered out by the caller. This
+behavior is to avoid tight coupling between this function and other SQL files that might need to
+define additional Mathesar-specific schemas as our codebase grows.
+
+Each returned JSON object in the array will have the form:
+  {
+    "oid": <int>
+    "name": <str>
+    "description": <str|null>
+    "table_count": <int>
+  }
+*/
+SELECT jsonb_agg(schema_data)
+FROM (
+  SELECT 
+    s.oid AS oid,
+    min(s.nspname) AS name,
+    min(d.description) AS description,
+    COALESCE(count(c.oid), 0) AS table_count
+  FROM pg_catalog.pg_namespace s
+  LEFT JOIN pg_catalog.pg_description d ON
+    d.objoid = s.oid AND
+    d.objsubid = 0
+  LEFT JOIN pg_catalog.pg_class c ON
+    c.relnamespace = s.oid AND
+    c.relkind = 'r'
+  WHERE
+    s.nspname <> 'information_schema' AND
+    s.nspname NOT LIKE 'pg_%'
+  GROUP BY s.oid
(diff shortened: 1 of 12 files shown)


seancolsen:
Note this is stacked on top of #3597

db/sql/0_msar.sql
```diff
@@ -662,6 +662,45 @@ SELECT EXISTS (SELECT 1 FROM pg_attribute WHERE attrelid=tab_id AND attname=col_
...
+    min(s.nspname) AS name,
+    min(d.description) AS description,
+    COALESCE(count(c.oid), 0) AS table_count
+  FROM pg_catalog.pg_namespace s
+  LEFT JOIN pg_catalog.pg_description d ON
+    d.objoid = s.oid AND
+    d.objsubid = 0
+  LEFT JOIN pg_catalog.pg_class c ON
+    c.relnamespace = s.oid AND
+    c.relkind = 'r'
```
mathemancer, line 695: I think this filtering move warrants a comment about why it's in the `JOIN` rather than `WHERE` clause. It's not completely nonstandard, but it's not super common or obvious either. I spent longer than I'm proud of working through what would go wrong if it were just in the `WHERE` clause.

mathemancer (changes requested):
Tests look great, overall organization is also nice.

I have some changes to request for the SQL code itself; see my specific line comments.

db/sql/0_msar.sql
```diff
@@ -662,6 +662,45 @@ SELECT EXISTS (SELECT 1 FROM pg_attribute WHERE attrelid=tab_id AND attname=col_
...
+    "description": <str|null>
+    "table_count": <int>
+  }
+*/
+SELECT jsonb_agg(schema_data)
+FROM (
+  SELECT 
+    s.oid AS oid,
+    min(s.nspname) AS name,
+    min(d.description) AS description,
```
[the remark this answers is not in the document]
seancolsen: Fixed in d85df60c3af1c8d36a73ffa961dd266bb3745a39

I originally wanted to use `any_value` but (to my surprise) this function was only just recently [added](https://www.postgresql.org/docs/current/release-16.html#RELEASE-16-FUNCTIONS) in PostgreSQL 16.

I'm accustomed to the behavior in MySQL which basically automatically applies `any_value` any time you don't explicitly use an aggregate function.

It's helpful to know what sort of patterns you prefer to use for cases like this. Thanks. I'll try to use additional grouping columns going forward.

db/sql/0_msar.sql
seancolsen: Fixed in e79f31c2e1f4a36b4cb68a67bcea92d49bb917fb

> it's not super common or obvious

For what it's worth: I tend to use the `ON` clause for filtering conditions any time I'm using `LEFT JOIN`. So I stuff things in the `ON` clause quite often, as a pattern. Within the `WHERE` clause I typically only reference tables from `FROM` and `JOIN` clauses — or sometimes I'll reference the pk of a `LEFT JOIN` table when I'm doing an exclusion join.

There might be exceptions to that pattern. But I just mention this to give you a window into my thought process when writing queries. Like, if I see a non-pk column from a `LEFT JOIN` table referenced in `WHERE`, it'll raise my eyebrows and I'll look for a bug.

seancolsen:
Ready for re-review, @mathemancer

db/sql/0_msar.sql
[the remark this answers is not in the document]
mathemancer: > I originally wanted to use `any_value` but (to my surprise) this function was only just recently [added](https://www.postgresql.org/docs/current/release-16.html#RELEASE-16-FUNCTIONS) in PostgreSQL 16.

Innnteresting. I could swear there was something like that before, but maybe I dreamt it.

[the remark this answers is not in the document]
mathemancer: I suspect you're correct that the join would be more performant, but I note that the `psql` command `\d` uses the `obj_description`, so it's probably not _that_ bad.

db/sql/0_msar.sql
```diff
@@ -662,6 +662,46 @@ SELECT EXISTS (SELECT 1 FROM pg_attribute WHERE attrelid=tab_id AND attname=col_
...
+    "description": <str|null>
+    "table_count": <int>
+  }
+*/
+SELECT jsonb_agg(schema_data)
+FROM (
+  SELECT 
+    s.oid AS oid,
+    s.nspname AS name,
+    obj_description(s.oid) AS description,
```
mathemancer, line 687: I'm sorry to go around again for such a small thing, but I think you were correct in a previous conversation (or comment; can't remember) when you noted that we should namespace these things since a (perhaps foolish) user could have created a function with the same name in some context. So, it'd be better to use `pg_catalog.obj_description`.

mathemancer (changes requested):
I have one tiny change to request. If you're okay with that, feel free to merge after namespacing the `obj_description` function.

db/sql/0_msar.sql
seancolsen: Cool. Makes sense. Fixed in f3289641fcf8f33ebe2c64aaf7d6c98dbf9e7bd4

Anish9901:
@seancolsen I noticed that this PR doesn't have any python test for testing the `schemas.list_` rpc function, it only has one for testing the expected name of the endpoint. Any reason why that may be?

seancolsen:
@Anish9901 It didn't seem worth testing to me. Do you think it should have a test? What would you want to assert in that test? I'm open to it certainly. But also trying to avoid too much boilerplate in order to move quickly. For reference, we discussed this in our last [meeting](https://tldv.io/app/meetings/66460318b384e50013a175b9) (starting at about 48:20 and my take-away from that discussion was that the test coverage that matters the most to us right now is the test coverage at the SQL layer.

Anish9901:
Ok, I just went through the video, as per my understanding, I don't think we agreed on not adding the python tests for the rpc endpoints rather I think we decided on not having end-to-end tests which would call up the db and hence tend to be slow. FWIW, I think we should at least have one test for making sure that the sql function is called with appropriate parameters, this is relatively low effort. I'd like to hear @mathemancer's opinion on whether or not we want tests that check the wiring of the functions that are called from within the rpc functions.

mathemancer:
As I'm looking through the PR with @Anish9901 's concerns in mind:

- Given that we're not doing any E2E or even full-backend-call testing, we need to be extra conscientious about testing that the wiring between tests is as expected, and that tests are mangling things appropriately.
- For the schema listing RPC function, therefore, we could add a test to:
  - Make sure the call to the underlying python-layer function is as expected, and
  - The list comprehension produces expected output, given some monkeypatched result from the python-layer function.

This would be similar to, for example, the test for `columns.list_`. It's kind of a pain to put together, but it does give us something to break if we accidentally typo something in the function logic that we might not otherwise notice.

PR #3599, opened 2024-05-17
Implement `tables.list` rpc endpoint

### LINKED ISSUE #3600: Implement `tables.list` RPC method

Anish9901:
Fixes #3600

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
@mathemancer, @seancolsen As of now the `tables.list` endpoint returns the schema oid with each table_info json blob, however, it is debatable whether or not we want to have that, as`tables.list_` filters tables based on a schema oid rather than database id. In other words, we already have the schema oid that will be returned in each json object returned by `tables.list` endpoint. Any thoughts about whether or not we want to include schema oid in each json object?

seancolsen:
@Anish9901 

> whether or not we want to include schema oid in each json object?

Good question. I would lean towards keeping it. I don't see much harm in having it there. And the data structures in the front end _do_ actually appear to use that property as returned from the REST API. It's possible it would be easy to refactor that stuff out of the front end. But it would take me some time to investigate deeply enough to form a definitive answer.

mathesar/rpc/tables.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Information about a table.
+
+    Attributes:
+        oid: The `oid` of the table in the schema.
+        name: The name of the table.
+        schema: The `oid` of the schema where the table lives.
+        description: The description of the table.
+    """
+    oid: int
```
mathemancer, line 21: I'm okay with this, but we should chat at some point with @seancolsen about preferences w.r.t. specific ID names instead of just `id` in this type of case. I went with `id` rather than `attnum` for column info, for example. I don't really have a strong preference, but we should choose one way and stick with it.

mathemancer: As I'm looking at this more, I think `id` is more consistent with referring to the `schema` and using its (o)id.

mathemancer (approved):
I noticed one detail that we should discuss, but I prefer merging this and then fixing it once we've decided on the convention to use.

PR #3620, opened 2024-06-12
Implement `schemas.add` RPC method

### LINKED ISSUE #3619: Implement `schemas.add` RPC method

seancolsen:
Fixes  #3619

## Notes

I went around in circles a bit with this PR, which is why it took a while. I'll describe my thought process chronologically in hopes that it might be useful as these same problems could potentially apply to other DB objects as well...

1. First I proceeded to remove the `if_not_exists` parameter from `msar.create_schema` because I didn't want the additional complexity in there. I wanted to default to _not_ using `IF NOT EXISTS` for the sake of simplicity. This is in line with comments in [this meeting](https://tldv.io/app/meetings/665897c0678bae00133db210?transcript=true&video=true) (starting at 21:10) on which @mathemancer and @Anish9901 were supportive. I wanted to raise an exception if the supplied schema name already existed.

1. But then I noticed that our codebase actually made use of that `if_not_exists` parameter in a few places for internal schemas and creating ephemeral schemas on the fly for type inference. Ugh!

1. My initial instinct was to modify the IF NOT EXISTS code locations by resorting to calling raw SQL from within the service layer (instead of calling functions). I had a tricky time figuring out how to do that cleanly with SQLAlchemy. I imagine it's possible, but it didn't seem to fit our patterns, and I didn't want to fiddle with it too much.

1. After some hemming and hawing, I decided to retain support for `if_not_exists`, in `msar.create_schema` while also adding support for setting descriptions on schemas within the same function. This choice sent me down a rabbit hole, making me wish I had trusted my early intuition on `if_not_exists` more strongly.

1. The `msar.create_schema` function got more and more complex as I discovered more edge cases and inconsistencies. If it can return an existing schema, then we need to make sure that `NULL` description values don't overwrite existing description. Ok fine. Then for the front end's sake, we should actually be fetching and returning the description. Ok. But that means the return value gets much more complex. It should be an object instead of a simple oid. And if it's an object, it ought to match the structure of the objects returned from `msar.get_schemas`, meaning it should supply a `table_count` property too. Ugh. So I started modifying `msar.get_schemas` to accept a `sch_oid` filter parameter which would allow me to compose that function to reliably get the full schema details in a consistent manner. But then I just threw up my hands and said, "this is ridiculous!"

1. So I decided to split the function into two:

    - `msar.create_schema_if_not_exists(sch_name text)`

      and

    - `msar.create_schema(sch_name text, description text DEFAULT '')`

    This means you can use `IF NOT EXISTS` _OR_ you can supply a description — but you can't do both at the same time. And at the API layer there's no way to use `IF NOT EXISTS`.

    This approach has some limitations. But compared to cramming all that logic in one function, it's much simpler and less prone to weird edge cases and bugs.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,7 +14,7 @@ from db.engine import add_custom_types_to_ischema_names, create_engine as sa_cre
 from db.types import install
 from db.sql import install as sql_install
 from db.schemas.operations.drop import drop_schema_via_name as drop_sa_schema
-from db.schemas.operations.create import create_schema as create_sa_schema
+from db.schemas.operations.create import create_schema_if_not_exists_via_sql_alchemy
 from db.schemas.utils import get_schema_oid_from_name, get_schema_name_from_oid
 
 from fixtures.utils import create_scoped_fixtures
@@ -210,7 +210,7 @@ def create_db_schema(SES_engine_cache):
         if schema_mustnt_exist:
             assert schema_name not in created_schemas
         logger.debug(f'creating {schema_name}')
-        create_sa_schema(schema_name, engine, if_not_exists=True)
+        create_schema_if_not_exists_via_sql_alchemy(schema_name, engine)
         schema_oid = get_schema_oid_from_name(schema_name, engine)
         db_name = engine.url.database
         created_schemas_in_this_engine = created_schemas.setdefault(db_name, {})
--- a/db/schemas/operations/alter.py
+++ b/db/schemas/operations/alter.py
@@ -27,10 +27,8 @@ def comment_on_schema(schema_name, engine, comment):
     Change description of a schema.
 
     Args:
-        schema_name: The name of the schema whose comment we will
-                     change.
-        comment: The new comment. Any quotes or special characters must
-                 be escaped.
+        schema_name: The name of the schema whose comment we will change.
+        comment: The new comment.
         engine: SQLAlchemy engine object for connecting.
 
     Returns:
--- a/db/schemas/operations/create.py
+++ b/db/schemas/operations/create.py
@@ -1,26 +1,53 @@
-from db.schemas.operations.alter import comment_on_schema
-from db.connection import execute_msar_func_with_engine
+from db.connection import execute_msar_func_with_engine, exec_msar_func
 
 
-def create_schema(schema_name, engine, comment=None, if_not_exists=False):
+def create_schema_via_sql_alchemy(schema_name, engine, description=None):
     """
-    Creates a schema.
+    Creates a schema using a SQLAlchemy engine.
 
     Args:
         schema_name: Name of the schema to create.
         engine: SQLAlchemy engine object for connecting.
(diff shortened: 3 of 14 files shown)


mathemancer (approved):
Okay, this PR has the required functionality and doesn't seem to add any bugs. So, we'll merge it. However, I have some feedback on the SQL functions.

I disagree with your statement that the needed logic is so complicated that it can't all be in one function. I suspect you got tangled up using the actual `IF NOT EXISTS` SQL, because it doesn't give any feedback w.r.t. whether the schema already existed, or was newly created. This makes the branching for whether to add a comment or not cumbersome.

From another perspective, though, there is only one decision to make: Attempt all the logic (i.e., create the schema and comment), or none of it. All logic should be skipped if both are true:

- The `if_not_exists` flag (is true), and
- the the schema already exists.

In this case, we can't create the schema, don't want to comment on it, and don't want to throw an error. So, we just exit the function.

Now, suppose the opposite. I.e., at least one of the checks above fails. Then we should attempt to create the schema and comment on it, in that order. 

- If the schema already exists, we'll throw an error and exit (before attempting to comment)
- If it doesn't already exist, we'll create it and comment on it.

While this is technically more branching, we don't actually have to implement or think about it. We're just letting nature take its course w.r.t. error handling. We also don't need the `IF NOT EXISTS` SQL clause anymore. Here's an example of what I mean, demonstrated via a function that would be (I think) an improvement:

```sql
CREATE OR REPLACE FUNCTION
msar.create_schema(
  sch_name text,
  description text,
  skip_preexisting boolean DEFAULT false
) RETURNS TEXT AS $$/*
Create a schema with an optional comment, returning its OID.

Args:
  sch_name: The unquoted name of the schema to be created.
  description: a comment to add to the schema, un-(quoted/escaped).
  skip_preexisting: Whether to skip the function logic if the schema already exists.
*/
DECLARE
  schema_oid oid  := to_regnamespace(quote_ident(sch_name));
  schema_preexists boolean := schema_oid IS NOT NULL;
BEGIN
  IF NOT (skip_preexisting AND schema_preexists) THEN
    -- No need for 'IF NOT EXISTS', since we want to throw an error here if the schema exists.
    EXECUTE format('CREATE SCHEMA %I', sch_name);
    schema_oid := to_regnamespace(quote_ident(sch_name));
    PERFORM msar.comment_on_schema(schema_oid, description);
  END IF;
  RETURN schema_oid;
END;
$$ LANGUAGE plpgsql;
```

I've changed `if_not_exists` to `skip_preexisting` since I think it's clearer.

As for why not just have multiple functions:
- The single function puts all branching in one spot where it's visible.
- I think the single function will be more extendable over time (e.g., if we want to create a schema and apply default permissions for it someday).

PR #3623, opened 2024-06-13
Implement `table.import` RPC endpoint

### LINKED ISSUE #3622: Implement `tables.import` RPC method

Anish9901:
Fixes #3622

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/tables/operations/import_.py
```diff
@@ -0,0 +1,92 @@
...
+    file_path,
+    column_names,
+    header,
+    delimiter=None,
+    escape=None,
+    quote=None,
+    encoding=None
+):
+    conversion_encoding, sql_encoding = get_sql_compatible_encoding(encoding)
+    schema_name, table_name = get_fully_qualified_relation_name(table_oid, conn).split('.')
```
mathemancer, line 57: This is brittle and will fail if a schema or table name contains `'.'`.

db/tables/operations/import_.py
```diff
@@ -0,0 +1,92 @@
...
+    formatted_columns = sql.SQL(",").join(sql.Identifier(column_name) for column_name in column_names)
+    copy_sql = sql.SQL(
```
mathemancer, line 59: Why pull apart the qualified table name just to put it back together?

mathemancer (changes requested):
Okay, some parts of this are great:
- The RPC-layer stuff looks good
- I like the handling of string table creation (with caveats).

But, there's some brittleness, mostly around string wrangling of identifier names. I suppose the quickest fix would be just to sort out those strings more carefully (i.e., the table and schema names) and test them.

A better long-term solution would be to make a DB-layer PL/pgSQL function to compose and execute a query to create the string type input table, and then compose and execute the `COPY` query, given just the
- table name,
- schema oid,
- comment,
- column name list (derived from csv headers)
- dialect/encoding info (either separate args or JSONB)

You'd run that function using `Cursor.copy`.
It might be possible to do it all in a single DB call rather than multiple (not sure).

The reason that would help with robustness would be that you'd remove one layer of translation when quote/escape wrangling of both identifiers and dialect information. I'm quite concerned about trying to make sure everything is properly set up in Python before sending it to the DB.

Anish9901:
> You'd run that function using `Cursor.copy`.

That's an interesting idea @mathemancer, I wasn't aware that `Cursor.copy` could be used to execute functions similar to `Cursor.execute`. I'll give it a shot.

mathemancer:
> > You'd run that function using `Cursor.copy`.
> 
> That's an interesting idea @mathemancer, I wasn't aware that `Cursor.copy` could be used to execute functions similar to `Cursor.execute`. I'll give it a shot!

I double-checked since I wasn't sure. You can't actually run multiple operations using `Cursor.copy`, so you'd still have to create the table, then do the copy operation as a separate call. I still think it would be more robust to do the escape/quote wrangling on the DB, but now I'm much less convinced it's worth the effort to move the logic there. Use your judgement w.r.t. whether you want to do the string composition in python (but much more carefully) or on the DB.

mathemancer (changes requested):
This still has quote-/escape-wrangling problems. It's quite complicated. See my specific comments. Overall, I think you should choose one of two methods:

1. Return everything un-quoted/escaped from the DB function, and use `sql.Literal` or `sql.Identifier` for every single literal or identifier in Python to handle the quoting and escaping, or
2. Build the `COPY` statement in the `prepare_table_for_import` function, and return it instead of the current JSON. Then, just use that statement with the `Cursor.copy` call.

Anish9901:
@mathemancer I ended up choosing the 2nd option since that seemed more robust here are some sample returns from the SQL function:

**For a TSV file:**
```py
{
    'copy_sql': 'COPY public."tsv import 2" ("Center", "Status", "Case Number", "Patent Number", "Application SN", "Title", "Patent Expiration Date") FROM STDIN CSV HEADER DELIMITER \'\t\' QUOTE \'"\' ENCODING \'UTF8\'',
    'table_oid': '2162234'
}
```

**For a CSV file:**
```py
{
    'copy_sql': 'COPY public."cSv import 1" ("Integer", "Boolean", "Text", "Decimal") FROM STDIN CSV HEADER DELIMITER \',\' ENCODING \'utf-8\'',
    'table_oid': '2162409'
}
```

mathemancer (approved):
Okay, looks good to me. Thanks for all your hard work on this!

PR #3633, opened 2024-06-18
Quoting refactor

seancolsen:
As discussed during [today's meeting](https://tldv.io/app/meetings/667192c048d77e001377fe04?transcript=true&video=true) (starting at 4:13).

From our hack MD meeting agenda:

> I want to revisit our discussion around quoting SQL values. Here's what I'd like to propose:
> 
> - In the `msar` schema, names in function arguments should be UNQUOTED. (This should already be satisfied.)
> - _And_ in the `msar` schema, names in function _return values_ should be also UNQUOTED. (This is not currently satisfied, but would only require small changes to satisfy.)
> 
> Adopting the second standard above would allow us to proceed incrementally with our quoting refactor. And we could simplify some code as we go. This approach is in line with the way I've been structuring my PRs thus far. Basically we could consider the `__msar` namespace deprecated and we could seek to rewrite functions into the `msar` namespace as we go, fully adopting the escape-on-output pattern within the `msar` schema.
> 
> Previously we agreed that we didn't want to mix patterns. Indeed **I** was vocal about this. But my opinion here has changed after some time working with the code. The realization I've come to is that we _already_ have multiple patterns in some places. And I've noticed bugs due to it. My proposal seeks to slightly move (and clarify) the boundaries around where each pattern is appropriate. Names in arguments and return values would be exclusively quoted within `__msar` functions. And names in arguments and return values would be exclusively _unquoted_ within the `msar` schema. Then, the more we wean ourselves off the `__msar` schema, the more surmountable the quoting refactor would eventually be.
> 
> To bring our SQL into conformance with the second standard above, I would:
> 
> - Fix the following functions which currently return quoted values.
> 
>     - `msar.get_column_name`
>     - `msar.get_column_names`
>     - `msar.get_constraint_name`
>     - `msar.get_relation_name_or_null`
> 
>     For each such function we could either: move it into the `__msar` schema; or we could remove the quoting on the return value (refactoring call sites as needed). We could decide between these two strategies on a per-function basis depending on which strategy appears easiest.
>     
> - Move the following functions into the `__msar` schema.
> 
>     - `msar.get_duplicate_col_defs`
>     - `msar.process_col_def_jsonb`
>     - `msar.process_con_def_jsonb` 
> 
>     This is because they return `__msar.col_def` and `__msar.con_def` types which contain quoted values. These functions are not called from the service layer, so this move seems fine to me. Conveniently, those custom types are already in the `__msar` namespace, as I would expect since they contain quoted values.
> 
> - Rename `msar.build_unique_column_name_unquoted` to `msar.build_unique_column_name`
> 
>     This is because the fact that the column name is unquoted should be evident from the fact that it's in the `msar` schema.
> 
>     It's interesting to note here that this function calls `msar.get_fresh_copy_name`, and it's also is _called by_ `msar.build_unique_fkey_column_name`. Both of those functions already return unquoted names (without making the lack of quoting clear in the function name or documentation). This is great example of the waters being even muddier than I first realized when I raised this topic. Multiple patterns are already here! I'm out to separate the wheat from the chaff.
> 
> - Rename `msar.get_fully_qualified_object_name` to `msar.build_qualified_name_sql`
> 
>     This makes it clearer that the return value is an SQL fragment (quoted) and not an (unquoted) name.
> 
> The above changes are easy enough that I'd like to do them now, in a dedicated PR.

I followed my plan above except for `msar.get_fully_qualified_object_name`. I ended up moving that one into `__msar` because, well, I just don't like it. Grumble grumble. I don't think we actually _need_ to be passing around fully qualified table names as SQL. So I'd like us to refactor that out when we refactor __msar out. We could debate this later. It's a small point. That function doesn't get called from very many places. But I don't like the way it encourages us to set up other functions that accept its output as input. I think we can reverse the order of composition in some cases and achieve a much cleaner result. But all of that is for later.

I also added some developer-facing documentation to the SQL directory which documents the new standard as well as some other things too.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
Ok I also added 779af5e1b2bf07327594b964344a20b9baa4ee41 with one more change. I found myself actually wanting a function `msar.get_relation_name` which returns the unquoted table name (without the schema name). So I changed the existing function name to be clear that it returns a _qualified_ name.

db/sql/00_msar.sql
```diff
@@ -185,35 +130,21 @@ $$ LANGUAGE plpgsql;
...
-*/
-BEGIN
-  RETURN format('%s.%s', sch_name, obj_name);
-END;
-$$ LANGUAGE plpgsql RETURNS NULL ON NULL INPUT;
-
-
-CREATE OR REPLACE FUNCTION
-msar.get_fully_qualified_object_name(sch_name text, obj_name text) RETURNS text AS $$/*
+msar.build_qualified_name_sql(sch_name text, obj_name text) RETURNS text AS $$/*
```
Anish9901, line 133: Since it returns a quoted value, shouldn't this be in the `__msar` namespace instead of `msar`?

db/sql/00_msar.sql
```diff
@@ -282,7 +213,7 @@ $$ LANGUAGE SQL RETURNS NULL ON NULL INPUT;
 
 CREATE OR REPLACE FUNCTION
 msar.get_column_name(rel_id oid, col_id integer) RETURNS text AS $$/*
-Return the name for a given column in a given relation (e.g., table).
+Return the UNQUOTED name for a given column in a given relation (e.g., table).
```
Anish9901, line 216: I think we should move this to `__msar` because we use it much more frequently than similar functions like `get_column_names`, `get_constraint_name`, or `get_relation_name_or_null`. Forgetting to quote it is much more likely.

db/sql/00_msar.sql
seancolsen: I don't have a strong opinion about this. I think `build_qualified_name_sql` could actually fit into `msar` or `__msar`. In 7f09220662bda25727dee32e3a9f643c3c54002c I've moved it to `__msar` though since you suggest that.

- I think it's okay in `msar` because the name of the function makes it clear that it's returning an SQL fragment, not a name. I think it's okay to return SQL fragments from functions as long as it's clear that it's a SQL fragment at not a name. In the case of this fully qualified name, there is no way to return it as a string without the quotes, so it has to be quoted. And in general our `msar` schema should allow for that, since eventually we won't have `__msar` anymore.

    That being said, I would like to avoid passing around fully-qualified names as single strings, just because it is so much _like_ a name that I think if falls into a kind of gray area that I'd like to avoid if possible.

- So with an eye towards removing the `build_qualified_name_sql` function eventually, I think it's fine to place it in `__msar` because that signals that it's deprecated.

I want to be clear: it's not that `__msar` is for "anything containing quoted stuff" and `msar` is for "nothing at all quoted ever". For example, in my [Record Summary PR](https://github.com/mathesar-foundation/mathesar/pull/3561/files), I wrote a function `msar.build_record_summary_query_from_template` which returns an entire SQL select query as a string. That string absolutely must contain quoted values. There's no way around it! But that's okay because there's no way to "unquote" that SQL. _(As a side note I should mention that when I authored that PR, my understanding of the purpose of `__msar` was that it was for private functions not called by the service layer, so I did define some functions within `__msar` which I'll now go back and move to `msar`.)_

db/sql/00_msar.sql
seancolsen: > because we use it much more frequently

This gives me the impression that you'd like to quote names which are frequently used and not quote names which are not frequently used. Am I interpreting your rationale correctly?

If so, I really strongly disapprove of this line of thinking. In the developer-facing documentation within this PR, I wrote:

> **If it _can_ be unquoted, then it _should_ be unquoted.**

I like this rubric because it has low ambiguity and high conformance with industry best-practices concerning security.

I'm okay with functions within `__msar` operating on quoted names because (as of of PR) this schema is effectively deprecated. It only exists as a matter of technical debt. And after our _full_ quoting refactor (sometime next year) it won't exist at all, with its contained functions and types being refactored into `msar` with unquoted names. I don't want to put things in `__msar` for now just because they're convenient or because we might forget to quote things. I'd rather put as much as we can into `msar` (with unquoted names) so that we can work incrementally towards chipping off at this technical debt.

Does this make sense? I'm worried that I'm not understanding something your comment. Happy to chat about this on a call to clarify.

Anish9901: I don't have a strong opinion on this. Initially, I was under the impression that you were open to moving these functions to the `__msar` namespace since you had mentioned this: 

> For each such function we could either: move it into the __msar schema; or we could remove the quoting on the return value (refactoring call sites as needed). We could decide between these two strategies on a per-function basis depending on which strategy appears easiest.

However, I do think I understand your rationale behind keeping it in `msar`, the rationale being adopting the escape-on-output pattern. But, in this specific case, it doesn't seem like we are passing the values of the function(i.e. the column name) around and we are immediately quoting it once we make the call to the function. Therefore, I thought moving it to `__msar` might make sense.

I do have some questions about how we'll go on to remove the `__msar` schema from our codebase, things like what would happen to `__msar.exec_ddl`, what would the future refactor for functions like `__msar.get_column_names`, `__msar.get_relation_name` would look like, etc. but probably this isn't the right place to discuss those.

db/sql/00_msar.sql
```diff
@@ -291,13 +222,13 @@ Args:
...
+SELECT attname::text FROM pg_attribute WHERE attrelid=rel_id AND attnum=col_id;
 $$ LANGUAGE sql RETURNS NULL ON NULL INPUT;
 
 
 CREATE OR REPLACE FUNCTION
 msar.get_column_name(rel_id oid, col_name text) RETURNS text AS $$/*
-Return the name for a given column in a given relation (e.g., table).
+Return the UNQUOTED name for a given column in a given relation (e.g., table).
 
 More precisely, this function returns the quoted name of attributes of any relation appearing in the
```
Anish9901, line 233:
```suggestion
More precisely, this function returns the unquoted name of attributes of any relation appearing in the
```

seancolsen: Thanks! Fixed in a15f6fefbf7d269f5039b16bfa9c1fbb85a63f4e

db/sql/00_msar.sql
```diff
@@ -310,13 +241,13 @@ Args:
   rel_id:  The OID of the relation.
   col_name:  The unquoted name of the column in the relation.
 */
-SELECT quote_ident(attname::text) FROM pg_attribute WHERE attrelid=rel_id AND attname=col_name;
+SELECT attname::text FROM pg_attribute WHERE attrelid=rel_id AND attname=col_name;
```
Anish9901, line 244: We'd probably also need a refactor for qualifying the catalog table with `pg_catalog`

seancolsen: Yeah, I see that as unrelated to this PR though.

Anish9901: Yep, definitely not worth pursuing in this PR.

db/sql/00_msar.sql
seancolsen: > it doesn't seem like we are passing the values of the function(i.e. the column name) around and we are immediately quoting it

Correct. And because this PR removes the quoting from the `get_column_name` return value, it will allow us to incrementally refactor those call sites in order to continue gradually pushing that quoting down deeper towards the SQL queries whenever we're working with that code.

I did move some functions into `__msar` in this PR, but only because it would have been an unreasonable amount of work to change fully unquote their return values. For example, the function `get_duplicate_col_defs` (previously in `msar`, now in `__msar`) returns `__msar.col_def[]`. That _type_ contains quoted values. Ugh! We'll need to refactor that type and all its usages in order for `get_duplicate_col_defs` to fit cleanly into `msar`. That's a lot of work.

And yes, there will be a bunch of other small decisions to make when it comes time to do this whole refactor. I'll be keen to provide guidance on those decisions. For example:

- For `exec_ddl`, I want that function to go away! It should just be inlined. I feel pretty strongly about this. I've already talked to Brent about this. I can explain more if needed when the time comes.
- For `get_column_names`, it's only used in order to construct `con_def` instances, so it will be straightforward to refactor in conjunction with refactoring the type. It's _work_. But it won't require a ton of decision or redesign.

mathemancer (commented):
Looks good to me.

mathemancer (approved):
Previous review was meant to be an approval.

PR #3672, opened 2024-07-10
Improve tables metadata

### LINKED ISSUE #3671: Improve tables metadata
I need some changes to tables metadata in order to continue with the front end implementation of the RPC transition.

I'm open to discussing this, but here's what I'd like:

- Add two new RPC methods: `tables.list_with_metadata` and `tables.get_with_metadata`. These RPC methods should:
    1. Fetch tables data from the user database as the `list` and `get` methods currently do.
    1. Then they should call out to the internal DB to retrieve the metadata for each table being listed.
    1. Each returned table object should then have a `metadata` property containing all metadata values. If no metadata record exists, then the `metadata` property should be `null`.
    
        For example:

        `tables.list_with_metadata`

        ```json
        [
          {
            "oid": 18351,
            "name": "foo",
            "schema": 2200,
            "description": null,
            "metadata": {
              "import_verified": null,
              "column_order": [
                1,
                4,
                2,
                3
              ],
              "record_summary_customized": null,
              "record_summary_template": null,
            }
          },
          {
            "oid": 66125,
            "name": "bar",
            "schema": 2200,
            "description": "hi there",
            "metadata": null
          }
        ]
        ```

- For the `tables.metadata.patch` RPC method:
    - Fix bug where the request fails when no metadata record exists yet. I basically want this to function like `UPSERT` (i.e. `INSERT ... ON CONFLICT DO UPDATE ...`) instead of `UPDATE ...`.
    - Rename the method to `tables.metadata.set`. This is to better reflect the behavior described above.
    - Rename the `metadata_dict` parameter to `metadata`.

        As I side note, I've seen this "dict" terminology a few places in the RPC API and I'd prefer that we avoid it. A dict is a python construct, and might be confusing to people consuming the API in the future. The equivalent construct in JSON is "object". But if we can leave out those words entirely, I think that would be even better, keeping things simpler. I don't want to be too picky about naming at this stage, but I mention this because it would serve to aliveate confusion for naming of the property that gets _updated_ to align with the naming of the property that is _returned_ from the two new RPC methods I've requested above.

- In the `TableMetaData` model:
    - Remove all default values for metadata fields. Here's why I want this change...

        Say I use the RPC API to set a `column_order` value for a table. If no metadata record exists for the table yet, then I don't want the service layer to automatically set values for other metadata fields like `import_verified`. I want those fields to be initialized as `NULL`. With `import_verified` for example, the default is currently `False`. That's bad because it means that if a user re-orders columns in a pre-existing table, the table will be marked as an unverified import.

seancolsen:
Fixes #3671

## Notes

- I didn't implement `tables.get_with_metadata` (as specified in the issue). I think this is fine for now. I'm not 100% sure I'll actually _need_ `tables.get_with_metadata`. I'll implement it later if I do find that I need it.

- In 243823f8ce7be40199b4d137ad6b4268da89ec10 I removed `test_tables_meta_data_set`. This is the type of test that I have advocated for _not writing_ (in a meeting recently). I still maintain this having this sort of test in the codebase does us more harm than good at this stage. It slowed me down while making this PR. I wanted to remove the return value from `set_table_meta_data` and from `tables.metadata.set`. I don't want a return value there because I see no need for API consumers to utilize a return value. I want to keep these things as simple as possible. In removing the return value, the test broke. And I didn't have an easy way to refactor the test to fix the assertion. So I just deleted it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models/base.py
```diff
@@ -138,3 +138,11 @@ class Meta:
...
             )
         ]
+
+
+table_metadata_fields = {
+    'import_verified',
+    'column_order',
+    'record_summary_customized',
+    'record_summary_template',
+}
```
mathemancer, line 148: Not needed. See my other comment.

mathesar/utils/tables.py
```diff
@@ -90,13 +90,9 @@ def get_tables_meta_data(database_id):
...
 
-def patch_table_meta_data(table_oid, metadata_dict, database_id):
-    metadata_model = TableMetaData.objects.get(database__id=database_id, table_oid=table_oid)
-    alterable_fields = (
-        'import_verified', 'column_order', 'record_summary_customized', 'record_summary_template'
+def set_table_meta_data(table_oid, metadata, database_id):
+    TableMetaData.objects.update_or_create(
+        database=Database.objects.get(id=database_id),
+        table_oid=table_oid,
+        defaults={f: v for f, v in metadata.items() if f in table_metadata_fields}
```
mathemancer, line 97: You can just pass the `metadata` dictionary directly. I can think of no cases where the behavior would be different. E.g., if you have extra keys in that dictionary, they'll be silently ignored.

mathemancer (changes requested):
Mostly okay, but I have a change to request:

Please remove the Table Metadata key set from the models file (and its use in the tables utilities file). It's not needed, and will inevitably get out-of-sync with either the model, or the relevant `TypedDict`, or both.

mathesar/utils/tables.py
mathemancer: Apparently it's a bit more subtle than this. I'd still like to simplify this to match a more typical example. Given that the dictionary to be passed is documented and indeed typed, I want to avoid maintaining these fields in so many places.

mathemancer (approved):
Okay, in the interest of speed and breaking things, I've made the change I suggested. @seancolsen , if you're okay with this, go ahead and merge. If not, we can go back and forth on it.

mathesar/rpc/tables/metadata.py
```diff
@@ -77,26 +86,22 @@ def list_(*, database_id: int, **kwargs) -> list[TableMetaData]:
...
-@rpc_method(name="tables.metadata.patch")
+@rpc_method(name="tables.metadata.set")
 @http_basic_auth_login_required
 @handle_rpc_exceptions
-def patch(
-    *, table_oid: int, metadata_dict: SettableTableMetaData, database_id: int, **kwargs
-) -> TableMetaData:
+def set_(
+    *, table_oid: int, metadata: TableMetaDataBlob, database_id: int, **kwargs
+) -> TableMetaDataRecord:
```
Anish9901, line 98:
```suggestion
) -> None:
```

seancolsen: Fixed in 3396114bd98f0898b93943ea5c8c75b3fbc6ee67. Thanks!

mathesar/utils/tables.py
seancolsen: >  I can think of no cases where the behavior would be different. E.g., if you have extra keys in that dictionary, they'll be silently ignored.

Here are some cases:

- For an existing TableMetaData record, pass a `metadata` object like: 

    ```json
    {
      "table_oid": 99
    }
    ```

    The associated table is changed.

- For an existing TableMetaData record, pass a `metadata` object like: 

    ```json
    {
      "created_at": "1999-01-01 00:00:00"
    }
    ```

    The creation timestamp is changed.

- For a nonexistent TableMetaData record, pass a `metadata` object like:

    ```json
    {
      "id": 99
    }
    ```

    This creates a new record with the supplied id, messing up the PostgreSQL sequence, preventing any subsequent records from being created through the normal process

For stability's sake, I would prefer that the API not allow any of the above behavior.

I'd also like to point out that this field validation was already in place before my PR, right [here](https://github.com/mathesar-foundation/mathesar/blob/develop/mathesar/utils/tables.py#L95-L97). All I did was move the list of fields closer to the other code that it mirrors. The reason I moved it is precisely because I share your concern about that code getting out of sync. I actually hemmed and hawed about this for a while during this PR. I tried a played around with a few different approaches, basically kind of hating all of them. I'm finding that there is really an absurd amount of duplication and boilerplate with the patterns we're using. I don't blame us for this. I blame python. For example in `rpc/tables/metadata.py` there are _eight_ occurrences of `record_summary_customized`. To me that's crazy. It's crazy that we can't easily be more expressive about our intent with these fields.

I don't like having an additional list of metadata fields in `mathesar/models/base.py` either (as I originally had in this PR when I opened it). I tried two other approaches to maintaining the field validation while avoiding the need to maintain a list of fields.

- I tried finding a way to do some runtime introspection on the `TableMetaData` class, to get a list of field names set on that class but not its ancestors. And I wanted to then manually ignore fields like `database` and `table_oid`. That got kind of messy and started to bump up against the limits of my immediate knowledge of Python metaprogramming. I imagine this approach is possible. I just didn't want to continue spending time trying to figure out a clean way to do this.

- I also tried moving the `TableMetaDataBlob` TypedDict class lower down in the module graph. I tried putting it in `mathesar/models/base.py`. Then I imported it from there to `mathesar/utils/tables.py` so I could iterate over the fields for the sake of validation. But I couldn't figure out how to appropriate incorporate the _documentation_ of that class into the rendered output in MkDocs. I wrestled with this for quite a while actually. Probably at least 1 hour. I read over the [mkdocstrings docs](https://mkdocstrings.github.io/usage/) trying to figure out a way to do this. I imagine it's possible. But again, I didn't want to spend more time on it.

Ultimately I chose to prioritize application stability (validating the metadata fields) and development velocity (abandoning the above approaches) at the expense of code duplication. I still think that's the right balance to strike.

Given the counter examples I posed at the start of this comment, do you still maintain that we don't need any field validation here?

Do you have any better ideas for reducing code duplication?

Given that the list of fields was already in place before this PR, do you think that this code duplication is important enough to address in _this_ PR?

seancolsen: We discussed this in our technical beta meeting today and decided to take Brent's approach for now but to hopefully come back to this topic later and find a better way to validate RPC request inputs in general.

PR #3711, opened 2024-07-29
Homepage UI

### LINKED ISSUE #3687: Implement changes to home page

pavish:
Fixes #3687 

This PR implements home page changes based on the new permissions designs

Note:
* The "collaborator" count on the card is not displayed. The store for it will be implemented in upcoming PRs and the count will be added on the UI once that is done.
* The colors and layout are based on the existing UI and not the figma designs. This was decided in the design meetings.
* Existing UI for the welcome page and empty states are unchanged. (The designs & colors are slightly different on the Figma file).

Reference: [Figma file](https://www.figma.com/design/ajkaJt5zyd8HUV2do34yx2/Permissions-UX?node-id=0-1&t=lEgvMSUe6mLwyIBn-1)

Screenshots:
* Screenshot 2024-07-31 at 5 03 36 AM
* Screenshot 2024-07-31 at 5 04 18 AM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/api/rpc/databases.ts
```diff
@@ -2,17 +2,100 @@ import { rpcMethodTypeContainer } from '@mathesar/packages/json-rpc-client-build
...
 
-export interface Database {
+export interface DatabaseResponse {
   id: number;
   name: string;
   server_id: Server['id'];
 }
 
+/**
+ * TODO: Figure out a better place to move classes like these.
```
seancolsen, line 12: I like this idea, but I have some different opinions regarding the rules. It would be nice to chat about this on a call.

Also, it's nice seeing this sort of proposal written down, but process-wise I don't like using code comments as a discussion channel. I'd prefer to use a wiki PR or github issue.

mathesar_ui/src/api/rpc/databases.ts
```diff
@@ -2,17 +2,100 @@ import { rpcMethodTypeContainer } from '@mathesar/packages/json-rpc-client-build
...
+ *
+ *  changeProperty(newValue: string): Model {
+ *    // Do not mutate propreties
+ *    this.property = newValue;
+ *    return this;
+ *  }
+ * }
+ * ```
+ */
+export class Database implements DatabaseResponse {
```
seancolsen, line 71: I'd like to put this code somewhere else, and I support putting it in `@mathesar/models/database`. That move doesn't necessarily need to happen in this PR though.

mathesar_ui/src/api/rpc/databases.ts
```diff
@@ -2,17 +2,100 @@ import { rpcMethodTypeContainer } from '@mathesar/packages/json-rpc-client-build
...
+  readonly id: number;
+
+  readonly name: string;
+
+  readonly server_id: number;
+
+  readonly server_host: string;
+
+  readonly server_port: number;
```
seancolsen, line 80: Can we use a single `server: Server` property instead of three separate `server_*` properties?

mathesar_ui/src/api/rpc/databases.ts
```diff
@@ -2,17 +2,100 @@ import { rpcMethodTypeContainer } from '@mathesar/packages/json-rpc-client-build
...
+
+  constructor(databaseResponse: DatabaseResponse, server: Server) {
```
seancolsen, line 82: I'm curious why you chose a class for `Database` (instead of an interface). I'd be inclined to use an interface since they're simpler and we don't have any methods. Maybe you foresee adding methods in the future? With an interface we'd lose the server id validation in the constructor, but that doesn't seem important to me. I don't have a strong opinion here though. Mostly just curious.

mathesar_ui/src/systems/databases/create-database/ConnectExistingDatabase.svelte
```diff
@@ -0,0 +1,108 @@
...
+  export let onCancel: () => void;
+  export let onSuccess: (db: Database) => void;
+
+  const databaseName = requiredField('');
+  const host = requiredField('localhost');
+  const port = requiredField(5432);
+  const role = requiredField('');
+  const password = requiredField('');
+  const installationSchemas = requiredField<InstallationSchema[]>(['internal']);
+  const form = makeForm({ databaseName, installationSchemas });
```
seancolsen, line 34: I think `makeForm` should include all the form fields. That way the "Create Database" button doesn't get enabled until all the fields are filled out. Was there a reason you left some of the fields out of the form?

seancolsen (commented):
I added some inline comments. Also:

- I didn't like the global CSS you had in `ConnectDatabaseModal` so I pushed aa42c5273a842ef2ab01166ca705960d88d70c90 to refactor it out.

    I've said this before, but I'd prefer that we avoid global CSS whenever it's easy to do so. And this is a case where I think it's pretty trivial to avoid.
    
    The changes I pushed _did_ alter the spacing of some things in minor ways, but I can see that the spacing in your version didn't match Figma pixel-for-pixel so I think we're okay. In fact, I'd say that the Figma spacing is actually a bit wonky.

    In contrast, the global CSS in `InstallationSchemaSelector` is fine with me. Ideally we wouldn't have the global CSS there either, but that's harder to change, so I see it as an okay compromise to make for the sake of speed.
    
    I've not been super happy with the abstractions that Svelte gives us to handle styling like in `InstallationSchemaSelector`. I'm curious if it will improve in Svelte 5. In particular I'm wondering if [snippets](https://svelte-5-preview.vercel.app/docs/snippets) will help. For example it would be neat if `CheckboxGroup` could accept an optional snippet for its `ul.options` element. Then we could pass in a snippet containing an entire `ul` element (with a CSS class). If this actually works, then I think it could become an incredibly powerful way to build low-level components which can be easily styled. I'm looking forward to playing with this at some point.

- I also made a few other minor changes in aa42c5273a842ef2ab01166ca705960d88d70c90:

    - I changed the left-most grid column to be `auto` instead of hard-coding it as `100px`. (That way it's more adaptive to other code changes, and not reliant on `px` units, which I prefer to avoid.)

    - I removed superfluous `<div class="icon-holder">` element and  `class="content-holder"` attribute, simplifying the markup.

I'd like to give it one more quick look before we merge.

mathesar_ui/src/systems/databases/create-database/ConnectExistingDatabase.svelte
pavish: Good catch. I missed them while adding the fields. https://github.com/mathesar-foundation/mathesar/pull/3711/commits/e6a70d922c0f6527acf96bb96f60d646824dbc63 fixes this.

mathesar_ui/src/api/rpc/databases.ts
pavish: We could resolve this in the proposal PRs and on call. I've removed the discussion code comments.

mathesar_ui/src/api/rpc/databases.ts
pavish: I'd like to make any change related to this after we figure out the store structures from the proposal PRs.

mathesar_ui/src/api/rpc/databases.ts
pavish: We could resolve this in the proposal PRs and on call.

seancolsen (approved):

PR #3722, opened 2024-08-03
Proposal for tree-like front end stores structure

seancolsen:
This PR is more of an RFC — not actual code to be merged.

I wrote this code up rather quickly to capture some of my thoughts/ideas/opinions/preferences following a 1:1 conversation that @pavish and I had at the end of our [2024-08-01 beta progress check in meeting](https://tldv.io/app/meetings/66ab94cd78de8e0013496793). And I put it in a PR against the actual codebase to show that the types all check out.

@pavish, I invite you to take a look at this ahead of our next call if you like. I hope that much of this will make sense on its own, given what I already expressed in our call. But a good place to begin reading would be the `App.ts` file I added. As I see it, an instance of that class would encapsulate all the shared state in the app, and this would be the _only_ imported store used anywhere (if we were to eventually adopt this pattern completely). That is the top of the "tree". From there it's just stores all the way down.

I hope this code (along with yours too) will help us both point to some concrete examples as we continue our discussion.

My hunch is that you'll have a knee-jerk reaction against this tree pattern, just because it's so different from the other patterns that I've seen you push for. And if that's the case, then I hope you'll give some thought to formulating specific, pragmatic concerns with it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/PROPOSAL/App.ts
+++ b/mathesar_ui/src/stores/PROPOSAL/App.ts
@@ -1,0 +1,56 @@
+import { type Readable, derived } from 'svelte/store';
+
+import {
+  ImmutableMap,
+  WritableMap,
+  collapse,
+  unite,
+} from '@mathesar/component-library';
+
+import type { TODO } from './apiTypes';
+import type { Database } from './Database';
+import type { Server } from './Server';
+
+function buildDatabases(
+  servers: Readable<ImmutableMap<Server['id'], Server>>,
+): Readable<ImmutableMap<Database['id'], Database>> {
+  const unsortedDatabases = collapse(
+    derived(servers, ($servers) =>
+      unite(
+        [...$servers].map(([, server]) =>
+          derived(server.databases.derivedValues(), (d) => [...d]),
+        ),
+      ),
+    ),
+  );
+  const sortedDatabases = derived(unsortedDatabases, (d) =>
+    d.flat().sort((a, b) => a.name.localeCompare(b.name)),
+  );
+  return derived(
+    sortedDatabases,
+    (d) => new ImmutableMap(d.map((database) => [database.id, database])),
+  );
+}
+
+export class App {
+  servers: WritableMap<Server['id'], Server>;
+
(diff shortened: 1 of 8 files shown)


pavish:
@seancolsen 

The proposed design does a good job of modelling data and keeping the store structure data oriented.

I believe there was a misunderstanding in our last call regarding the structure. I like a tree structure and prefer it where required, we're not on opposite terms on that. What I'd like to figure out is state management, primarily retention for items that are not required in the global state. Here are my concerns:

* Items on imported stores are global (I'm using global to denote long lived objects on the app context).
* The proposed design encourages contributors to store state of almost all database objects on the imported store which could lead to stale states, and unnecessary data retention risking memory leaks.
* While we would be identifying and clearing/resetting the data appropriately, I do not think we can expect the same from our contributors, which would add review & maintenance overhead.
* More than preventing bad code, I'd like the patterns to nudge contributors towards being careful of where they're storing data.
* Some top-of-my-head scenarios:
  - The list of databases, schemas, tables & explorations are all required globally for enabling navigation and for preloading data. Storing them on the imported store is good.
  - The list of collaborators is something that is only required in the settings tab on the Database Page. Storing them on the imported store would be pointless unless we need them in other places.
  - Tablular data (columns, constraints, records, joinable columns etc.,) is required only when displaying the table (& probably other larger components/page in the future). Retaining them on the imported store is not quite useful.
  - Should we refetch the tabular data when opening a table page?
    - Ideally, that'd be the case atleast for properties like columns & records inorder to make sure that we're reflecting upto date information.
    - It's best suited to store information like these that get refetched locally when & where they're required than globally.
  - Moving a column to another table would force us to refetch the other table to prevent stale states.

I'd raised a PR for a proposal that separates the context of the UI state & the data model. I've also added some high level thoughts on the PR. Please take a look at it and we could have a call to discuss how to proceed with the store structures.

pavish:
I'm closing this PR since we've come to an agreement on the store structures.

PR #3723, opened 2024-08-05
Fix "No current database" error

### LINKED ISSUE #3720: "No current database" error when moving to home page from schema page
## Description
* Navigate to the schema page (`/http://localhost:8000/db/<db_id>/<schema_id>/` where tables are listed.
* Click on the Mathesar icon in the breadcrumbs (It redirects to `/`).
* Notice an error being thrown on the console:
   ```
  Uncaught (in promise) Error: No current database
    at databases.ts:110:17
    at sync (chunk-46BDKJIK.js?v=a71937be:71:22)
    at Array.<anonymous> (chunk-46BDKJIK.js?v=a71937be:82:9)
    at Object.set (chunk-46BDKJIK.js?v=a71937be:30:35)
    at DatabasesStore.clearCurrentDatabaseId (databases.ts:96:28)
    at handleUnmount (DatabaseRoute.svelte:22:20)
    at run (chunk-WAIAKUYY.js?v=a71937be:19:10)
    at Array.forEach (<anonymous>)
    at run_all (chunk-WAIAKUYY.js?v=a71937be:25:7)
    at destroy_component (chunk-WAIAKUYY.js?v=a71937be:1672:5)
   ```
* The app crashes after that i.e. no other action is possible until a full reload.
* We discussed throwing this error in https://github.com/mathesar-foundation/mathesar/pull/3651, however, we missed testing the scenario of the unmounting of DatabaseRoute during review.

## Some possible causes
* This error is related to [this line](https://github.com/mathesar-foundation/mathesar/blob/2cc122319bb3d1af0504c91124ee9d4fb2bc2ad8/mathesar_ui/src/stores/databases.ts#L86) and [here](https://github.com/mathesar-foundation/mathesar/blob/2cc122319bb3d1af0504c91124ee9d4fb2bc2ad8/mathesar_ui/src/routes/DatabaseRoute.svelte#L22).
* The `DatabaseRoute` is unmounted while visiting `/`, which sets `databasesStore.currentDatabaseId` to `undefined`.
* This in turn results in `databasesStore.currentDatabase` being `undefined`.
* There are two possible causes for this:
* The most possible cause:
   - The `tables` store uses the `currentDatabase` store. The tables store is initialized when visiting the schema page.
   - This keeps the `currentDatabase` store in use.
   - Since the Svelte stores within `tables` store are long living and live on the highest app context, they never get destroyed. This is by design. (Eg., `currentTablesData` derived store).
   - This keeps `currentDatabase` subscribed even when moving to the base `/` route.
* Another less-possible cause:
  - Svelte derived store seems to perform a `sync` action during unsubscribing if the store had been set a value before (i.e. inited), which forcibly syncs the `currentDatabase` store throwing the error since the value is undefined.
    ```
      const unsubscribers = stores_array.map((store, i) => subscribe(store, (value) => {
        values[i] = value;
        pending &= ~(1 << i);
        if (inited) {
          sync();
        }
      }, () => {
        pending |= 1 << i;
      }));
    ```
    Refer [this line](https://github.com/sveltejs/svelte/blob/24aff3032d7af26162fb6a319b07131d21739e9c/src/runtime/store/index.ts#L192) in the svelte repository.

## Expected behavior
* There should be no error.

seancolsen:
Fixes #3720

## Notes

I think I said this on our call, but I'm not surprised that the approach of throwing errors led to a bug. I _had_ tested this approach empirically, but apparently not with sufficient rigor to uncover the bug you reported.

With this the new approach in this PR, I'm not 100% sure that it won't ever lead to bugs. But I've tested it with the reproduction setup from your report, and it appears to be okay.

I figure that if we do end up discovering a bug due to this approach we can determine at that point whether it's worth it to introduce a different pattern, potentially sacrificing the ergonomics I was after for the benefit of type-safety.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish (approved):
I double checked and it turns out the cause is the less expected option (2) `Svelte derived store seems to perform a sync action during unsubscribing if the store had been set a value before (i.e. inited), which forcibly syncs the currentDatabase store throwing the error since the value is undefined.` from [the issue](https://github.com/mathesar-foundation/mathesar/issues/3720).

I'm fine with not worrying about type safety for `currentDatabase` especially since we'll be refactoring this once we have the new store structures.

PR #3751, opened 2024-08-12
Records grouping bugfix

### LINKED ISSUE #3742: Front end can't match groups with records
Previously, when grouping records using the REST API, each group object returned had a `result_indices` field. The front end relied on this field to match groups up with their corresponding result records.

The RPC-based records grouping added in #3721 doesn't seem to provide any way for the front end to match groups to their records.

To minimize implementation time, ideally, we'd use the same structure — a `result_indices` (array of numbers) within each group object.

I can see that the new RPC API _does_ provide a `results_eq`, but that doesn't fully allow the front end to match groups with records. For example if I'm grouping an email column by email domain, then `results_eq` will show only the domain, and the front end won't be able to know what the domain is for the emails within each record.

mathemancer:
Fixes #3742 

This adds a new field to the `grouping` response to collect result indices.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@Anish9901 This PR is dependent on #3749 , and shouldn't be merged until after that one. But, it's ready for review.

mathemancer:
@seancolsen This is probably worth a quick glance for behavior if you have time.

Anish9901 (approved):
Looks good to me!

seancolsen (changes requested):
Structurally, this is good.

However, as far as I can tell, the `result_indices` was using zero-based indexing in the REST API and now is using 1-based indexing in this PR.

From an API consumer's perspective, I think 0-based indexing would be more intuitive. And it's also how the front end already expects this data to be formed. Do you think you could change it to be 0-based?

mathemancer:
> Structurally, this is good.
> 
> However, as far as I can tell, the `result_indices` was using zero-based indexing in the REST API and now is using 1-based indexing in this PR.
> 
> From an API consumer's perspective, I think 0-based indexing would be more intuitive. And it's also how the front end already expects this data to be formed. Do you think you could change it to be 0-based?

Hmmmmmm. I took that to be primary key values. So, that's what we're returning.

mathemancer:
Okay, I made the change. Should be the same as the old way now.

seancolsen:
Corresponding front end PR: #3755

PR #3754, opened 2024-08-13
Records delete bugfix

### LINKED ISSUE #3752: Modify `records.delete` RPC method to permit deleting records via stringified primary key values
From a table with a numeric primary key column, I can delete records via an RPC request like this:

```json5
{
  "jsonrpc": "2.0",
  "method": "records.delete",
  "id": 0,
  "params": {
    "database_id": 1,
    "table_oid": 692359,
    "record_ids":[
      1
    ]
  }
}
```

But, with the REST API, the front end has actually been sending _stringified_ primary key values. So it would be easier if I could send this:

```json5
{
  "jsonrpc": "2.0",
  "method": "records.delete",
  "id": 0,
  "params": {
    "database_id": 1,
    "table_oid": 692359,
    "record_ids":[
      "1"
    ]
  }
}
```

Initially, I thought: _"hmmm, sending pk values in the correct type would be more idiomatic, so why not change this on the front end?"_ But unfortunately this doesn't appear to be an easy change to make. It's possible of course. But it's hard to tell how much time. I spent 30 minutes and just ended up uncovering a bunch of problems fanning out across different modules.

So if this would be easier to change on the backend, I'd rather do that.

mathemancer:
Fixes #3752 

This PR lets the caller submit record IDs to be deleted in any type, and then coerces them to text (along with the actual primary key column) before comparing in the `DELETE`'s `WHERE` clause.

**Technical details**

A side-effect of this PR is that we can no longer support some more exotic primary key column types (e.g., composite types, array types), whereas the previous version would have.

We consider this to be a reasonable compromise at the moment, since we don't really officially support tables with non-integer primary keys anyway. However, because the new behavior is a bit inconsistent (and may be confusing), we should readdress this at some point, and try to come up with a better solution.

Note that we do _not_ coerce the submitted keys to the type of the primary key column, since that conversion could be lossy, and result in deleting more records than intended.

When submitting primary keys to which we apply custom formatting (e.g., datetime types), they should be formatted when submitted.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/00_msar.sql
+++ b/db/sql/00_msar.sql
@@ -3337,7 +3337,9 @@ INSERT INTO msar.expr_templates VALUES
   ('uri_scheme', 'mathesar_types.uri_scheme(%s)'),
   ('uri_authority', 'mathesar_types.uri_authority(%s)'),
   -- Email part getters
-  ('email_domain', 'mathesar_types.email_domain_name(%s)')
+  ('email_domain', 'mathesar_types.email_domain_name(%s)'),
+  -- Data formatter which is sometimes useful in comparison
+  ('format_data', 'msar.format_data(%s)')
 ;
 
 CREATE OR REPLACE FUNCTION msar.build_expr(rel_id oid, tree jsonb) RETURNS text AS $$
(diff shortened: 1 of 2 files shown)


mathemancer:
@Anish9901 This shouldn't be merged until after #3751 .

seancolsen:
Front end counterpart: #3757

seancolsen (approved):
Works for me!

seancolsen:
Even though @Anish9901 hasn't yet reviewed, I'm merging this because I have a bunch of other work that I'd like to stack on top of it so that it's easier for Pavish to review while I'm gone.

If we need to make changes later, I figure we can do those in a follow-up PR.

PR #3769, opened 2024-08-20
Implement `roles.add` RPC endpoint

### LINKED ISSUE #3670: roles.add
* API adds a new role

Request parameters:
```
database_id: Database['id'];
name: string;
login: boolean;
password?: string; // Should be present if login is true
```

Response:
`RoleInfo` class (Refer `mathesar/rpc/roles.py`)

Anish9901:
Fixes #3670 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/rpc/roles.py
```diff
@@ -53,6 +54,10 @@ class RoleInfo(TypedDict):
     description: Optional[str]
     members: Optional[list[RoleMember]]
 
+    @classmethod
+    def from_dict(cls, d):
+        return cls
+
```
mathemancer, line 60: This is broken. You need to instantiate the `cls` on `d`. Given that, you don't actually need this method. If you don't want validation (fair when returning info from the DB), you can just use `RoleInfo(role)` where you call this.

db/sql/00_msar.sql
```diff
@@ -1013,30 +1061,89 @@ $$ LANGUAGE SQL STABLE RETURNS NULL ON NULL INPUT;
...
+  rolename: The name of the role to be created, unquoted.
+  password_: The password for the rolename to set, unquoted.
+  login_: Specify whether the role to be created could login.
+*/
+DECLARE
+  sch_name text;
+  mathesar_schemas text[] := ARRAY['mathesar_types', '__msar', 'msar'];
+BEGIN
+  CASE WHEN login_ THEN
+    PERFORM create_basic_mathesar_user(rolename, password_);
```
mathemancer, line 1132: This needs a namespace to work in all contexts.

db/sql/00_msar.sql
```diff
@@ -1013,30 +1061,89 @@ $$ LANGUAGE SQL STABLE RETURNS NULL ON NULL INPUT;
...
+  ELSE
+    PERFORM __msar.exec_ddl('CREATE ROLE %I', rolename);
+    PERFORM __msar.exec_ddl(
+      'GRANT CREATE, TEMP ON DATABASE %I TO %I',
+      current_database()::text,
+      rolename
```
mathemancer, line 1138: I wouldn't hold up the PR on this, but we've been changing to just using the following pattern instead

```suggestion
    EXECUTE format('CREATE ROLE %I', rolename);
    EXECUTE format(
      'GRANT CREATE, TEMP ON DATABASE %I TO %I',
      current_database()::text,
      rolename
```

db/sql/00_msar.sql
```diff
@@ -948,6 +948,54 @@ FROM (
 $$ LANGUAGE SQL STABLE;
 
 
+CREATE OR REPLACE FUNCTION msar.get_role(rolename text) RETURNS jsonb AS $$/*
```
mathemancer, line 951: We should make it super explicit that this function only returns information about roles which are _direct_ members of the given role. This is in contrast to `pg_has_role(oid, oid, privilege)`. To be honest, I'd rather use something that calls that system function, and just get the direct and indirect members all at once.

mathemancer: After discussing with @pavish , we don't need indirect members where this function is used. With that said, this means the term `"members"` as used here is different from the standard usage in PostgreSQL, so we should make sure that's noted in the documentation.

mathemancer (changes requested):
There are some issues that keep this from working. See line comments for details.

Additionally, please write a couple of SQL tests to make sure the functionality is working (these would have caught one of the issues), and a couple of mocked call-of-the-RPC-function tests for the `roles.add` and `roles.list_` methods (these would have caught the other breaking problem).

db/sql/00_msar.sql
```diff
@@ -948,6 +948,54 @@ FROM (
...
+  'create_role', r.rolcreaterole,
+  'create_db', r.rolcreatedb,
+  'login', r.rolcanlogin,
+  'description', pg_catalog.shobj_description(r.oid, 'pg_authid'),
+  'members', rolemembers.members
+)
+FROM pg_catalog.pg_roles r
+  LEFT OUTER JOIN rolemembers ON r.oid = rolemembers.oid
+WHERE r.rolname = rolename;
+$$ LANGUAGE SQL STABLE;
```
pavish, line 996: We already have a `msar.get_roles` function which has the same query & json response. Can we modify the function arguments to accept an optional filter to add to the where clause and re-use the same instead of duplicating it?

@mathemancer What are your thoughts on the approach?

db/sql/00_msar.sql
pavish, line 1138: The user/role should not have the `CREATE` privilege on the database while creating the role, which both `create_basic_mathesar_user` and the code here grants.

A tangential question:
I was under the impression that we would allow `mathesar_schemas` to be used by `PUBLIC` which will not require us to have to explicitly grant access to each role. Is that not the case? What about roles created outside of Mathesar and configured here? Do those roles not have access to our schemas? (Also a question to @mathemancer)

pavish (changes requested):

Anish9901:
@pavish @mathemancer I found out that creating a new role automatically grants `CONNECT` and `TEMPORARY` privileges to the role, no matter weather the role is login or non-login. I also tried revoking all privileges from the created roles which had no effect.

db/sql/00_msar.sql
Anish9901: Specified in the description in `role_info_table()`

pavish:
@Anish9901 `CONNECT` and `TEMPORARY` are good with me.

mathemancer:
> @pavish @mathemancer I found out that creating a new role automatically grants `CONNECT` and `TEMPORARY` privileges to the role, no matter weather the role is login or non-login. I also tried revoking all privileges from the created roles which had no effect.
> 
> You could also notice this in the sql test in the PR `test_create_role()`

Huh. Well, I looked into this in the docs and found [this](https://www.postgresql.org/docs/16/ddl-priv.html):

> PostgreSQL grants privileges on some types of objects to PUBLIC by default when the objects are created. No privileges are granted to PUBLIC by default on tables, table columns, sequences, foreign data wrappers, foreign servers, large objects, schemas, tablespaces, or configuration parameters. For other types of objects, the default privileges granted to PUBLIC are as follows: CONNECT and TEMPORARY (create temporary tables) privileges for databases; EXECUTE privilege for functions and procedures; and USAGE privilege for languages and data types (including domains). The object owner can, of course, REVOKE both default and expressly granted privileges. (For maximum security, issue the REVOKE in the same transaction that creates the object; then there is no window in which another user can use the object.) Also, these default privilege settings can be overridden using the [ALTER DEFAULT PRIVILEGES](https://www.postgresql.org/docs/16/sql-alterdefaultprivileges.html) command.

@pavish What are the UI implications (if any) of this? I think we should avoid altering those default privileges unless we really need to (even on internal DBs) and we should definitely not do that on non-internal DBs.

pavish:
@mathemancer 

> What are the UI implications (if any) of this? I think we should avoid altering those default privileges unless we really need to (even on internal DBs) and we should definitely not do that on non-internal DBs.

The UI would not be affected by this. We currently do not support altering default privileges for objects and have no intentions of doing that later.

mathemancer (approved):
Okay, LGTM.

db/sql/00_msar.sql
mathemancer: I think this would be the right thing to do. We'd have to add default privileges to them.

pavish (approved):
Looks good to me!

PR #3798, opened 2024-08-28
Modals and Forms Visual Updates

ghislaineguerin:
### Modals Update

I've been improving modals across Mathesar to ensure consistency, better usability, and adherence to design guidelines. Here's a summary of the changes and improvements:

#### Alignment and Spacing Issues
- **Modal Headers:** The prominence of modal headers has been increased to ensure balance with the footers, as there was insufficient visual weight at the top.
- image
- The modal title size is now consistent and larger than content headings. This issue was especially noticeable in the Constraints modal, where the headers inside the modal were larger than the modal header.
image
- **Close Button Alignment:** The close button is now correctly aligned with the start of the title in all modals.
- image
- **Collapsing Headers:** In Constraint modals, collapsing headers were adjusted to maintain a consistent height, regardless of whether a button is present.
- image

#### Styling and Contrast Issues
- **Form Field Labels:** Improved the contrast of form field labels to distinguish them clearly from body text.
- image
- **Message Boxes:** Updated the style of message boxes to have a less boxy appearance, improving the flow of content and reducing vertical separations that make content harder to scan.
- image
- **Inline Forms and Descriptive Button Labels:** Inline forms were used to add foreign or unique keys, and redundant titles were removed. Instead, the save button labels were updated to indicate the action, such as adding a new object.
- image

#### Consistency Enhancements
- **Input Box Style:** Ensured consistency in using boxed input groups, applying the same style across different modals, such as import modals and database creation modals.
    - image
- **Modal Descriptions:** Standardized the placement and presentation of descriptions.xw

### Create Link Updates

The "Create Link" modal has been revised to enhance clarity and maintain consistency with other forms in the app.

- **Outcome Box Update**: Each outcome now has its own collapsible section containing a single, concise outcome statement. Previously, the outcome box listed all the input boxes for naming every element involved, which cluttered the interface. If the user agrees with the default outcomes, they won't need to interact with these input boxes. This change reduces visual distractions and makes it easier to review the outcomes which was the main goal for this special message box.
- **Updated Labels**: Labels have been revised to follow the new naming convention for new elements. For example, fields now use phrasing like "Name of new column" instead of "New column name".
- image
- image

## Guidelines
https://hackmd.io/0MNSBOSoRUC0cnYEqMj9eA?edit


mathesar_ui/src/systems/table-view/constraints/NewFkConstraint.svelte
```diff
@@ -190,7 +194,7 @@
     onProceed={handleSave}
     onCancel={onClose}
     size="small"
-    proceedButton={{ label: $_('add') }}
+    proceedButton={{ label: $_('Add Foreign Key Constraint') }}
```
seancolsen, line 197: i18n keys should be in snake_case

mathesar_ui/src/systems/table-view/table-inspector/column/column-extraction/ExtractColumnsModal.svelte
```diff
@@ -300,10 +300,10 @@
           text={targetTableName
             ? $_('columns_removed_from_table_added_to_target', {
                 values: { count: $columns.length },
-              })
+              }) + ' A link column will be created to maintain the relationship between records.'
```
seancolsen, line 303: When you finish putting all your strings into dict.json, make sure you don't have any post-translation string concatenation like this.

mathesar_ui/src/App.svelte
```diff
@@ -141,6 +144,10 @@
     height: 100vh;
   }
 
+  p {
+    line-height: 1.5;
+  }
```
seancolsen, line 149: This seems a little odd to me. I'm fine with changing the line height. But I wouldn't expect us to customize it specifically for paragraphs. We don't even use `<p>` very often in Mathesar. So I'm worried that this CSS will lead to inconsistent style.

Can you explain why you set this here?

I'd typically expect to see `line-height` set on a very high-level element like `html` and then for it to sometimes be overridden for elements like headings.

seancolsen (commented):
Looks good to me so far! I added a couple small comments.

mathesar_ui/src/systems/table-view/constraints/NewFkConstraint.svelte
ghislaineguerin: I will update, thanks.

mathesar_ui/src/App.svelte
ghislaineguerin: The blocks of text were too tight and I was concerned about making this a global change in case it affected buttons or other elements like the pills. I will consider your feedback and try adjusting it globally.

kgodey (commented):
(I only looked at the PR description, not the code.)

This looks great @ghislaineguerin, very nice work!  Just one small piece of feedback.

366027524-64007861-a1c9-49c8-bc11-fa57eee93d06

mathesar_ui/src/systems/schemas/AddEditSchemaModal.svelte
```diff
@@ -60,12 +60,13 @@
   getInitialName={() => schema?.name ?? ''}
   getInitialDescription={() => schema?.description ?? ''}
   saveButtonLabel={schema ? $_('save') : $_('create_new_schema')}
+  namePlaceholder="Eg. Personal Finances, Movies"
```
seancolsen, line 63: This new string needs to be translated

mathesar_ui/src/systems/table-view/constraints/NewUniqueConstraint.svelte
```diff
@@ -174,7 +176,7 @@
   <CancelOrProceedButtonPair
     onProceed={handleSave}
     onCancel={handleCancel}
-    proceedButton={{ label: $_('add') }}
+    proceedButton={{ label: $_('Add Unique Constraint') }}
```
seancolsen, line 179: This translation key should be in snake_case

mathesar_ui/src/systems/table-view/link-table/LinkTableForm.svelte
```diff
@@ -263,49 +267,50 @@
...
-              <NewColumn
-                base={{ name: $mappingTableName }}
-                baseWhich="mapping"
-                {target}
-                field={columnNameMappingToTarget}
-              />
-            {/if}
+          <p>
+            <RichText
+              text={'We will create a new table [mappingTable].'}
```
seancolsen, line 272: This text needs to be translated

mathesar_ui/src/systems/table-view/link-table/LinkTableForm.svelte
```diff
@@ -263,49 +267,50 @@
...
+              {#if slotName === 'mappingTable'}
+                <Pill table={{ name: $mappingTableName }} which="mapping" />
+              {:else if slotName === 'targetTable'}
+                <Pill table={target} which="target" />
+              {:else if slotName === 'baseTable'}
+                <Pill table={base} which="base" />
+              {/if}
+            </RichText>
+          </p>
+          <Field field={mappingTableName} label={$_('Name of New Table')}>
```
seancolsen, line 284: This translation key should be in snake_case

mathesar_ui/src/systems/table-view/link-table/LinkTableForm.svelte
```diff
@@ -263,49 +267,50 @@
...
+            </span>
+          </Field>
+          {#if $mappingTableName}
+            <NewColumn
+              base={{ name: $mappingTableName }}
+              baseWhich="mapping"
+              target={base}
+              targetWhich="base"
+              field={columnNameMappingToBase}
+              label={$_('Name of New Column {number}', {
```
seancolsen, line 300: This translation key should be in snake_case

mathesar_ui/src/systems/table-view/link-table/LinkTableForm.svelte
```diff
@@ -263,49 +267,50 @@
...
+                values: { number: 1 },
+              })}
+            />
+            <NewColumn
+              base={{ name: $mappingTableName }}
+              baseWhich="mapping"
+              target={isSelfReferential ? base : target}
+              targetWhich={isSelfReferential ? 'base' : 'target'}
+              field={columnNameMappingToTarget}
+              label={$_('Name of New Column {number}', {
```
seancolsen, line 310: This translation key should be in snake_case

mathesar_ui/src/systems/table-view/link-table/NewColumn.svelte
```diff
@@ -9,7 +9,7 @@
 
   import Pill from './LinkTablePill.svelte';
 
-  const label = 'Column Name';
+  export let label = 'Name of New Column';
```
seancolsen, line 12: This string should be translated. I recognize that it wasn't translated before. But it would be good to make that change here. If there's a good reason for _not_ translating it, then we ought to add a code comment explaining why.

seancolsen (commented):
I've not thoroughly review the changes from a stylistic or aesthetic perspective because I don't expect to have strong opinions about that and I trust Ghislaine's judgement here.

But I did add some critique on the code pointing out some i18n issues.

seancolsen:
Heads up that in c23d6730aaa5657fa3dce97f1ffb184e799c19ae I resolved some merge conflicts with `Collapsible.scss` that seemed straightforward to me. I wanted to get the code updated in the branch in order to merge into the dependent PR #3805 so that I could test it.

seancolsen:
Some more things I noticed:

- This UI was more concise before:

    | Before | After |
    | -- | -- |
    | image | image |

    I think the simpler version is clearer and easier to read.

- Some UI element within the Create Link dialog is not responding well to narrow viewports, causing overflow. 

    

    Prior to these changes, this UI shrunk to narrow viewports quite well.

    

    Not that I think this is particularly important right now. But it would be good to trace the root cause of this change since it's possible that it's a low-level change that could have other consequences as well.

ghislaineguerin:
@seancolsen 
I agree with your review of the outcome box in the Create Link modal. My initial idea was to consolidate the actual outcomes at the top so that the user is not distracted by all the visual clutter. However, adding the help text to the input boxes didn't really accomplish that. I've been thinking more, and I've pushed a new version that aims to make the user goal of "knowing what changes" more efficient. I think we should aim to show outcomes that we feel are reasonable, and if the user wants to change them, they can interact. Otherwise, the main goal is fulfilled for most.

seancolsen:
@ghislaineguerin cool. Is this ready for re-review then?

kgodey:
Has my review comment also been addressed?

ghislaineguerin:
@seancolsen Yes, it is ready now. 

@kgodey Yes, It is solved because I kept the inline layout for the new column input. I've moved away from using the help descriptions inside outcomes as it fractures the outcome text. The label is now clearly differentiated from the content. See below.

image

kgodey:
Looks good @ghislaineguerin but are we losing the colors for the table names?

ghislaineguerin:
@kgodey we only had colors in the context for the Create Link dialog.

kgodey:
Ah right, sorry @ghislaineguerin, wasn't paying enough attention.

ghislaineguerin:
Closing this PR since we decided to consolidate all changes into one:. See conversation here: https://matrix.to/#/!UnujZDUxGuMrYdvgTU:matrix.mathesar.org/$yxAeaL_S4E1_HgmnpeZadGTOG9cTKzWOsFeRRkIE0XI?via=matrix.mathesar.org&via=matrix.org

PR #3808, opened 2024-09-06
Add summaries to self

mathemancer:
Fixes #3806 

This PR adds summaries for records retrieved, patched, or added when the parameter `return_record_summaries` is set to `True`.

**Technical details**

It also simplifies the record patching and adding SQL functions at the cost of a (very minor) performance hit.  I added and modified tests to make sure things were wired up, but only in a basic way due to time constraints.

## Checklist
 
- [X] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [X] My pull request targets the `develop` branch of the repository
- [X] My commit messages follow [best practices][best_practices].
- [X] My code follows the established code style of the repository.
- [X] I added tests for the changes I made (if applicable).
- [X] I added or updated documentation (if applicable).
- [X] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3802, opened 2024-08-30
Change response structure for record summary

Anish9901:
Fixes #3801

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3811, opened 2024-09-10
Propagate RPC record summary changes to front end

seancolsen:
## Notes

- The corresponding back end PRs are:
    - #3761
    - #3802
    - #3808

- This PR removes quite a bit of code because previously record summaries were being rendered on the front end in several places by taking the fields + the template to produce a string. Now the architecture is _much_ simpler, doing all the rendering on the backend.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/stores/table-data/tabularData.ts
```diff
@@ -39,9 +39,19 @@ export interface TabularDataProps {
...
     typeof processColumn
   >[0]['hasEnhancedPrimaryKeyCell'];
+  /**
+   * When true, load the record summaries associated directly with each record
+   * in the table. These are *not* the record summaries associated with linked
+   * records. Instead they are the summaries of the records themselves. By
+   * default, we don't load these summaries because it's a performance hit. But
+   * we need them for the records within the record selector.
+   */
+  loadIntrinsicRecordSummaries?: boolean;
```
pavish, line 49: @seancolsen The only place we seem to require intrinsic record summaries is when selecting a record via the record selector which updates a cell in the parent table. However, the RPC method `records.patch` already returns the required record summary within `linked_record_summaries`. Could we use that directly instead of requesting the intrinsic record summaries?

It is possible that I may not be seeing additional places/cases where this is required. If so, could you specify where/how we need them?

pavish (dismissed):

mathesar_ui/src/stores/table-data/tabularData.ts
pavish: Ah.. I see that we close the Record selector immediately after picking the record, and the summary is required while the parent record is being saved.

pavish (approved):
Looks good!

PR #3879, opened 2024-09-25
Fix updating of table name

### LINKED ISSUE #3867: Intermittent failure to display updated table name
I'm observing a tricky bug when editing a table name via the table inspector. The name is consistently updated in PostgreSQL. But in the UI sometimes new name is updated and sometimes not. I can't seem to identify any correlation between failure and specific reproduction steps. It's also difficult because it doesn't seem to be perfectly random either. For example, I had it succeed 25 times in a row. Then refreshed the page and had it fail 25 times in a row. I've also seen it succeed on the first try and fail on the second. And I've seen it fail on the first try too. My hunch is that there's some sort of race condition in the front end code.

One more clue: I've noticed that sometimes the entire table refreshes after I update its name — as in, the _records_ reload too. And I _think_ it's those times when the renaming succeeds. I'm not positive about this though.

seancolsen:
Fixes #3867

- This was a really interesting bug!

- The bug _seemed_ to be intermittent, but I just wasn't paying close enough attention to my reproduction steps.

    After observing some unexplained variance in behavior I decided to hammer at it a bit in an attempt to see how often it was failing vs succeeding. I started out with a table named `a` and proceeded to rename it `b` and `c` and so on. All actions succeeded. When I got to `z`, I refreshed the page and started over. (I _should_ have renamed the table _back to `a`_ before refreshing the page, but that was an oversight.) So when I started with the table named `z` after refreshing the page, renaming the table to `a` failed. And renaming it to `b` and `c` and so on also failed. This was very strange to me. So many sequential successes followed by so many sequential failures!

- Now that I understand the cause of the bug, I can explain its behavior in deterministic terms: renaming a table would succeed when the new name was sorted after the old name. Otherwise it would fail. I find this hilarious!

- The cause of the bug comes back to how we've been dealing with sorting tables. For a while we've had some code comments atop the file warning readers that this module is kind of a mess. I had written a comment:

    > I suggest having a derived store that does the sorting.

    I think that approach would have helped avoid this bug.

- In fixing this bug I considered using `SortedImmutableMap`, but I don't think that would really help here. Consider this:

    ```ts
    new SortedImmutableMap(
      (v) => [...v].sort(([, a], [, b]) => a.localeCompare(b)),
      [
        [1, 'x'],
        [2, 'y'],
        [1, 'a']
      ]
    );
    ```

    That gives a map like: `{ 1 => "x", 2 => "y" }`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/tables.ts
+++ b/mathesar_ui/src/stores/tables.ts
@@ -237,11 +237,14 @@ function updateTableStoreIfPresent({
   });
   const tablesStore = tablesStores.get([schema.database.id, schema.oid]);
   if (tablesStore) {
-    tablesStore.update((tablesData) => {
-      const tables = sortTables([...tablesData.tablesMap.values(), fullTable]);
+    tablesStore.update(({ requestStatus, tablesMap }) => {
+      tablesMap.set(fullTable.oid, fullTable);
+      const newTablesMap = new Map(
+        sortTables(tablesMap.values()).map((t) => [t.oid, t]),
+      );
       return {
-        ...tablesData,
-        tablesMap: new Map(tables.map((t) => [t.oid, t])),
+        requestStatus,
+        tablesMap: newTablesMap,
       };
     });
     schema.setTableCount(get(tablesStore).tablesMap.size);


pavish (approved):
Nice!

kgodey:
Neat.

PR #3882, opened 2024-09-25
Associate explorations with schemas at API layer

### LINKED ISSUE #3880: Modify API to make explorations aware of their associated schema
## Context

1. In the UI, each exploration is associated with one schema by virtue of being presented "within" a schema.
1. In Mathesar 0.1.7, that association was handled by virtue of each `Exploration` referring to a `Table` which then referred to a `Schema`. All of those associations were contained within the internal database.
1. Mathesar 0.2 onwards no longer has that `Table` model. Each exploration weakly references a table oid in the user database. And we need to look inside the user database to see what schema that table lives within.
1. The change in data model presents a small problem, given our current UX.
1. At some meeting a few months ago we had decided to solve this problem within the front end. The front end would fetch all the explorations for a database and then filter them out.
1. However, now that I'm in the code I can see that taking that approach will require significant front end refactoring. Pavish and I had a [call](https://tldv.io/app/meetings/66f4095a1812d600131d14e2) about this today and decided that it would be best to solve this problem in the service layer instead of in the front end.

## Changes needed

Here's what we'd like:

- `explorations.list`

    - Add an optional `schema_oid` parameter to filter the list of explorations by schema.

    - Modify return value to include a `schema_oid` property.

- `explorations.get`

    - Modify return value to include a `schema_oid` property.

- `explorations.add`

    - Modify return value to include a `schema_oid` property.

To handle these changes, the service layer would need to call out to the user database for these two RPC methods.

seancolsen:
Fixes #3880

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/rpc/explorations.py
+++ b/mathesar/rpc/explorations.py
@@ -6,6 +6,8 @@ from typing import Optional, TypedDict
 from modernrpc.core import rpc_method, REQUEST_KEY
 from modernrpc.auth.basic import http_basic_auth_login_required
 
+from db.connection import exec_msar_func
+
 from mathesar.models.base import Explorations
 from mathesar.rpc.exceptions.handlers import handle_rpc_exceptions
 from mathesar.rpc.utils import connect
@@ -29,6 +31,7 @@ class ExplorationInfo(TypedDict):
         database_id: The Django id of the database containing the exploration.
         name: The name of the exploration.
         base_table_oid: The OID of the base table of the exploration on the database.
+        schema_oid: The OID of the schema containing the base table of the exploration.
         initial_columns: A list describing the columns to be included in the exploration.
         transformations: A list describing the transformations to be made on the included columns.
         display_options: A list describing metadata for the columns in the explorations.
@@ -39,6 +42,7 @@ class ExplorationInfo(TypedDict):
     database_id: int
     name: str
     base_table_oid: int
+    schema_oid: int
     initial_columns: list
     transformations: Optional[list]
     display_options: Optional[list]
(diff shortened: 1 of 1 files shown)


seancolsen:
@Anish9901 this is stacked on #3855. Please review but don't merge if that PR is still open.

mathesar/rpc/explorations.py
```diff
@@ -6,6 +6,8 @@
 from modernrpc.core import rpc_method, REQUEST_KEY
 from modernrpc.auth.basic import http_basic_auth_login_required
 
+from db.connection import exec_msar_func
+
```
Anish9901, line 10: We never import `exec_msar_func` in the endpoint definition files. Please create appropriate function under the `db/` namespace and use them here.

Anish9901 (changes requested):
- I think it would be better if we stored `schema_oid` in the `Explorations` model instead of requesting the `schema_oid` everytime, especially since we don't allow moving tables within schemas through mathesar. This would also make listing and filtering much easier.

- If you do feel strongly about the way you've implemented it then, I'd prefer to getting the schema_oid in the `from_model` method for `ExplorationInfo` instead, you'd do this by 
   - Defining `get_relation_namespace_oid` function in `db/tables/operations/select.py`.
    - Altering `from_model` to take `conn` as an argument.
    - Calling `get_relation_namespace_oid` defined in select.py for the `schema_oid` field.

seancolsen:
Abandoning in favor of #3892

PR #3897, opened 2024-09-30
Reduces frontend caching, fixes a few bugs

pavish:
Fixes #3864
Fixes #3877
Reduces frontend caching: Only caches current schema, tables, and explorations
Fixes duplicate requests for `databases.get`
Fixes common data: Null scenarios, removes abstract types
Fixes some minor translation issues (see commits)

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3909, opened 2024-10-01
Fix bug when updating table twice

### LINKED ISSUE #3867: Intermittent failure to display updated table name

### LINKED ISSUE #3908: Unable to drag columns after dragging one column
## Description
* Open a table
* Drag and move a column. Notice that it moves to the new location -> Good.
* Try to drag and move another column. Notice that nothing happens -> Bad.
* See video:

https://github.com/user-attachments/assets/76ff984e-e741-4eb0-95ee-6164a959d49f

## Expected behavior
* We should be able to drag and move columns.

seancolsen:
Fixes #3867
Fixes #3908

## Notes

- I think this bug was introduced in #3897
- When reviewing that PR, the optional storage of `databaseId` and `schemaOid` within `TablesData` seemed a little weird to me, but I just went with it, especially since the PR seemed to be just in incremental refactor of an already messy module.

- At some point when we have time, I'd like to see:

    - A `TablesStore` class which manages the storage of multiple tables within one schema. It would have:

        - A non-reactive back-reference to its containing `Schema` object.
        - A reactive map of tables.
        - Public methods for CRUD on tables.

    - A `SchemaRouteContext` class (similar to `DatabaseRouteContext`) which holds a `TablesStore` instance.

    That way, any time you have a `SchemaRouteContext`, you get everything you need. And when you don't have a `SchemaRouteContext` you don't get _any_ of it. And all the tables CRUD would be nicely encapsulated into the `TablesStore` class.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/tables.ts
+++ b/mathesar_ui/src/stores/tables.ts
@@ -203,13 +203,13 @@ function putTableInStore({
     $tablesStore.databaseId === schema.database.id &&
     $tablesStore.schemaOid === schema.oid
   ) {
-    tablesStore.update(({ requestStatus, tablesMap }) => {
-      tablesMap.set(fullTable.oid, fullTable);
+    tablesStore.update((tablesData) => {
+      tablesData.tablesMap.set(fullTable.oid, fullTable);
       const newTablesMap = new Map(
-        sortTables(tablesMap.values()).map((t) => [t.oid, t]),
+        sortTables(tablesData.tablesMap.values()).map((t) => [t.oid, t]),
       );
       return {
-        requestStatus,
+        ...tablesData,
         tablesMap: newTablesMap,
       };
     });


pavish (approved):
@seancolsen Nice! Looks good!

> any time you have a SchemaRouteContext, you get everything you need. And when you don't have a SchemaRouteContext you don't get any of it

Yes, this is precisely how I had intended to structure the route contexts and list of objects they need. The `DatabaseRouteContext` will contain the `SchemasStore`, and the `SchemaRouteContext` will contain the `TablesStore` and the `ExplorationsStore`. This way we'll no longer have a need for imported stores.

For navigation purposes on the header, I'd expect the Route level components to set the stores in the breadcrumbs.

I decided to not go through with these changes in the recent model refactoring since we use the `schemas`, `currentSchema`, `tables`, and `currentTables` stores in several places and I didn't want to increase scope.

I think we can discuss this again when we have the bandwidth to make the changes.

PR #3968, opened 2024-10-16
Fix browser console warnings

### LINKED ISSUE #3929: Browser console warnings in production: No listeners found when dispatching event 'focus'.
## Description
* Open the developer tools, and then open a table.
* Notice these warnings:
* Screenshot 2024-10-04 at 5 04 51 PM
* These should not be thrown in production.
* We need to **identify why a custom event is thrown and there are no listeners to perform actions on it**.

seancolsen:
Fixes #3929

## Notes

- This PR removes the code that throws console warnings. I don't think we need the warnings. I think it's okay if an event gets called on an EventHandler when not listeners are attached.

- The EventHandler subclass is defined in `SheetSelectionStore.ts`, and its event is (expectedly) called `'focus'`. A `SheetSelectionStore` instance dispatches this event to pass a message to its containing scope that the active cell should be imperatively focused via browser DOM APIs.

- The event dispatch code is in `SheetSelectionStore.ts`

    ```ts
    this.selection = new PreventableEffectsStore(new SheetSelection(), {
      focus: () => this.focus(), // which calls `void this.dispatch('focus');`
    });
    ```

    Due to the way that `PreventableEffectsStore` works, the focus event gets dispatched (almost) whenever `this.selection` changes.
    
    (I say "almost" because it's also possible to update the selection while _preventing_ that effect, which is necessary in a couple places and probably not relevant to explain here.)

- The event handler function is in `Sheet.svelte`:

    ```ts
    onMount(() =>
      selection?.on('focus', async () => {
        if (!sheetElement) return;
        await tick();
        focusActiveCell(sheetElement);
      }),
    );
    ```

    Without that event handler code, the active cell would not get focused when the selection changes via the keyboard arrow keys.

- The reason the event gets dispatched before it has any listeners/handlers is that we're holding a `SheetSelectionStore` instance higher up in the component tree and passing it down into `Sheet`. So the selection ends up getting updated before the `Sheet` component is mounted. I think this is fine and we can safely ignore it.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/common/utils/EventHandler.ts
+++ b/mathesar_ui/src/component-library/common/utils/EventHandler.ts
@@ -40,13 +40,7 @@ export default class EventHandler<Events extends Record<string, unknown>> {
     value?: Events[EventName],
   ): Promise<void> {
     const callbacks = this.listeners.get(eventName);
-    if (!callbacks) {
-      // eslint-disable-next-line no-console
-      console.warn(
-        `No listeners found when dispatching event '${String(eventName)}'.`,
-      );
-      return;
-    }
+    if (!callbacks) return;
     await Promise.all(
       [...callbacks.values()].map((callback) => callback(value)),
     );


pavish (approved):
Sounds good w.r.t removing the warning.

PR #3983, opened 2024-10-23
Drop all msar functions at start of install SQL

seancolsen:
This PR modifies the SQL install process to ensure that all functions within the `msar` schema are dropped at the start. It handles all types of functions and all overloads. This ended up actually being a bit tricky to write. I tried a number of dead-end approaches that failed to handle various edge cases before settling on this (admittedly rather simple) code that seems to work.

I discussed this on a call with @mathemancer and he was interested in this idea. I'm opening a dedicated PR for this, and I'll be basing other work on top of it so that I don't need to write a bunch of DROP FUNCTION statements as I change parameter structure of existing functions.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


Anish9901:
@seancolsen Out of curiosity, why did you decide against dropping the `msar` schema instead?

seancolsen:
> why did you decide against dropping the `msar` schema instead?

Right. Very good question!! There's a good deal of nuance here...

I had been holding out hope that we'd end up dropping the whole schema at the start of the install step. It sounds like perhaps you were too. Golly it would simplify a lot things! I'm still open to that approach, and it could be worth considering in earnest.

My initial motivation (months ago) for considering the "drop all functions" approach was that Brent had expressed some concern about dropping msar because (in theory) the user could have things like msar types as column types — in which case dropping msar with CASCADE would delete their data. Personally, I'm 0% concerned about that scenario. I'm also 0% concerned that a user would have an msar function used within a trigger on their user data (which could also pose similar risks). But I'm very motivated to find a way to stop bumping into these dastardly function overloads (_overlords?_) that tend to linger in my dev DB and bork things up. And if Brent doesn't want to drop msar, then I want to find a middle ground that will speed up development without incurring his skepticism.

Additionally, in the course of my work on record summary templates, Brent had some strong opinions that the user's record summary metadata should live inside msar (instead of in the internal database). In theory, I really like the idea of putting user metadata inside msar. I think I'd wholeheartedly lean toward doing that from the start, were we to begin building Mathesar afresh today. So it was with an eye towards **eventually allowing the storage of user inside msar** that I chose to drop all functions instead of drop the schema. 

Incidentally, once I began mentally walking myself down the road of putting record summary template metadata inside msar, I began to identify a number of bumps along that road. We'd need to figure out permissions on metadata and migrations of metadata. Those are solvable problems, but it's not like they're entirely straightforward. I made a case to Brent that they're problems we should spend time trying to solve _right now_. I'm interested in potentially moving _all_ the metadata into the user database at some point, but doing it in conjunction with a sufficiently robust permissions and migrations functionality. Such functionality might even require additional user-facing permissions features.

So I ended up (tentatively) convincing Brent to entertain an approach to record summary templates where they are still stored in Django and just passed into the DB layer via JSON args. But Brent's point remains — storing metadata in msar might be worth pursuing, perhaps even soon. So I didn't want to close the door to that approach by dropping msar. In theory we could use a different schema like `msar_metadata` or so. But personally, I'd be much more inclined to _reduce_ the number of Mathesar specific schemas, not increase that number.

mathemancer (approved):
I'll go ahead and merge this since you seem to need it for your current work, @seancolsen . I'll make a PR with a more full-featured versions where we can discuss further.

PR #4015, opened 2024-11-05
Use mocked `exec_msar_func` calls for RPC tests

Anish9901:
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/tests/rpc/columns/test_c_base.py
+++ b/mathesar/tests/rpc/columns/test_c_base.py
@@ -4,14 +4,17 @@ This file tests the column RPC functions.
 Fixtures:
     rf(pytest-django): Provides mocked `Request` objects.
     monkeypatch(pytest): Lets you monkeypatch an object for testing.
+    mocked_exec_msar_func(mathesar/tests/conftest.py): Lets you patch the exec_msar_func() for testing.
 """
+import json
 from contextlib import contextmanager
 
+from db.columns import _transform_column_alter_dict, _transform_column_create_dict
 from mathesar.rpc import columns
 from mathesar.models.users import User
 
 
-def test_columns_list(rf, monkeypatch):
+def test_columns_list(rf, monkeypatch, mocked_exec_msar_func):
     request = rf.post('/api/rpc/v0/', data={})
     request.user = User(username='alice', password='pass1234')
     table_oid = 23457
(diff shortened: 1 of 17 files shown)


mathesar/tests/rpc/columns/test_c_base.py
```diff
@@ -144,23 +100,24 @@ def mock_connect(_database_id, user):
...
+    mocked_exec_msar_func.fetchone.return_value = [1]
     actual_result = columns.patch(
         column_data_list=column_data_list,
         table_oid=table_oid,
         database_id=database_id,
         request=request
     )
+    call_args = mocked_exec_msar_func.call_args_list[0][0]
+    transformed_column_data = [
+        _transform_column_alter_dict(column) for column in column_data_list
```
mathemancer, line 113: Generally, I'd prefer avoiding using this function in this test. The reason is that it makes it harder to modify the underlying function (e.g., making it so that we don't even need the transform function), since it breaks this test for unrelated reasons.

mathesar/tests/rpc/columns/test_c_base.py
```diff
@@ -123,11 +76,14 @@ def mock_column_info(_table_oid, conn):
             'valid_target_types': None
         }
     ]
+    mocked_exec_msar_func.fetchone.return_value = [expect_col_list]
     actual_col_list = columns.list_(table_oid=23457, database_id=database_id, request=request)
+    call_args = mocked_exec_msar_func.call_args_list[0][0]
     assert actual_col_list == expect_col_list
+    assert call_args[2] == table_oid
```
mathemancer, line 83: I think you should check all the arguments. That way, if someone fatfingers a function name down the road (i.e., messes up the string describing the SQL function), or accidentally switches the order of the connection and function name, this will catch it. Same for the other similar tests.

Also, we should use something more identifiable than `True` for the mock connect yield. Maybe something like:

```python
class MockConnection:
    pass
```

mathemancer (changes requested):
Good start, but:

You should check the whole call to `exec_msar_func`, including the connection and function name. This helps these tests catch a whole bunch of easy-to-make errors as we modify the python wrappers.

mathesar/tests/rpc/columns/test_c_base.py
Anish9901: > I think you should check all the arguments. That way, if someone fatfingers a function name down the road (i.e., messes up the string describing the SQL function), or accidentally switches the order of the connection and function name, this will catch it.

@mathemancer These conditions are already covered by `test_db_wrapper`. Not only will it check whether the 2nd argument in each `exec_msar_func` call is a string, it'll also verify that a function specified in the string exists on the underlying database. This would mean that we would also catch any accidentally switches of order between the connection and function name.

mathesar/tests/rpc/columns/test_c_base.py
Anish9901: I see your point @mathemancer, but, for the case you're describing I think I would prefer if the test _did_ break, since the way the function under test used to work has changed.

If we _were_ to not use these functions in these tests however, here are some of the possible ways I see to achieve that:
1. Not checking `assert call_args[3] == json.dumps(transformed_column_data)` (Doesn't seem ideal).
2. Hardcoding the result from the function in the test (Test would still fail if the behavior of the underlying function changes).
3. Using the transformed result as an argument to API call in the test (Test fails since the underlying function expects the argument to be not transformed).

mathesar/tests/rpc/columns/test_c_base.py
mathemancer: Ahhhh you're right. I forgot about that.

mathesar/tests/rpc/columns/test_c_base.py
mathemancer: I prefer (2).

- It clearly specifies the actual form that needs to be made to the DB layer.
- It verifies that no changes in the transformation have happened unintentionally, for example if someone were to introduce a bug in the `_transform_column_alter_dict` function itself.
- It detaches the test from the internal implementation of the function being tested as much as possible.

I agree the test should break if we align the API `column_data_list` with the form of the SQL layer `column_data_list`, but we _don't_ want it to break because (for example) the function being used in the test is now missing, which it would in that case with this test as currently written. We want it to break because we expect the input and output to be different, but they're now the same. I.e., we want the test assertion to say "Hey! this JSON isn't what I expected", not "Hey! There's a private function missing!".

Hopefully that makes sense.

mathemancer (approved):
Okay, LGTM!

PR #4026, opened 2024-11-13
Users RPC transition

### LINKED ISSUE #3993: Move `User` model manipulation to RPC endpoint
## Problem

One of our few remaining REST endpoints is `/api/ui/v0/users/`. Keeping that endpoint requires keeping around DRF, and in particular, requires keeping around our DRF error handling machinery. This adds an astonishing amount of complexity and code to our repo.

## Proposed solution

We should move Mathesar user-wrangling functionality to RPC functions. The current list of endponts mapped to functions (from the wiki) is:

| Endpoint                                                    | HTTP Method | Function                                  |
| --                                                          | --          | --                                        |
| `/api/ui/v0/users/{userId}/`                                | DELETE      | `users.delete`                            |
| `/api/ui/v0/users/{userId}/`                                | GET         | `users.get`                               |
| `/api/ui/v0/users/{userId}/`                                | PATCH       | `users.patch`                             |
| `/api/ui/v0/users/{userId}/password_reset/`                 | POST        | `users.password.revoke`                   |
| `/api/ui/v0/users/`                                         | GET         | `users.list`                              |
| `/api/ui/v0/users/`                                         | POST        | `users.add`                               |
| `/api/ui/v0/users/password_change/`                         | POST        | `users.password.replace_own`              |

## Additional context

We elected to forgo moving these endpoints as part of the architectural transition to save time.
```[tasklist]
### Tasks
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4000
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4001
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4002
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4003
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4004
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4005
- [ ] https://github.com/mathesar-foundation/mathesar/issues/4006
```

### LINKED ISSUE #4000: Implement `users.delete` RPC function
We need to implement a `users.delete` RPC function to replace the `DELETE /api/ui/v0/users/{userId}/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.delete')
@http_basic_auth_superuser_required
@handle_rpc_exceptions
def delete(
        *,
        user_id: int,
) -> None:
```

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

### LINKED ISSUE #4001: Implement `users.get` RPC function

### LINKED ISSUE #4002: Implement `users.patch` RPC function
We need to implement a `users.patch` RPC function to replace the `PATCH /api/ui/v0/users/{userId}/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.patch')
@http_basic_auth_login_required
@handle_rpc_exceptions
def patch(
        *,
        user_id: int,
        user_info: SettableUserInfo,
        **kwargs
) -> UserInfo:
```

A superuser can patch anyone's user, and a user can patch their own user.

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

### LINKED ISSUE #4003: Implement `users.password.revoke` RPC function
We need to implement a `users.password.revoke` RPC function to replace the `POST /api/ui/v0/users/{userId}/password_reset/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.password.revoke')
@http_basic_auth_superuser_required
@handle_rpc_exceptions
def revoke(
        *,
        user_id: int,
        password: str,
) -> None:
```

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

### LINKED ISSUE #4004: Implement `users.list` RPC function
We need to implement a `users.list` RPC function to replace the `GET /api/ui/v0/users/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.list')
@http_basic_auth_login_required
@handle_rpc_exceptions
def list() -> list[UserInfo]:
```

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

### LINKED ISSUE #4005: Implement `users.add` RPC function
We need to implement a `users.add` RPC function to replace the `POST /api/ui/v0/users/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.add')
@http_basic_auth_superuser_required
@handle_rpc_exceptions
def add(
        *,
        user_info: SettableUserInfo,
) -> UserInfo:
```

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

### LINKED ISSUE #4006: Implement `users.password.replace_own` RPC function
We need to implement a `users.patch` RPC function to replace the `POST /api/ui/v0/users/password_change/` endpoint.

The function signature should be:

```python
@rpc_method(name='users.password.replace_own')
@http_basic_auth_login_required
@handle_rpc_exceptions
def replace_own(
        *,
        user_id: int,
        old_password: str,
        new_password: str,
        **kwargs
) -> None:
```

A user can only use this function to modify their own password.

See [the wiki](https://wiki.mathesar.org/projects/2024/architecture-transition/rpc/#functions) for more context.

Anish9901:
Fixes #3993
Fixes #4000 
Fixes #4001 
Fixes #4002 
Fixes #4003 
Fixes #4004
Fixes #4005 
Fixes #4006

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
Ooo how fun! I love seeing a PR with some front end work too from you @Anish9901 🙂

This is mostly great, but I pushed b34300e92d1f743ab9c151a4e294e096312e72b7 to fix a small issue...

In `WritableUsersStore`, the `request` property was previously a `CancellablePromise` because that's was was returned from `api.users.list()`. The fact that it's a `CancellablePromise` is pretty much the whole reason we're storing at as a class variable here — so that we can cancel it later. But the new  `api.users.list()` function doesn't return a `CancellablePromise`. You don't get that until you call `run()` on it. So I changed some things to make sure that `request` stayed as a `CancellablePromise` and then I added the cancellation call back in which you removed.

_(As a side note, this `.run()` calling was my design and after working with it, I don't actually like this design, but that's another story — it's just to say that I think my poor design is part of what led you astray here.)_

I've reviewed the front end code and done a small amount of playing with the app to verify that things work okay. Looks good to me!

mathesar/rpc/users.py
```diff
@@ -0,0 +1,140 @@
...
+@http_basic_auth_login_required
+@handle_rpc_exceptions
+def patch(
+    *,
+    user_id: int,
+    user_info: SettableUserInfo,
+    **kwargs
+) -> UserInfo:
+    user = kwargs.get(REQUEST_KEY).user
+    if not (user.id == user_id or user.is_superuser):
```
mathemancer, line 106: Did you try to use the `set_authentication_predicate` decorator for this? I think it would be cleaner and more reusable to try to use that, or set up a custom auth decorator for the "is self or superuser" case.

mathemancer: It would also ensure that if the upstream package changes the response for failure of these decorators, that this function doesn't behave differently.

mathesar/rpc/users.py
```diff
@@ -0,0 +1,140 @@
...
+@handle_rpc_exceptions
+def replace_own(
+    *,
+    user_id: int,
+    old_password: str,
+    new_password: str,
+    **kwargs
+) -> None:
+    user = kwargs.get(REQUEST_KEY).user
+    if not user.id == user_id:
```
mathemancer, line 127: Same here. Try to use a custom auth decorator. You can just add one that guarantees that there's a `user.id` in the request that equals the `user_id`. Then, we have an added benefit of having a failure if a developer for some reason names the parameter something unexpected (i.e.,  something other than `user_id`)

mathesar/utils/users.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    user.set_password(user_def["password"])
+    user.save()
+    return user
+
+
+def update_user_info(user_id, user_info, requesting_user):
+    if not requesting_user.is_superuser:
+        user_info.pop("is_superuser", None)
+    User.objects.filter(id=user_id).update(
```
mathemancer, line 34: I'm suspicious of having an auth-related step (making sure a non-superuser can't elevate their privileges) in this function, rather than the RPC layer function. Did you consider putting it there? We don't want to have to look in multiple files to validate privilege logic in this case, unless there's a really good reason.

Also, it seems like `User.objects.get` be better here. Why did you go with the `filter` function? Specifically, I think we'd want to throw an Exception if there's no matching object. Maybe I'm missing something.

Finally, can't you just explode the `user_info` dict for this update call? E.g.,

```
User.objects.get(id=user_id).update(**user_info)
```

mathesar/utils/users.py
```diff
@@ -0,0 +1,62 @@
...
+
+def delete_user(user_id):
+    User.objects.get(id=user_id).delete()
+
+
+def change_password(user_id, old_password, new_password):
+    user = get_user(user_id)
+    if not user.check_password(old_password):
+        raise Exception('Old password is not correct')
+    user.set_password(new_password)
```
mathemancer, line 53: Same here. I don't think this layer should be authenticating against the user's old password. That should be the RPC layer (unless I'm missing something).

mathemancer (changes requested):
Some changes to request. See specific comments. The overall points are:

- Try to use the built-in `django-modern-rpc` auth system for the `is_self` check.
- Avoid spreading "auth-like" checks across multiple files.

Question: Was doing this full-stack the reason that you needed to do all endpoints at once? I could see how that would happen if there are dependencies in the front end between the old REST endpoints. I'd generally prefer reducing scope on a PR like this by reducing the number of endpoints transitioned at once, rather than skipping tests and docs.

mathesar/rpc/users.py
```diff
@@ -0,0 +1,140 @@
...
+def add(*, user_def: UserDef) -> UserInfo:
+    user = add_user(user_def)
+    return UserInfo.from_model(user)
+
+
+@rpc_method(name='users.delete')
+@http_basic_auth_superuser_required
+@handle_rpc_exceptions
+def delete(*, user_id: int) -> None:
+    delete_user(user_id)
```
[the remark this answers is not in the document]
Anish9901: This is intentional, we only allow superusers to delete accounts, including their own apparently, although not through the UI.

mathesar/utils/users.py
Anish9901: > I'm suspicious of having an auth-related step (making sure a non-superuser can't elevate their privileges) in this function, rather than the RPC layer function. Did you consider putting it there?

I did consider putting it there but thought that it would be better to add all the logic related to updating user in a single function instead of splitting it, this way we won't forget to check whether the requesting_user is a superuser, if we decided to use the `update_user_info` function elsewhere. But oh well..., we could also use the RPC functions directly instead. I've updated it as per your request.

> Also, it seems like User.objects.get be better here. Why did you go with the filter function? Specifically, I think we'd want to throw an Exception if there's no matching object. Maybe I'm missing something.

Django advises against doing this: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.update
We _could_ raise an exception based on the number of rows changed on the db that is returned by `update()` though.

> Finally, can't you just explode the `user_info` dict for this update call? E.g.,
> 
> ```
> User.objects.get(id=user_id).update(**user_info)
> ```

We can't use `update` on a User object or any object we receive with `get`, we can only use it with `filter`. Yep, I know... kinda stupid. But I think its because `update` performs a SQL update instead of loading the result in memory first, contrary to the way `get` works.

mathesar/rpc/users.py
Anish9901: I think it is next to impossible to use `set_authentication_predicate` for testing `is_self` because, for `is_self`, we would require `user_id` which is a parameter that we pass through the request. The way django_modern_rpc is set up, the permission checks set up through `set_authentication_predicate` happen before the RPC function's `args` and `kwargs` are populated for RPC function execution.

So, even something like this doesn't work:
```
@wraps(func)
    def wrap(*args, **kwargs):
        user_id = kwargs.get('user_id')
        wrapper = auth.set_authentication_predicate(http_basic_auth_check_user, [user_is_self_or_superuser, user_id])
        return wrapper(func)(*args, **kwargs)
```
I also tried getting the `user_id` using the request, but `request.POST` results in `None` when printed inside `http_basic_auth_check_user()`.

I finally settled on writing a custom decorator but without using `set_authentication_predicate`.

Anish9901:
> Question: Was doing this full-stack the reason that you needed to do all endpoints at once? I could see how that would happen if there are dependencies in the front end between the old REST endpoints.

Nope, They seemed to be small and less complex to implement thats why I did it in one shot.

> I'd generally prefer reducing scope on a PR like this by reducing the number of endpoints transitioned at once, rather than skipping tests and docs.

I'll keep that in mind next time.

mathemancer (changes requested):
Okay, I think I figured out what's bothering me about this PR w.r.t. the janky auth stuff. The functions need to change. 

As designed, they're kind of left over from the REST style, but a major goal of the RPC change is to leave that stuff behind. Instead of the current setup for `patch`, we should emulate the password change functions: One for 'self', and another for 'other'. Crucially, we don't need a `user_id` passed to patch our own user: That is leftover thinking from the REST endpoint. So, we should have a `patch_self` function that doesn't take a `user_id`, and which any user can access, and a `patch_other` function that _does_ take a `user_id`, and which only a superuser can access. Also, the `replace_own` function shouldn't take a `user_id` at all! This is completely unneeded, and only serves to force us to make that awkward auth check to make sure that a user can only modify their own password. Finally, I think these functions should take primitive arguments rather than a `SettableUserInfo` object. The reason is that it makes it a bit clearer what the function is actually allowed to modify, and it's a bit safer. The `patch_self` function shouldn't even take a `is_superuser` parameter. 

The front end change should be pretty easy for the `patch` stuff, since the function that calls is already aware of whether the user is modifying themselves or another. You just need to call the appropriate RPC function at that point.

Sorry to keep going around on this stuff, but I think it's pretty important to avoid adding any ad-hoc auth pieces if we can avoid it.

Bonus idea: Given that we'd be splitting out `patch_self` and `patch_other`, we should be able to wrap up the password change logic into the same function, since that's actually just a patch. The only reason we had it separate in the first place is because of the self vs. other distinction. This is optional, but it would simplify the module a bit.

Anish9901:
@seancolsen I've made some changes for patching users as suggested by Brent, in 7b31e81, can you please review them?

mathemancer (approved):
Okay, thanks for all your work on this. I'm happy with it once @seancolsen signs off.

PR #4030, opened 2024-11-19
Send null for string like columns that don't allow empty strings when cell is cleared

### LINKED ISSUE #3784: Send `null` for string-like columns that don't allow empty strings when cell is cleared
## Description
* Open or create a table with a column that allows string input that performs some validation.
   - Date & Time
   - Date
   - Email
   - URI etc.,
* Let's say we have a URI column.
* Edit a cell and enter some string., eg., `"test"`.
* This will result in an error. That's expected.
* Now edit the cell again and remove everything clearing the cell.
* This will result in an error again, when it shouldn't.
* That's because the frontend sends the patch request with an empty string `""`.
* (The user can still right click and set to null, but that's not always obvious in this situation).
* For such columns that have custom validation and do not allow empty strings, the frontend should send `null` instead of empty string `""`.

aniketkulkarni17:
Related to #3784 

**Technical details**

- When a cell is cleared, the frontend will send "null" instead of "" to the backend, thus allowing empty values for the following column datatypes:
  - Date
  - Date and time
  - Time
  - Email
  - URI

**Screenshots**

- Before
Screenshot 2024-11-19 214938
Screenshot 2024-11-19 214954

- After
Screenshot 2024-11-19 214852
Screenshot 2024-11-19 214907

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/records.ts
+++ b/mathesar_ui/src/stores/table-data/records.ts
@@ -592,7 +592,16 @@ export class RecordsData {
     const cellKey = getCellKey(rowKey, column.id);
     this.meta.cellModificationStatus.set(cellKey, { state: 'processing' });
     this.updatePromises?.get(cellKey)?.cancel();
-
+    const nullableColumnTypes = new Set([
+      "mathesar_types.email",
+      "date",
+      "timestamp with time zone",
+      "mathesar_types.uri",
+      "time with time zone",
+    ]);
+    if (nullableColumnTypes.has(column.type) && record[column.id] === "") {
+      record[column.id] = null;
+    }
     const promise = api.records
       .patch({
         ...this.apiContext,


aniketkulkarni17:
@seancolsen does this fit the requirement? Please let me know if any changes are required.

seancolsen (changes requested):
Thanks for working on this, @aniketkulkarni17. This kind of works, but has some issues...

## Issues

- For Email and URI columns, the front end _is_ sending `null` to the API (good). But it doesn't _display_ `NULL` to the user afterwards. If you click Refresh, _then_ you see `NULL`, but that's not great UX.

- Database types like `time without time zone` are no included in your string set and thus don't receive this fix. We could potentially address that by adding more database types to your string set. But I'd rather fix this at a lower level.

- This PR has a failing test because the code in your changes is not formatted correctly. We use [Prettier](https://prettier.io/) to automatically format code. You can [read more in our docs](https://github.com/centerofci/mathesar/blob/master/mathesar_ui/README.md#formatting) about using it. Personally, I have VS Code configured to format on save which I love.

    ```
    docker exec -it -w /code/mathesar_ui mathesar_service_dev npx prettier --write .
    ```

    I'd like you to take a different direction with the overal implementation of this PR, but when doing so, make sure to run Prettier at some point in your workflow and commit the changes it makes. The changes from Prettier don't need to be in a separate commit but it's also fine if they are.

## New direction

I said I'd like you to fix this at a lower level, and part of the reason is that the UI for inputting a cell value actually gets used in a lot of places. Not just within the sheet, but also the record page, the default value setting, the filter conditions on the table page, the filter transformations within the data explorer, the record selector etc. Some of those might be unaffected by this bug due to higher-level logic that handles null values, but suffice to say: the clean way to handle this is to do it at a low level, per-type.

_(As a disclaimer, I'll say that this work touches on some of the messiest and hard-to-work-with parts of our front end codebase, so, sorry about that! You're going to end up on somewhat of a wild goose chase.)_

Recall that I [said](https://github.com/mathesar-foundation/mathesar/issues/3784#issuecomment-2486038819):

> Interestingly, the Duration type actually already works as expected.

Let's dig deeper to see _why_ it works for Duration...

1. Go to [src/stores/abstract-types/type-configs/](https://github.com/mathesar-foundation/mathesar/tree/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/stores/abstract-types/type-configs) and you'll see one file for each [UI type](https://wiki.mathesar.org/engineering/glossary/ui-types/). In the code, a "UI Type" is called an "AbstractType". But the idea is that each UI type potentially maps to _multiple_ database types. And one database type maps to one UI type. The default object exported from each of those files defines the UI type. Look at the `cellInfo.type` property, and you'll see the name of the cell type that we use to work with this UI type. So in [duration.ts](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/stores/abstract-types/type-configs/duration.ts) we see a `cellInfo.type` of 'duration'. Good. _(But other UI types, like Email won't necessarily line up one-for-one like this!)_

1. Now if you look in the [src/components/cell-fabric/data-types/](https://github.com/mathesar-foundation/mathesar/tree/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/components/cell-fabric/data-types) directory, you'll see files which match up with those cell types. So within [duration.ts](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/components/cell-fabric/data-types/duration.ts) (same file name, different directory) you'll see that [getInput()](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/components/cell-fabric/data-types/duration.ts#L46) is returning a `FormattedInputProps` component with props that [specify](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/components/cell-fabric/data-types/duration.ts#L24) that the `formatter` is to be a `DurationFormatter`. Ok cool. So, what is a `DurationFormatter`, and how does it work?

1. In [src/utils/duration/DurationFormatter.ts](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/utils/duration/DurationFormatter.ts), we're specifying a [parse()](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/utils/duration/DurationFormatter.ts#L170) function. That's the entry point to the logic that translates the string the user entered into a standardized value. Notice that within [ParseResult](https://github.com/mathesar-foundation/mathesar/blob/develop/mathesar_ui/src/component-library/formatted-input/FormattedInputTypes.ts#L14) the value can be `T | null`, meaning that the return value of the parse function should be `null` to represent a `NULL` cell value. Cool. So where is the logic that returns `null` for an empty string input?

1. Higher up in `DurationFormatter.ts`, we have [this](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/utils/duration/DurationFormatter.ts#L23-L26)

    ```ts
    let cleanedInput = userInput.trim();
    if (cleanedInput === '') {
      return null;
    }
    ```

    Boom! That' exactly what we want to see. Not only does it translate an empty string to a `null` value, it also trims the string first. Brilliant!

So, what I'd like you to do to fix this issue is walk through the same steps above but for the problematic _UI Types_ (not database types). So if we start with the Date type, we'd go to step `1.` and navigate to `src/stores/abstract-types/type-configs/`. See that `cellInfo.type` is `date`. Then go to `src/components/cell-fabric/data-types/` and open `date.ts`, find the component returned from `getInput()`, look at its props, and eventually get to `DateTimeFormatter` which has a [parse](https://github.com/mathesar-foundation/mathesar/blob/cbabca3af698c9d5e72322216e5beb3ff428fcfd/mathesar_ui/src/utils/date-time/DateTimeFormatter.ts#L16-L39) function that doesn't handle empty strings correctly. Wonderful. _Here_ is where we need to make the fix for the Date type.

We could consider adding a new abstraction like some sort of intermediate formatter that automatically handles empty strings, and then we'd use _that_ formatter as a basis for the custom formatters used in these types. But I'm not sure that's worth it at this stage.

URI and Email are going to be a little trickier, so leave those for last. We'll probably need to add a new `email.ts` file in `src/components/cell-fabric/data-types/`. I'm not entirely sure.

seancolsen:
@aniketkulkarni17 are you still planning to work on this?

zackkrida:
@aniketkulkarni17 I've converted this PR to a draft while it awaits update. If you'd like to resume work please comment here, otherwise the PR may be closed or picked up by a maintainer depending on priorities. Thank you.

seancolsen:
I'm closing this now, but we can re-open it if you'd like to resume work @aniketkulkarni17

PR #4034, opened 2024-11-21
Fix two typos and add a spellcheck

### LINKED ISSUE #4033: Two typos in the docs folder
## Description
There are two typos in the `docs/docs` folder of the repository.

## Expected behavior
No typos.

## To Reproduce
Open the files and observe bug.

## Environment
N/A; happens on all devices.

## Additional context
Not exactly a bug, but the pull request template said to open an issue before submitting the PR.
There is a pull request to fix this: #4034.

TheGiraffe3:
Fixes #4033

In this pull request, I added a spellcheck and fixed two typos found by it.

## Checklist
  index.md`).
- [x] My commit messages follow the [best practices][best_practices].
> - [x] My code follows the established code style of the repository.

I think so; it's not exactly code.
> - [x] I added or updated documentation (if applicable).

The whole pull request is about updating documentation.
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (commented):
Thanks for this, @TheGiraffe3.

I would definitely like to merge the specific spelling corrections in this PR.

And I also love the idea of adding some tooling to catch spelling mistakes!

I don't love the idea of leaning so hard on GitHub actions though, because it's hard to run workflows like this locally, so you don't end up seeing the test failures until its in CI. It's not necessarily a show-stopper, but it would be nice to have something that we could run locally too. Ideally something that integrates with mkdocs so that it throws errors when running `build --strict`. For example [mkdocs-spellcheck
](https://github.com/pawamoy/mkdocs-spellcheck) looks promising, though I've not tried it. Looks like we'd be able to configure ignored words directly within mkdocs.yml, which would be convenient. Another approach would be finding a separate python-based markdown linting tool and adding a local script to run it. I wouldn't be surprised if there might even be some tools that perform more elaborate grammar checking. If we could get a local spell-checking solution set up, then we'd still add a GitHub action that runs that script to ensure that PRs with spelling errors fail.

As a side-note: If we could use such a tool to enforce some of our other [markdown code standards](https://wiki.mathesar.org/engineering/markdown/) that would be nice too. I did some [research](https://github.com/mathesar-foundation/mathesar/issues/3194) a while back into potentially adopting [mdformat](https://mdformat.readthedocs.io/en/stable/index.html) to auto-format our markdown. I ran into a snag. I could be worth picking this work back up though.

If you'd be up for taking a stab at getting a docs spell-checker working locally _too_, then that would be really nice. If not, I can understand. Let me know. This PR definitely adds value as-is so I'm inclined to merge it if you're not up for digger deeper into the local spell-checking use-case.

TheGiraffe3:
It's possible to configure ignored words in a separate file, which can be placed anywhere; I just didn't include that here because it wasn't necessary.

I'll look into the possibility of a local spellcheck, though it may take a few days.

TheGiraffe3:
It looks like the author of that mkdocs spellcheck isn't interested in adding a GitHub actions workflow without funding. 
With this one, it is possible to add an ignore file in the same directory, or add the ignored words in-line.
I've tried to convert the mkdocs spellcheck into an action that would be runnable locally as well as on GitHub, but I've run into a few snags; I'm not sure it will be possible to get something set up that runs both on GitHub and on a desktop.

Thoughts?

seancolsen:
@TheGiraffe3 I'm closing this PR in favor of #4045 where I've gone with [mkdocs-spellcheck](https://github.com/pawamoy/mkdocs-spellcheck) instead. That approach has the following benefits over this PR:

- It doesn't require any extra GitHub actions to be configured. We already have an action [here](https://github.com/mathesar-foundation/mathesar/blob/bd1fe16d4ffddddc4c94352b287bfe2929f0f9cd/.github/workflows/test-docs.yml) which attempts to build the docs site. The action will fail if any errors or warnings are encountered at build time. With the new plugin, spelling errors will result in build errors. In general I'm quite wary of any increase in complexity within the CI pipeline. I've seen it gum up development before when people end up sinking time into troubleshooting CI failures.

- It checks the _built_ content instead of the _source_ content. I was surprised by this approach, but it actually seems to do a better job. For example it didn't report `**V**alue` as a spelling error, presumably because it's able to convert "**V**alue" to "Value". And it also reported a number of other spelling errors emanating from source content in Python files. Pretty neat!

- It can be run locally.

#4045 also includes the specific spelling fixes in this PR too.

Thanks for your work on this though! Great idea!

TheGiraffe3:
Fine with me, glad that I was able to help.

PR #4065, opened 2024-12-12
Execute default expr when not dynamic.

### LINKED ISSUE #3716: API returns column default value formatted differently from submitted
## Steps to reproduce

1. Use the front end to set a default value for a text column. In the value field, enter:

    ```
    foo bar
    ```

1. Observe that the default is set up as expected in PostgreSQL.

1. Expect the "Default Value" configuration form for the column to display a default value of `foo bar`, exactly as-entered.

1. Instead, observe that the default value displayed in the form (as is returned from the API) is:

    ```
    'foo bar'::text
    ```

This bug presents usability problems for users wishing to _modify_ an existing default value.

The default value _can_ be modified, but it must first be extracted from within the text casting expression.

## Considerations

I think a proper fix for this bug may require some re-thinking of the API return value schema when fetching columns. I'm not sure we'd be able to fix this while sticking to using a single string here.

mathemancer:
Fixes #3716 

This adds logic that executes the stored default expression for a column and then formats it properly for the given type whenever it's not dynamic.

**Technical details**

This should bring the current logic up to parity with the older behavior, or better. Since we now apply the format function to the default before putting it in the JSON for return, we (for example) return Date-type defaults in the same ISO format as the column values. The front end seems to happily consume this.

I modified a test to make sure the new behavior is working.

Note that when the default _is_ dynamic (e.g., `now()` or `GENERATED ALWAYS AS IDENTITY`), we don't attempt to execute it. This is for a number of reasons, but the most important is that we don't want to burn sequence values when listing the columns of a table.

**Screenshots**

### Before
image

### After
image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/00_msar.sql
+++ b/db/sql/00_msar.sql
@@ -934,6 +934,46 @@ WHERE has_privilege;
 $$ LANGUAGE SQL STABLE RETURNS NULL ON NULL INPUT;
 
 
+CREATE OR REPLACE FUNCTION
+msar.describe_column_default(tab_id regclass, col_id smallint) RETURNS jsonb AS $$/*
+Return a JSONB object describing the default (if any) of the given column in the given table.
+
+The returned JSON will have the form:
+  {
+    "value": <any>,
+    "is_dynamic": <bool>,
+  }
+
+If the default is possibly dynamic, i.e., if "is_dynamic" is true, then "value" will be a text SQL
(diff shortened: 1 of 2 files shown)


seancolsen (approved):
Works for me 👍🏻

PR #4077, opened 2024-12-13
Left-align text within number inputs

### LINKED ISSUE #2057: Number fields look awkward when right-aligned on the record page
## Design problem

- The NumberInput component right-aligns its contents for the following reasons:
    - It's consistent with numbers being right-aligned within cells.
    - It allows the user's cursor to remain in a consistent position while entering numbers, whereas left-aligning the contents causes the cursor to jump around as grouping separators are automatically added/removed from the number while the user types. @pavish Has expressed a strong opinion to keep numbers right-aligned for this reason.

- On the Record Page, the right-aligned numbers can look a bit weird because they are so far from their corresponding labels, and they're out of alignment with the other fields on the page.

    image

## Possible solutions

- Left-align numbers on the Record Page. Pavish argues against this, to avoid the cursor moving as grouping separators are automatically added to the number. If we do choose to left-align numbers on the Record Page, the we should also consider how to best align numbers for other usages of the NumberInput component (e.g. Record Selector search, default value entry, filter condition).
- Reduce width of the field
- Keep it as-is
- Something else?

seancolsen:
Fixes #2057

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish I recall you having some concerns about left-aligning the text within number inputs because the cursor can sometimes jump around when editing due to the automatic grouping separators. During usability testing, I've noticed that the right alignment of numbers can be confusing for users, so I think that left-alignment has a more optimal set of tradeoffs, all things considered. I'm open to continuing to debate this if you feel strongly about it.

pavish (approved):
I personally find the cursor jumping around because of the grouping separators distracting when typing. However, I agree that it does look odd to keep number right aligned in the record page.

I don't think the cursor issue is worth discussing at our current stage, we can come back to it at a later point in time.

PR #4031, opened 2024-11-19
Comprehensive documentation improvements

seancolsen:
## Overview

This PR:

- Adds a _lot_ of new user-focused docs content to our docs guide
- Cleans up a bunch of other stuff in the docs guide too
- Makes changes in the UI to make some things easier to document and reference
- Adds lots of small help bubbles in the UI, many of which link out to the new content in the docs guide
- Adjusts terminology in the UI to make things more consistent and clear
- Makes some small UX adjustments in a few places too

## High level demo of docs content changes

https://github.com/user-attachments/assets/72cf6774-b90c-477a-81c4-f981f6d151a7

## Demo of UI changes

https://github.com/user-attachments/assets/10773541-48dd-4884-ba19-a56de028ec6c

## Notes

- `Docs:` commits are docs content
- Other commits are UI changes

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
have the right to submit it under the open source license
indicated in the file; or

(b) The contribution is based upon previous work that, to the best
of my knowledge, is covered under an appropriate open source
license and I have the right under that license to submit that
work with modifications, whether created in whole or in part
by me, under the same open source license (unless I am
permitted to submit under a different license), as indicated
in the file; or

(c) The contribution was provided directly to me by some other
person who certified (a), (b) or (c) and I have not modified
it.

(d) I understand and agree that this project and the contribution
are public and that a record of the contribution (including all
personal information I submit with it, including my sign-off) is
maintained indefinitely and may be redistributed consistent with
[...]

PR #4089, opened 2024-12-17
Improve UX when creating new items

### LINKED ISSUE #2490: Prompt users to name tables during creation
## Problem
When users create a new table in Mathesar, an automatic name is generated, and users are not prompted to enter a name themselves. Based on usability testing, this can lead to confusion in future steps, such as when deciding the direction of table links.

## Proposed Solution
We need to update the workflow for creating a new table to prompt users to enter a name for the table before it is added to their schema.
This could happen in a modal, similar to when a new schema is created.

## Additional context
https://hackmd.io/@mathesar/B1aXELfpj

### LINKED ISSUE #4054: Highlight newly-created items via UI animations
Let's consider the flow of adding a new schema from the database page. After the user submits the modal form and the modal closes, the user is still on the database page looking at the list of schemas. We'd like to use some CSS animation on the newly-created schema card to help the user locate it amongst the list of schemas. We should also scroll it into view if necessary.

The animation should be useful, subtle, and tasteful. It is to be improvised during implementation.

We should apply this same animation when other entities are created too. For example:

- Roles
- Collaborators
- Databases and tables (in conjunction with #4055)

### LINKED ISSUE #4055: Don't "open" newly created databases/tables
## Current behavior

- Our UX is inconsistent when creating new things.

    - When creating a new Schema the user remains on the database page, still viewing the list of schemas.
    
    - When creating a new Database or Table, we navigate to the corresponding page after creating it.

## Desired behavior

- Consistently follow the pattern currently used when creating schemas — don't "open" new things after creating them

- When creating a new database, stay on the home page looking at the list of databases.

- When creating a new table, stay on the schema page looking at the, list of tables.

## Notes

- #4054 will help improve the UX when the new thing ends up inside a large list of other things.

seancolsen:
### Prompts users to name tables during creation

Fixes #2490

### Makes behavior consistent after creating most things

Fixes #4055

The pattern is now to remain on the view of _all_ the things — not to "open" the new thing you just created.

The one exception is Users because the workflow and UX would require more changes there.

### Adds a new low-level `highlightNewItems` svelte action

- This is the real meat of the PR!

- See the demo video below to understand how it works.

- There were lots of tricky challenges in building this, and I'm  pretty happy with how it turned out. I ended up using `IntersectionObserver`, `ResizeObserver`, `MutationObserver`, `requestAnimationFrame`, and `@starting-style`. Phew!

- I'll note that `@starting-style` is still [not widely supported](https://caniuse.com/mdn-css_at-rules_starting-style), but I think that's okay here. Users with older browsers won't see the highlight or the hint, but these are not crucial features. They can still use Mathesar without them. And at this point all major browsers have recent versions with sufficient support, so it's only a matter of time before necessary support is wide.

### Highlights newly-added things

Fixes #4054

- This PR highlights newly-added:

    - ✅ **Databases**
    - ✅ **Schemas**
    - ✅ **Tables**
    - ✅ **Child roles**
    - ✅ **Privileges**

- It does _not_ highlight newly-added:

    - ❌ Users
    
        We have a page-based workflow for adding and editing users. We'd need larger changes in order to bring this workflow into consistency with the rest of the app, and I didn't want to take that, both for the sake of implementation scope, and for the sake of minimizing design changes/discussions.

    - ❌ Roles, Collaborators
        
        I want to get #4031 merged before messing with that UI again.

        Also Implementation will require some additional complexity within `new-item-highlighter` in order to handle a row that contains multiple top-level elements being added together.

    - ❌ Rows, Columns — because implementation would be tricky

    - ❌ Filter, Sort, Group, Record summary template fields, exploration transformations

        I don't think highlighting is necessary here because the newly-added items appear right next to the button to add the thing, so it's pretty clear to the user which one was just added.

    - ❌ Links, Constraints

        I skipped these for the sake of time because they seem lower-priority
        

## Demo

https://github.com/user-attachments/assets/e168ee52-bec7-4c28-a150-90f7c8a7c8b3

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@ghislaineguerin you might be interested in looking over the UX here as well.

pavish (approved):
@seancolsen Nicee!! 😎

Looks good to me!

PR #4095, opened 2024-12-19
Change order of filters in the filters section

### LINKED ISSUE #4094: Change order and labels of filtering comparisons
## Current behavior

- When filtering a text column, the available comparisons are as follows:

    - `equals`
    - `is empty`
    - `is not empty`
    - `contains`
    - `starts with`

    image

## Desired behavior

1. Change the order of the filtering comparisons by moving `contains` to the top. This would make it the default comparison for text types.

1. Change the filter comparison names to include text describing their case sensitivity.

    The result would be the following filtering comparisons:

    - `contains (case insensitive)`
    - `equals (case sensitive)`
    - `is empty`
    - `is not empty`
    - `starts with (case insensitive)`

1. When actually filtering records, the case sensitivity of all comparisons should remain the same.

1. We should make sure the same fix is applied to text-_like_ columns as well (URI and Email).

## Rationale

During user testing, we've noticed that users seem to expect text filtering to be case insensitive by default. Re-ordering the filtering comparisons is a quick way to solve this problem while still preserving case sensitivity for "equals".

## Additional context

This issue originally began as #4051

sharath-1517:
Fixes #4094

**Technical details**

- I have ordered the filters according to the issue mentioned.
- Added truncate component to the dropdown to avoid inconsistent text overflow.

**Screenshots**

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sharath-1517:
Hey @seancolsen I have changed the ui to be truncated as I mentioned in the issue conversation. But I have to yet figure out the filtering operations.

seancolsen:
@sharath-1517 Can you clarify this please?

> But I have to yet figure out the filtering operations.

Is this PR ready for review? Or do you still have more work to do on it? I'm no sure what you mean by "filtering operations", given that the issue is clear about _not_ changing any of the logic used to filter records.

sharath-1517:
Yes absolutely @seancolsen the PR is ready to review. I mentioned "But I have to yet figure out the filtering operations" because I thought that I should work on the filtering operations as well after seeing these points below,
github-1

Seems like that's not how it was and I have misunderstood the task. You can go on with the review for this PR. Thanks for your response and sorry for the inconvenience.

seancolsen (changes requested):
Thanks @sharath-1517

Your application of `<Truncate>` looks good.

Unfortunately there are still two problems here...

## Labels need to be type-aware

When filtering a non-text column via "equals", the label for the comparison should read "equals" not "equals (case sensitive)". I would imagine this to be relatively straightforward to fix, though I don't know the exact steps. I'd like you to take a stab at addressing this issue.

## User confusion will still arise

Hmmmm... This is where this change gets a little trickier than I initially thought. This PR kind of helps, but it doesn't fully address the [original intent](https://github.com/mathesar-foundation/mathesar/issues/4051) of the issue. The problem here was that users expected filtering to be done in a case _insensitive_ manner but the "happy path" to applying a filter results in the filter being performed in a case-sensitive manner.

Here's why this PR doesn't fully solve the problem:

1. Say I click "Filter" to open the filtering drop down.
1. Then I click "Add New Filter".

    At this point, a filter condition will be added, with the first column chosen by default (say `id`). In most cases that first column will be a numeric column, so it won't have a "contains" comparison. So "equals" will be chosen as the comparison by default.

1. Now, if I change `id` to some other text-like column, say `name`, then I'm right back to this same problem — we have case-sensitive filtering when we ought to have case-insensitive filtering.

What _does work_ about this PR is that if I begin from the column header, open the context menu, and choose "Filter Column", _then_ I get "contains" by default. That's great!

So here's what I think we should do...

When you open the "Group" dropdown, do you see how the "Add New Grouping" button opens a dropdown to select a column? I think we should implement that same UX for Filter. Let's also do it for Sort too, just to make them all consistent. This way, when you add a new filter condition, you'll begin with a specific column, not just the first column. I think this is actually an improvement anyway, because the user is unlikely to _want_ to filter or sort by `id`.

Do you think you could try to figure out how to do that?

sharath-1517:
Hey @seancolsen thanks for getting back! Yeah sure I will try these two issues that you have mentioned above. I might get certain follow up questions on the go. I'll make sure I'll keep you posted as often as possible. Thanks again!

sharathvgts:
> @sharath-1517 heads up that we are aiming to have this issue resolved in the next couple weeks. If I don't see some significant progress here by the end of the week, I'll need to take it over myself. Sorry for the short notice on this. I should have made the time sensitivity more clear in the original issue.

Hey @seancolsen. I will surely give you an update before this weekend. Thanks for letting me know!

sharath-1517:
Hey @seancolsen I have tried the `grouping` like dropdown in `filtering`. I used `Grouping` components code and it's dependencies as base and tried changing like the same for filters. I want to make sure I am going in the right track. Could you take a look at it please?

`PR Ready for review`

sharath-1517:
`PR Ready for a review`

seancolsen:
@sharath-1517 reached out to me via Matrix DM about this PR on 2025-01-10 and we had a quick chat where I explained that I would be closing it.

Here's an except from our conversation:

> I spent some time reviewing your PR yesterday and there is quite a lot that still needs to be done to handle this correctly. I was hoping I might be able to push a few small changes to account for any remaining work, but there are a number of problems with this PR. We're under a time crunch right now, pushing towards out beta release. Normally, I'd take the time to explain everything wrong with the PR and give you an opportunity to fix it. But unfortunately I don't have time for that right now. Explaining everything wrong would probably take me more time than it would for me to fix it myself. Sorry about this. I'm planning to circle back it it next week. It's possible that we won't have time to include this in the beta release, in which case I might not spend time on it next week. And in that case I'd leave the PR open and we might have more time to dig into the problems together. I have some higher-priority beta-blocking tasks that I'm working on currently, and I'll need to see how long these take me to complete before I can get my head back into this PR.

So I'll be opening a separate PR to fix this issue.

Thus closing.

PR #4109, opened 2025-01-07
Don't highlight databases/tables/schemas when filtering

seancolsen:
This fixes a small bug I noticed today: Before this PR, when you would filter a list of databases, schemas, or tables and then clear the filter, the items would be highlighted as if they were new. This is an oversight of mine when implementing #4089. Oops! This PR fixes it.

The fix is arguably somewhat hacky (and contains some code repetition). But I think it's an okay compromise.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@pavish you had previously pushed 93bf60f08cd3e57d2129d74e1998770b01444c83, merging develop into this PR. I just (intentionally) wiped that out with a force push. Sorry if this messes anything up.

This PR had accumulated some merge conflicts. I set about resolving them without noticing that you had pushed more commits to this PR. Incidentally, the merge conflicts here were actually not very straightforward to resolve. They required me to manually make changes to multiple files, some even outside the set of conflicting files. Only upon trying to push my resolutions up did I notice your prior merge commit, preventing me from pushing. Rather than reconciling all of that history, I chose to force-push my merge up for the sake of time.

It would be great if you could please review this soon.

pavish (approved):
@seancolsen I do not think exposing an `enabled` boolean is the best way to achieve this.

1. We need to be aware of this and probably do it everywhere we use the `highlightNewItems` action where we require modifying the entire array.
2. There are bound to be scenarios we would miss and cannot easily handle.
    - Here's an example of a scenario not handled:
      - Open the home page. Refresh the browser (to clear our stores).
      - Now throttle the network speed to slow down requests.
      - Open a database.
      - Notice the skeleton loader for schemas.
      - Once it's loaded, all the items are highlighted.
    - There's a hacky partial fix for not highlighting everything when the database changes (from the breadcrumb) by using the {#key} tag, which does not cover this.

Considering all the ways the child elements could change (filtering, breadcrumb, loading etc.,), a better approach would be to keep highlighting disabled by default, and only enable it upon specific user actions (Eg., adding or updating a schema).

Here's a top-of-my-mind approach that might work better for us.
1. Have a controller for the highlightNewItems schema, that encapsulates the enabled boolean.
2. Enable highlighting using the controller upon specific user actions.
    ```
    async function save() {
      schemaHighlightController.enableOnce();
      await createSchema(database, { name, description });
    }
    ```
3. Create a method `enableOnce`, which would enable highlighting and automatically disable it after the mutationObserver runs exactly once, which would be when the child elements change after the user action.

Having said that, the code works, and I'm going to approve this PR (since we have to get the beta out soon). I'll leave it upto you to decide if you'd like to tackle this now or create an issue instead and handle it after beta.

seancolsen:
> keep highlighting disabled by default, and only enable it upon specific user actions

Yeah, I spend a good deal of time hemming and hawing about this while implementing the `highlightNewItems` utility. Basically there's a "declarative approach" (what I've implemented), and an "imperative approach" (what you're suggesting). With both approaches, there's a decent risk for things to fall through the cracks. The declarative approach risks over-highlighting. And the imperative risks under-highlighting. (For example, I see it as a feature that when you rename a schema, the schema is automatically highlighted when it ends up in a different position in the list.) I can see advantages and disadvantages to both sides. But I'll admit that the declarative approach has turned out to be more problematic that I originally predicted. I'd be open to exploring the imperative approach, and I like the controller-based design you've suggested. It might not even be that much work to add! But yeah, I don't really want to spend time on that at the moment. I'd rather not even open a ticket for it. This is the sort of thing I'd be inclined to keep in the back of our minds and revisit if we continue noticing friction with the declarative approach. All of this is to say: I agree, and would be keen to try going down that road later if needed.

> We need to be aware of this and probably do it everywhere we use the `highlightNewItems` action where we require modifying the entire array.

I don't think we need to bother with `enabled` in all cases. For example, I don't see anything wrong with `highlightNewItems` in `ModifyRoleMembers.svelte` or `AccessControl.svelte` because there's no filtering functionality there. But yes, we do need to be aware of this behavior. If we choose to add filtering there, then that might be a good time to consider adding a controller.

Issue #4051, opened 2024-12-06
Change records text filtering to be case insensitive


seancolsen:
## Steps to reproduce

1. Use our library sample data.
1. Go to the Publishers table page.
1. Filter by Name equals `Crown` (in title case).
1. Expect and observe one result. Good.
1. Change the filter value from `Crown` to `crown` (in lowercase).
1. Expect to see the same one result.
1. Observe no results. Bad.

mathemancer:
Some notes:
- `Contains` is already case insensitive in the front end
- `Starts with` has a case insensitive version in the back end
- I think we should avoid making filtering into a "fuzzy search" Between this and #4052 , I'm worried that's the direction you're heading with these issues. If we want fuzzy searching of table rows, we should make that its own feature. 
- We could always put a little check box next to the filter name "Starts with" to toggle case sensitivity

seancolsen:
@mathemancer

> `Contains` is already case insensitive in the front end

Ah **ha**! Ok, this is a very good point. I didn't realize that!! I have a new idea now (described below).

> We could always put a little check box next to the filter name "Starts with" to toggle case sensitivity

Yeah, I this would be my ideal solution. But it's actually not that easy on the front end. The filtering UI is pretty heavily abstracted and adding a little check box would mean an entirely different _structure_ for filter conditions. It's possible of course, but not exactly low-hanging-fruit, so not the sort of stuff I'm looking for while polishing things pre-beta.

---

For reference, this is the current behavior:

![image](https://github.com/user-attachments/assets/9590e53d-1782-4649-8baa-244cf86e72ab)

The filtering comparisons are:

- `equals`
- `is empty`
- `is not empty`
- `contains`
- `starts with`

**Here's what I think we should do:**

1. Change the order of the filtering comparisons by moving `contains` to the top. This would make it the default comparison for text types and would address the problem I've observed during user testing.

1. Change the filter comparison names to include text describing their case sensitivity.

The result would bhe the following filtering comparisons:

- `contains (case insensitive)`
- `equals (case sensitive)`
- `is empty`
- `is not empty`
- `starts with (case insensitive)`

How does that sound to you, @mathemancer?

Anirudh2112:
Hello @seancolsen and @mathemancer, I would like to contribute to this

mathemancer:
@seancolsen Agreed on all points. Sounds great to me.

mathemancer:
[...]

PR #4131, opened 2025-01-14
Fix UX issue with case-sensitive filtering

### LINKED ISSUE #4094: Change order and labels of filtering comparisons

seancolsen:
Fixes #4094

## Notes

- Much to my chagrin, the linked issue doesn't do an adequate job describing the original _problem_. Even the issue it supersedes — #4051 — doesn't explain the original problem from the user's perspective. Sorry! 😬

    To understand what's going on here, I think it would be worth spending seven minutes watching this [call](https://tldv.io/app/meetings/677e92c83b23c40013f0146c) (from 37:49 and 45:09). In the call, I explain the problem to @ghislaineguerin and get her approval for the direction I took with the fix here.

- Originally, I had wanted to change the comparison labels to be clearer about whether they are case sensitive. However, in implementing this I eventually decided that I'd rather keep the labels as-is for now.

    I still think it would be nice to communicate the case sensitivity of string comparisons to the user. But there ended up being a number of points stacked against changing the labels once I get into the code and looked at the result of changing it.
    
    Because the label change was my idea originally, I'm hoping it's okay to rescind it without thoroughly explaining all the nuance around my rationale, just for the sake of time. This PR achieves my original goal — the default user flow for filtering a text column is performed in a case-insensitive manner. And so I don't think it's important to address the labeling right now.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/abstract-types/operations/filtering.ts
+++ b/mathesar_ui/src/stores/abstract-types/operations/filtering.ts
@@ -79,7 +79,6 @@ const equalityFiltersResponse: AbstractTypeFilterDefinitionResponse[] = [
 // This is the API response expected from the server
 // Might be better if we can have this with the types endpoint
 const filterResponse: AbstractTypeFilterDefinitionResponse[] = [
-  ...equalityFiltersResponse,
   {
     id: 'contains_case_insensitive',
     name: 'contains',
@@ -100,6 +99,7 @@ const filterResponse: AbstractTypeFilterDefinitionResponse[] = [
     },
     hasParams: true,
   },
+  ...equalityFiltersResponse,
   {
     id: 'uri_scheme_equals',
     name: 'URI scheme is',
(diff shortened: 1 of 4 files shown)


pavish (approved):
Looks good!

Issue #3423, opened 2024-01-26
Attempting to add connection with both Library Management and Movie Collection schema results in a 502


pavish:
## Description
* I'm consistently able to reproduce this on my test setup on GCP, not locally.
* When I try to add a new connection to a DB and check both "Library Management" and "Movie Collection" schemas in the "Schemas to Install' field, it results in a 502 Bad Gateway.
* Endpoint: https://pavish-testing.mathesar.dev/api/ui/v0/connections/create_from_scratch/
* This subsequently results in a followup issue: https://github.com/mathesar-foundation/mathesar/issues/3420

## Hunch
* Could it be that caddy is killing the request if it takes too long to complete?

## Expected behavior
* The request should succeed.

Anish9901:
I tried to troubleshoot this issue, here are the findings:

- Gunicorn has a default timeout for requests which is 30s.
- Adding both the datasets at the same time took about **25s** locally but takes **3.1 mins** for a remote db.
- This time could vary depending on where the remote db exists i.e. latency between the server hosting mathesar and the remote db server.

Possible solutions:
- Increase the timeout for requests to something large enough.
- Remove 'Movie Collection' and introduce a smaller dataset.
- Make the data loading pipeline efficient for large enough datasets like the 'Movie Collection'.

kgodey:
@Anish9901 thanks for the update!

> Make the data loading pipeline efficient for large enough datasets like the 'Movie Collection'.

Could you provide more detail on what this would entail?

Anish9901:
> Could you provide more detail on what this would entail?

One way I could think of is to use both .sql and .csv, we use .sql to set up the schema, tables, and FKs and use csv to bulk load data using `COPY` as it is supposed to be much more efficient than regular `INSERTS`.

If this works, we should do it for all our datasets as it would also make it efficient to start a new demo instance for users.

Anish9901:
I just tried this and observed major improvements.

Adding both datasets now takes:
- **12s** instead of **3.1 mins** for a **remote db**.  _(93.5% improvement than before)!!!_
- **9s** instead of **25s** for a **local db**. _(64% improvement than before)!!!_

kgodey:
That's awesome, @Anish9901! Very nice work.

PR #4142, opened 2025-01-20
Add beta dataset loaders to setup RPC functions

### LINKED ISSUE #4138: Add new default data sets to Mathesar
We've made some new data sets for use cases on our new website. We'd like to add them to Mathesar and allow users to import them when they create a new database (like we have with books and movies now).

Related PR: https://github.com/mathesar-foundation/mathesar-data-playground/pull/20

As part of this, we should also remove our movies data set, but leave the books one in.

mathemancer:
Fixes #4138 

This completes the frontend and backend work for #4138 . It adds loaders for 6 new datasets to the database setup functions.

**Technical details**

The new keys to load the datasets are documented in the setup function docstrings.

**NOTE**: While the reviewer should double-check that the functions are actually working, we'll need to do _extensive_ QA on these datasets themselves at some point before releasing beta.

- Check for typos in table/column names
- Check for actual data in each table (I found and removed one that was missing data already)
- Check that each table can be inserted into as expected (i.e., I got all the sequence value-setting correct)

This PR includes the front end parts to let a user choose some data sets, and also removes the Movie Collection data set:

image

@kgodey @ghislaineguerin With this amount of sample data sets, I'd argue we need to reorganize the pane a bit. Maybe two sections: "Mathesar System Schemas", and "Sample Data Sets" to make it clear what's going on. I think with this many checkboxes it's pretty cluttered. Regardless, this is working for the moment.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@pavish You could probably just add the front end changes to this PR before merging if you'd like. I thought about doing it myself, but decided it was probably less confusing for you to just make those changes.

mathemancer:
@seancolsen Pavish is out tomorrow, so I'm reassigning this to you. There is a simple front end change to add, the only thing that keeps me from doing it myself is that I'm not sure whether I should hand-edit the data set descriptions in the translation area.

Anish9901 (changes requested):
@mathemancer I am not so confident with the approach this PR takes for loading datasets. Specifically, I don't think we should be using multiple `INSERTS` at all for loading data into tables, we should be using `COPY FROM`. The reason I am saying this is because we have ran into issues where loading our Movies dataset, which relied on similar sql dumps for loading data, resulted in gunicorn timeouts for remote servers. You can find more details about it in #3423 & PR #3448. There is a handy python script in the PR #3448 `demo/install/dumpcsvs.py` that would allow you to dump all the data of different tables in a schema into respective csvs.

Even though the approach mentioned above is quite overkill for the size of the datasets we are loading as compared to our Movies dataset, I think it would be worth it in terms of speed and efficiency. Especially now that we would allow users to add 7 datasets at once & loading datasets would be one of the first thing that a user might do after installing mathesar.

Another, although less preferable solution would to use a single `INSERT` statement per table similar to the ones used in `library_without_checkouts.sql`.

mathemancer:
> @mathemancer I am not so confident with the approach this PR takes for loading datasets. Specifically, I don't think we should be using multiple `INSERTS` at all for loading data into tables, we should be using `COPY FROM`. The reason I am saying this is because we have ran into issues where loading our Movies dataset, which relied on similar sql dumps for loading data, resulted in gunicorn timeouts for remote servers. You can find more details about it in #3423 & PR #3448. There is a handy python script in the PR #3448 `demo/install/dumpcsvs.py` that would allow you to dump all the data of different tables in a schema into respective csvs.
> 
> Even though the approach mentioned above is quite overkill for the size of the datasets we are loading as compared to our Movies dataset, I think it would be worth it in terms of speed and efficiency. Especially now that we would allow users to add 7 datasets at once & loading datasets would be one of the first thing that a user might do after installing mathesar.
> 
> Another, although less preferable solution would to use a single `INSERT` statement per table similar to the ones used in `library_without_checkouts.sql`.

@Anish9901 The movies collection data set that inspired doing all the custom work for `COPY` rather than `INSERT` has more than 500,000 rows of data in it. The largest data set in this PR has fewer than 500 rows. If they load all of these datasets, it's fewer than 2000 rows. At this scale, we should prioritize compatibility and simplicity over performance. To put a finer point on it, the difference between `COPY` and `INSERT` performance for data at the scale in this PR is completely irrelevant, would be imperceptible to the user.

As a compromise, I did some performance testing with the difference between combined `INSERT` statements vs. per-row `INSERT`s. I found minor performance improvements in cases of extreme latency (e.g., 1000ms) between the server and DB.

Since I already did that work, I'm including it in the PR. However, in the future, we should probably just stick with the `INSERT` statements as they come from `pg_dump` unless we really have nothing else to work on, and the performance is _demonstrated_ to be a problem (as it was for the movies collection).

mathemancer:
Update: I'm going to attempt the front end work myself, having gotten some answers on the Matrix channel about how to deal with the translation stuff.

mathemancer:
Okay, update 2: 

- @seancolsen You just need to review now, and make sure I didn't screw anything up in the front end.
- @Anish9901 Same re-review needed as before
- @kgodey @ghislaineguerin Double-check that we're okay with the amount of noise in that pane with this many sample data schemas before we merge.

mathesar_ui/src/systems/databases/create-database/InstallationSchemaSelector.svelte
```diff
@@ -17,13 +17,23 @@
     // These strings are not translated because these are the names that will
     // appear in the Mathesar UI after installation.
     internal: 'mathesar_types, __msar, msar',
+    bike_shop: 'Bike Shop',
```
seancolsen, line 20: Translation keys should all be snake_case

seancolsen (changes requested):
Hey @mathemancer just skimming this over breakfast w my kids in case it can speed things up. This does need to be changed since it seems like there are i18n strings that don't appear in the JSON file. I can push these a little changes later, but if you can do it, it might be faster since my day is a little crazy today.

mathesar_ui/src/systems/databases/create-database/InstallationSchemaSelector.svelte
seancolsen: Ahh shoot. Sorry. Nevermind.

seancolsen: This should be fine

seancolsen (approved):
Ok sorry, I was moving too fast. I think this looks good. I've not run it yet, but I think that's okay. I'll QA later. I'll leave it to you to merge.

Anish9901 (approved):
I'll take 1000ms of performance improvement :) Thanks for making the changes @mathemancer!

ghislaineguerin (approved):
@mathemancer I think it's fine to keep as is for now (long list). As long as the action buttons below are always visible.

seancolsen:
@ghislaineguerin 

> As long as the action buttons below are always visible.

Yes, the action buttons are always visible

Issue #4172, opened 2025-01-24
Confusing upgrade docs instructions


pavish:
## Description
- Open docs for 0.2.0: `releases/0.2.0`.
- Notice that the Summary says users would have to reinstall and we don't support upgrades.
- Notice that there's a section for Upgrades in the same page.
- The Upgrade section mentions that users will lose all user accounts and explorations.
   - Calling this an "Upgrade" is confusing to me.

## Expectation
  - Be clear on what we support reg., upgrades in the Summary.
  - Do not have a title called "Upgrade", since we don't actually perform any upgrade.
    - This is the most confusing part for me.
  - Perhaps rename it to: 'Recover the databases you created from previous installations'.

seancolsen:
@pavish note that @mathemancer and I are both actively working on the upgrade instructions in #4168

mathemancer:
@pavish please check the docs in the PR @seancolsen mentioned and close or update thisjissuejas appropriate.

pavish:
The changes in https://github.com/mathesar-foundation/mathesar/pull/4168: https://github.com/mathesar-foundation/mathesar/pull/4168/commits/51435f4df4695b64ebab851240a004721b199190 alleviate my concerns.

pavish:
> Yeah, I'm not sure what to call it. From our perspective, "upgrade" isn't correct. But I think most of our users consider Mathesar to be a single unit, with the "user databases" as part of it. From that perspective, this operation is an "upgrade that loses some (meta) data", and we should manage expectations with that in mind.

@mathemancer I think with the updated documentation mentioning the effects of upgrade, the word "upgrade" doesn't seem confusing.

PR #4191, opened 2025-01-28
Fix install from scratch instructions, improve upgrade instructions

### LINKED ISSUE #4173: Issues in setting up Mathesar from scratch
## Description
Some errors I encountered while setting up Mathesar from scratch, while following the docs for 0.2.0:

1. `requirements-prod.txt` is no longer present. We need to use `requirements.txt`.
    ```
    pip install -r requirements-prod.txt
    ```
     will fail.

2. `gettext` should be mentioned in the pre-requisites.
```
(mathesar-venv) root@14a43afd0798:/etc/mathesar# python3 manage.py compilemessages
CommandError: Can't find msgfmt. Make sure you have GNU gettext tools 0.15 or newer installed.
```

3. I was unable to get gunicorn serving Mathesar. I need to debug this further.

pavish:
Fixes #4173

Addresses #4172 

This PR:
- Removes quotes and spaces from the docs surrounding environment variables.
  - This is important since we are now using `os.environ.get` instead of `decouple_config`. See discussion in Matrix.
- Adds `ALLOWED_HOSTS` to the list of available configurations.
- Fixes incorrect instructions for install from scratch.
- Improves the recovery instructions for databases from previous installations.
   - The docs for upgrade are for deleting previous django db and recreating it.
   - This PR aims to guide the user to note down the database details and reconfigure them in Mathesar after installing again.

Note:
- I've left the gunicorn & caddy systemd service creation instructions as they are, since now they work properly as the environment variables are fixed (i.e removal of spaces & quotes).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


docs/docs/administration/install-from-scratch.md
```diff
@@ -223,19 +225,20 @@ Then press <kbd>Enter</kbd> to customize this guide with your domain name.
...
+1. Create a media directory for storing user-uploaded media
 
     ```
-    python -m mathesar.install --skip-confirm | tee /tmp/install.py.log
+    mkdir .media
     ```
 
 
-1. Create a media directory for storing user-uploaded media
+1. Install Mathesar functions on the database:
```
mathemancer, line 235: This description is no longer accurate. This step now runs the Django migrations and collects static files.

docs/docs/administration/install-via-docker-compose.md
```diff
@@ -110,7 +110,7 @@ Restart the docker containers for the configuration to take effect.
 If you'd like to use an external PostgreSQL server for Mathesar's internal database, you'll need to do the following:
 
 
-1. On the existing database server, [create a new database](https://www.postgresql.org/docs/current/sql-createdatabase.html) for Mathesar to store its metadata.
+1. On the external PostgreSQL server, [create a new database](https://www.postgresql.org/docs/current/sql-createdatabase.html) for Mathesar to store its metadata.
```
mathemancer, line 113: I think this change makes things more, rather than less, confusing. I.e., "external PostgreSQL server for Mathesar's internal database..."

docs/docs/releases/0.2.0.md
```diff
@@ -481,137 +482,120 @@ We will assume you're using our provided`docker-compose.yml` file. If you have h
...
+
+    -- Create a new mathesar_django database to use with 0.2.0
+    -- Set owner to whichever role you use to run Mathesar
+    CREATE DATABASE mathesar_django OWNER mathesar;
+
+    -- Change current database to the new mathesar_django database
+    \c mathesar_django;
+
+    -- Drop the temporary database
+    DROP DATABASE mathesar_temp;
```
mathemancer, line 512: Changing from the database you originally logged into to another and dropping from there causes errors for me:

```
[brent@minismod ~]$ psql -h localhost -U mathesar -d mathesar_django
Password for user mathesar: 
psql (17.2, server 13.18 (Debian 13.18-1.pgdg120+1))
Type "help" for help.

mathesar_django=# \l
                                                       List of databases
      Name       |  Owner   | Encoding | Locale Provider |  Collate   |   Ctype    | Locale | ICU Rules |   Access privileges   
-----------------+----------+----------+-----------------+------------+------------+--------+-----------+-----------------------
 mathesar_django | mathesar | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           | 
 postgres        | mathesar | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           | 
 template0       | mathesar | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           | =c/mathesar          +
                 |          |          |                 |            |            |        |           | mathesar=CTc/mathesar
 template1       | mathesar | UTF8     | libc            | en_US.utf8 | en_US.utf8 |        |           | =c/mathesar          +
                 |          |          |                 |            |            |        |           | mathesar=CTc/mathesar
(4 rows)

mathesar_django=# CREATE DATABASE mathesar_temp;
CREATE DATABASE
mathesar_django=# \c mathesar_temp
psql (17.2, server 13.18 (Debian 13.18-1.pgdg120+1))
You are now connected to database "mathesar_temp" as user "mathesar".
mathesar_temp=# DROP DATABASE mathesar_django;
WARNING:  terminating connection because of crash of another server process
DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
HINT:  In a moment you should be able to reconnect to the database and repeat your command.
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
!?> 
```

That's why I had them log out and back in.

docs/docs/releases/0.2.0.md
```diff
@@ -481,137 +482,120 @@ We will assume you're using our provided`docker-compose.yml` file. If you have h
...
     ```
 
-1. Log out with `CTRL+D`, and log into your new `mathesar_django` database:
+1. Log out with `CTRL+D`.
 
+1. Stop Mathesar.
     ```
-    docker exec -it mathesar_db psql -U mathesar -d mathesar_django
+    docker compose -f /etc/mathesar/docker-compose.yml down
```
mathemancer, line 519: We've avoided changing names for any of the docker compose services specifically so they _don't_ have to stop Mathesar for upgrade steps.

Were you getting errors when upgrading without stopping the old services? If so, what were they?

docs/docs/releases/0.2.0.md
```diff
@@ -481,137 +482,120 @@ We will assume you're using our provided`docker-compose.yml` file. If you have h
...
     ```
 
-1. Then run:
-
-    ```sql
-    DROP DATABASE mathesar_temp;
+1. Rename the old docker-compose file.
```
mathemancer, line 522: Good idea.

docs/docs/releases/0.2.0.md
```diff
@@ -481,137 +482,120 @@ We will assume you're using our provided`docker-compose.yml` file. If you have h
...
+    CREATE DATABASE mathesar_django OWNER mathesar;
 
-1. Run Django migrations
+    -- Change current database to the new mathesar_django database
+    \c mathesar_django;
 
-    ```
-    python manage.py migrate
+    -- Drop the temporary database
+    DROP DATABASE mathesar_temp;
```
mathemancer, line 580: I didn't check in the "install from scratch" case, but I suspect this will break there too.

docs/docs/releases/0.2.0.md
```diff
@@ -5,8 +5,9 @@
 Version 0.2.0 is Mathesar's first **beta** release and features a brand new [access control system](#permissions), massive [performance improvements](#performance), and an important [security fix](#data-files).
 
 !!! caution ""
-    **We do not fully support upgrading from previous versions to `0.2.0`.**  For more details see [below](#upgrading).
-
+    **We do not support upgrading from previous versions to `0.2.0`.**
```
mathemancer, line 8: I think this is a bad way to frame this. You want this section to give someone who's currently running 0.1.7, and who wants to run 0.2.0 a way to proceed. This framing is far too scary and negative. You should at least add some material here along the lines of "if you're currently running 0.1.7, and you'd like to move to 0.2.0, you need to reinstall Mathesar, and then reconnect."

There's nothing accurate about the word "recover", and it's scary for users. Their data won't be deleted, and it won't go anywhere. If they're connected via a different client, they won't even notice anything via that other client. "reconnect" or something would be far better.

docs/docs/releases/0.2.0.md
```diff
@@ -456,23 +457,23 @@ The following PRs represent work done to reduce technical debt, update dependenc
 [#4152](https://github.com/mathesar-foundation/mathesar/pull/4152 "Release 0.2.0")
 [#4170](https://github.com/mathesar-foundation/mathesar/pull/4170 "Rpc performance")
 
-## Upgrading to 0.2.0 {:#upgrading}
-
-Upgrading to version 0.2.0 will:
+## Recover databases from a previous installation {:#recover}
```
mathemancer, line 460: This title is inaccurate, and also scary for users. Maybe "reconnect"

mathemancer (commented):
I think this reworking of the docs leaves users without a way to proceed if they're running 0.1.7, and they'd like to be running 0.2.0. I.e., you've removed the concept of "upgrade", and replaced it with nothing. Even if you don't want to call it "upgrade", you need to have some obvious sequence of steps that a user should follow in that use case, even if those steps come with warnings about potential data loss, etc.

docs/docs/releases/0.2.0.md
pavish: > Their data won't be deleted, and it won't go anywhere.

Depends on what we call 'data', their PostgreSQL databases won't go anywhere but they lose all of Mathesar data including connected database configuration.

I agree perhaps "Recover" might not be the best wording, but I'm strongly against using "Upgrade". I'm onboard with workshopping better terms.

docs/docs/releases/0.2.0.md
pavish: We are instructing the user to do a `docker compose up` later down and I don't think we can assume the user wouldn't face any issues during that process. I felt it is simpler to ask the user to stop and start after new installation.

docs/docs/releases/0.2.0.md
pavish: Modified there as well.

docs/docs/releases/0.2.0.md
pavish: I've used the term "Migrating to 0.2.0 from previous versions". I think this is a reasonable middleground.

docs/docs/releases/0.2.0.md
pavish: I've updated the PR to use: "Migrating to 0.2.0 from previous versions".

docs/docs/administration/install-from-scratch.md
pavish: Fixed in https://github.com/mathesar-foundation/mathesar/pull/4191/commits/2b90b066713c4f4edf7eb02c5c0fac9e0df328fc

docs/docs/administration/install-via-docker-compose.md
pavish: I've removed the term `existing/external` and used `your PostgreSQL server` instead. I think this would be more clear.

Here's the commit: https://github.com/mathesar-foundation/mathesar/pull/4191/commits/78095c78c8b522a7b31a86c5e08f1cefb0ab4f8e

seancolsen (approved):
Looks great! The new ~~upgrade~~ **migration** instructions read very well!

I added a couple small commits, cleaning up some markdown formatting.

PR #4255, opened 2025-02-12
Copy suggestions for passwordless roles in Mathesar

zackkrida:
@seancolsen and @mathemancer:

Here are some suggestions for the "minimum viable" approach to passwordless roles in Mathesar. The main changes are:

1. Add help text below password inputs which states "Can be left blank if your role doesn't use a password"
2. Replace the concept of "Stored Role Passwords" with "Stored Roles". The difference is subtle but the password _itself_ becomes an implementation detail / subfeature rather than the top-level concept presented to users.
 
### Connecting a DB

Screenshot From 2025-02-12 11-45-03

- Q: What if the user tries to connect a DB that requires a password but they do not supply one, because our submit button "lights up" without it?
  - A: The user will see an error indicating that the connection failed. This combined with the blank password field should be a sufficient indication that they've missed their password. 

### Disconnecting a DB

Screenshot From 2025-02-12 11-44-42

### Managing stored roles

Screenshot From 2025-02-12 11-45-32
Screenshot From 2025-02-12 11-45-53
Screenshot From 2025-02-12 11-45-47
Screenshot From 2025-02-12 11-45-41

### Updated docs

Screenshot From 2025-02-12 11-46-46


seancolsen (approved):
Just to give a quick top-level review:

@zackkrida this is brilliant. Thank you for identifying an easy fix here. I've not looked through the entire diff, but I skimmed your screenshots and I think it's great! With the smaller details, I trust that either: you've handled them here already, or @mathemancer will handle them in his PR, or we'll catch them in QA, or they won't be a big deal anyway.

zackkrida:
Concerning small details, I'm pretty confident that I've covered every link and reference to "stored role passwords" here and revised them to "stored roles", but we should definitely double check as part of the review of @mathemancer's PR, assuming these changes are merged in.

mathemancer (approved):
I found a few lingering references in the flow for removing a stored role. I'm going to note them here for reference after merging this.

image

image

image

And in the flow for disconnecting the last DB:

image

I think the help text is probably fine there, but we should consider changing the text for the checkbox

For reference:

```sh
$ git grep "stored.*password" mathesar_ui/src/i18n/languages/en/dict.json

mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_password": "Remove Stored Password",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_password_for_identifier": "Remove Stored Password for [identifier]?",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_password_help": "Removing this stored password will prevent collaborators assigned to this role from accessing databases on server ''{server}''.",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_role_passwords": "Remove stored role passwords",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_role_passwords_help_1": "Because you are deleting the last database associated with its containing server, this option is present.",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_role_passwords_help_checked": "With this box [bold](checked), the passwords that you've stored for PostgreSQL roles on this database's server will be deleted from Mathesar's internal database.",
mathesar_ui/src/i18n/languages/en/dict.json:  "remove_stored_role_passwords_help_unchecked": "With this box [bold](unchecked), the passwords will remain in Mathesar's internal database, but you will not have a way to edit them or remove them unless you re-connect another database associated with the same server.",
mathesar_ui/src/i18n/languages/en/dict.json:  "see_docs_for_stored_role_password_help": "See our [docsLink](documentation) for more information on stored role passwords.",
mathesar_ui/src/i18n/languages/en/dict.json:  "stored_password": "Stored",
mathesar_ui/src/i18n/languages/en/dict.json:  "stored_password_removed": "The stored password has been removed.",
mathesar_ui/src/i18n/languages/en/dict.json:  "stored_role_passwords": "Stored Roles",
mathesar_ui/src/i18n/languages/en/dict.json:  "stored_role_passwords_help": "Mathesar stores roles that you would like to use when authenticating with PostgreSQL to work with data. Stored roles can then be assigned to users via collaborators.",
mathesar_ui/src/i18n/languages/en/dict.json:  "update_stored_password": "Update Stored Role",
mathesar_ui/src/i18n/languages/en/dict.json:  "update_stored_password_for_role": "Update stored role [role]",
mathesar_ui/src/i18n/languages/en/dict.json:  "upgrade_specify_a_role_help": "To perform the upgrade, Mathesar needs the name and password for the role which owns the \"msar\" and \"__msar\" schemas. In most cases, Mathesar can find this information by looking in its stored role passwords. But depending on how you have chosen to configure your PostgreSQL roles and privileges, you may need to specify a role to perform the upgrade. Role names and passwords provided here will be not be stored.",
```

To continue, I'll

- Approve and merge this, since I like (and it seems others do too) the general approach.
- Make a few more changes on top of this after it's merged into my PR to handle the references I noted above.

Thanks for your help with the copy, @zackkrida !

PR #4257, opened 2025-02-12
Persist column width

### LINKED ISSUE #3262: Persist customized column widths
See sub-issues.

### LINKED ISSUE #3263: Front end changes to persist customized column widths

seancolsen:
Fixes #3263
Fixes #3262

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/systems/table-view/tableViewUtils.ts
+++ b/mathesar_ui/src/systems/table-view/tableViewUtils.ts
@@ -1,0 +1,10 @@
+import type { ProcessedColumn } from '@mathesar/stores/table-data';
+
+export function* getCustomizedColumnWidths(columns: Iterable<ProcessedColumn>) {
+  for (const column of columns) {
+    const width = column.column.metadata?.display_width;
+    if (!width) continue;
+    const entry: [number, number] = [column.id, width];
+    yield entry;
+  }
+}
--- a/mathesar_ui/src/api/rpc/columns.ts
+++ b/mathesar_ui/src/api/rpc/columns.ts
@@ -1,3 +1,4 @@
+import { defaultColumnWidthPx } from '@mathesar/geometry';
 import { rpcMethodTypeContainer } from '@mathesar/packages/json-rpc-client-builder';
 
 export type BooleanInputType = 'checkbox' | 'dropdown';
@@ -104,6 +105,9 @@ export interface RequiredColumnMetadata {
   duration_min: DurationUnit;
 
   duration_max: DurationUnit;
+
+  /** The pixel width of the column */
+  display_width: number;
 }
 
 /** The column metadata values, types as we get them from the API. */
@@ -125,6 +129,7 @@ export const defaultColumnMetadata: RequiredColumnMetadata = {
   date_format: 'none',
   duration_min: 's',
   duration_max: 'm',
+  display_width: defaultColumnWidthPx,
 };
 
 /**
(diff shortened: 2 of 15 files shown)


seancolsen:
@pavish please review but don't merge until #4248 is merged.

mathesar_ui/src/systems/table-view/tableViewUtils.ts
```diff
@@ -0,0 +1,10 @@
+import type { ProcessedColumn } from '@mathesar/stores/table-data';
+
+export function* getCustomizedColumnWidths(columns: Iterable<ProcessedColumn>) {
+  for (const column of columns) {
+    const width = column.column.metadata?.display_width;
+    if (!width) continue;
+    const entry: [number, number] = [column.id, width];
+    yield entry;
+  }
+}
```
pavish, line 10: @seancolsen I'm really curious about the decision to use generators for utility functions such as these. I’m not seeing the benefit in terms of reduced complexity or improved performance in this case. Is it because you'd like to follow/establish a pattern? I'd love to hear your perspective.

pavish (approved):
Looks good

mathesar_ui/src/systems/table-view/tableViewUtils.ts
seancolsen: I don't feel too strongly about this, but I guess I've just been increasingly _"thinking in terms of iterators"_ over the past couple of years. The flexibility of iterators makes for nice building-blocks at the lower levels, in my opinion. There are _some_ cases where they can perform better, but as you point out, this is not one of those cases.

- Perhaps you'd write this instead?

    ```ts
    export function getCustomizedColumnWidths(columns: Iterable<ProcessedColumn>) {
      const entries: [number, number][] = [];
      for (const column of columns) {
        const width = column.column.metadata?.display_width;
        if (!width) continue;
        entries.push([column.id, width]);
      }
      return entries;
    }
    ```

    If I were reviewing this code, I'd have no problems with it! It's quite similar to my code. I have a slight preference for the generator version, because to my mind there is just ever so slightly "less going on" inside the function. That's subjective though! And I'm certainly not advocating for any sort of code style standard in favor of generators.
    
    If you'd like to advocate for a code style standard _against_ generators, I'd be open to discussing it. I recognize that most JS/TS devs don't tend to use generators. In my experience they're are quite common in Python, I would guess because they've been around much longer.

- Perhaps you'd write this?

    ```ts
    export function getCustomizedColumnWidths(columns: ProcessedColumn[]) {
      return columns
        .map((c) => [c.id, c.column.metadata?.display_width ?? undefined])
        .filter(([, w]) => w !== undefined) as [number, number][];
    }
    ```

    This is nice and tidy, but it requires a type assertion, which I don't like. We could get rid of the assertion with a custom type guard, but that just gets too complex and messy for this simple case in my opinion.

    TypeScript 5.5 adds [Inferred Type Predicates](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#inferred-type-predicates) which are designed to solve exactly this problem. My understanding of this new feature is that we'd (hopefully) be able to write this:

    ```ts
    export function getCustomizedColumnWidths(columns: ProcessedColumn[]) {
      return columns
        .map((c) => [c.id, c.column.metadata?.display_width ?? undefined])
        .filter(([, w]) => w !== undefined);
    }
    ```
    
    Assuming that works, then it would probably tip the scale for me to prefer this version over all the others (including the code in this PR). I've been excited about this TS feature for a while now!
    
    
All in all, it doesn't matter too much to me. If there's a reason you _don't_ like generators, I'd be curious to hear. But I want to be cautious of getting too deep into a debate if's not something you feel strongly about.

PR #2773, opened 2023-03-31
Allow data to be copied from Mathesar and pasted elsewhere

seancolsen:
Fixes #1688

## Notes

- This PR provides a partial implementation of the "copy" functionality that I've [proposed](https://github.com/centerofci/mathesar-wiki/pull/90) in the [Clipboard Interaction Specs](https://github.com/centerofci/mathesar-wiki/blob/clipboard-specs/design/specs/clipboard-interactions.md). **It would be good to read the spec before reviewing this PR.**

- Scheduling of review

    - If we had ample time, I'd be inclined to get the Clipboard Interaction specs merged before merging this partial implementation. Instead, I'm doing the two in parallel. Why? Because I actually started with the implementation in order to get a good sense of the available browser APIs and their limitations. Getting my hands dirty with this first part of the implementation made the spec process much clearer than it would have been if I had written the specs before writing any code. I'm hoping that the "copy" portion of the spec is sufficiently straightforward that the code in this PR will require minimal alteration after final approval of the spec. So I think it's best to go ahead and try getting this merged as-is even though we're still hammering out the spec.

- In this PR:

    - You can copy data from Mathesar as TSV plain text. This allows you to paste it into other applications like LibreOffice Calc, Google Sheets, and Airtable.

    - You can copy cells from a table sheet or an exploration sheet.

    - The cell values are copied as _formatted values_. This means that FK cells are copied as their record summaries, data cells are copied as their formatted dates and so on.

- What's missing:

    - This PR does not implement any paste functionality.

    - This PR does not write the Mathesar-specific representation to the clipboard when copying. That's okay because we haven't implemented paste yet.

    - This PR does not have adequate handling of newlines/tab characters in cell content. As noted in the spec, I'm interested in utilizing a 3rd party lib for that, so I'd like to do a bit more research before going into that implementation. I'll track that work in tickets after merging this PR and merging the design spec.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the `master` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
[...]

Issue #2383, opened 2023-01-27
Client-side aggregations from cell selection


seancolsen:
## Problem

- Spreadsheets have functionality to display aggregate info based on the selection of cells

    ![image](https://user-images.githubusercontent.com/42411/215113348-376ee3d3-82d7-4b15-b9e2-aa091fb0cc0e.png)

- Mathesar lacks any such functionality

## UX Design goals

- We should figure out how to give the user similar functionality to shown above.

## Possible solutions

- In #2342 we would like to add a "Cell" tab to the table inspector. The Data Explorer already has such a tab. The Cell tab within the inspector (in the Table Page, Exploration Page, and Exploration Editor) could display aggregate info.

## Considerations

- The Record Page has table widgets which don't allow the table inspector. If the user wants to display selection-based aggregate info, is that possible from within a table widget?
- When we display aggregate info, how do we decide which types of aggregations to show?
- It seems that types which we could conceivably aggregate (e.g. Number, Date, etc) are transmitted via the API as strings. In most cases the front end will be able to parse those strings, but in some cases the front end parsing will fail. For example, Postgres can handle numbers with much higher precision than JS. What is our expected behavior in this case?
- Are there performance concerns with this feature? Does the front end to parse the cell values as the user forms their selection? This seems fine for small selections, but it's common for the user to select all the cells in a column. Do we need to parse the values when fetching and cache them? Would this present performance concerns during load time? Could we offload that parsing/caching work to a web worker to be done asynchronously on load?

## Next steps

- If we decide to move forward with this feature we will need UX design specs and an implementation RFC

seancolsen:
The maintainers team met today to discuss this issue. We decided to approve the feature at the product level, with the recognition that it still requires UX design before implementation.

@mathemancer expressed some concern that this feature would not be useful enough to justify its inclusion in Mathesar. He worries that giving user too many ways to compute aggregates will become confusing.

I and @ghislaineguerin both made a case for this because we rely on this feature regularly when using spreadsheets.

To help address @mathemancer's concern, we'd like to note that the UX design should take care to contain this feature to a small part of the app where users are unlikely to confuse its purpose with our other grouping and summarizing features.

seancolsen:
Towards the UX design process, here's a quick mockup of what I have in mind:

![image](https://github.com/mathesar-foundation/mathesar/assets/42411/6149e69f-237d-47dd-817d-ef32dc0b5736)

Issue #936, opened 2022-01-03
Add keyboard shortcuts for setting a cell to NULL


seancolsen:
## Desired behavior

1. Select a cell.
1. Press `Shift+Delete` or `Shift+Backspace`.
1. The cell value will be set to NULL, with Mathesar exhibiting the same loading, success, and error behavior as it currently does when the user updates a value within a cell.

## Notes

- The keyboard shortcuts should only be active when in **select** mode -- in **edit** mode there should not be any keyboard shortcuts for the user to set the cell value to NULL.

- If the column has a NOT NULL constraint, then an error toast message should show before a request is sent to the server, and no changes should be made.

    The toast message should read:
    
    > `<column_name>` does not allow `NULL`.
    
    Within the toast message, the The column name should be stylized with the `Identifier` component and NULL should be stylized with the `Null` component.

- If the cell is already NULL, then no request should be sent to the server and no toast message should display.

## Additional context

- [NULL Values spec](https://wiki.mathesar.org/design/specs/specs-null-values)
- [Discussion thread](https://github.com/centerofci/mathesar-wiki/pull/13#discussion_r770537811) with some differing opinions about whether to use `Delete` or `Shift+Delete`.

nk183:
hi @seancolsen can I work on this

kgodey:
Go ahead @nk183, thanks!

seancolsen:
@nk183 Are you still working on this? We'll be working to close out this milestone soon, so we'll need to re-assign it to a core team member if we don't get a PR by 2022-04-19. Thanks.

AbhinavOhri:
hi @seancolsen  can I work on this?

Anish9901:
Assigned to you @AbhinavOhri. Thanks!

AbhinavOhri:
hi @Anish9901, I need some help on this. I have figured out the part of how to run the keyboard shortcut but I am unable to figure the file where I should make the desired changes and how to check whether the edit mode is enabled or not.

AbhinavOhri:
I have completed setting up the keyboard shortcut. But I could not figure it out  the "NOT NULL constraint" segment , can you please guide me on that?

pavish:
@AbhinavOhri Each `Column` instance has a `nullable: boolean;` property. If this is true, it means that the column accepts `null` as a valid value in the database. If it's false, it means that the column does not allow null.

[...]

PR #4258, opened 2025-02-13
Support pasting data into cells from system clipboard

### LINKED ISSUE #3143: Implement paste into cells
```[tasklist]
### Tasks
- [ ] https://github.com/centerofci/mathesar/issues/2844
- [ ] https://github.com/centerofci/mathesar/issues/2812
```

seancolsen:
Fixes #3143

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
I'm moving this to 0.2.2 since I've encountered some unexpected roadblocks during implementation, and during today's [product meeting](https://tldv.io/app/meetings/67b5e3e2e5b9270013a3c11b) we agreed that this doesn't need to block 0.2.1.

seancolsen:
At the very least I need a review from @pavish concerning the front end code.

But, since this work had a lot of interesting edge cases concerning usability, I'd also like to invite everyone else on the team to look at this if you're interested @zackkrida @ghislaineguerin @mathemancer @Anish9901 @kgodey.

Note this is for 0.2.2 so we're in no rush to merge it. I'd like to have some other eyes on this PR for the sake of QA and usability testing. If you want to take this for a spin and see if you can identify any unexpected behavior, that would be really helpful! And if you just want to watch the videos and chime in with thoughts regarding usability, that would be great too!

zackkrida:
@seancolsen great PR and videos, the power of the functionality and any trade-offs you made to keep down scope are very clear. I have some thoughts, I apologize in advance about the somewhat random order of them:

## Casting and compatibility between different cell types

I _love_ how all of this works, no complaints here. Really well-thought out and implemented. 

## On the friction of the modal

I think the inconvenience added by the modal is a good thing for the cmd+v shortcut for one simple reason: We do not yet have undo/redo for table edit and insert operations. Edits can be quite destructive without the ability to undo them!

I would propose introducing a modifier key to allow for a simple way to bypass the modal. This would allow for "power" users to avoid the modal for speedy operations, and is also unlikely to be accidentally discovered by users. The most likely flow is that they attempt a standard paste with ctrl+v and then would see a hint in the modal that makes them aware of the modifier shortcut to bypass the modal:

image

That would allow us to avoid questions about where and how to persist the state for a "Don't show this again" checkbox and additional logic to check which "mode" the user is in to choose to trigger the modal or just paste the changes.

## Future wishlist

Finally, while playing with this functionality it made me think quite a bit about additional functionality I would like to see in Mathesar. In no particular order:

- Backspace shortcut to clear selected cell contents
- Display the count of selected cells/rows in the footer of the table view
- "Copy/paste cell value" options in the right click context menu (as you mentioned)
- Undo/redo for cell edits

I don't think we have GitHub issues for _any_ of those, yet, and I'd be happy to scaffold some new issues.

ghislaineguerin:
@seancolsen Looks good!
I agree with @zackkrida on the modal and I think it's a good idea to have some sort of command that allows to bypass it. We can make it more discoverable by also including it in the contextual menu and providing both paste options with the keyboard command displayed.

I know that the scope for this was to copy the data, but I was wondering if there are any concerns about it interfering later if we decide to support copying a row or a column. For example, in Table Plus, a row can be copied and pasted as a new record (it creates an INSERT sql statement for that record). I can see that being useful when there's a repetitive data entry task. Although that might be addressed by adding a duplicate row/column option to the contextual menu, it wouldn't allow users to replace all the content of an existing row.

This, of course, does not affect the copying of data, which looks good to me.

seancolsen:
@zackkrida

> a modifier key to allow for a simple way to bypass the modal

Neat idea! From a UX perspective, I like it!

From an implementation perspective, this would unfortunately require a non-trivial amount of additional work. If it were as simple as `if (event.modifiers.shift) needsConfirmation = false;` then I'd be inclined to go ahead and tack that on to this PR. But here's the thing: when I implemented the _copy_ functionality (almost 2 years ago, in #2773), I chose to rely on [ClipboardEvents](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent), which don't give us any control over modifier keys. If the user does <kbd>Ctrl</kbd>+<kbd>V</kbd> then a ClipboardEvent is emitted, but if the user does <kbd>Shift</kbd>+<kbd>Ctrl</kbd>+<kbd>V</kbd>, then no ClipboardEvent! Drat!

The good news is that there's a [newer Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard). It was not widely supported two years ago but now seems to be viable! Moving to the newer browser API would require modest research, a (very light) refactor of our clipboard system, and extensive cross-browser testing. It would come with unknowns, so it's possible there could be dragons lurking here. But (theoretically) the newer API would also allow us to implement paste via context menu (as @ghislaineguerin mentioned), which I'd really like to have! I could potentially take this on after this PR, still for 0.2.2. I'll leave it to you, @zackkrida to decide whether that ought to be a priority for 0.2.2. I'd want to budget 1 day for the initial research phase where I POC the new API and make sure it actually suits our needs. After that I'd have a better idea of the time needed to actually finish it. But it would probably be another 0-4 days of remaining work, as a rough guess.

> Backspace shortcut to clear selected cell contents

This is #936. Perhaps you'd care to weigh in there?

> Display the count of selected cells/rows in the footer of the table view

You might be interested in #2383 for this.

> "Copy/paste cell value" options in the right click context menu

Yeah, it's probably worth having a ticket to track this. I'll let you scaffold that if you're inclined.

> Undo/redo for cell edits

This is #2298

seancolsen:
@ghislaineguerin

Hmmm, I'm not entirely sure what your asking. You mentioned "copy" a few times. Note that this PR doesn't change anything about copying (which we already have). It only affects _pasting_.

I _think_ you're asking about how we could potentially issue `INSERT` queries in response to user paste actions.

In the [original specs](https://wiki.mathesar.org/design/specs/clipboard-interactions/#pasting-cells-to-mathesar), I had written:

> If the paste data extends beyond the rows displayed in the table, then new rows are added as needed.

But I didn't implement that behavior in this PR. Why? A few reasons:

- I wanted to cut scope to account for the unexpected challenges that I encountered on this PR.
- @mathemancer mentioned the idea of a paste operation potentially being confusing if it were to do an UPDATE _and_ and INSERT at the same time. This situation would be a user pasting into existing rows but the sheet not having enough rows to handle the size of the clipboard data. If the user happens to be near the end of the sheet only due to pagination, then would the user actually expect new records to be inserted? Or would the user expect records on the subsequent page to be updated? In considering this question more, I thought it'd be best to err on the side of avoiding the situation altogether. So in this PR, I show a toast error in that scenario. I don't think it's actually a very important scenario for Mathesar to handle.
- Being able to INSERT in response to a paste action (without UPDATE) would absolutely be useful. This is where the user's selection is entirely _beyond_ the sheet — in the "placeholder" row. I think pasting into the placeholder row would be pretty clear: users would expect it to insert new records. And it wouldn't update any existing records. I'd like to have this functionality! But (in addition to scope concerns) I didn't want to put that in this PR because it's actually hard to _get_ your selection into the placeholder row! You might _try_ clicking a cell in the placeholder row, but—_psych_—that won't select the cell! ...it will just insert a new blank record. If you select a cell in the last record and then press the down arrow key, _then_ you'll have a cell in the placeholder row selected. But that's a pain. Before we set about implementing INSERT via paste, I'd like to re-think our placeholder row a bit. I've [noticed](https://groups.google.com/a/mathesar.org/g/staff/c/yF34JjmY6MQ/m/vVfTwPe-AgAJ) it being a source of confusion during user testing. I'm interested in exploring a new UX wherein the user could click into the placeholder row without inserting a new record. The INSERT query would only happen once the user _changes_ something in the row, e.g. via editing a cell or pasting data. This will require more discussion though. (And I don't want to actually _have_ that discussion in this PR.) I'll raise it later on a separate thread.

> I'm thinking in the future we could offer different copy modes for the data as well, like plain text, csv, html, markdown table, and also "copy column" or "copy row".

Copying as a markdown table is a cool idea!

I don't quite understand what "copy column" and "copy row" would do, and you mentioned "copy row" a couple times. But I'm not sure this is the right time/place to flesh that out. If it's important to you, maybe open a separate issue where we could discuss these features there?

ghislaineguerin:
@seancolsen thanks for the thorough explanation. I am mentioning copy in the context of the data grid mostly. I know this PR doesn't touch it, but was mostly thinking of updating a row or column via paste and referenced copy as part of that user flow, since in order to do so the user must have a selection to copy first.

And yes, 100% agree with placeholder row needing some rethinking. I had the same experience during user tests.

pavish (approved):
@seancolsen This looks great!

I think the error states and edge cases are well handled.

I have a couple concerns:
1. I do not think copying foreign key values should result in the record summary string. I think it should always only be the value stored in the column.
    - It's essential that the user understands that the record summary is not the content stored on the db, but a summary of the actual referenced row. We already display the values differently on the UI, with a pill.
    - What happens when the foreign key is a text column referencing a text column from another table? How would copying such a value into a text column work? (Right now, record summaries will fail for it, but this is a question I'd like us to think about).
    - We could make this behaviour configurable in the future, but for now, I think it should directly copy the underlying value.
2. While copying empty string, we should take into account if the column has a not null constraint. If it does have a not null constraint, and it supports strings, we should paste an empty string instead of null.

I don't think either of these comments need to block this PR. I'm going to go ahead and merge it in. I'll create an issue to track these.

PR #4260, opened 2025-02-13
Export schemas

### LINKED ISSUE #4250: Allow exporting schemas
## Description

It would be useful to have a feature that allows users to export database schemas directly from Mathesar. This would potentially help users with portability and backup of their Postgres data.

## Proposed solution

- Implement a button on the schema page that allows users to export schema.
- Implement APIs that would call `pg_dump` for dumping the schemas along with its contents.

Anish9901:
Fixes #4250

Lets you export your schemas along with its table contents.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/views/export.py
```diff
@@ -71,6 +79,55 @@ def stream_table_as_csv(
...
+    with connect(database_id, user) as conn:
+        schema_name = get_schema(schema_oid, conn)['name']
+        pg_dump_cmd = [
+            'pg_dump',
+            '-h', conn.info.host,
+            '-p', str(conn.info.port),
+            '-U', conn.info.user,
+            '-d', conn.info.dbname,
+            '-n', sql.Identifier(schema_name).as_string(),
+            '-O'  # Don't include owner info in the dump
```
mathemancer, line 97: I think you should add the `--inserts` flag here. We should prioritize compatibility over performance, unless the performance really becomes a problem for someone.

I'd also like you to look into whether there's a flag to add that will get rid of the `SET` commands at the start of the file. Those can also cause problems when loading, and aren't usually needed.

mathemancer (changes requested):
Overall, pretty good. I have some changes to request for the actual `pg_dump` command.

mathesar/views/export.py
Anish9901: > I'd also like you to look into whether there's a flag to add that will get rid of the SET commands at the start of the file.

There is no such flag, unfortunately, users would have to take those statements out manually. Apparently, postgres includes those statements to make sure that upon restore, the server behaves exactly the same as when the dump was created. 

Given this, and the fact that we use `GENERATED BY DEFAULT AS IDENTITY` for `id` columns, would it make sense for us to prioritize compatibility over performance? The performance for someone dumping their local database and restoring it on a remote server would be terrible if there's a lot of data. Assuming this is a fairly common use case for someone using our product.

mathemancer (approved):
LGTM.

For context, for anyone watching, we're going to assume that we'll have `postgres-client` installed on the service layer for the foreseeable future.

mathemancer:
@pavish we'll wait to merge this till you review the front end chagnes.

pavish (changes requested):
@Anish9901 

- The export fails whenever the schema has at least one table that the user does not have select privileges on. For this scenario, I'd either expect the "Export" button to not be enabled, show a valid error, or export only the tables that the user has access to.
  - I do think there might be additional cases where export would fail due to permission related issues and we have to test it all.
- I'd like the Export button to be placed within a dropdown menu. For eg., Similar to the dropdown in the Database Page which contains the Disconnect Database option.
  - The chances that the user would export the entire schema as a SQL file is low that it does not warrant us placing the Export button in such a prominent location.

Anish9901:
@pavish I agree with your critique,

> "Export" button to not be enabled, show a valid error, or export only the tables that the user has access to.

As far as export behavior is concerned, out of these three, I'd prefer the first option. I don't think this feature would be of much use to the user if we only dump a subset of tables from their schema. They could do that anyway with per table CSV export.

> I'd like the Export button to be placed within a dropdown menu. For eg., Similar to the dropdown in the Database Page which contains the Disconnect Database option. 

I agree.

That said, it would take me a while to figure out these changes on the frontend, and given that we want to cut our release tomorrow, would you be able to take over the frontend changes required? So that we can get this in before the feature freeze?

zackkrida:
@Anish9901 I do think @pavish's feedback should block this PR from inclusion in the current release if it can't be resolved by end of tomorrow. If it can, great!

I also think the triple-dot menu that triggers a dropdown, like that on the DB page, is also a good choice for the design. 

Lastly, when I think about use cases for functionality like this, I can imagine there are users who would _only_ want to export the structure of a schema and not just the data, depending on the scenario. It would be nice to offer both options in the future. There's an incongruity in the current approach where we offer csv data export at the table-level but sql data export at the schema level and both of these are labeled with the same "Export" button. I could see that being confusing. But I think it can be optimized in the future.

pavish:
@Anish9901 The UI changes are not my primary concern, I could get it done in this PR. However, I'm not entirely sure about our product strategy with exporting schemas. 

The fact that export would fail even if one table is owned by another user is not a good idea. I'd rather have the functionality working for all users, and export only the data that the user has access to. Until we have that figured out, the other reasonable option is to limit this functionality to database superusers only.

I'd rather wait a couple days to resolve this (and get it into the next release) rather than rushing the PR just to get it in this release.

pavish (approved):

zackkrida:
@Anish9901 @pavish @mathemancer:

I also think the PR _could_ be merged in it's current state. I really appreciate the improvements you've made, @pavish. However, I think it is best to **draft this PR and save it for the next release** for the following reasons:

It isn't clear to me yet who this functionality is for and what goal its intended to serve. When we _do_ include it in a release I'd like to be able to give a clear impression of the value it offers to our users. Furthermore, if some discussion from the team and a refinement of this PR could make it _significantly more_ useful to users, I'd like to take the extra time to do so before we include it in a release. 

I'll draft this now just so it's clear that it shouldn't be merged in the short term.

Anish9901:
@pavish @zackkrida Here are some of my thoughts:

### The problem with exporting a subset of tables as SQL:

- When dumping a subset of tables, using `-t` flag, the information about the schema structure gets lost.
    - This means not only the schema structure definition is lost, but any objects other than the table, like functions, custom types, etc which are stored within the schema, that the table might depend upon would be lost. Without these objects, the restore might fail, and the dump will be useless.
- If someone only has `SELECT` only on a referer table, but not on the referent(s), the information about the constraints is still included in the dump. And the restore would eventually fail if the referent table(s) are not present on target db. 
- For dumping a table successfully, you not only need `SELECT` on that table you also need, but you also need `SELECT` on the sequences used by that table.
- Given that a lot of our users want READ ONLY setups, and the fact that we don't even support granting permissions on the sequences through mathesar, this feature would be useless for someone with only `SELECT` on a table.
- Exporting CSV on the other hand requires only `SELECT` privilege on the table and is more portable, and accessible for someone wanting to use it in other applications(e.g. Excel, Airtable).
- All this to say, that if anyone with limited privileges on the db wants their data out of postgres, SQL is probably not a format that would be suitable for them. Also, having "schema exports" without schema structure definition makes no sense to me.

### Who is this feature useful for?
- Anyone trying out mathesar for the first time or a long time user who has their data in one of our internal dockerized db wanting to move their data to a more robust postgres platform would use this feature.
- Admins wanting a regular backup of their data, that can be easily restored. Given that we are still in beta, and are prone to make mistakes, it would induce confidence in users if backups can be taken easily.

### How can we extend this in the future?
- Allowing to restore these dumps easily through mathesar (would require design and probably would take more than a couple release cycles).
- Allowing to export all the accessible tables as CSVs within a zip/tar file in one click from the schema page.
- [Brent's idea] Have some metadata that would allow us to restore mathesar specific metadata along with the dump, e.g. Explorations.

zackkrida:
@Anish9901 I'm going to close this. Based on some of our past discussions, we will want to refine how we're allowing users to export from Mathesar before moving forward.

We can either re-open this in the future, or use it as inspiration for a future implementation. Thanks for this work!

PR #4317, opened 2025-03-08
Fix problem with timezones and date/time keywords

### LINKED ISSUE #1694: Entering 'now' on a datetime column without timezone support shows different time from what user expects
## Description
* Edit a datetime column with type 'timestamp without time zone'
* Notice the Date picker showing the client local time in the 'now' button. Eg., `19/09/2022 20:54`
* The request sent to the server contains 'now'.
* The server response contains the value with a timezone attached (belonging to the server). Eg., `2022-09-19T15:24:54.468139 AD`
* The frontend discards the timezone since the column is configured not to contain time zone information and displays `19/09/2022 15:24` while the user expects `19/09/2022 20:54`.

Note: For reproducing this bug, the browser timezone should be different from the Mathesar DB's timezone.

## Behaviour when direct value is entered
* Instead of 'now', type in `19/09/2022 20:54`.
* Notice the request sent to the server being `2022-09-19 20:54:00`.
* The response sent from server after saving is `2022-09-19T20:54:00.0 AD`.
* This is displayed correctly because the timezone offset here is 0.

## Expected behavior
* The value expected by the user needs to be shown to the user.

## Discussion - options
* Should the frontend send the time for 'now' instead of now? This will result in other postgres keywords unable to translate to the client timezone.
* Should the frontend also consider the timezone for this column? But in this scenario, a different client in a different timezone will see a different time, even though this column is supposed to only show the same static datetime values everywhere. This would go against the usefulness of this column type.
* Can the backend take the client timezone information from request, and store & return the value accordingly in responses? [This would make it a lot easier for the frontend]

seancolsen:
Fixes #1694

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/components/cell-fabric/data-types/components/date-time/DateTimeInput.svelte
+++ b/mathesar_ui/src/components/cell-fabric/data-types/components/date-time/DateTimeInput.svelte
@@ -115,7 +115,7 @@
     bind:value
     isRelative={allowRelativePresets}
     {type}
-    {formattingString}
+    {formatter}
     on:change={close}
   />
 </AttachableDropdown>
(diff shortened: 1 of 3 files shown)


pavish (approved):
Looks good to me!

PR #4329, opened 2025-03-13
support uuid casts and inference

### LINKED ISSUE #4312: Implement backend functions for UUID support
_TODO: This issue needs an expanded description._

As part of #4310 we need to add supporting backend code to Mathesar which enables us to cast columns to and from UUID to other types. 

We also need logic to generate pseudorandom UUIDv4 values with [gen_random_uuid](https://www.postgresql.org/docs/current/functions-uuid.html#FUNCTIONS-UUID).

Anish9901:
Fixes #4312

Adds casting functions and inference for the uuid type.

**UUID inference in action:**

Screenshot 2025-03-13 at 4 59 28 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


db/sql/05_msar.sql
```diff
@@ -3642,13 +3643,23 @@ Args:
 DECLARE
   default_expr text;
   raw_default_expr text;
+  easy_dynamic_defaults jsonb := '[
+    "now()",
+    "current_timestamp",
+    "current_date",
+    "current_time",
+    "gen_random_uuid()"
+  ]'::jsonb;
```
mathemancer, line 3652: I don't like adding these dynamic defaults in this way:

- It's brittle to transmit this info as strings (in this direction)
- We don't need it for any current or desired functionality in this PR or cycle.
- We already have the default generation for pkey columns handled through other means. We don't need to offer users the option to set dynamic defaults by typing them in at this time.

Please remove (or argue in favor of keeping) this.

mathemancer (changes requested):
I think this is good, except for the dynamic default setting functionality.

db/sql/05_msar.sql
Anish9901: Removed,
It is mentioned in the issue description that we want this functionality,
> We also need logic to generate pseudorandom UUIDv4 values with [gen_random_uuid](https://www.postgresql.org/docs/current/functions-uuid.html#FUNCTIONS-UUID).

FWIW, I only included the defaults that were super easy to set and require no modification, I wouldn't want the users to type these defaults as a string either, I assumed we would have some way for the user to choose defaults based on their column types on the frontend.

> We already have the default generation for pkey columns handled through other means.

That's true but I don't think there would be a way for someone to set defaults for a non-pk uuid column which the user might create via import, its debatable whether we want to allow this though.

mathemancer: Long-term, we do want to handle this, but I'd like to have a coherent product strategy first for how we want to deal with dynamic defaults in general, and then apply that to the UUID case.

PR #4337, opened 2025-03-13
Refactor records store to improve readability

pavish:
This PR refactors the records.ts file:
- `Row`  is now a discriminated union of different classes
- `identifier` is used as the defacto field to uniquely identify rows.
- `rowIndex` is moved to `display.ts` as `rowNumber`.
    - A `DisplayRowDescriptor` is used to wrap Rows and only the record rows that need to display the rowNumber contain this property.
- `rowKey` is removed and no longer used.
    - `identifier` replaces `rowKey`.
    - The type `RowKey` is still present and refers to the `identifier`.
- Fixes a bug with the rowNumber when there are a combination of persisted & drafted new records.
- Implements a minor UI change to display the number of newly saved records & unsaved records.

Note:
- I raised this PR to make it easier to work on PK related issues assigned to me. As such, I would be stacking further PRs on top of this one.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (changes requested):
This is definitely an improvement!

In the future, I'd really appreciate having an opportunity sync up before launching into refactors like this. Since I had my head in this precise problem quite recently, there are a number of things I would have done differently with the refactor. But I don't think it's worth spending extra time to change them now, on top of the time you've already spent.

That said, I have very little work time left this cycle, so I'm not going to spend a ton of time critiquing this. I want to get it merged and move forward. I've focused most of my effort on QA.

I found some bugs:

- When I try to paste cells, I don't see the cell update. And I see "Saving Changes" at the top of the sheet. But it never resolves.
- After grouping the sheet, I get a JS error when clicking on a data cell.

During QA phase, I think it would be good to instruct our QA testers to hammer at row-related features pretty hard.

Questions:

- Why the changes to `getGloballyUniqueId`? What was wrong with it before?

pavish:
@seancolsen This is ready for review again.

> In the future, I'd really appreciate having an opportunity sync up before launching into refactors like this.

I agree, we should be doing that for any significant refactors. I proceeded with this PR since I wasn't making any architectural changes and it was faster for me to proceed and then sync up for an actual architectural refactor, when we have the bandwidth.

> When I try to paste cells, I don't see the cell update. And I see "Saving Changes" at the top of the sheet. But it never resolves.

Fixed in https://github.com/mathesar-foundation/mathesar/pull/4337/commits/b33da21c30e94017856a1843bd84e48052a1895d.

It was yet another case of using the primary key instead of a row level identifier.

I didn't like the fact that `paste.ts` handled checking the primary key column and using that as part of the blueprint. Sheet is a common utility and as such, I disagree that it should be handling this. I've moved that part to `records.ts`, and used `row.identifier` for the key, which simplifies this flow and keeps concerns separate.

> After grouping the sheet, I get a JS error when clicking on a data cell.

Fixed in https://github.com/mathesar-foundation/mathesar/pull/4337/commits/b6c18136969ef6081a2bfc6b6f504024b0b7cf0f.

seancolsen (approved):
Looks good. Thanks @pavish.

PR #4344, opened 2025-03-15
Revert custom line-height on Dropdown

### LINKED ISSUE #4332: Short line height in dropdown label causes text clipping
This has been bugging me.

There are lots of way to reproduce this. Here's one:

1. Add a role named `jumpy_quail_grabs_yogurt`.
1. Configure a collaborator with that role.
1. Observe text clipped:

    Image

There are a handful of places throughout the codebase that have `line-height: 1;`. In fixing this issue, I think we probably ought to audit these and see if other places need the same fix. My understanding is that line-height shouldn't go below 1.2.

I put this in 0.2.2 because the problem is somewhat glaring on the UI I'm building for #4313

Image

seancolsen:
Fixes #4332

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/dropdown/Dropdown.scss
+++ b/mathesar_ui/src/component-library/dropdown/Dropdown.scss
@@ -9,7 +9,6 @@ button.dropdown.trigger {
     text-overflow: ellipsis;
     white-space: nowrap;
     box-sizing: content-box;
-    line-height: 1;
     min-height: 1em;
     flex-grow: 1;
     > *:not(:last-child) {


seancolsen:
@pavish in #4332 you said:

> I wanted the dropdown to have the same height as other input fields such as buttons and textbox

Cool, I like this goal.

I'm continuing the conversation here, now specifically within the context of the changes I'd like to make.

I'm confused by your comment though. Because your addition of `line-height: 1` in `Dropdown.scss` seems to have made the heights _less_ consistent, not more.

On `/administration/users/1/` we have Input, Select, and Button components. I measured their height across three browsers at 100% and 200% zoom. Here is what I found (before this PR):

- Element heights (in px):

    | Element | Firefox       | Chrome  | Safari      |
    | --      | --            | --      | --          |
    | Input   | 34 (33.5 @2x) | 30.8    | 33          |
    | Select  | 28            | 28      | 28          |
    | Button  | 30.8          | 30.8    | 30 (31 @2x) |

    _@2x = when zoom = 200%_

This PR makes `Select` use the same height as `Button`.

- In Chrome, all three heights become identical.

- Firefox and Safari are a little weirder because:

    - They appear to be using a minimum line-height for `<input>` elements, perhaps for the sake of a11y.
    - Page zoom seems to have some subtle effects that I don't fully understand.

    I got sucked into a rabbit hole researching the difference in line-height behavior for `<input>` elements across different browsers, but I think if we just go off of Chrome it'll simplify this discussion.

Even in FF and Chrome, this PR would bring the height of `Select` _closer_ to the height of `Input`, which seems like a win. It still wouldn't be identical, but I don't want to spend time perfecting it right now when I'm ultimately just trying to solve the problem of clipped text.

Will this work for you, @pavish?

In #4332 you said:

> we should make that change for all input elements.

I'm not seeing a need for other changes. From what I can tell, other input elements already have line-height of 1.2 (via `inherit`). If you could point out a specific problem that this PR creates, I'd be very curious to see!

PR #4348, opened 2025-03-17
Allow editing PK cell when inserting rows, when PK column does not have a dynamic default set

### LINKED ISSUE #4245: Allow `INSERT` of primary key values
## Description
When entering data into a table that has e.g only a primary key and no other columns, the key cannot be set.

Image
 

## Expected behavior
It is possible to set the key

## To Reproduce
Create table
```sql
CREATE TABLE "Table" (
    "key" TEXT NOT NULL,

    CONSTRAINT "Table_pkey" PRIMARY KEY ("key")
)
```
Try and create a row in Mathesar

## Additional context

- [Matrix conversation](https://matrix.to/#/!vXLxAqmrJWsDMWPSpo:matrix.mathesar.org/$O8-PSUuAvh9EDAM3KK61SM7KjePZWR_86pkATNv3EN4?via=matrix.mathesar.org&via=matrix.org&via=t2bot.io)

## Similar issues

- #2685

pavish:
Fixes #4245

- Has some subtle UX changes w.r.t client side errors on rows. This video should demonstrate that.

https://github.com/user-attachments/assets/3a5c75af-aff5-4abc-8f44-d7f9e7deccd9
- Client side cell errors now show an '*' to indicate that the values are required.
  - This should be sufficient to get the message across.
  - The Row control shows the list of columns that do not accept null.
- There are several existing bugs with cell selection in our `develop` branch when it comes to adding new rows.
  - I tried to attempt fixing some of them in this PR, but it seemed to require a lot more effort than I had hoped. I'll be creating issues for them for us to tackle later. 
  - I would eventually like the first editable cell to focus and enter edit mode when attempting to add a new row.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
@seancolsen Please review code
@zackkrida Please review the changes to the UX

zackkrida:
@pavish I spent some time this morning playing with this and I like all of the design decisions you've made concerning cell background and the required asterisks.

Two small thoughts:

- Not very related but as part of this PR could you add `word-wrap: break-word;` to the error container so we don't have the the overflow issue visible in the last frame of the video? If this isn't as small as a change as it seems to me, could you create a new issue for it?
- What would you think if we added some additional text next to "New records will be repositioned on refresh." which also said something like "Required fields are marked with an asterisk (*)."? It could be a good way to use existing negative space in the UI to give the user additional information.

pavish:
@zackkrida 

> Not very related but as part of this PR could you add word-wrap: break-word; to the error container so we don't have the the overflow issue visible in the last frame of the video? If this isn't as small as a change as it seems to me, could you create a new issue for it?

Sure, it's added in https://github.com/mathesar-foundation/mathesar/pull/4348/commits/8e6d905a18173a6118842d04fb8ba1c6fa689f52.

> What would you think if we added some additional text next to "New records will be repositioned on refresh." which also said something like "Required fields are marked with an asterisk (*)."? It could be a good way to use existing negative space in the UI to give the user additional information.

The negative space here exists because the size of the row is higher. For tables with just a couple columns, it would look like this.

Screenshot 2025-03-18 at 8 17 52 PM

I think using `*` is a well-established pattern for required fields in forms and users would be able to understand it quite clearly, and we show the message while hovering on the row-control.

I'd leaning towards not adding the additional text. However, if you have a strong opinion, I'm fine with adding it. What do you think?

zackkrida:
@pavish 

> The negative space here exists because the size of the row is higher. For tables with just a couple columns, it would look like this.

Good point, I'd agree it's not a good use of the space. No strong opinion from me.

seancolsen (approved):
Looks good!

PR #4375, opened 2025-03-26
Fix bug with renaming columns on import

### LINKED ISSUE #4372: Custom column names at import aren't saving
## Description

If you import data into Mathesar and rename the columns using the inputs in the "Table preview", the names you add will not be saved, and instead the existing names (when "Take the first row as column names" is set) or auto-generated names (when "Set generic column names, keeping the first row as data") will be used.

## Expected behavior

This is a regression and the prior behavior, where names were saved, is ideal.

## To Reproduce

⚠️ See updated repro steps [below](https://github.com/mathesar-foundation/mathesar/issues/4372#issuecomment-2753013392)

1. ~~Start an import with some sample data, for example:~~
   ```csv
    id,first_name,last_name,age
    1,Sarah,Smith,20
    2,Jane,Wong,34
    3,Ethan,Park,18
    4,Michael,Leigh,22
   ```
2. ~~Attempt to rename the columns in the table preview and complete the import.~~
3. ~~Observe that the column names were not saved.~~

## Environment

On develop/Mathesar 0.2.2

seancolsen:
Fixes #4372

## About the underlying problem

In one of the many steps we perform when confirming an import, the front end hits `columns.patch` in order to update the column names and types. Prior to this PR, it sent `name`, `type`, and `type_options` for every column in the table. So in Mathesar 0.2.1 even for the auto-generated PK column (which we don't allow the user to cast), the front end was still saying:

> yes DB-layer, please change the type of the `id` column to "integer"

The backend was able to handle this just fine. It should be a no-op after all.

But for some reason the backend is _not_ able to gracefully handle that same sort of request when made against columns that originated from the CSV and were subsequently configured to have IDENTITY defaults. (Interestingly, I noticed that if you choose a pre-existing numeric column and take the extra step of changing the default to None, then this bug is not present.)

I haven't dug into the backend code, but just from looking at the error message, I can spot this:

```sql
ALTER TABLE "Table 7"
ALTER COLUMN a DROP DEFAULT,
ALTER COLUMN a TYPE integer USING msar.cast_to_integer(a),
ALTER COLUMN a SET DEFAULT ::integer,
--                         ^^^^^^^^^ 👈 Here is the syntax error
ALTER COLUMN name DROP DEFAULT,
ALTER COLUMN name TYPE text USING msar.cast_to_text(name)
```

Perhaps one could argue this is a backend bug. I don't know.

## About the fix

In the end I chose to fix this on the front end by just never sending `type` and `type_options` _for PK columns_. The front end does not allow the user to change those settings for PK columns, and I think it ought to stay that way.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/pages/import/preview/importPreviewPageUtils.ts
+++ b/mathesar_ui/src/pages/import/preview/importPreviewPageUtils.ts
@@ -1,6 +1,6 @@
 import { dataFilesApi } from '@mathesar/api/rest/dataFiles';
 import type { DataFile } from '@mathesar/api/rest/types/dataFiles';
-import type { Column } from '@mathesar/api/rpc/columns';
+import type { Column, ColumnPatchSpec } from '@mathesar/api/rpc/columns';
 import { getCellCap } from '@mathesar/components/cell-fabric/utils';
 import type { Schema } from '@mathesar/models/Schema';
 import type { Table } from '@mathesar/models/Table';
@@ -91,17 +91,34 @@ export function buildColumnPropertiesMap(
   );
 }
 
+function finalizeColumn(
+  { id, type, primary_key, type_options }: Column,
+  name: string | undefined,
+): ColumnPatchSpec {
+  return {
+    id,
+    name,
+
+    // For most columns we include type information so that users can modify
+    // column types during import.
+    //
+    // But for PK columns we don't want to send these details to the backend. In
+    // most cases it wouldn't matter if we sent the type details, because we
+    // disable the type config form elements on the front end and it would be
+    // theoretically be a no-op to send the type details that we got back from
+    // the server. However in [#4372] we had a slippery bug that seemed best to
+    // fix on the front end by avoiding sending type details for PK columns.
+    //
+    // [#4372]: https://github.com/mathesar-foundation/mathesar/issues/4372
+    ...(primary_key ? {} : { type, type_options }),
+  };
+}
+
 export function finalizeColumns(
   columns: Column[],
   columnPropertiesMap: ColumnPropertiesMap,
 ) {
   return columns
-    .filter((column) => columnPropertiesMap[column.id]?.selected)
-    .map((column) => ({
-      id: column.id,
-      name: columnPropertiesMap[column.id]?.displayName ?? '',
-      type: column.type,
-      type_options: column.type_options,
-      metadata: column.metadata,
-    }));
+    .filter((c) => columnPropertiesMap[c.id]?.selected)
+    .map((c) => finalizeColumn(c, columnPropertiesMap[c.id]?.displayName));
 }


seancolsen:
@pavish I'd like you to review front end code.

@mathemancer I'd like you to read over my assessment of the problem and solution to make sure it seems sane to you.

zackkrida (approved):
@seancolsen your code makes sense and the comments are useful. I see that column `metadata` is no longer being returned for _any_ column regardless of whether it is a primary key. Are you certain that wasn't being used for anything? If so, LGTM.

seancolsen:
> I see that column metadata is no longer being returned for any column regardless of whether it is a primary key. Are you certain that wasn't being used for anything?

My, what a keen eye you have, Zack!

Yes I'm certain that removing `metadata` is safe here. Here's why:

1. In the back end, we can look at the [columns.add RPC handler function](https://github.com/mathesar-foundation/mathesar/blob/3c095f9b1cf7e85c859bf1fc424c22e756f00198/mathesar/rpc/columns/base.py#L282). 
    - It's [SettableColumnInfo](https://github.com/mathesar-foundation/mathesar/blob/3c095f9b1cf7e85c859bf1fc424c22e756f00198/mathesar/rpc/columns/base.py#L118) parameter type does not have a `metadata` property.
    - But being that we current do not use tooling to enforce types on the back end, we can quickly trace this function body down too, and see that `column_data_list` only every gets passed to the DB layer. So there is effectively no way it would be using an undocumented `metadata` property. 

1. On the front end, this means that our [ColumnPatchSpec](https://github.com/mathesar-foundation/mathesar/blob/3c095f9b1cf7e85c859bf1fc424c22e756f00198/mathesar_ui/src/api/rpc/columns.ts#L191) type is correct — it does not have a `metadata` property. Because we _do_ use a type-checker on the front end, we can be quite confident we're safe! Yay TS! 🙂

---

Read on, only if you're curious...

Ok, so then why was `metadata` there?

- It was likely left over from and old REST API that _did_ support metadata changes in the same API as column type/name modification. Now we handle this sort of combination via RPC batching.

Ok, so then why was there no TS error _before_ my change?

- I actually removed metadata because TS complained at me after my changes. Gosh I ♥ TS!

- It's kind of fun to see why there was no TS error before....
    
- Prior to my change, [finalizeColumns](https://github.com/mathesar-foundation/mathesar/blob/3c095f9b1cf7e85c859bf1fc424c22e756f00198/mathesar_ui/src/pages/import/preview/importPreviewPageUtils.ts#L117) was returning an array of `ColumnPatchSpec` _subtypes_ — objects that conformed to the `ColumnPatchSpec` interface but had an extra "metadata" property too. That subtype got fed into [updateTable](https://github.com/mathesar-foundation/mathesar/blob/3c095f9b1cf7e85c859bf1fc424c22e756f00198/mathesar_ui/src/stores/tables.ts#L224). Because type substitution is contravariant in function parameters, there was no TS error here. A subtype can be substituted for a supertype as a function argument.
    
- After my change, I added `finalizeColumn`. In doing so, I added an explicit return type. (I could have left this to inference, but it seemed like a good idea here.) But function return types are _covariant_! So I can't substitute a subtype in place of a supertype. I had to make the return value conform to `ColumnPatchSpec` _exactly_.

Anyway, sorry for the long winded comment. Honestly, writing this was an opportunity for me to shore up my understanding of variance. I only just began learning about this after seeing the TS variance annotations feature added to TS somewhat recently.

zackkrida:
Thanks @seancolsen. That was an enjoyable read.

mathemancer (commented):
I agree we should also fix this behavior so that sending type details is a no-op in the back end.

Issue #4313, opened 2025-03-06
Choose or set primary key when importing data


zackkrida:
Blocked by #4311 and #4312.

As part of #4310 we need to enhance Mathesar's import flow to allow users to:

- Choose the type of Mathesar's default primary key column
- Choose an existing column as the PK instead of using Mathesar's default primary key 
    - Restrict the user to choose from columns that seem "valid" (matching Mathesar's UUID or INTEGER column types)
    - Make it clear that any semantic meaning will be lost (i.e, future integers will be sequential based on the current max value, UUIDs will be pseudo-random.)

## Example mockups

_These are very hasty in-browser mockups to convey the **general idea** of this work. **Do not consider them final implementations.**_

> The user choosing an existing ID column as their PK
![Image](https://github.com/user-attachments/assets/f0115b33-924e-437b-98ec-58a29b3d446b)

>  The using choosing the UUID type for the default PK column
![Image](https://github.com/user-attachments/assets/e1d14770-1608-4f9f-a9ba-7702e46854ac)

>  The using choosing the standard ID type for the default PK column
![Image](https://github.com/user-attachments/assets/2e6ffc2f-5831-487a-863d-59b58ac2ebfb)

zackkrida:
@pavish any updates on how the flow here should behave based on your conversations with @mathemancer?

pavish:
@zackkrida 

The default selected option would be "Let Mathesar add a Primary Key Column" and it would be an integer column named `id` as we currently have. Brent and I are agreed on the flow for modifying it as described in the issue.

We had a few concerns/UX suggestions:
1. Move the Primary Key section closer to the table & differentiate it from the other two sections.
   - Any modification in the Primary key section does not & should not affect the whole table.
     - Eg., The names & types of other columns are not changed.
   - This is in contrast to the other sections, where modifications affect the entire table.
     - Eg., modifying header setting or the column inference setting updates the entire table again with different column names and types.
2. Hide the Primary key section by default within an "Advanced" section. 
    - We do not expect most users to get into the weeds of configuring the PK to this extent.

zackkrida:
@pavish that all sounds great to me with one exception: what is the reason to not support uuid as an option for "Let Mathesar add a primary key column"? Not doing so creates an inconsistency in the import and table creation behavior and prevents users from using the same type of primary key consistently across imported tables and tables created in Mathesar. I strongly think we should support it.

pavish:
@zackkrida 

[...]

PR #4376, opened 2025-03-26
Fix CSS issue with inline layout of LabeledInput component

### LINKED ISSUE #4373: Primary key import settings should render inline-block
## Description

In #4313 we added a new block of settings to configure the primary key at import time:

Image

There is a small CSS bug on the form elements in this section. Each label + field combination is displaying `block` and fills 100% width of the form. This means that the whitespace at the end (right side) of each field is clickable. 

## Expected behavior

The negative white space to the right of each field should not be interactive.  

## To Reproduce

In `develop` / Mathesar 0.2.2

## Additional context

Fixing this globally would require making our `.labeled-input` containers `inline-block`, which would break many of the other forms in Mathesar. It's likely that this will be best-fixed with a local override and custom style for the `CollapsibleFieldset` component or even for this singular implementation of said component.

CC @seancolsen for frontend approval and advice for how to proceed.

seancolsen:
Fixes #4373

## Notes

Hopefully this isn't too reckless, but I've made this PR against 0.2.2 even though it makes a  far-reaching change to a low-level component.

I think it's safe though.

The new CSS just says:

> Don't make a `LabeledInput` component wider than it needs to be.

This affects a lot of UI, but in a very small way. It means the click targets are more contained. Clicking on empty space no longer targets a `<label>` element that contains an `<input>` element. The problem was especially evident on the UI I added in #4313, but it was all over the app. Ironically enough it's also present here:

Image

That overly-large click target is on the same page as the problem mentioned in the linked issue, and it's been there a _long_ time. But it's just less annoying because the associated input is just text, so it doesn't open a dropdown.

I did a pretty thorough audit of our usage of `LabeledInput` and a lot of spot-checking. In addition to fixing the linked issue directly, here are some other areas of Mathesar that now have narrower `<label>` elements so as to avoid turning negative space into unnecessarily large click targets:

 

 

 

 

 

There are more as well, but I think this should demonstrate the change well enough.

---

And here's an example of one that is effectively unchanged to the user because its content is so wide:

Image

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/labeled-input/LabeledInput.scss
+++ b/mathesar_ui/src/component-library/labeled-input/LabeledInput.scss
@@ -42,6 +42,15 @@
     }
   }
 
+  &.layout-inline,
+  &.layout-inline-input-first {
+    // This is to avoid having the `<label>` element take up the full width of
+    // its parent. If that happens it can become an annoyingly large click
+    // target. The problem is especially evident when a `Select` component is
+    // used as the input element.
+    max-width: max-content;
+  }
+
   &.layout-inline .label-content {
     display: flex;
     flex-direction: row;


seancolsen:
@pavish please review code.

@zackkrida I'd like you to take a look at this too, just to make sure you're comfortable merging this into 0.2.2 this late in the release cycle. It comes with some risk. So if you'd prefer to bump this to 0.2.3, I can cherry-pick it over to a new branch. That way this change would undergo more testing before being released. Personally I feel good about forging ahead with it, but I'll leave the final call to you.

zackkrida (approved):
@seancolsen your solution works great, I've clicked around pretty extensively and I'm comfortable including this in 0.2.2.

PR #4411, opened 2025-04-14
fix matrix_tests

### LINKED ISSUE #4405: `matrix_tests` job should fail if any dependency fails
## Description

Apparently, the behavior of GH actions has changed since we set up our test-running strategy in this PR: #3254 . In that PR, lots of effort went into testing that the matrix_tests job failed whenever a dependency failed. The problem is that now the matrix_tests job, which is supposed to collect upstream successes or failures, is now skipped rather than failed if a dependent job fails:

Image

This is a big problem, since it means we need to add a new branch protection status check rule every time we add a new version to test in our testing matrix. This is unacceptable and too difficult to maintain. Currently, for example, it seems we could merge with failing python tests even after @zackkrida 's recent changes to the branch protection rules.

Here's an issue and a dirty hacky solution to the problem: https://github.com/actions/runner/issues/2566 . You can also see that there's lots of other repos that have crashed into the same problem.

## Expected behavior

The `matrix_tests` job should fail if any dependency job fails. Then, we should update GH branch protection rules so that the check is on the `matrix_tests` rule, not on each dependency.

## Additional context

We noticed the problem in the unfortunate merging of the broken PR #4397 .

Anish9901:
Fixes #4405

The matrix-tests job now always results in a failure if "any" of the upstream job fails and returns success if "all" upstream jobs succeed or are skipped.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
Looks good so far!

But, I think we should add the other required checks to the `matrix_tests` collector. Maybe we should then rename it. Things to add to the collector:

- Check front end code format
- Check front end types
- Audit front end code
- Run Python linter
- Run front end tests
- Run front end linter

_Avoid_ adding the `Find unused code` job, since that's just a warning (at the moment).

mathemancer (approved):
Okay, LGTM. Nice work!

Issue #4370, opened 2025-03-24
Improve type casting and inference


mathemancer:
## Problem

Our current type-wrangling system is showing its age:
- Type inference is tied to casting
- Inference is therefore less powerful and slower than it should be
- Casting, on the other hand, is less flexible than it should be.

PR #4426, opened 2025-04-24
Pass casting options when importing

mathemancer:
Related to #4370 

Passes the details inferred for casting a text column to a type during import back to the
- `tables.get_import_preview` function when previewing a table, and
- `columns.patch` function when saving the table.

**Technical details**

No visible functionality change in this PR, since this is the front end work to take advantage of back end performance improvements which aren't yet implemented.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (commented):
Neat to see PR from you with so much TS! 🙂

This all looks mostly fine to me.

- I got confused when reading this though:

    ```ts
    export type ColumnCastOptions = Record<string, unknown>;
    ```

    I spent quite a while trying to track down some more information about this type in order to understand what it is and what it does.

    - Working my way forwards, I went looking at where this value _goes_. I ended up looking in python code but I didn't see any `cast_options` properties noted in `PreviewableColumnInfo` or `SettableColumnInfo`. I assume this is code that will come later. Yes? In general I don't love the idea of putting types into the front end before the corresponding API changes are made on the backend. But I'm okay with it in this case because I can see how it wouldn't break anything.

    - Working my way backwards, I went looking at where this value _comes from_. I ended up noticing that `data_modeling.suggest_types` returns a `details` property which appears to be the same value. So I pushed efe514f344cbae66d3c49cfe9e48106c9457b239 to make that connection clearer to other devs reading this code. I spent a little while trying to trace down some more info about that `detail` property down into the service layer and the DB layer but I eventually gave up.

    - I inspected the runtime behavior while importing a CSV and noticed that a numeric column got this value

        ```json
        {
          "decimal_p": ".",
          "mathesar_casting": true
        }
        ```

        Okay, that helps me understand this a little better.

    I'd like to see some changes that help other devs (even future me) avoid this goose chase in the future.

    I would even consider doing this:

    ```ts
    /**
     * (Insert description here to explain what's going on.)
    */
    export type ColumnCastOptions = unknown;
    ```

    It doesn't seem like the front end ever needs to index into that object, so typing it as `unknown` papers over some of the complexity here and indicates to front end devs that they needn't concern themselves with the inner-workings of this value at all. And a code comment explaining what it is at a high level would help to make it even more explicit that this only relevant to back-end stuff.

    Basically, if I see a type like `Record<string, unknown>` I want to understand what the entries represent and what kinds of things are used as keys. To be clear: seeing `unknown` is perfectly fine because _that_ tells me it's something I don't need to worry about.

mathemancer:
@seancolsen Actually, while I prototyped the frontend code, @pavish was the implementer of this PR. Should be ready for re-review, regardless.

seancolsen (approved):

PR #4452, opened 2025-05-09
Fix CI

Anish9901:
This PR fixes the following error that was occurring for the past couple of days in our CI(probably due to rate limits):
```
failed to solve: python:3.10-bookworm: failed to resolve source metadata for docker.io/library/python:3.10-bookworm: failed to authorize: failed to fetch oauth token: Post "https://auth.docker.io/token": EOF
```
- Instead of pulling the python & postgres docker images several time for each job instance, we cache all the versions of the image that we test against, and we load it separately for each job instance.
- The `cache-images` job runs only when backend tests are required and create a new cache only when there is a cache miss.
- `cache-images` is not set to be a `need` for backend test jobs as most of the time the cache won't be a miss as the key is deterministic, and there is no point in waiting for `cache-images` jobs to succeed given that there isn't going to be a cache miss.
- Under the cases when cache is a miss, however, a new cache will be created during the same run by `cache-images` and we'll have to manually re-run the failed jobs.
- Any changes in `'**/Dockerfile', '**/api_tests/Dockerfile', '**/Dockerfile.caddy', '**/Dockerfile.devdb'` would invalidate the previous cache and would require a new cache to be created as the cache key is depends on their hash.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer (changes requested):
I think the cache-missing-hitting logic has too many bells and whistles, and we should remove it. It's also a bit faulty (see below). Instead, just grab a fresh version of each needed python and/or postgres image for each CI run, when any dependent jobs are needed. See:

https://docs.docker.com/build/ci/github-actions/share-image-jobs/

The goal here isn't to cache between runs, it's to reuse the same image for each job in a given run. In particular, we want to make sure that we're testing against the version of a given image tag when running the pipeline. The image _tag_ is associated with different actual images over time. For example, see the commit history of the official Dockerfile used to build the image(s) tagged with `python:3.13-bookworm`.

https://github.com/docker-library/python/blob/90aa427282f3f8283652c97011a511a77ea699b8/3.13/bookworm/Dockerfile

pavish (changes requested):
@Anish9901 I agree completely with @mathemancer.

`actions/cache` caches content across different workflow runs. This is not a good idea since the actual underlying images might change.

The cleanest approach is to save the images needed as tar files using `docker save` and upload the artifacts before any of the checks, then while running tests, download the artifacts and load the images using `docker load`.

Anish9901:
@mathemancer @pavish, your suggested solution is the first thing I tried as can be seen here: 28f10f8
the problem with  that approach is that it requires us to sync all the downstream jobs to wait for the download and upload of all the images to complete which is around 2.2GB, this sync and some other overheads slows down the CI by 2-2.5x. This is something which I would like to avoid if possible.

I agree that the images might differ in between runs but I don't think this is that significant, we are only caching python and postgres here, which I would expect to be atleast functionally stable. Also the changes to the images are not that often.

<p float="left">
  
  
</p>

To ensure that our cache always has the latest images between runs however, I have added a SHA check that verifies whether the SHA of the latest images on docker hub matches the SHA of the images in our cache. If there is a mismatch, we pull new images.

If you have any simpler ideas that would help keep our CI time ~5 minutes, I’d love to hear them.

mathemancer:
I do have some ideas, some of which will speed things up in general. Try to:

1. Use an initial job to define relevant matrices so we can avoid copying arrays everywhere,
2. Use those matrices to pull relevant images and upload tarballs as artifacts _in parallel_,
3. Use the same matrices for consuming those artifacts in relevant testing jobs.

Bonus points: Actually build relevant Mathesar docker images in step (2) and save _those_ as artifacts so we don't have to repeatedly build the image in each runner. This saves time and money, and makes the whole thing more robust, since we don't have to get each runner to build successfully; it'll happen at the beginning of the workflow.

- [GH Action showing parallel build/cache](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-an-output-to-define-two-matrices)
- [Docker example of reusing images](https://docs.docker.com/build/ci/github-actions/share-image-jobs/)

After setting that up, it should be possible to just remove the cache-checking logic entirely.

pavish (approved):
@Anish9901 I think the updates look good.

I can see the value of the changes @mathemancer is proposing but I think the pressing issue right now is to get CI working to unblock other PRs. I think this should be picked up as a separate issue.

@mathemancer Would you be open to merging this in and letting @Anish9901 work on the changes you're proposing separately?

mathemancer (approved):
As @pavish says, we can get this merged and make a separate issue for fixing this to be more maintainable long term.

PR #4453, opened 2025-05-11
Persist custom column widths in explorations

### LINKED ISSUE #1966: Improve type safety for array indexed access.
## Problem

In this code, `myNumber` has type `number`, when its runtime value is actually `undefined`:

```ts
const numbers = [3, 5, 7, 11, 13];
const myNumber = numbers[100];
```

This can lead to bugs, ([example](https://github.com/centerofci/mathesar/pull/1876#discussion_r1027411805)).

## Potential solutions

- `(A)` Enable [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) within tsconfig. This would likely lead to many type errors that we'd need to track down and fix.

- `(B)` Begin incrementally adopting a practice of using [Array.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) which has type safety by default. The only problem is that it hasn't been supported in Safari for that long, so we'd need to set up a polyfill for it.

- `(C)` Use type assertions to explicitly widen the result of an indexed access type. For [example](https://github.com/centerofci/mathesar/blob/512dab229eeb30801b0aa35419f29987ab35ffdd/mathesar_ui/src/systems/record-selector/RecordSelectorTable.svelte#L132),

    ```ts
    const row = records[index] as RecordRow | undefined;
    ```

    I have been trying to do this whenever indexing an array. But it's easy to forget, and it also sacrifices type safety in the case the the type of the array actually changes due to some other refactor.

@pavish and @rajatvijay I'm curious to hear your thoughts on which of the above approaches you might be interested in adopting, if any.

### LINKED ISSUE #4463: Persist custom column widths in explorations
As described in the [Exploration Display Improvements](https://wiki.mathesar.org/projects/table-query-integration/specs/exploration-display-improvements/)

### Establish a means to persist column display options inside explorations

Several other requirements described within this spec involve storing column display options inside an exploration. Since explorations do not currently have any persistent display options, storing them will require a new data structure.

#### Considerations

- In the table page, column display options are associated with column _attnums_. But we can't do that in explorations because one table column can occur multiple times, and some result column won't correspond to particular table column at all.

- In the data explorer, columns can be renamed, removed, and (eventually) rearranged. Ideally we'd like a mechanism to persist column display options which is resilient against such changes to the query definition.

#### How column display options will be stored

- We'll store display options within the `display_options` JSON field that already exists on the `Explorations` model. (We've not used that field thus far, but it appears to be exposed to the API already, so no backend changes would be necessary here.)

- Display options will reference result columns using **multiple attributes**.

- Here is an example `display_options` JSON value which would store display options for two result columns:

    <details>
    <summary>Example <tt>display_options</tt> value:</summary>

    ```json5
    {
      "2": { // 👈 result column index (zero-based)
        "result_column": {
          "name": "date_of_birth", // 👈 result column name
          "index": 2,
          "type": {
            "name": "timestamp with time zone"
          }
        },
        "display_options": {
          "date_format": "us",
          "time_format": "12hr"
        }
      },
      "5": {
        "result_column": {
          "name": "emails",
          "index": 5,
          "type": {
            "name": "_array",
            "item_type": {
              "name": "bigint"
            }
          }
        },
        "display_options": {
          "width": 250
        }
      }
    }
    ```

    </details>

- The front end will have a reconciliation process which runs every time the query runs.

-  The reconciliation process will look at each display options entry and do its best to find a matching entry in the list of result columns. The presence of multiple column attributes will help the reconciliation algorithm maintain resilience against various small mutations to the result set such as renamed and re-arranged columns.

- The reconciliation process will produce a _corrected_ `display_options` mapping in the event that something has changed within the query or the DB.

- If the corrected display options differ from the stored display options, the front end will update its stored display options. If the exploration is unsaved, those updated display options will continue to persist only in the browser memory. But if the exploration is saved in the internal database, then the front end will _silently_ update the saved exploration in the internal database. The user won't see any unsaved changes or notice that the saved exploration has been updated.

### Persist column widths

Save column widths inside the exploration.

Rationale:

- Of course this is helpful for users, but it's not strictly necessary as a path to DML in explorations. This feature is specified as a requirement because I think it will actually also help us to test out our column display option persistence mechanism because it's the only display option that should be possible for _all_ column types. Plus, it's possible to change column width even not sidebars are visible. It serves as a nice stress test for technical and UX concerns regarding column display options.

seancolsen:
Primarily:

Fixes #4463

Also:

Fixes #1966 (via polyfill)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
@seancolsen I'd like to take some dedicated time to review this PR. I've not started reviewing due to other tasks that are higher on my list, I'll get to this in a day or two. Please let me know if/when this PR might start blocking you, I'll prioritize it right away.

mathesar_ui/src/systems/data-explorer/displayOptions.ts
```diff
@@ -0,0 +1,336 @@
...
+  // between the un-typed display options at the API layer and the
+  // strongly-typed display options in the front end.
+}
+
+export function makeColumnAnchor(
+  resultColumn: QueryResultColumn,
+  index: number,
+): ColumnAnchor {
+  return {
+    name: resultColumn.alias,
```
pavish, line 155: The term `name` here is confusing for me since we use the term alias everywhere else, and `name` is ambiguous and could refer to the alias, display name, and the column's actual name.

mathesar_ui/src/systems/data-explorer/QueryManager.ts
```diff
@@ -196,17 +206,39 @@ export default class QueryManager extends QueryRunner {
...
 
+  private async reconcileDisplayOptions(serverResponse: ExplorationResult) {
+    const reconciliation = reconcileDisplayOptionsWithServerResponse(
+      this.getQueryModel().display_options,
+      serverResponse,
+    );
+    if (!reconciliation.hasChanged) return;
+
+    const newDisplayOptions = reconciliation.newValue;
+    const shouldAutoSave = !get(this.queryHasUnsavedChanges);
```
pavish, line 217: The only scenario I could think of, where `shouldAutoSave` would be true here, is if the column type changes on the database. Is there any other scenario where this would be true?

pavish (approved):
@seancolsen Looks good!

The display options weren't updated in the query being sent to the server when the query isn't saved. While that should be fine in this scenario, I noticed we were creating the request json in two places, one for the ephemeral query and one for the persisted query.

I added a commit https://github.com/mathesar-foundation/mathesar/pull/4453/commits/2f1ab6cfbce85a2140939bb7c8bdddd986829866 to remove this duplication.

seancolsen:
Thanks @pavish.

> I added a commit https://github.com/mathesar-foundation/mathesar/commit/2f1ab6cfbce85a2140939bb7c8bdddd986829866 to remove this duplication.

I appreciate you adding the commit yourself.

In this case, I had actually removed `display_options` from `AnonymousExploration` intentionally. This is in preparation for removing it from the back end types too. The idea here (as we discussed on a call briefly) is that `explorations.run` should not accept any display options and should not return any display options. The back end currently does nothing with the `display_options` it receives, and it only ever returns empty `display_options` back to the front end. Its presence in the API seems vestigial.

So going forward, is it okay with you if I move `display_options` back into `AddableExploration`?

PR #4488, opened 2025-05-19
Fix docs links

zackkrida:
Small PR to fix a docs typo. Already deployed. I can't remember if this should be against master, develop, or both :sweat_smile:


kgodey:
Master, then merge master back to develop as usual. See #4225 for an example.

zackkrida:
Ah that's right. Thanks @kgodey! This is superseded by #4489.

PR #4481, opened 2025-05-16
Hotfix for SQLAlchemy connecting to DBs over UNIX sockets

kgodey:
Fixes #4480 by @zackkrida 

Relevant SQLAlchemy docs:
- [PostgreSQL Unix Domain Connections](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#unix-domain-connections)
- [engine.URL](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.engine.URL)

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #4492, opened 2025-05-20
Move hostname = None into socket conditional

zackkrida:
This PR updates our socket connection bugfix from #4481, which unfortunately introduced a regression where explorations on DBs _not_ connected via Unix socket were broken.

## Testing instructions

Run Mathesar locally and verify you can work with explorations on non-socket and socket (if you can, I will test this as well) connections.

### DIFF
--- a/db/deprecated/engine.py
+++ b/db/deprecated/engine.py
@@ -24,7 +24,7 @@ def create_future_engine(
     query = {}
     if hostname.startswith("/"):
         query = {"host": hostname}
-    hostname = None
+        hostname = None
     conn_url = URL.create(
         "postgresql",
         username=username,


seancolsen (approved):
I've tested this with a TCP postgres connection and it worked for me.

I didn't test with a socket connection because I haven't been following that work and didn't want to spend time getting up to speed on how to set it up right now. I can perform this additional testing if you like, @zackkrida.

I also looked over the diff, and it looks sane to me.

Anish9901 (approved):
Looks good, I tested some basic DML and data explorer with a TCP connection and a socket connection and both worked without errors.

zackkrida:
Thanks @Anish9901 and @seancolsen! I've tested with both connection types as well. I'm going to merge this into the release branch.

PR #4506, opened 2025-05-24
Experimental MCP server

zackkrida:
This PR adds an experimental [MCP](https://modelcontextprotocol.io/introduction) server to the Mathesar codebase. This server can be integrated locally with LLM-supporting IDEs like Cursor and allow AI agents to make tool calls against a running Mathesar instance. 

See the README for installation instructions. Check out the branch and follow the instructions there to get started. 

## Some description of the code

This is very much a proof-of-concept and probably unsuitable for production. It's about 500 lines of python in a single file, along with one new RPC method which reads all of Mathesar's RPC methods from the registry and dumps their signatures into an API response. 

This is used by the MCP server to dynamically register all of Mathesar's RPC methods as tools the LLM can access. MCP servers currently support fewer than 40 tools, and Mathesar has ~96 RPC methods, so this server registers each _group_ of methods (users, tables, analytics, etc.) as a single tool which can then call all of the RPC functions in that namespace. 

I had a previous version which registered all individual tools, and Cursor would only recognize the first 40 in a seemingly-random order.


zackkrida:
This is moving to a standalone repository.

PR #4507, opened 2025-05-26
Add `is_joined_column` to exploration return

### LINKED ISSUE #4502: Improve explorations.run return value to indicate which columns are from the base table
## Current behavior

- The `explorations.run` API return value contains a `column_metadata` blob. Each of the values contains information about a column within the query.
- The front end needs to know which columns originate from the base table and which columns don't. It needs to distinguish between these different columns in order to implement [DML on base tables](https://wiki.mathesar.org/projects/table-query-integration/specs/data-explorer-dml/updating-base-table-cells/). And later, when we have DML on related tables, we'll still need to maintain this distinction because the UX will likely be different for the base table vs related tables.
- Currently the front end has no easy way to tell the difference between these sorts of columns based on the `explorations.run` return value alone.

## Desired behavior

- Each value in the `column_metadata` blob should contain a new boolean property that distinguishes between base table columns and non-base-table columns.
- I don't care too much about the name of this property. Some ideas:
    - `is_base_table_column`
    - `is_from_base_table`

## Additional context

- Brief [Matrix discussion](https://matrix.to/#/!PoPCLRmlBoTeaUwiZy:matrix.mathesar.org/$_nqgbJ7e_Z662uxcOxI0ZyTUGq9klyYhSzXJbc9JADI?via=matrix.mathesar.org)

mathemancer:
Fixes #4502 

Adds an `is_joined_column` attribute to each `column_metadata` blob in the response to calls to `explorations.run` (or `explorations.run_saved`). This boolean is `false` if:

- The column was added in the "Base Table Column" section of the UI, or
- The column is a grouping or aggregation of such a column

Otherwise, this is false.

**Technical details**

I didn't write tests for this, since testing explorations is a bit sketchy. But, the logic and LOC are obvious and minimal, respectively.

We should make up for this with a scenario test involving explorations until such time as we move the exploration code to a more maintainable/testable state.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/deprecated/queries/base.py
+++ b/db/deprecated/queries/base.py
@@ -345,7 +345,7 @@ class InitialColumn:
         """
         A base column is an initial column on a query's base table.
         """
-        return self.jp_path is None
+        return not self.jp_path
 
     def __eq__(self, other):
         """Instances are equal when attributes are equal."""
--- a/mathesar/utils/explorations.py
+++ b/mathesar/utils/explorations.py
@@ -184,6 +184,7 @@ def _get_exploration_column_metadata(
             "type_options": sa_col.type_options,
             "metadata": ColumnMetaDataRecord.from_model(column_metadata) if column_metadata else None,
             "is_initial_column": True if initial_column else False,
+            "is_joined_column": not initial_column.is_base_column if initial_column else None,
             "input_column_name": input_column_name,
             "input_table_name": input_table_name,
             "input_table_id": initial_column.reloid if initial_column else None,
@@ -191,10 +192,7 @@ def _get_exploration_column_metadata(
         }
     for alias, metadata in exploration_column_metadata.items():
         input_alias = metadata["input_alias"] or alias
-        metadata["is_base_table_column"] = (
-            exploration_column_metadata[input_alias]["input_table_id"]
-            == exploration_def["base_table_oid"]
-        )
+        metadata["is_joined_column"] = exploration_column_metadata[input_alias]["is_joined_column"]
     return exploration_column_metadata
 
 


seancolsen (changes requested):
Sorry, but this still doesn't give me what I need.

I'll take a step back and explain the problem that the front end has...

## Goal

- The front end needs to present different editing UX for different sorts of columns. Here are the options I can think of so far:

    - **No editing**: This is similar to what we currently have in the data explorer, though the cell background would be gray to indicate that the cell is read-only.

    - **Direct editing**: Here, you edit directly within the cell, like the current table page. For FK columns, you open the record selector to modify the FK value. (From these FK cells we'll also eventually be able to open the "modal record page" for the related record.)

    - **Many-to-one editing**: You can't edit directly in the cell. Double-clicking opens the modal record page, focusing a specific input. The modal record page will also allow the record _reference_ to be changed (but this UX is still TBD).

    - **One-to-many editing**: One cell contains many related records. You can add/remove/list related records as well as edit individual fields of specific related records.

    - **Many-to-many editing**: Similar to one-to-many, but optimized for many-to-many so as to require fewer clicks. We don't plan to implement this until we've released and evaluated the one-to-many editing UX first.

- For the first iteration of DML, the front end needs to figure out which columns should support "direct editing" and present that UX to users. Any columns which would eventually support other kinds of editing would be kept to "no editing" for now.

## Examples

### Hierarchical categories

image

| Column | The editing UX we ultimately want |
| -- | -- |
| id           | no editing |
| name         | direct editing |
| parent_name  | many-to-one editing |

### Duplicate titles

This shows the titles associated with at least 4 movies

image

The _ultimate_ UX here is debatable.

- The `title` column could potentially be "direct editing" or "no editing".

- The remaining columns should definitely be "no editing" in the short term. Eventually they could potentially become "one-to-many" with some restrictions. That is: we could potentially allow some interactions like listing the movies records within the aggregation, but there will likely be some nuance to this UX. The point is: the UX should definitely _not_ be "direct editing" at any point, now or in the future.

### Movies + genres

image

| Column | The editing UX we ultimately want |
| -- | -- |
| id           | no editing |
| title        | direct editing |
| release_date | direct editing |
| genres       | many-to-many |


## This PR

My first critique served to delineate the editing UX for the columns in the "hierarchical categories" example. The changes you've since pushed _do_ allow me to handle that case! But I'm having trouble with the other two.

The front end gets `is_initial_column` and `is_joined_column` booleans for each column, so I can use those in conjunction.

- Looking at **Duplicate titles**, I get:

    | column | `is_initial_column` | `is_joined_column` | desired UX |
    | -- | -- | -- | -- |
    | title            | false | false | "no editing" (or maybe "direct editing") |
    | count            | false | false | "no editing" |
    | min_release_date | false | false | "no editing" |
    | max_release_date | false | false | "no editing" |

- Looking at **Movies + genres**, I get:

    | column | `is_initial_column` | `is_joined_column` | desired UX |
    | -- | -- | -- | -- |
    | id           | false | false | "no editing" |
    | title        | false | false | "direct editing" |
    | release_date | false | false | "direct editing" |
    | genres       | false | true | "no editing" (short term), "many-to-many" (long term) |

I hope you can see here that there's no way for me to use these two boolean values to determine the UX.

## Next steps

- Should we discuss this on Wednesday?

- Or, would you like to propose an API response structure that you think would address the needs of the front end?

Issue #4510, opened 2025-05-27
Implement Django db models for forms, and form management endpoints


pavish:
## Description
- This issue tracks implementing:
  - The Django DB models following the form structure from the [Forms parent issue](https://github.com/mathesar-foundation/mathesar/issues/4509).
  - Form management RPC methods needed for creating, listing, retrieving, and deleting a form.

## Out of scope of this issue
- Public sharing and related models/logic
- Permission management
- Form submission endpoint
- Lookup columns & record summary related logic

### 1. Forms Django DB models and necessary migrations.
   - The decision of whether separate models for `Form` and `FormField` are required, is left to the implementer.

#### `forms.add`
- Adds a form
- The following fields should be filled in at the backend service layer, they're not supplied by the frontend:
    - `owner`
    - `created_at`
    - `updated_at`
    - `schema_oid` (if we plan to persist schema oid)
    - `target_table_oid` within `ForeignKeyField` (it could be sent from the frontend, if it'll make work easier on the backend).
- Performs the following validation before saving:
    - All columns specified in `nested_fields` should belong to the table with the oid:
      - `target_table_oid` (in case of foreign key parent fields)
      - or,`linked_table_oid` (in case of reverse foreign key parent fields).

#### `forms.get`
- Retrieves a saved form
- Provides an additional `field_source_analysis_map`. This is a map of all field ids with their respective column definitions. This might extend to provide additional information in the future based on frontend needs.
- Response structure:
    ```typescript
    interface FormDefintionGetResponse {
      form_defintion: FormDefinition,
      field_source_analysis_map: Record<FormField['id'], {
        column: ColumnInfoWithMetadata, // same as column information present in columns.list_with_metadata
      }>
    }
    ```

#### `forms.replace`
- Replaces a form with the new content
- (using the term replace instead of patch, because each user save updates the entire form definition)
- The input would be the same one provided to `forms.add` with an additional form id, `id: FormDefinition['id']`.

#### `forms.list`
- Retrieves a list of saved forms - no pagination required as of now.

#### `forms.delete`
- Deletes a form - no permission management, as part of this issue.

mathemancer:
@pavish I thought we were skipping the owner thing for now. Am I misremembering?

mathemancer:
> > I thought we were skipping the owner thing for now. Am I misremembering?
> 
[...]

PR #4525, opened 2025-06-02
Form models

Anish9901:
Related to #4510

Adds Django models required for implementing forms.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/models/base.py
```diff
@@ -283,6 +283,57 @@ class Explorations(BaseModel):
...
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
+    message = models.CharField(null=True)
+    redirect_url = models.URLField(null=True)
+    submit_label = models.CharField(null=True)
+
+
+class FormField(BaseModel):
+    id = models.CharField()
+    attnum = models.SmallIntegerField()
+    related_form = models.ForeignKey('Form', on_delete=models.CASCADE)
```
pavish: This should not be an entire sub-Form, it's only a list of FormFields.

Anish9901: The way I understand this is that each form can have multiple form fields. I added a foreign key on `Form` model, So, anytime we list forms we can filter the related FormFields by form ids. Am I missing something?

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
+    message = models.CharField(null=True)
+    redirect_url = models.URLField(null=True)
+    submit_label = models.CharField(null=True)
+
+
+class FormField(BaseModel):
+    id = models.CharField(primary_key=True)
+    attnum = models.SmallIntegerField()
+    table_oid = models.PositiveBigIntegerField()
```
Anish9901: `table_oid` is only required when the field/column refers another table.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+
+
+class SubmissionSettings(BaseModel):
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
+    message = models.CharField(null=True)
+    redirect_url = models.URLField(null=True)
+    submit_label = models.CharField(null=True)
+
+
+class FormField(BaseModel):
```
Anish9901, line 305: This model is intended to store information for all kinds of columns(`Scalar`, `Foreign key`, or `Reverse Foreign key`). Let me know if this model is missing any required fields or any fields that can be removed.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+        choices=[
+            ("scalar_column", "scalar_column"),
+            ("foreign_key", "foreign_key"),
+            ("reverse_foreign_key", "reverse_foreign_key")
+        ],
+    )
+    record_summary_template = models.JSONField(null=True)
+    label = models.CharField(null=True)
+    help = models.CharField(null=True)
+    placeholder = models.CharField(null=True)
```
Anish9901: What does `placeholder` stores, do we need it?

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    id = models.UUIDField(primary_key=True)
+    name = models.CharField()
+    version = models.IntegerField()
+    database = models.ForeignKey('Database', on_delete=models.CASCADE)
+    base_table_oid = models.PositiveBigIntegerField()
+    schema_oid = models.PositiveBigIntegerField()
+    description = models.CharField(null=True)
+    header = models.OneToOneField('FormHeader', on_delete=models.DO_NOTHING)
+    submission = models.ForeignKey('SubmissionSettings', on_delete=models.CASCADE)
+    is_public = models.BooleanField(default=False)
```
Anish9901, line 294:
```ts
interface PublicShareSettings {
  is_enabled: boolean;

  // Future consideration
  slug: string;
  password?: string; // write-only on API
}
```
I've flattened this to `is_public` we can make a separate model for `PublicShareSettings` when we add `slug` or `password`.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+
+class SubmissionSettings(BaseModel):
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
+    message = models.CharField(null=True)
+    redirect_url = models.URLField(null=True)
+    submit_label = models.CharField(null=True)
+
+
+class FormField(BaseModel):
+    id = models.CharField(primary_key=True)
```
Anish9901: @pavish Why does this `id` need to be a string?

Also, we cannot name a field `id` without it being a `primary_key`, so if this is customizable we should consider renaming it.

Anish9901 (commented):
@pavish @mathemancer
- I haven't added any fields that are for future considerations, we can add them in the _future_
- Let me know if I missed any fields or if any fields are redundant and can be removed.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
 
+class Form(BaseModel):
+    id = models.UUIDField(primary_key=True)
+    name = models.CharField()
+    version = models.IntegerField()
+    database = models.ForeignKey('Database', on_delete=models.CASCADE)
+    base_table_oid = models.PositiveBigIntegerField()
+    schema_oid = models.PositiveBigIntegerField()
+    description = models.CharField(null=True)
+    header = models.OneToOneField('FormHeader', on_delete=models.DO_NOTHING)
```
pavish: This adds a FK on the form table, and when the form itself is deleted, nothing happens to the FormHeader. That should not be the case.

Here's what I'd expect:
- When Form is deleted, the related FormHeader row should also be deleted.
- A FormHeader row should always have a related Form row.

Ideally, an FK would be present on the FormHeader table, referencing the Form table with "on delete Cascade", and the fk column would also have a NOT NULL constraint.

On a side note, I'm not sure if FormHeader needs to be a model, I'd like the RPC method to return it in the structure I've specified, but it could be stored in a flat manner on the Form table itself.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    submission = models.ForeignKey('SubmissionSettings', on_delete=models.CASCADE)
```
pavish: Same comment as above for FormHeader. SubmissionSettings should have the FK to Form and not the other way around. Each Form has it's own SubmissionSettings, and when a Form row is deleted, it's SubmissionSettings row should also be deleted.

Also, this should be an OneToOneField. We should also enforce that the SubmissionSettings is always present for a Form (either at the service layer or in a db trigger).

Similarly, I would be okay with it being a flatted out in the Form table and enforcing the json structure only on the RPC methods.

mathesar/models/base.py
pavish: The interfaces provided on the forms issues are Typescript types and correspond only to the json structure in RPC responses, and do not specify the underlying model.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    is_public = models.BooleanField(default=False)
+
+
+class FormHeader(BaseModel):
+    title = models.CharField()
```
pavish: `title` needs to be a JSON field.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    subtitle = models.CharField(null=True)
```
pavish: `subtitle ` needs to be a JSON field.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+
+
+class SubmissionSettings(BaseModel):
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
+    message = models.CharField(null=True)
```
pavish: `message` needs to be a JSON field.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    is_public = models.BooleanField(default=False)
+
+
+class FormHeader(BaseModel):
+    title = models.CharField()
+    subtitle = models.CharField(null=True)
+
+
+class SubmissionSettings(BaseModel):
+    role = models.ForeignKey('ConfiguredRole', on_delete=models.CASCADE)
```
pavish: This **should not cascade** and delete the Form and/or SubmissionSettings row when a ConfiguredRole is deleted. This fk field should be nullable.

mathesar/models/base.py
pavish: The model seems to be missing the fields needed to construct `nested_fields`.

`nested_fields` should be present for the `ForeignKeyCreateAllowed` and `ReverseForeignKeyField` types. 

It also misses`create_label` and `allow_create`.

Please also refer the TypeScript definitions `ScalarColumnField`, `ForeignKeyField`, and `ReverseForeignKeyField` on how each field kind differs and the fields required for each of them. All of these inherit the BaseField.

mathesar/models/base.py
pavish: In the future, we would like the user to send a REST/RPC request directly to fill a form. These ids will serve as the keys in the request being sent, so it's important to be able to customize them. However, I agree that this should not be string.

We could follow the same pattern we follow for form id here, it can be an `uuid`.

In the future, we would have a separate unique field to represent the request key (similar to how we handle slug in the Form model), and we'd just copy this value over as its default.

Short answer:
- This should be a Primary Key.
- This needs to be an UUID v4.

mathesar/models/base.py
pavish: Thanks for flagging this.

It is meant to store [placeholders for user inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/placeholder).

We can remove it to reduce scope, our latest UX design direction doesn't use it.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+            ("scalar_column", "scalar_column"),
+            ("foreign_key", "foreign_key"),
+            ("reverse_foreign_key", "reverse_foreign_key")
+        ],
+    )
+    record_summary_template = models.JSONField(null=True)
+    label = models.CharField(null=True)
+    help = models.CharField(null=True)
+    placeholder = models.CharField(null=True)
+    needs_validation = models.BooleanField(default=False)
```
pavish: This is not accurate. This should be either contain a list of settings for what kind of validations to perform, or we should just have an `is_required` field here. 

The `BasicValidation` interface contains an `is_required` field which is a type of validation which specifies if we require the end-user to enter a value. This is a form level validation performed in the frontend/service layers.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    table_oid = models.PositiveBigIntegerField()
+    related_form = models.ForeignKey('Form', on_delete=models.CASCADE, related_name='form_field')
+    kind = models.CharField(
+        choices=[
+            ("scalar_column", "scalar_column"),
+            ("foreign_key", "foreign_key"),
+            ("reverse_foreign_key", "reverse_foreign_key")
+        ],
+    )
+    record_summary_template = models.JSONField(null=True)
```
pavish: Let's remove `record_summary_template` in this PR. I'd like to add it along with `lookup_columns` when the UX for it is confirmed.

mathesar/models/base.py
pavish: Yes, you're right with how the FK should work. I mistook this as something representing `nested_fields`, which seems to be missing.

pavish (changes requested):
@Anish9901 I've added some comments. I'd primarily like to flag that the fields needed for storing `nested_fields`, `create_label`, and `allow_create` are missing.

I assume this might be because you've attempted to translate the TypeScript types in the issue directly into the db models, and the union types may not have come across clearly. The TS types in the issue represents the structure passed between the frontend and the backend, and not the underlying db schema.

Please take a look at the types for `ForeignKeyField` and `ReverseForeignKeyField` again. Let me know if it'd help if we went through this together over a call.

I've found TS to be the most expressive when it comes to representing types, but if it doesn't work for communicating intent in GH issues, we can consider augmenting it with more information.

mathesar/models/base.py
mathemancer: I agree with @pavish on this one. Just bring the form header stuff into the form model itself.

mathesar/models/base.py
mathemancer: @pavish I can see a reason to keep this separate: So a user can use the same submission settings for multiple forms (a likely use case), and manage that without manipulating the forms themselves. _However_, I also see that it would be simpler in the meantime to just bring it into the `Form` itself, especially given our current security model.

mathesar/models/base.py
mathemancer: Agreed.

mathesar/models/base.py
mathemancer: Agreed. I'd rather have both or neither in the model.

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
...
+    message = models.CharField(null=True)
+    redirect_url = models.URLField(null=True)
+    submit_label = models.CharField(null=True)
+
+
+class FormField(BaseModel):
+    id = models.CharField(primary_key=True)
+    attnum = models.SmallIntegerField()
+    table_oid = models.PositiveBigIntegerField()
+    related_form = models.ForeignKey('Form', on_delete=models.CASCADE, related_name='form_field')
```
mathemancer: I'd prefer you just call it `form`, and the `related_name` should just be `fields`.

mathesar/models/base.py
mathemancer: @pavish The validation should only be performed in one layer or the other, no? Which should it be?

mathesar/models/base.py
mathemancer: After chatting with @pavish, It's okay to make the `id` an integer to fit in with other models. _However_: Please add a `key` field of a string type, to hold the `key` for the field when submitting. This should be unique per-form, but not globally (as an `id` would be).

mathesar/models/base.py
```diff
@@ -283,6 +283,60 @@ class Explorations(BaseModel):
     description = models.CharField(null=True)
 
 
+class Form(BaseModel):
+    id = models.UUIDField(primary_key=True)
```
mathemancer: I had a chat with @pavish about this. Basically, using our random security token as the `id` makes it difficult to invalidate that token. This is because it'd also invalidate references to the id, e.g., from `FormField`s.

Instead, just use an incrementing integer ID here, as we usually do, and add an additional `token` field of `UUID` type. That way, an admin can revoke access to the form by changing the `token` field, without breaking DB-layer references. It's important that any error thrown by the form submission endpoint with an invalid token cuts off execution and just returns the fact that there exists no form with that `id, token` pair.

@pavish please confirm that I'm correctly relaying the conclusion from our call.

mathemancer (changes requested):
Overall, I agree with most of @pavish's comments. I noticed an issue with the `UUID` type for `Form.id` and its intended purpose. See specific comments about `Form.id` and `FormField.id`, and needed new attributes.

Separately: Let's not merge this into develop until we cut the 0.3.0 release branch, since we may change these models as we iterate on the UI, and I'd like to avoid a useless migration for users.

mathesar/models/base.py
pavish: @mathemancer 

> The validation should only be performed in one layer or the other, no? Which should it be?

If there's a NOT NULL constraint on the underlying column, that'll error out at the db layer. This field has nothing to do with that.

This is for columns that allow NULL on the db, but the user wants to ensure that the end-user specifies data for it while filling the form.

This check happens on the frontend for providing immediate feedback, and also the service layer as part of validating incoming submission requests.

mathesar/models/base.py
pavish: > please confirm that I'm correctly relaying the conclusion from our call

Yes, this is what we need.

One additional note that we discussed: Also add a unique constraint on the `token` column (even though it's an UUID). This is because we may want the user to be able to customize this in the future and having a unique constraint does no harm at the moment.

mathesar/models/base.py
Anish9901: > This is not accurate. This should be either contain a list of settings for what kind of validations to perform, or we should just have an `is_required` field here.
> 
> The `BasicValidation` interface contains an `is_required` field which is a type of validation which specifies if we require the end-user to enter a value. This is a form level validation performed in the frontend/service layers.

Oh, I misunderstood `is_required` as being a field that specifies whether additional validation is required as it was mentioned within `BasicValidation` interface. I understand now that it is supposed to be an attribute of a form field.

zackkrida:
Just noting that as discussed in a call today, we'll create a "forms" feature branch to merge this into rather than merging into "develop".

mathesar/models/base.py
```diff
@@ -283,6 +283,56 @@ class Explorations(BaseModel):
...
+            ("reverse_foreign_key", "reverse_foreign_key")
+        ],
+    )
+    label = models.CharField(null=True)
+    help = models.CharField(null=True)
+    readonly = models.BooleanField(default=False)
+    styling = models.JSONField(null=True)
+    is_required = models.BooleanField(default=False)
+    # foreign_key/reverse_foreign_key related settings
+    parent_field = models.ForeignKey('self', on_delete=models.CASCADE, related_name='child_fields', null=True)
```
Anish9901, line 323: I've added `nested_fields` but named it `child_fields` instead, as it seemed more clearer.

mathemancer (approved):
Once @pavish is happy, I'm happy.

mathemancer:
> Just noting that as discussed in a call today, we'll create a "forms" feature branch to merge this into rather than merging into "develop".

I went ahead and made the feature branch `forms_feature_1`, and set up `develop`-like branch protection for any branch in the repo named something matching `forms_feature_*`.

pavish (changes requested):
@Anish9901 The changes look great to me.

One thing I missed in the previous review, the form fields (`fields` and `child_fields`) need to be ordered when displaying visually to the user.

We will need an `order` integer field in the `FormField` model for that.

This is not explicitly mentioned in the issue, since these are presented in an array. I'll update the issue to account for that.

Anish9901:
> We will need an `order` integer field in the `FormField` model for that.

Added in a559ab8, @pavish.

PR #4533, opened 2025-06-04
Fix install script to accept unix domain socket based connections

### LINKED ISSUE #4512: Unix domain sockets do not work with the direct installation method
## Description
- Install using the direct installation method
- Provide a url for connecting via unix socket
- The installer mentions that the connection failed and asks for a different connection string

## Expected behavior
- The installer should work properly for unix socket based connections

pavish:
Fixes #4512

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/setup/process_env.py
+++ b/setup/process_env.py
@@ -90,7 +90,7 @@ def validate_pg_env_variables(env_vars):
 
         return True
     except Exception as e:
-        sys.exit(f"Invalid: Unable to connect to the database. {e}")
+        sys.exit(f"Unable to connect to the database. {e}")
 
 
 def construct_env_vars_from_connection_string(connection_string):


setup/process_env.py
```diff
@@ -77,43 +77,23 @@ def generate_secret_key(length=50):
...
+            password=env_vars.get("POSTGRES_PASSWORD", "")
+        ):
+            pass
 
-def validate_pg_connection_string(connection_string):
-    """
-    Validate the provided PostgreSQL connection string.
+        return True
+    except Exception as e:
+        sys.exit(f"Invalid: Unable to connect to the database. {e}")
```
mathemancer: This isn't correct. The variables could be correct, but some other issue could cause a connection failure. It seems that the error type is always `psycopg.OperationalError` (argh). Maybe just remove "Invalid: " from the error.

setup/process_env.py
```diff
@@ -169,24 +137,30 @@ def main():
...
+            "POSTGRES_PORT": env_vars.get("POSTGRES_PORT", ""),
+            "POSTGRES_DB": env_vars["POSTGRES_DB"],
+            "POSTGRES_USER": env_vars["POSTGRES_USER"],
+            "POSTGRES_PASSWORD": env_vars.get("POSTGRES_PASSWORD", "")
+        }
 
-    updated_content = update_env_lines(lines, updates)
-    sys.stdout.write(updated_content)
+    if validate_pg_env_variables(pg_env_vars):
+        updates.update(pg_env_vars)
```
mathemancer, line 161: Doesn't this overwrite their password in the file if they fatfinger a command line argument? That should not happen. Sorry I didn't catch this before.

setup/process_env.py
```diff
@@ -169,24 +137,30 @@ def main():
...
 
-    if len(sys.argv) > 1 and sys.argv[1].strip():
-        connection_string = sys.argv[1].strip()
+    connection_string = sys.argv[1].strip() if (len(sys.argv) > 1 and sys.argv[1].strip()) else None
+
+    if connection_string:
+        pg_env_vars = construct_env_vars_from_connection_string(connection_string)
+
     else:
         required_pg_keys = ["POSTGRES_HOST", "POSTGRES_DB", "POSTGRES_USER"]
```
mathemancer, line 146: Why are we still requiring a `POSTGRES_HOST` variable? It's not needed for making a unix socket connection. Neither are the other two. What's the point of this validation, other than baking in assumptions that may be faulty?

E.g., if they set _no variables_ and provide no connection string, and we just let `psycopg` do its thing, the default behavior will be to use the unix socket at the default location on the default port, etc. That seems preferable.

setup/process_env.py
```diff
@@ -169,24 +137,30 @@ def main():
...
 
-    pg_updates = validate_pg_connection_string(connection_string)
-    updates.update(pg_updates)
+        pg_env_vars = {
+            "POSTGRES_HOST": env_vars["POSTGRES_HOST"],
+            "POSTGRES_PORT": env_vars.get("POSTGRES_PORT", ""),
+            "POSTGRES_DB": env_vars["POSTGRES_DB"],
+            "POSTGRES_USER": env_vars["POSTGRES_USER"],
+            "POSTGRES_PASSWORD": env_vars.get("POSTGRES_PASSWORD", "")
+        }
```
mathemancer, line 158: I don't think we should save any of these, unless they specifically came from a connection string. It doesn't make sense to me to save empty or vacuous values.

mathemancer (changes requested):
I think the overall method of connecting in this PR is much cleaner than the previous version.

However, You're building in lots of assumptions that seem unwarranted to me, and then validating that those assumptions are met. I see no reason to add a bunch of empty string variable values to their environment file, and also no reason to throw errors just because some variable we expect doesn't exist. What's the value of that for the user? We should instead try to minimize assumptions, and let `psycopg` do its best (which is usually great) with whatever info is provided.

setup/process_env.py
mathemancer: Talked to @pavish . We'll handle this edge case in the next PR.

setup/process_env.py
mathemancer: Talked to @pavish this is to support faulty assumptions in our models.

setup/process_env.py
mathemancer: Also supporting underlying assumptions.

setup/process_env.py
mathemancer: Changed after chatting with @pavish .

PR #4534, opened 2025-06-04
Add pagination to record selector

### LINKED ISSUE #4522: Remove or increase record selector row limit
## Problem

I don't always know what to search for in the record selector when associating linked records. I want to be able to scroll through records.

## Proposed solution

Ideally, just show all records in the record selector, with pagination. Otherwise, at least increase the limit a _lot_.

seancolsen:
Fixes #4522

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@mathemancer please review back end code changes.

@pavish please review front end code changes + UX.

@ghislaineguerin please review the UX changes here.

@zackkrida I'm inviting you to review UX too, since your head has been in this problem. If you don't want to spend time on this, that's fine — just unassign yourself.

zackkrida (approved):
Thanks @seancolsen—I definitely want to spend the time on this! I’ve reviewed the changes and overall, this is great! It cleanly addresses the problem and lets users view more than 10 records without losing context. Crucially, users can browse _all_ available records without needing to have any "known values" about the record they're looking for.

One non-blocking enhancement I’d suggest is making the total record count and the min/max range more visible, especially whenever the page size exceeds 10. I understand the intent behind hiding this info was to save vertical space, especially when the “+ Create Record” button appears. However, I believe we should prioritize displaying the record-count metadata in the modal footer. I think it's important context for users that can inform their behavior. I can elaborate that rationale if needed, but for now here's why I think it's okay to do from a UI perspective:

- When there are 100 or fewer options, pagination doesn’t appear at all, so there’s no extra footer height to worry about.
- Large sets already take up more space. Once you go beyond 100 records, pagination kicks in, showing users up to 90 records more than the previous limit. Since we're already using significantly more vertical space in that scenario, dedicating a small portion of it to "Records 1–100 of 2,300" and placing "+ Create Record" just above it still represents a huge net win in visible rows.
- Filtering, which displays the button, already implies a smaller result set. If someone filters and triggers the "+ Create Record" button, they’re already narrowing down a large list to find a specific item. At that point, the result set should already be small enough that losing one or two additional rows for the button isn’t a big concern.

Anyway, take that or leave it :man_shrugging:. Barring any technical concerns from anyone else, LGTM.

seancolsen:
Thanks for the feedback, @zackkrida. I implemented your suggestion in 23f173fa3c0d5d3e1d1c26010310ebd7bec637b5.

- I moved the stats from the page jumper to the window footer.

    

    A couple minor adjustments also visible in the above screenshot:

    - I simplified the help text below the page jumper input.
    - I added added locale formatting to the stats numbers (so they display with grouping separators as necessary).

- I was able to get pretty fancy with fluid layout in the footer. It rarely takes up more space than before! Here are all the layout possibilities:

    - **Stats only**

        

    - **Stats + pager**

         
        
    - **Stats + add button**

         
        
    - **Stats + add button + pager**

         

        


- Also, I changed it so that when the results are loading, the stats are hidden (in order to avoid showing incorrect information). They still contribute to the layout though (in order to avoid layout shift).

    


All-in-all, this is definitely way better than before! Thanks for suggesting this! I also like how this solves the problem of "scrolling discoverability" that I mentioned in the demo video.

zackkrida:
@seancolsen it looks great. Very cool!

pavish (approved):
Looks good to me!

PR #4536, opened 2025-06-05
Add `forms.add` RPC endpoint

Anish9901:
Related to #4510
<table>
<tr>
  <td>
<details open>
  <summary><b>Request:</b></summary>

```json
POST http://localhost:8000/api/rpc/v0/

{
  "jsonrpc":"2.0",
  "id":0,
  "method":"forms.add",
  "params":{
    "form_def":{
        "name": "form1",
        "version": 1,
        "database_id": 1,
        "schema_oid": 18145,
        "base_table_oid": 18152,
        "header_title": {"title": "text"},
        "fields": [
            {
                "key": "k1",
                "attnum": 10,
                "index": 1,
                "kind": "foreign_key",
                "label": "Authors",
                "target_table_oid": 18146
            },
            {
                "key": "k2",
                "attnum": 2,
                "index": 2,
                "kind": "scalar_column",
                "label": "First Name",
                "parent_field_key": "k1"
            },
            {
                "key": "k3",
                "attnum": 3,
                "index": 3,
                "kind": "scalar_column",
                "label": "Last Name",
                "parent_field_key": "k1"
            }
        ]
    }
  }
}
```
</details>
</td>
  <td>
<details>
  <summary><b>Response:</b></summary>

```json
HTTP/1.1 200 OK
Date: Sun, 15 Jun 2025 15:23:20 GMT
Server: WSGIServer/0.2 CPython/3.13.4
Content-Type: application/json
X-Frame-Options: DENY
Content-Length: 2218
Vary: Accept-Language, Cookie
Content-Language: en
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Cross-Origin-Opener-Policy: same-origin

{
  "id": 0,
  "jsonrpc": "2.0",
  "result": {
    "id": 80,
    "created_at": "2025-06-15T15:23:20.501Z",
    "updated_at": "2025-06-15T15:23:20.501Z",
    "token": "7c976e44-e8cb-4d3b-8fab-b45201e752f3",
    "name": "form1",
    "description": null,
    "version": 1,
    "database_id": 1,
    "schema_oid": 18145,
    "base_table_oid": 18152,
    "is_public": false,
    "header_title": {
      "title": "text"
    },
    "header_subtitle": null,
    "submit_role_id": 1,
    "submit_message": null,
    "redirect_url": null,
    "submit_label": null,
    "fields": [
      {
        "id": 184,
        "key": "k1",
        "attnum": 10,
        "form_id": 80,
        "index": 1,
        "kind": "foreign_key",
        "label": "Authors",
        "help": null,
        "readonly": false,
        "styling": null,
        "is_required": false,
        "parent_field_id": null,
        "target_table_oid": 18146,
        "allow_create": false,
        "create_label": null
      },
      {
        "id": 185,
        "key": "k2",
        "attnum": 2,
        "form_id": 80,
        "index": 2,
        "kind": "scalar_column",
        "label": "First Name",
        "help": null,
        "readonly": false,
        "styling": null,
        "is_required": false,
        "parent_field_id": 184,
        "target_table_oid": null,
        "allow_create": false,
        "create_label": null
      },
      {
        "id": 186,
        "key": "k3",
        "attnum": 3,
        "form_id": 80,
        "index": 3,
        "kind": "scalar_column",
        "label": "Last Name",
        "help": null,
        "readonly": false,
        "styling": null,
        "is_required": false,
        "parent_field_id": 184,
        "target_table_oid": null,
        "allow_create": false,
        "create_label": null
      }
    ],
    "field_col_info_map": {
      "k1": {
        "column": {
          "id": 10,
          "name": "Author",
          "type": "integer",
          "default": null,
          "nullable": true,
          "description": null,
          "primary_key": false,
          "type_options": null,
          "has_dependents": false,
          "current_role_priv": [
            "SELECT",
            "INSERT",
            "UPDATE",
            "REFERENCES"
          ],
          "metadata": null
        },
        "error": null
      },
      "k2": {
        "column": {
          "id": 2,
          "name": "First Name",
          "type": "text",
          "default": null,
          "nullable": true,
          "description": null,
          "primary_key": false,
          "type_options": null,
          "has_dependents": false,
          "current_role_priv": [
            "SELECT",
            "INSERT",
            "UPDATE",
            "REFERENCES"
          ],
          "metadata": null
        },
        "error": null
      },
      "k3": {
        "column": {
          "id": 3,
          "name": "Last Name",
          "type": "text",
          "default": null,
          "nullable": true,
          "description": null,
          "primary_key": false,
          "type_options": null,
          "has_dependents": false,
          "current_role_priv": [
            "SELECT",
            "INSERT",
            "UPDATE",
            "REFERENCES"
          ],
          "metadata": null
        },
        "error": null
      }
    }
  }
}
```
</details>
</td>
</tr>
</table>

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar/utils/forms.py
```diff
@@ -0,0 +1,84 @@
...
+            or user_dbrm.user.is_superuser
+            or submit_role.name in parent_names
+        ), "Insufficient privileges for selected submit_role"
+    return submit_role
+
+
+def validate_related_fields(tab_attn_map, user_dbrm, submit_role):
+    user_dbrm.cofigured_role = submit_role  # DANGEROUS!!!
+    with user_dbrm.connection as conn:
+        assert related_fields_exist(tab_attn_map, conn), "Invalid related fields"
```
Anish9901: > Performs the following validation before saving:
> - All columns specified in `nested_fields` should belong to the table with the oid:
>   - `target_table_oid` (in case of foreign key parent fields),
>   - or,`linked_table_oid` (in case of reverse foreign key parent fields).

Is my assumption to use the submit role for validating related fields correct here? As I understand, the person filling the form must have access to these fields, validating using submit_role here ensures the same.

On another note, since the user can pass any configured role as submit_role for a form, this could potentially lead to privilege escalation, however, in our case we only call `validate_related_fields` after calling `validate_submit_role`, which makes sure user actually has access to the submit_role either directly or indirectly. Is there a safer way we can do this?

Do we _really_ need this kind of validation? If possible, I would like to avoid this, and have this kind of validation happen on the frontend instead.

mathesar/rpc/forms.py
```diff
@@ -0,0 +1,185 @@
...
+        kind: Type of the selected column(scalar_column, foreign_key, reverse_foreign_key)
+        label: ???
+        help: ???
+        readonly: Specifies whether the selected field is readonly.
+        styling: ???
+        is_required: Specifies whether a value for the field is mandatory.
+        parent_field_id: The Django id of the Field set as parent for related fields.
+        target_table_oid: The OID of the related table.
+        allow_create: Specifies whether adding new records is allowed within related fields.
+        create_label: ???
```
Anish9901: @pavish can you help me describe the fields marked with `???`.

mathesar/rpc/forms.py
```diff
@@ -0,0 +1,185 @@
...
+        token: A UUIDv4 object used to identify a form uniquely.
+        name: The name of the form.
+        description: The description of the form.
+        version: The version of the form.
+        database_id: The Django id of the database containing the Form.
+        schema_oid: The OID of the schema where within which form exists.
+        base_table_oid: The table OID based on which a form will be created.
+        is_public: Specifies whether the form is publicly accessible.
+        header_title: ???
+        header_subtitle: ???
```
Anish9901: Here as well.

mathesar/utils/forms.py
pavish: > Is my assumption to use the submit role for validating related fields correct here? As I understand, the person filling the form must have access to these fields, validating using submit_role here ensures the same.

Yes, this is correct.

> since the user can pass any configured role as submit_role for a form, this could potentially lead to privilege escalation

The user cannot pass any configured_role as submit_role. Please refer to the ownership section from this comment: https://github.com/mathesar-foundation/mathesar/issues/4509#issuecomment-2966883219

> Do we really need this kind of validation? If possible, I would like to avoid this, and have this kind of validation happen on the frontend instead.

What is the complexity in performing this validation while saving the form? I do think we'd require it, however we don't have to block this PR for the validation. Can you create another issue for this and we can discuss it there?

mathesar/rpc/forms.py
pavish: This is the header and the subtitle that's displayed to the end-users on the rendered form. This is different from the form name.

mathesar/rpc/forms.py
pavish: Label is the `label` content displayed for the form inputs, similarly help denotes the help content the end user needs to see.

`styling` stores information on the visual appearance of the inputs. (Please refer to this comment: https://github.com/mathesar-foundation/mathesar/issues/4509#issuecomment-2966883219, it should explain things in more detail).

`create_label` is the label shown for the section in the UI where the user is creating a new record, instead of choosing one for the FK.

mathesar/utils/forms.py
Anish9901: >The user cannot pass any configured_role as submit_role. Please refer to the ownership section from this comment: https://github.com/mathesar-foundation/mathesar/issues/4509#issuecomment-2966883219

Yes, this is exactly how it works, the validation happens in `validate_submit_role`. I guess I'm not that comfortable letting an admin choose any arbitrary role configured on the db even when they might not be a member of one or have one configured. But again, admins can give themselves any role in the admin section so I guess its fine.

> What is the complexity in performing this validation while saving the form? I do think we'd require it, however we don't have to block this PR for the validation. Can you create another issue for this and we can discuss it there?

This already works, validation happens in `validate_fields`.

mathesar/rpc/forms.py
Anish9901: added in 7f4ad7b

mathesar/rpc/forms.py
Anish9901: added in 7f4ad7b

mathesar/rpc/forms.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @classmethod
+    def from_model(cls, form_model, field_col_info_map=None):
+        def rm_keys(model, keys):
+            return {k: v for k, v in model.__dict__.items() if k not in keys}
+        return {
+            **rm_keys(form_model, ('_state')),
+            "fields": [rm_keys(f, ('_state', 'created_at', 'updated_at'))for f in form_model.fields.all()],
+            "field_col_info_map": field_col_info_map
+        }
```
mathemancer: This should be explicit. I.e., just do:
```python
dict(
    a = model.a,
    b = model.b,
    ...
)
```
The way you have it now is clever, but means that if we add metadata to the underlying model, we'll have to make sure to also remove it here. That ties this API (which should be concerned with FE expectations) too closely to the model implementation.

Also, you should use your `FieldInfo` class to get the field info from the field models.

mathesar/rpc/forms.py
```diff
@@ -0,0 +1,190 @@
...
+        create_label: ???
+    """
+    key: str
+    attnum: int
+    index: int
+    kind: Literal["scalar_column", "foreign_key", "reverse_foreign_key"]
+    label: Optional[str]
+    help: Optional[str]
+    readonly: Optional[bool]
+    styling: Optional[dict]
```
mathemancer, line 166: It seems like these dictionaries are always things the back end will never have to know about the form of. Is that correct? E.g., no reason for the back end to get information from the `styling` dictionary.

mathesar/utils/forms.py
```diff
@@ -0,0 +1,121 @@
...
+from mathesar.models.base import (
+    Form, FormField, Database, ConfiguredRole, UserDatabaseRoleMap, ColumnMetaData
+)
+
+
+def validate_submit_role(user_dbrm, submit_role_id=None):
+    with user_dbrm.connection as conn:
+        submit_role = (
+            ConfiguredRole.objects.filter(id=submit_role_id).first()
+            or user_dbrm.configured_role
```
mathemancer, line 19: This logic doesn't seem right. If they submit a role ID which doesn't exist, it should produce an error. The current logic will silently use their own configured role. Instead, just use `ConfiguredRole.objects.get` and let nature take its course.

mathesar/utils/forms.py
```diff
@@ -0,0 +1,121 @@
...
+from django.db import transaction
+
+from db.forms import fields_exist, get_oid_col_info_map
+from db.roles import get_current_role_from_db
+from mathesar.models.base import (
+    Form, FormField, Database, ConfiguredRole, UserDatabaseRoleMap, ColumnMetaData
+)
+
+
+def validate_submit_role(user_dbrm, submit_role_id=None):
```
mathemancer: Just call this `get_submit_role`. It returns a non-boolean. The validation is incidental.

db/forms.py
```diff
@@ -0,0 +1,34 @@
+import json
+from db import connection as db_conn
+
+
+def fields_exist(tab_attn_map, conn):
+    """
+    Utility function to determine that form fields passed through the forms API actually exists on the db.
+
+    tab_attn_map should have the following form:
+    {
```
mathemancer: Delete this function please.

mathemancer (changes requested):
Overall, not too bad, but:

- At this stage, the only thing this function should validate is that the user is allowed to run queries as the chosen `submit_role`.
- All other validations should be deleted.

The current function has an absolute firehose of back-and-forth between the service and the DB, meaning saving a form definition will take way too long. The only relevant validations are:

1. Is the user allowed to use this role?
2. Can the chosen role run an `INSERT` statement as defined by the form?

The first is done in this PR; fine. The second should be deferred until we have the `INSERT` statement building working. Any other validation style is going to bake in assumptions that are (at best) only temporarily correct, along with all the back-and-forth between the service and DB.

Side note: It's not okay to modify the role associated with a `UserDatabaseRoleMap` object passed to a function. If at some point that map gets used somewhere else, unexpected behavior will certainly result.

mathesar/rpc/forms.py
Anish9901: That is true, its here just for documentation purposes, since we _are_ storing this information in our models

mathesar/utils/forms.py
Anish9901: fixed

Anish9901:
@pavish as per @mathemancer's request I've removed `validate_fields`.

> Side note: It's not okay to modify the role associated with a UserDatabaseRoleMap object passed to a function. If at some point that map gets used somewhere else, unexpected behavior will certainly result.

Not passing the model around anymore, getting it right before using it.

db/forms.py
```diff
@@ -0,0 +1,21 @@
+import json
+from db import connection as db_conn
+
+
+def get_oid_col_info_map(tab_attn_map, conn):
```
mathemancer, line 5: In the long run, we should make this part of the overall `get_form_db_details` (or similar) function. That function will:

- Build the actual insert query
- Check that it works
- Bundle up any required info (for adding, modifying, or any other use of the form) and return it.

Most (maybe all) of that logic should run on the DB to avoid traffic to-and-fro.

For now, however, I'm fine with keeping this since we need to have the API response lined up for Front End development.

db/sql/05_msar.sql
```diff
@@ -5595,3 +5595,54 @@ BEGIN
   );
 END;
 $$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION
+msar.get_oid_col_info_map(oid_attn_map jsonb)
```
mathemancer, line 5601: Sorry to think of this so late in the game, but:

The DB layer shouldn't return this as a map, since we want a function (eventually) that just returns info about columns of a query result or insert statement, some of which may or may not have OIDs. I acknowledge that it's a stretch for inserting specifically, and I'm okay with leaving this as a shim until the point where we've actually built the `INSERT` for forms, but please return an array of `col_info` blobs with the `attnum` and `table_oid` _on the inside_, rather than as keys (hope I'm making sense. I'm rushing before I jump on a boat).

mathesar/utils/forms.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def get_oid_attnums_map(form_model):
+    oam = defaultdict(list)
+    fields = {field.key: field for field in form_model.fields.all()}
+    for field in fields.values():
+        table_oid = field.parent_field.target_table_oid if field.parent_field else form_model.base_table_oid
+        oam[table_oid].append(field.attnum)
+    return oam
+
```
mathemancer, line 38: In the medium term, I suspect this logic will move down to the DB.

mathemancer (approved):
Okay, I think I'm satisfied with this for now. Once we have the function for actually _running_ a form-based insert, we should assess how much of this metadata-getting logic should move down to the DB layer to sync up with the actual inserting logic.

mathemancer:
@Anish9901 I think @pavish still needs to look, correct? If I'm wrong, just merge. Otherwise, wait for his approval.

mathemancer:
Fat-fingered. Sorry.

mathesar/utils/forms.py
```diff
@@ -0,0 +1,113 @@
...
+
+def get_field_col_info_map(user, form_model):
```
mathemancer: Just realized a problem while reviewing #4554 . 

This shouldn't take the user as an arg. It should instead use the `submit_role` for that form. In particular, things like the permissions associated with columns need to reflect the form's role, not the current user don't they? In that case, just make this a method of the `Form` model. I.e., the call should be something like: "Dear `Form`, please tell me about yourself."

mathemancer: Maybe _maybe_ you could have a method for non-public forms with no associated `submit_role` in the future that would involve the user's role, but I don't think that is needed at the moment.

mathemancer (changes requested):
Sorry but I noticed a problem while reviewing #4554 .

mathesar/utils/forms.py
Anish9901: We _are_ switching user's configured role with `submit_role` before opening a connection.

mathemancer: Then why is the user relevant at all?

Anish9901: ok, added a connection property within the `Form` model itself.

Anish9901:
As discussed on matrix with @mathemancer, https://github.com/mathesar-foundation/mathesar/pull/4536#pullrequestreview-2941794094 will be handled separately, here is the issue tracking the same: [#4561](https://github.com/mathesar-foundation/mathesar/issues/4561)

pavish:
@Anish9901 and @mathemancer Please go ahead and merge this PR when the review is done. I will be requesting/making changes as I work on the frontend.

PR #4556, opened 2025-06-17
ARCHIVE: Original Row Seeker Prototype

pavish:
This is a PR for a record selector alternative that I'd like to use for forms (and possibly the table page as well). I'm just using the name "Row seeker" to differentiate from our existing record selector.

This is currently functional, but with bugs and not production ready. I'd like the team to try this out. I'm expecting feedback on the UX, whether this aligns with the product direction, and usability concerns. 

This is a draft because it's not meant to be merged.

## Selection
- Records in the Row seeker are **sorted by their record summary**
- The selected record is highlighted and the list scrolls to make it visible when the dropdown opens
- Functionality similar to a select menu
- Keyboard navigable, users can use up+down arrow keys, and enter to select the record they need
- Screenshot 2025-06-17 at 2 50 27 PM

## Providing context
- The Record summary is first presented in the list.
- Following the record summary, the user sees 3 columns we choose as default.
  - This can be user configurable in the future.
  - Currently, it's just the first three non-pk columns.
  - We can take a number of metrics into account to sensibly choose these fields. Eg., if fields are already in record summary, based on column ordering on the base table etc.,
- Screenshot 2025-06-17 at 2 50 27 PM

## More context when the user desires
- The user can expand each record to view upto 10 columns.
- Eventually, I'd like the user to configure these "lookup columns". They are available to view as well to filter from.
- Screenshot 2025-06-17 at 2 50 40 PM

## When additional context is not necessary
- For simple tables with just one column, we do not show the default fields.
- The user will still however have the option to view all fields by expanding the record.
- We can optimize this further by not showing defaults if all columns except the PK are already part of the record summary.
- We could also entirely remove the expander.
- Screenshot 2025-06-17 at 2 52 47 PM

## The absolute simple version
- You can view this while testing by changing `mode` to `basic` in the RowSeekerController. 
- Just the summaries.
- Screenshot 2025-06-18 at 1 16 03 PM

## Searching through the list
- As mentioned above, the list is sorted by the record summary.
- The user can **type to search the record summary**.
- Screenshot 2025-06-17 at 2 51 17 PM

## One click filter
- Clicking on a value from the "lookup fields" filters by that value.
- Screenshot 2025-06-17 at 3 13 30 PM
- Screenshot 2025-06-17 at 2 51 27 PM
- Screenshot 2025-06-17 at 2 51 37 PM

## Custom filtering
- The user can also manually add filters by clicking on the 'filter funnel/drilldown' icon.
- When they select a column, they'll be prompted to type a value to filter by that column.
- Note: **This is not fully functional yet.**
- The goal is to show the inputs according to the column types, as we show in the table filter dropdown.
-  For FK columns, we'd show a very simple row seeker, with just the record summaries in the input and the search bar. It'll function like a "typeahead" input.
- Screenshot 2025-06-17 at 2 51 50 PM
- Screenshot 2025-06-17 at 2 52 01 PM

## UI when there are multiple column filters and search
- Screenshot 2025-06-17 at 2 52 31 PM

## Pagination
- This uses the mini pagination component that was merged recently.
- Pagination is displayed when records > 200.
- This makes auto-scrolling to the current selected record not possible since it could be in any page. This is an existing problem with the record selector as well.
   - To overcome this, I'd like to display a static selected option above the search bar. I've not explored this yet.
- Screenshot 2025-06-17 at 2 53 56 PM

## Feature Discoverability
- We display an option to configure the record summary and lookup fields.
   - This is not implemented, but I'd like to open a modal with two panes when the user clicks this.
    - The left pane will contain the configurations for record summary and lookup fields which the user can modify.
    - The right pane will contain a preview of the dropdown. As the user modifies the configuration, they can see the changes right away.
- We also display the Table name in the footer for more context. The user can also navigate to it, if we want to enable that.
- Screenshot 2025-06-17 at 3 25 37 PM
- Other ideas: We can include "Edit" record option within each record when it's expanded.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
I'm expecting review from @zackkrida and @kgodey on the UX direction.

Everyone is welcome to test it out locally and provide suggestions.

zackkrida:
@pavish: I _love_ this experiment. Really excellent work, and I think this will solve a lot of problems and prepare Mathesar for future work we'd like to do in a number of ways. I could expand on those, but I instead want to offer one piece of feedback, which is almost a challenge, of sorts:

Have you seen uBlock Origin's user interface? Here's a quick video of it:

https://github.com/user-attachments/assets/c56d94db-8e12-4084-a3e9-3a90326081e4

What I'm getting at is likely obvious, but the UI defaults to the _simplest possible version_ and makes it very easy (if not a bit unclear) to increase the level of detail and control as-needed. I suspect, here, that the level of detail needed, by default, for many users and use cases _is_ to see a single-line text representation of linked records. I think record summaries are the best way to do that, and I suspect others might think single-columns are the best way. In either case, I think we could make the default presentation the simplest one while making it very easy to expand out to the more detailed view of seeing additional columns. That would also increase the overall information density of the result set. 

My main feedback though is that I definitely think this should become a part of Mathesar in our next release for both forms and the table view. I'd like to achieve something close to feature parity with the record selector, specifically in terms of _creating_ records, but I think that's something we'll be able to figure out.

pavish:
@zackkrida 

> What I'm getting at is likely obvious, but the UI defaults to the simplest possible version and makes it very easy (if not a bit unclear) to increase the level of detail and control as-needed. I suspect, here, that the level of detail needed, by default, for many users and use cases is to see a single-line text representation of linked records.

I completely agree. I think the "Absolute simple version" in the description should be the default.

However, we do need to solve the problem of choosing sensible columns as the default for record summary. The current approach of choosing the first text column for the record summary is not a good idea since the user would most probably see duplicated records and it wouldn't help them choose one (eg., the record summary is the first_name column). I'd be more inclined towards the default record summary being a combination of atleast 3 columns.

The filters could be even more simplified by having a "tag/log search" kind of UX (eg., GH's issue search). I did not implement this to get this prototype out quick, but this would be the most ideal UX.

I also had a call with @mathemancer a while back, and he seemed to have similar opinions as well.

> we could make the default presentation the simplest one while making it very easy to expand out to the more detailed view of seeing additional columns

I love this. This is what I tried going for but I realize that we could do that for everything other than the record summaries.

kgodey:
I also played around with this locally and +1 to everything @zackkrida said. This is awesome! @pavish let me know if you want to brainstorm about the UX for making this more "simplest possible + expandable".

seancolsen:
I played with this a bit.

Impressive work, @pavish! I like the direction this is heading!

- @zackkrida said:

    > make the default presentation the simplest one

    +1 on this!

- Pavish said:

    > I'd be more inclined towards the default record summary being a combination of atleast 3 columns.

    I'm not sure about this. I have some opinions, and I'd be curious to bounce some ideas around. But this doesn't seem to be the central subject here, so I'll hold off on launching into this topic. For now I'll just say: I'd like to be involved in any conversations that might result in changes to the behavior of the default record summary template.

- @zackkrida said:

    > become a part of Mathesar in our next release for both forms and the table view.

    I support working towards integrating this work into the table page eventually. But I feel strongly that we should not aim for that in the next release.

    Since Forms is an entirely new feature, I'm open to new designs within the forms system, even if they are quite rough. I'm inclined to ship something fast, get feedback, and polish it later. How polished does the row seeker need to be in order to integrate into forms? For me, the bar is very low! Also, the nature of the problem space for our first iteration on forms is very different from the table page. Since it's insert-only, and focused on anonymous users, the problem space is much more narrow than with the current table page. How polished would the row seeker need to be in order to _replace the record selector_?. Here, my bar is much higher! I'm not saying we shouldn't try. I'm just saying let's not block the release on this polish work.
    
    I think we ought to be able to ship forms without integrating the row seeker into the rest of the app. Then we could work on reconciling the two, ideally into a single construct that has all the power of both. But that will take more design and implementation than I think is necessary right now.

- I think the row seeker really shines in simple scenarios. For tables that don't have a lot of rows and columns, it seems preferable to the record selector. It obviously needs some more work for forms, but I don't think it's that far off.

    I'll describe some friction when using it on tables with a lot of rows or columns. But I think some friction when filling out a form is perfectly fine for now. As a tool primarily focused on _anonymous_ data collection, the forms system can have some friction in my opinion. My understanding of the "problem to solve" with forms is that we're targeting use-cases similar to Google Forms, which (to my knowledge) doesn't even have _any_ relational capabilities. So I'd expect the majority of forms to just be a single table. Perhaps there would be some very simple relationships in there, like tagging, where the row seeker currently shines. But my point is that I don't expect forms users to need the sort of power that the record selector offers over the row seeker.

- It took me a little time (and confusion) to understand the difference between searching and filtering. I don't have any ideas for how to improve this though.

- I expected filtering a FK field would allow me to search within the record summary for that field. I can see that it doesn't — it just filters on the raw FK value (i.e. an id number).

    This seems like a usability problem within the overall approach to filtering. I'm not sure what the solution is.
    
    The record selector solves this problem via nested record selectors. I understand that we're trying to avoid that nesting here, for the sake of simplicity. I think this in some contexts (e.g. an anonymous user filling out a form) this need for simplicity definitely outweighs the need for power.

- Filtering UX definitely needs more polish. For example, if I type text into a number field, I get a DB error. I can't edit my filters after I've added them. Multiple filter conditions on the same field seem to be applied via "OR" whereas filter conditions on different fields seem to be applied via "AND". I know you've alluded to some of these issues, and this draft is currently rough. I just want to express that overall I think the filtering feature is something we could potentially spend quite a bit of time polishing, depending on how deep we want to go.

- In terms of the overall design of the row seeker, the above three problems make me question the need for filtering _at all_. Like, what if we only had searching? I think this might actually be okay. Again, I'm seeking the simplest thing that will work _for forms_. There could be duplicate record summaries. Sure. I'm not convinced this "duplicate record summary problem" necessarily warrants the additional complexity of searching _and filtering_ though. Perhaps it does. I'm just tossing this out there because I think we should be seeking the absolute simplest solution that allows us to ship forms.

- In tables with a lot of rows (e.g. `movie_rentals.items`), listing records in the row seeker seems to be about 20x slower than in the record selector, according to my measurements. I don't think this is a problem worth solving for forms right now. For one thing it's probably fine if it's a bit slow. And for another thing, I don't expect people to to create forms that require the row seeker for really tall tables like this. But if we're going to use the row seeker in other contexts, we'll need to address this problem. My hunch is that it's because we're sorting by the record summary. I would expect that sort to require a sequential table scan (very expensive). I have some ideas (and opinions) about other approaches. When/if the time comes to address this performance problem, I'd like to be involved in planning the best strategy.

- Here's another change that I'd like to suggest. (This is more on the implementation-detail side, but it's somewhat structural and it affects UX.) The row seeker works great when its corresponding trigger element is near the top or the left of the viewport. But when the trigger is near the bottom or the right of the viewport, the positioning can result in the row seeker being entirely unusable. This problem is important to solve for forms.

    And here's another challenge: while we have thus far been optimizing Mathesar for _desktop_ usage, I think the form submission experience needs to be optimized for mobile usage too. So the row seeker needs to work well with a tiny viewport. To me, that means that (under certain circumstances) the row seeker needs to _cover_ its trigger. This makes the positioning logic more complex.

    Here's how the current positioning logic seems to work:

    1. The row seeker opens in loading state. At this point it's quite small.
    1. Popper appropriately positions the row seeker within the viewport so that all UI is visible.
    1. Then, when the row seeker finishes loading, it gets much bigger. How much bigger? That depends on the table structure and row count.
    1. When the size changes, the row seeker can end up off-screen.

    This problem is a little tricky. I can think of different approaches here.

    Here's the simplest solution I can think of (with minimal implementation time):

    1. When the row seeker is initially loading, show a loading spinner inside the trigger element. Don't open the dropdown at all.
    1. Then when it's done loading, open and position the dropdown.

zackkrida:
I've discussed the next steps for this PR with both @seancolsen and @pavish. The plan is for @seancolsen to put some time into the PR later this week with the following 

- Detach the row seeker it from the table page (linked record cell) -- It was only added for demonstration purposes in this 
- Reducing the functionality to the minimum necessary for forms:
    - search and display record summaries only, 
    - hide filtering controls
- Standardize colors (and make sure it looks good in dark mode)

I've assigned @seancolsen even though he isn't likely to start this for a few days.

seancolsen:
Pavish and I had a [call](https://tldv.io/app/meetings/68765f756a8e43001318a993) about this today and agreed that I would do the following:

1. Create a new branch for my work on top of this PR. Keep this PR open as an archive that we might come back to later.
1. Remove filtering feature
1. Remove expandable rows feature
1. Remove link to table
1. Remove config button
1. Improve initial loading state and dropdown positioning
1. Base work on top of forms PR. Open a new PR for the row seeker on top of the forms PR.
1. Integrate the row seeker into the form page.
1. Make the row seeker API accept a form id so that it work for anonymous users.
1. Revert changes to table page and record page, using the record selector there again.
1. Work with Kriti and Zack to get the final UX approved.
1. Work with Brent and Anish to get the backend changes approved.

zackkrida:
Excellent, thanks @seancolsen. I'll be here to help coordinate steps 11. and 12.!

seancolsen:
I've completed steps 1-7 (except for 6).

New PR is here: #4637

It is now stacked on top of the main Forms PR, #4609

Issue #4557, opened 2025-06-17
Figure out how to display forms on the schema page


zackkrida:
## Problem

As part of #4509 we will be adding forms to Mathesar, focused on data entry from anonymous users outside of one's organization (and therefore without their own user accounts). Despite the initial focus on external users, we need to determine how to present access to crucial parts of the forms UI to our users:

- Form builder: Access to build new and edit existing forms from a base table
- Form entry: How to view the working preview of a public form and submit data  

## Proposed solution

Our hunch is that the schema page is the most logical place to add this access, similarly to what we do for the data explorer:

![Image](https://github.com/user-attachments/assets/a7462dd3-17bb-4cbd-ade0-b1145cf9f77f)

However, this poses a problem. We'll now have _three_ related data types to manage on this page:

- Tables
- Data explorations
- Forms

We need to determine a way to present all of these alongside of each other.

Issue #4527, opened 2025-06-03
Allow creating new rows by pasting data


pavish:
## Problem
- Mathesar allows pasting into rows stored in PostgreSQL and "draft" rows that are created explicitly via the UI.
- Mathesar does not allow creating entirely new draft rows when pasting data. 

## Desired behaviour
- The user should be able to paste into the placeholder row, which should then create new rows from the pasted content.

## Additional context
- The PR https://github.com/mathesar-foundation/mathesar/pull/4505 allows pasting into draft rows that are explicitly added in the UI.
- The user request from the underlying issue https://github.com/mathesar-foundation/mathesar/issues/4420, in the [Matrix conversation](https://matrix.to/#/!vXLxAqmrJWsDMWPSpo:matrix.mathesar.org/$IjkoRc93WrU7YS2PuqqWw0429p8HvghLpV8aw-NXoU4?via=matrix.mathesar.org&via=matrix.org&via=t2bot.io), specifies this:
- > We have lots of low-tech users who keep track of their data locally in excel/sheets, and if they can just copy & paste into our mathesar front-end that would save us a lot of time. I am aware that you can export excel to csv but I cannot stress enough how low-tech some of our users are.
- The actual user requirement is to be able to create new rows when pasting data.

seancolsen:
I'm not entirely convinced that this is user-reported. The user said:

> you cannot paste in **new rows**

and

> For our users the ability to paste in **new rows** would be very useful

(emphasis added)

I think we ought to follow up after the 0.3.0 release to confirm. I mention this just because I'm not sure it's worth spending time on this feature unless it actually is important to a user.

seancolsen:
[...]

PR #4563, opened 2025-06-20
Standardize colors

### LINKED ISSUE #4550: Standardize colors and variables
## Problem

The way we use colors and color variables in the Mathesar UI is non-standard. In https://github.com/mathesar-foundation/mathesar/pull/4483, I introduced the beginnings of a new semantic system, but did not finish implementation due to time constraints. It only applies to dark mode, and many colors are still used semi-randomly.

This means the color system is now _more_ brittle due to another [competing standard](https://xkcd.com/927/).

## Proposed solution

Transition entirely to the new system in both dark mode and light mode. This should:
- Clean up tech debt
- Make it easier to develop new designs in the future without as much review
- Set us up for a proper design system
- Make my job easier while reviewing designs because I'll understand the design system
- Make it easier to see when we're adding new concepts or design ideas

## Additional context

I will update this issue with more specifics of the things I'm changing, both for my own reference and to give reviewers more context.

kgodey:
Fixes #4550 

This was the fastest way for me to accomplish #4550, since I needed to see the colors In the product to make it work. This pulls out literally all colors in the product and assigns them semantic meaning, and also standardizes relative color definitions. 

I'm marking this as a draft since I am not a frontend engineer and I do not know how idiomatic my code is.

Please note: I am not super-attached to the particular colors I chose, the main work here was extracting the colors and assigning them meaning. Light mode especially might need a little work on the colors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
Awesome work @kgodey!

This aligns so well with my original vision for the component library to also act as a design system. Thank you so much for doing this, I've been putting this off for so long and I'm super excited about this PR.

I've fixed a number of issues to make this PR close to mergeable - I've raised a separate PR targeting this branch: https://github.com/mathesar-foundation/mathesar/pull/4564. Each commit should provide context on why the change was made. I'm glad to discuss more on them if you want to. Please merge that PR into this branch once you've taken a look.

For @seancolsen and @zackkrida, please hold off on detailed code review until that PR is merged into this branch.

I've not yet completed my review w.r.t the colors and appearance. I have a lot of opinions on the colors and there are several places where improvements are needed. I'll get to it during the week. Having said that, I'm completely satisfied with the approach.

mathesar_ui/src/component-library/button/Button.scss
```diff
@@ -31,155 +31,225 @@
...
 
       &:hover,
       &.hover {
-        --button-background: linear-gradient(
-          to bottom right,
-          var(--neutral-200),
-          var(--neutral-300)
-        );
-        --button-border-color: var(--neutral-500);
+        background-color: var(--color-navigation-auxiliary-hover-10);
```
kgodey: @pavish I had left this as `color-database` because these only show up in the DB connection part of things. I would consider making then `color-primary` rather than introducing a new color here.

mathesar_ui/src/component-library/button/Button.scss
```diff
@@ -283,357 +388,157 @@
...
-  &.btn-outline-danger {
-    --button-background: transparent;
-    --button-border-color: var(--danger-border-color);
-    --button-color: var(--danger-color);
+  &.btn-link {
+    box-shadow: none;
+    padding: 0.3em;
+    height: auto;
+    min-height: 0;
+    color: var(--color-navigation-auxiliary);
```
kgodey: @pavish same with this, this only seemed to be used for DB-related things like adding constraints

mathesar_ui/src/component-library/common/styles/colors.scss
```diff
@@ -0,0 +1,170 @@
...
+
+// Special semantic colors
+$input-fill: hsl(212, 84%, 44.1%);
+
+// Color definition by usage
+$utility-colors: (
+  'brand': $tomato,
+  'brand-muted': hsl(hue($tomato), 40%, 50%),
+  'navigation': hsl(hue($wisteria), 10%, 70%),
+  'navigation-auxiliary': $amethyst,
```
kgodey: if we do want a new name for this, I'd call it accent-secondary and rename primary to accent-primary @pavish

kgodey:
@pavish Thank you! I'm glad you're satisfied with the approach.

Feel free to share any opinions on colors, I am not attached to the ones I picked and I think by the end I had spent too much time looking at them to make good choices.

pavish:
@kgodey I need to clarify, I am excited and satisfied with the color system re-organization approach and the way we're generating color specific variables. I'm still reviewing the color system and the colors themselves, and these are the steps I think we'd need before we could get this PR merged:
- We need to take some more time to make the design system consistent.
- We need to go through each page to test the colors for the dark theme.
- We need to improve the light theme colors.

W.r.t making the system more consistent, here are my thoughts as of now:
- semantic colors: which show intent (not purpose - which is good), 
- utility colors which include components/actions
  - they don't follow any specific pattern, we need to establish one
  - we need to make these utilities minimal, especially modify `highlight-*` utilities to denote purpose
- colors split by surface, text, and border:
  - surface has elevation specific styles but also has elements/components specific styles
  - we should separate out the elevation styles from the component styles - both can be within surface, but the component styles should depend on the elevation styles 
  - text & borders are mostly okay as is, since their representation is based on surfaces as a whole combining elevation & components - but we can improving them based on purpose/area of usage.
- there are no clear distinction between brand primary, secondary, tertiary/neutral colors - color-brand & color-primary compete over it

I think the next step would be to discuss these changes.

Having said that, this is already a massive improvement over our existing system, and I'd be glad to help improve just the colors and get this merged, even if we don't achieve perfection w.r.t the system.

mathesar_ui/src/component-library/button/Button.scss
pavish: I agree. I added this here to move the product specific utility colors to App.svelte and to tackle the colors as part of my further review.

I commented below to describe further steps we need to take.

pavish:
I'll try demonstrating the changes we need to improve design system consistency here:

#### 1. Clear distinction between brand and action colors
```scss
// Denotes branding specific colors
$brand-colors: {
  'regular': $tomato,
  'muted': hsl(...),
}

// Denotes action type - used in buttons, links etc.,
$action-colors: (
  'primary': map-get($brand-colors, 'regular'),
  'secondary': hsl(...),
  'link': hsl(...), // or tertiary
);
```

#### 2. Reduce utility colors and denote them by purpose
```scss
$utility-colors: (
  // Interaction
  'selection': hsl(...),
  'navigation': hsl(...),

  // visual highlights
  'highlight-subtle': hsl(...), // We can also try denoting this by usage type rather can role
  'highlight-neutral': hsl(...),
  'highlight-high': hsl(...),
);

// Product specific utilities are good as is.
```

#### 3. Surface colors - separate elevation and components
- We don't have to change the names of the finally generated variables, this is a change only in the scss file.
```scss
// All of these can be within --color-surface-*

// denotes elevation
$surface-elevation: (
  'base': hsl(...),
  'elevated-low': hsl(...), // We can also consider a number system like `elevation-1`
  'elevated-mid': hsl(...),
  'elevated-high': hsl(...),
  'elevated-top': hsl(...),
);

// denotes component specific surface controls - can depend on surface-elevation
// these denote visual components - not html tags
// we can further improve this by making it purpose based vs element based, but this is good
$surface-components: (
  'canvas': map-get($surface-elevation, 'elevated-low'),
  'card': hsl(...),
  'header': hsl(...),
  'control': hsl(...),
  'fieldset': hsl(...),
  'input': hsl(...),
  'supporting': hsl(...),
  'token': hsl(...),
);

// utility surface colors
$surface-special: (
  'fill-accent': hsl(...),
  'inset': hsl(...),
  'light': hsl(...),
);

// We should rename/remove `input-fill` to denote usage/purpose/or-component
```

#### 4. Border & text - we could make them depend on surface, and/or group by usage/purpose instead of components (if we have time)
```scss
// This has very little effect on code but offers better dx
// All these can be within `border-color-*` and `text-color-*`
// No need to change the names of the finally generated variables, this is a change only in the scss file.
// We can just leave the existing code as it for these, these changes are not very important.

$border-elevation: {
  'base': hsl(...),
  'elevated-low': hsl(...),
  'elevated-mid': hsl(...),
  'elevated-high': hsl(...),
  'elevated-top': hsl(...),
}

// The don't depend on one-another but can depend on border-elevation
$border-structure: (
  'container': hsl(...),
  'layout': hsl(...),
  'section': hsl(...),
  'grid': hsl(...),
  // ...,
);

$border-interactive: (
  'input': hsl(...),
  'control': hsl(...),
  'card': hsl(...), // I'm not entirely confident that card belong here
  // ...,
);

$border-decorative: (
  'divider': hsl(...),
  'shadow': hsl(...),
  'token': hsl(...),
);

// the structure, interactive, and decorative items can also be merged into `components` like with surface

$text-roles: (
  'primary': hsl(...),
  'secondary': hsl(...),
  'tertiary': hsl(...),
);

// These might even be auto-generated for each text type if we want them to be
$text-emphasis: (
  'faint': hsl(...),
  'strong': hsl(...),
  'highlight': hsl(...),
  'highlight-elevated': hsl(...),
);

$text-components: ( // components denote ui visual components, not html tags
  'label': hsl(...),
  'input': hsl(...),
  'icon': hsl(...),
  'visited-link': hsl(...),
  // ...,
);

// I think `disabled` should either be common to all these styles and auto-generated, or should be called something else.
```

#### 5. Simplify mixins
We can simplify the mixins by DRYing them up, but this is low priority.

kgodey:
@pavish This all makes sense to me. I'm happy to defer to your judgment here / hand the PR off to you (unless you particularly want my opinion on something). If there's anything I did that was confusing or you don't know the purpose of a variable, let me know!

My only note is that the three highlights cannot be subtle / neutral / high because they're the three colors used in the "Create Relationship" diagram so they have to be equally saturated.

I'll also defer to you and @seancolsen / @zackkrida on how much work to do before merging this.

seancolsen:
@kgodey I'm excited to take a look at this!

Heads up though that @zackkrida has told me I should prioritize work on #4557 and #4527 before I review this PR. This is in my queue though!

kgodey:
Sounds good @seancolsen, no rush from my end.

zackkrida:
This is a note for myself of some quick color issues I noticed, I don't think anyone has to make edits yet:

## 1

Screenshot From 2025-06-24 21-01-40

The color for the update indicator is plain white or black and should be changed to something more prominent

## 2

The accent color for "administrative" pages appears grayscale in both themes. It looks okay in light mode but I don't think it looks right in dark mode:

Screenshot From 2025-06-24 21-02-55
Screenshot From 2025-06-24 21-02-42

## 3

Screenshot From 2025-06-24 21-09-37

This tooltip in dark mode has a sort-of "aqua" blue/green theme that I don't find present anywhere else and feels incongruous with the rest of the colors.

kgodey:
FWIW all tooltips are the same teal color, the semantic "help" color.

seancolsen:
@kgodey I've begun reviewing this now. I'm expecting this review to take a while though.

@zackkrida had previously passed along this preference from you:

> She'd rather you (or any reviewer) just fix any small issues (layout shift on hover, contrast issues, etc.) rather than document them

Based on this preference, I'll push commits myself instead of requesting that you make changes. But in addition to those commits, I probably end up documenting any user-facing issues with screenshots too, just so the rationale is clear for any changes that I make.

kgodey:
Thanks @seancolsen, sounds good. And yeah, that preference is accurate (mainly because I'm not sure when I'll have a window to address review changes, so it seems most efficient for you and Pavish to make changes directly). 

I do want to review any actual color changes, though (although it is fine if you go ahead and make them and ask me to review after).

seancolsen (commented):
I'm mostly done reviewing this now.

I'm submitting this one comment as an official review, but I'll be dividing the substance of my review up into separate comments to follow shortly. I'm structuring my review this way so that I can come back later and update a single comment with more findings later if needed (as I did many times with the last styling PR).

Review status:

- [x] "I like..."
- [x] User-facing issues
- [ ] Code review

seancolsen:
## I like...

- more color!
- the overall aesthetics of the new colors
- having different colors for different entities (e.g. schema vs table)
- increased prominence of the cell selection
- the "Welcome to Mathesar" message on the home page
- the exploration icon in the nav bar's breadcrumb dropdown (Thank you for fixing!)
- moving "Edit Connection" into the falafel menu on the database page

seancolsen:
## User-facing issues

### input_font_size

I don't like this increase in font size for inputs.

image

When inputs have a different font size from the base text, I have found that there can be design and implementation issues that become more challenging.

For example, this increase in font size now causes text in table cells can move around when entering edit mode.

[Screencast_20250703_133620.webm](https://github.com/user-attachments/assets/bedc9e77-d9ce-40cf-b204-078229140352)

I would like to fix this by reverting the input font size to `1em`.

### focus_indicators

Some places that previously had visual focus indicators now have more subtle indicators, or in some cases no visible indicator at all.

For example:

- This is from the "New User" form. The first input is focused in both cases.

    | Before | After |
    | -- | -- |
    | image | image |

Other places I've noticed regressions in focus indicator:

- Lots of text inputs, generally
- Schema card
- Schema card falafel menu

### border_on_hover

The following UI elements have a border only when hovered. This causes layout shift.

- Table page > "Delete Table" button
- Schema Page > Exploration card
- Database Roles > Delete Role button


### feedback_button_focus_border

The feedback button has a border only when it's focused. This leads to some small layout shift that feels weird.

[Screencast_20250703_141354.webm](https://github.com/user-attachments/assets/b6127e6f-e825-4da0-8731-6ad44df34cbe)


### schema_card_bg_low_contrast

The contrast between the schema card background and the page background seems a bit too low. It's harder to distinguish the cards from the background.

image

### schema_card_active_bg

The schema card has an `:active` style. But what's especially strange is that the style gets applied when clicking on the falafel menu. That was not happening before.

[Screencast_20250703_142338.webm](https://github.com/user-attachments/assets/ef30ccba-bdff-4285-ad3b-9fd123fec87b)

I would like to remove the active style entirely. I don't think we need it.

### help_glow

The white glow around the help popover looks a bit strange to me.

Image

### active_cell_indicator

In light and dark mode, there is no longer a visible indicator for the active cell.

https://github.com/user-attachments/assets/c668d62c-2b98-44fd-89a3-6259faaf193f

### error_box_bg

I think the error boxes looked better with a red background

Image

### db_page_schemas_loading_state

On the database page, the schema list no longer has a visible loading indicator

https://github.com/user-attachments/assets/d3e5d391-14fe-4232-baa9-607abb8af98a

### modal_and_overlay_bg

In light mode, I liked the modal better with a white background because I think stands out better from the overlay. And I think the overlay should be more transparent (and perhaps darker too).

Image

In dark mode too, I think the overlay should be more transparent.

Image


### save_button_dot

The unsaved-changes indicator dot is not visible enough

| Before | After |
| -- | -- |
| Image | Image |


| Before | After |
| -- | -- |
| Image | Image |

### admin_menu_dot

The notification dot in the admin menu looks weird now.

| Before | After |
| -- | -- |
| Image | Image |

| Before | After |
| -- | -- |
| Image | Image |

### array_pill_bg

In light mode, I find the array pill background to be a bit too dark (leading to low contrast and reduced legibility).

Image

Dark mode is okay.

### null_inside_array_pill

In light and dark mode, a null array is confusing

| Before | After |
| -- | -- |
| Image | Image |

| Before | After |
| -- | -- |
| Image | Image |

### pk_cell_bg

PK cells should get a darker background to indicate that they cannot be edited.

Image

I _think_ dark mode is okay here (though I'm not sure how much distinction people would actually prefer when using dark mode). At the very least, the PK cells in dark mode seem to be on par with they way they were before this PR, so I'm inclined to keep them this way.

Image

### panel_resize_indicator

The panel resize indicator is harder to see 

Before:

https://github.com/user-attachments/assets/07142aa8-4338-4686-9145-1e337724e9e7

After

https://github.com/user-attachments/assets/6fe5b584-e628-4186-9c46-9522c55cca10

In light mode, it's visible but just barely. In dark mode I can't see it at all.

### record_summary_pill_bg

The record summary pill background seems too light. It looks especially bad when the cell is selected.

Image

Dark mode seems okay.

### error_toast

In light mode the error toast doesn't use error colors.

| Before | After |
| -- | -- |
| Image | Image |

### toast_timeout_indicator

Previously, the toast message had a reverse progress indicator to convey the timeout that will remove the toast message. This progress indicator is no longer visible (in light mode and dark mode).

| Before | After |
| -- | -- |
| Image | Image |





### input_error_bg

Previously, the input background would turn red when the input had an error associated with it. I think this was nice.

Image

### parent_record_selector_fk_cell_bg

When a nested record selector is open, the FK cell that opened it looks a bit weird. Dark mode is affected too.

Image

### import_preview_page_bg

These new backgrounds are a bit weird to me.

Image

- The page content previously had a background distinct from the page body. Now it has the same background color, but it has a border. This makes the border look a bit odd to me.

- The two fieldset elements on this page previously had the same background as the page content. Now they have a darker background. This seems weird. I can't find another place where a fieldset has a background. And the difference in backgrounds creates the appearance of a line through the fieldset legend, which I don't like.

### file_upload_hover_bg

This new hover background seems a bit intense. I'd prefer it to be more subtle.

https://github.com/user-attachments/assets/2f54e161-126b-47c0-a973-036555e4879b

### collapsible_fieldset_trigger_focus

When the trigger for a collapsible fieldset is focused, it seems to blend in with its background, at least on the import confirmation page.

Image

### disabled_checkbox

When checkboxes are disabled, their state becomes a bit harder to determine visually.

Image

I think the check mark should be black.

Dark mode seems okay here.

### badge_count_bg

The badge count text color is indistinguishable from its background.

| Before | After |
| -- | -- |
| Image | Image |

| Before | After |
| -- | -- |
| Image | Image |

### blue_green_color_combo

Aesthetically, I find this combination of colors to clash.

Image

### sort_condition_opacity

| Before | After |
| -- | -- |
| Image | Image |

### upgrade_cards

When a new version of Mathesar is available, the upgrade notification is not as prominent and visually appealing as it was before.

Image

Image

### login_page

The login page seems to be lacking style.

Image

seancolsen:
I'm still reviewing the code. I expect to have comments on that submitted tomorrow. After that I'll set about fixing the user-facing issues.

kgodey:
Thanks for the review, @seancolsen. Some comments:

> input_font_size

This was not an intentional change, sorry about that.

> focus_indicators

I did not verify contrast here, my primary goal was to make this a parameter that's easily changeable. Hopefully this is easy to fix by just changing the border token value for inputs and other affected areas.

> border_on_hover
> feedback_button_focus_border

Layout shifts were not an intentional change, I was just changing a lot of things at once and holding a billion details in my head.

> schema_card_bg_low_contrast

Agreed, hopefully this is also just a single token change.

> schema_card_active_bg

The falalel menu thing was not intentional, but please do not remove the active style entirely, it makes the UI feel more responsive when actually clicking on something changes appearance. 

> help_glow

You may need to change the light mode shadow values. I think dark mode probably needs more prominent shadows than light mode.

> active_cell_indicator

Not intentional.

> error_box_bg

There is still a red background, but the color token probably needs to be adjusted, it's too light.

> db_page_schemas_loading_state

Not intentional.

> modal_and_overlay_bg

Agreed about the overlay transparency and color. I would rather fix this by only changing the overlay appearance, rather than the modal background (because we have a consistent way of treating "elevation" in the UI, the modal background is part of that system).

> save_button_dot
> admin_menu_dot

I thought I fixed these, sorry. I would make a new semantic token for these that uses the brand color rather than reusing the danger semantic token here.

> array_pill_bg

I thnk that should just be a matter of updating the surface token color.

> null_inside_array_pill

Agreed, not intentional.

> pk_cell_bg

Yeah dark mode looks right here. In light mode, this should just be changing the color associated with the surface-disabled token.

> panel_resize_indicator  

This was supposed to use one of the help colors, I think. It was much more prominent for most of the time I was working on this PR, I didn't realize it regressed. Maybe use a different help token for this.

> record_summary_pill_bg

Agreed, I wanted to take another look at the colors once I had a little bit of distance because I was working on it too much at the time to be able to tell what looked good / bad anymore.

> error_toast
> input_error_bg
> badge_count_bg

This is probably the danger tokens being miscalibrated somehow, same as "error_box_bg"

> toast_timeout_indicator

Not intentional, sorry.

> parent_record_selector_fk_cell_bg

Not intentional

> import_preview_page_bg

Some of this was intentional, I found the previous version of the import page hard to process without more contrast, but I am not attached to the _specific_ colors, I just wanted more contrast between elements.

> file_upload_hover_bg

Agreed, that's weird, I thought hovers were just a standard change for all colors.

> collapsible_fieldset_trigger_focus
> disabled_checkbox
> sort_condition_opacity

Feel free to fix.

> blue_green_color_combo

I agree, this was going to be part of me taking a look at the colors again.

> upgrade_cards

Good to know, yeah, this should be fixed.

> login_page

I meant to review the Django templates and forgot, sorry about that.

seancolsen:
I have finished my review of the code in this PR and pushed some commits to address issues I noticed.

## Code issues I fixed

- In `EntityType.svelte` you had added some CSS with a syntax error (`var(-text-secondary)`). It turns out this component is not used anywhere. I removed the component via c45cfb496c9dbe351b572bf9f2175f84cac4a95f.

- In `SheetColumnHeaderCell.svelte` you had added some CSS with a syntax error (`var(font-weight-bold)`). I removed this line via 59e0cfd50012d5da5d0951e3606a1cf4b19b2ddf. Before this PR, this code was valid CSS (`var(--font-weight-bold)`), but that style actually had no affect because it was being overridden by style on a child element.

- In `SheetColumnHeaderCell.svelte` you had changed the font size for the column header cell. Previously it was `var(--sm1)`. You changed it to `1rem`. Upon comparing the results from the user's perspective, I think it looked better before (with a smaller font size). I'm guessing this change was a mistake, so I changed it back to the smaller size via a24f118d0d2ca165e0df7a93c54144abc2e69dce.

- I noticed that you made use of Svelte's [style props](https://v4.svelte.dev/docs/component-directives#style-props) feature in a number of places. Pavish and I debated how best to use this Svelte feature when it came out a few years ago. I was initially pretty gung-ho on using it, but Pavish cautioned against it. I've since come to see the pitfalls of this feature too — and embrace Pavish's opinion against using it. I could go into detail about these pitfalls, but it's unfortunately not super quick to explain, so I'm just going to hand-waive the rationale away for the time being. I left some of your style props in place. But there was a very low-level component, `NameWithIcon`, where you were accepting style props in a number of places throughout the codebase. I refactored that to use our pattern of putting CSS variables into an object-based component prop. I made that change via 0a30bb1f3306ddf10a697f2ca6eec5b9d1128467.

- In `GridTable.svelte` you changed `grid-template-rows` from `var(--Grid-table__template-rows)` to `1fr auto`. This change stood out to me because I was curious if that removed CSS variable was being used anywhere else. It wasn't. Good. But the new value of `1fr auto` also seemed odd to me. This doesn't appear to cause any issues with our current usage of `GridTable`. But I think in could, in theory, cause weird behavior if `GridTable` were to be used in a different context. Just to be safe, I changed this to `var(--Grid-table__template-rows, auto)` via c814f9725623829cd3db579cd256314456af3cd9. This is most likely inconsequential. I'm just trying to cover our bases here.

- I reverted some small changes you made in a few toast-related files via 7e0455f8f8959154a124d832157a75f9344047a0. As far as I can tell my change doesn't affect any behavior — it just keeps the code cleaner by maintaining component-specific CSS variables for clearer intent and looser coupling.

- In 5ba9cb3be8fdf7925da4a5805b1b5813cec4a39d I removed a now-outdated code comment. (This is not strictly related to any of your code changes.)

- In `Menu.scss` you had added some small margin atop the menu divider. This looked like a potential mistake to me. Plus, the user-facing appearance and behavior seemed better before this change. So I reverted it via c7a7506dd604fae63fe3c2825125ebd2c2355c38.

- In `Dropdown.scss` you had a `button.dropdown.trigger > .label .name` CSS selector.

    This does not follow component encapsulation because the Dropdown component's `.label` element doesn't contain a `.name` element — it only contains a slot that _could_ contain a `.name` element. This sort of encapsulation breakage could potentially lead to strange bugs in the future if left in place.

    I went searching for some usage of the Dropdown component that would satisfy this CSS selector. I found one example here:

    

    In toggling this CSS on/off, I _was_ able to observe an effect on the above example, but the effect wasn't discernable to the naked eye. The color of the text changed, but only a tiny amount.

    There could be other examples too. Perhaps there is an example somewhere in the app where removing this CSS would lead to a regression. I'm not certain. But I think the risk of removing this CSS is lower than the risk of keeping it. So I removed it via 8a8ca20754d08124714a0d6d4b850167a5c11b61.

    I have a hunch that you may have added this CSS intentionally but put it in the wrong place. If you happen to recall what problem this CSS was solving, then let me know and I'll try to figure out a better way to solve that problem.

- In `Collapsible.scss`, you had added this new CSS:

    ```scss
    .collapsible .collapsible-header > .btn-outcome {
      background-color: var(--semantic-outcome-inner-bg);
      border: 2px solid var(--semantic-outcome-inner-border);
      color: var(--semantic-outcome-inner-text);

      &:hover,
      &.hover {
        background-color: var(--semantic-outcome-inner-bg-hover);
        border: 2px solid var(--semantic-outcome-inner-border-hover);
      }
    }
    ```

    Similar to the above bullet point, this CSS does not follow component encapsulation, so I sought to remove it.

    This is special CSS to handle the case of an outcome button being inside a collapsible header. I'm reasonably certain that all such instances occur within `LinkTableForm.svelte` and `NewColumn.svelte`.

    In dafa7c0c8f347018f760fd6450ff2f8a6630a3b2 I refactored this CSS to use better component encapsulation.

    1. Before this PR:

        image

    1. After your changes:

        image

    1. After my follow-up adjustments:

        image

    You can see this had a small effect on the user-facing result. It's more work to make it look exactly like yours while following proper patterns. I can do that extra work if you think it's necessary. But for the sake of time I think this is good enough.

    I also had to remove this CSS which you added to `Collapsible.scss`

    ```scss
    .collapsible .collapsible-header .collapsible-header-btn {
      background-color: var(--surface-elevated-high);
      color: var(--text-header);
    }
    ```

    The reason here is that `background-color` and `color` were overriding the CSS set from `Button.scss` (via the Collapsible component's `triggerAppearance` prop) due to a higher CSS specificity. I had to remove those properties from `Collapsible.scss` to allow the `triggerAppearance` prop to work as it was designed.

    I also noticed you added:

    ```scss
    .collapsible .collapsible-content .label {
      color: var(--text-label);
    }
    ```

    I found an example of this being used within the "Create New Table" modal, but the CSS didn't seem to have an effect on the result. I'm wondering if this new CSS may have been a mistake. Since it doesn't follow component encapsulation, I removed it, for reasons similar to the above. Finger-crossed that this doesn't cause a regression.

- In `Button.scss`, you had added a `&.close-button` section. Again, this is an encapsulation problem. I was able to trace this class usage down and determine that it's limited exclusively to the close button inside `Window.svelte`. I tested the user-facing behavior in light mode and dark mode with your additional `&.close-button` CSS and without it, making sure to check all states across focus, active, and hover. Interestingly, the focus state was problematic _with_ your CSS — the contrast was too low for me to read the "X". Removing your `&.close-button` CSS had some subtle effects on the hover and active states, but they still look fine. I would include screenshots, but it would be kind of a lot of work (16 screenshots in total). So hopefully you can trust that I think it's fine without this CSS. I removed it via 60a0e7108597b8b17c2f77bac916c04842c29224.

- In `Button.scss`, you added a `.page` selector — another encapsulation problem.

    I think I see why you did this. Some of the CSS you added in `Pagination.scss` was not taking effect (because the `Button.scss` was winning due to higher CSS specificity). The Pagination component seems to be the only relevant code location here.

    I fixed this via 26e882ccba8b277e973087772a14f2ea9ef3eccd with a change that is also somewhat hacky. The user-facing behavior is the same as you had it, but the code will be less likely to cause issues down the road.

## Other comments

- I noticed that you're using the CSS [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) function a lot. This is neat, but it's a newer CSS feature, so I was mildly concerned that we could run into compatibility issues with older browsers. Caniuse [says](https://caniuse.com/mdn-css_types_color_color-mix) that **91.76%** of global users have browsers supporting this feature. In the past we've typically aimed for a higher level of support before adopting a new web feature — especially when implementing the feature in a manner that could render Mathesar unusable without support (as I think would be the case here). To put it another way: I think you'd agree that we wouldn't want 8% of Mathesar users to see some incorrect colors that might make text unreadable.

    With this potentially being a serious issue, I wanted to dig deeper into the browser support question. So I took a look at our Simple Analytics data.

    Based on our own analytics data, I think the browser support threshold is more like **99.7%** — very reassuring!
            
    That means that 99.7% of our _marketing site users_ have browsers which support the `color-mix` function — much higher than the 91.76% reported on caniuse! I think that's good enough.

    <details>
    <summary>Here are the steps I took to determine the 99.7% number</summary>

    1. I exported the analytics data for three months: April, May, and June as a parquet file.
    1. I loaded that into DuckDB via:

        ```sql
        create table events as select * from '2025-07-09_mathesar_org_datapoints.parquet';
        ```

    1. I inspected the browsers via:

        ```sql
        select browserName, count(*) num from events group by browserName order by num desc;
        ```

        ```
        ┌─────────────────────────┬───────┐
        │       browserName       │  num  │
        │         varchar         │ int64 │
        ├─────────────────────────┼───────┤
        │ Google Chrome           │  6890 │
        │ iOS Safari              │  1500 │
        │ Microsoft Edge          │  1259 │
        │ Safari                  │  1141 │
        │ Brave                   │  1096 │
        │ Firefox                 │   780 │
        │ NULL                    │   667 │
        │ Chrome Mobile           │   192 │
        │ Firefox Mobile          │   101 │
        │ Opera                   │    94 │
        │ DuckDuckGo              │    50 │
        │ Chrome for Android      │    38 │
        │ Opera GX                │    29 │
        │ Microsoft Edge Mobile   │    10 │
        │ Google Search Mobile    │     8 │
        │ Yandex                  │     6 │
        │ WebKit                  │     4 │
        │ Opera Air               │     3 │
        │ Microsoft Edge WebView2 │     3 │
        │ Safari Mobile           │     2 │
        │ Vivaldi                 │     2 │
        │ Opera Touch Mobile      │     2 │
        │ OperaMobile             │     2 │
        │ Google Search           │     2 │
        └─────────────────────────┴───────┘
        ```

    1. For the top handful of browsers, I dug deeper into the version info via:

        ```sql
        select browserVersion, count(*) as num
        from events
        where browserName = 'Chrome Mobile'
        group by browserVersion;
        ```

        ```
        ┌────────────────┬───────┐
        │ browserVersion │  num  │
        │    varchar     │ int64 │
        ├────────────────┼───────┤
        │ 135            │    63 │
        │ 134            │    17 │
        │ 133            │     8 │
        │ 138            │     3 │
        │ 131            │     1 │
        │ 137            │    35 │
        │ 132            │     3 │
        │ 136            │    62 │
        └────────────────┴───────┘
        ```

    1. I wrote a quick macro to determine browser support for the `color-mix` feature based on the browser name and version, consulting MDN for the compatibility data.

        ```sql
        create or replace temp macro supports_color_mix(browserName, browserVersion) as case
          -- technically Safari should be >= 16.2, but we don't have point-release data
          when browserVersion is null then null
          when browserName = 'Google Chrome'  then browserVersion >= 111
          when browserName = 'iOS Safari'     then browserVersion >= 16
          when browserName = 'Microsoft Edge' then browserVersion >= 111
          when browserName = 'Safari'         then browserVersion >= 16
          when browserName = 'Brave'          then
            case
              when browserVersion < 50 then NULL -- some alternate version scheme I don't understand
              else browserVersion >= 111 -- normal (Chrome-based) version scheme
            end
          when browserName = 'Firefox'        then browserVersion >= 113
          when browserName = 'Chrome Mobile'  then browserVersion >= 111
          when browserName = 'Firefox Mobile' then browserVersion >= 113
          when browserName = 'Opera'          then browserVersion >= 97
          else null
        end;
        ```  

    1. Then I queried it with this SQL:
        
        ```sql
        with
          s as (
            select supports_color_mix(browserName, browserVersion::real) as has_support
            from events
          ),
          b as (select * from s where has_support is not null)
        select sum(has_support) / count(*) * 100 from b;
        ```

        This gives **99.7** 

    Caveats:

    - This is actually measuring page _hits_, not unique users. I'm not sure how big of a deal that is though. It might actually be helpful to index more highly on users that are hitting the site more. I don't know.

    - I'm tossing out a number of analytics events for which I can't figure out the browser support.
    - For Safari 16, I'm erring on the side of optimism since we don't have full point-release version data.

    </details>

seancolsen:
Status:

- [x] List of things I like
- [x] User-facing issue discovery
- [x] Code review
- [x] Code fixes
- [ ] User-facing issue fixes
- [ ] Resolve merge conflicts

I'm done with my code review now. Tomorrow I'll begin work fixing the user-facing issues. This could end up taking me a few days. We'll see.

seancolsen:
@kgodey and @zackkrida I want to give a status update.

I'm working my way though fixing the user-facing issues in this PR, but it's going slower than I had hoped. I've pushed the commits I have so far and provided commentary on those fixes below.

But I mostly wanted to post this comment in order to highlight some friction I'm encountering on some issues. I describe it below.

## Friction

Most of the issues I've fixed so far have been straightforward. But I'll be honest: I've been tackling the easy ones first! As I'm starting to get into the harder ones, I'm hitting challenges which are requiring more time.

To serve as a concrete example, I documented my process in fixing a tricky issue. Perhaps there's a general approach to working with this new system that I'm not following correctly.

I fixed pk_cell_bg with a three-line commit. But it took me 2 hours. Oof. Here's my write-up:

### My process in fixing pk_cell_bg

In the `pk_cell_bg` issue. Disabled cells (like primary key cells) no longer appear with a dark background in light mode.

@kgodey said:

> this should just be changing the color associated with the surface-disabled token.

Okay. I went looking for a `--surface-disabled` CSS variable. I didn't find one. I see other `--surface-*-disabled` variables, but I didn't want to go grasping in the dark to change this without understanding it.

So I traced down the code which controls the background color of disabled cells. I quickly identified this code location in `App.svelte` as being pertinent:

```scss
body {
  --cell-bg-color-disabled: var(--surface-inset-muted);
}
```

I wanted to keep `--cell-bg-color-disabled` wired up as-is because its value affects exactly what I want to change. So I was faced with the decision of what to do about its value. Here was my thought process:

1. Does the combination of "surface", "inset", and "muted" _make semantic sense_ here? I'm not sure.

1. Should I change the value assigned to `--surface-inset-muted`? Probably not — because I can see it's also used for `.postgres-keyword`, which already looks good. So I won't touch that variable.

1. Maybe I should use `--surface-inset-disabled` instead? The "disabled" term does seem to make more sense than "muted", semantically speaking. After all, these are _disabled cells_.

    But when I use `--surface-inset-disabled`, the color looks too dark:

    Image

    Hmmm.

1. Should I change the value assigned to `--surface-inset-disabled`? Hmmm... I can see it's used in two other places. 

    - One place is in `LinkedRecordInput.svelte`, as the background for a disabled linked record input. That seems fishy from a semantic perspective, because "input" would be more appropriate than "inset" here. I wonder if that's a typo. Should I swap that CSS variable out too? It's tricky to even test this because it's hard to create a disabled linked record input. I end up testing it by creating a read-only role and viewing a record page with a linked record input. The user facing behavior is fine as-is:
    
        Image

        Okay. But it does seem like this variable ought to be changed to `--surface-input-disabled` so that it lines up with other inputs, semantically. I do that in e53eaca67c8dc13d0e3ecc4e80dc9ad30955420c. Now we're down to `--surface-inset-disabled` only being used in _one_ place...
    
    - The other place is the loading indicator for cells. That also seems fishy semantically (since "loading" state is independent of disabled "state").
        
        Perhaps I should swap _that_ out for a different CSS variable too? I go looking for `--surface-inset-loading`. That doesn't exist. Should I _add_ it? Hmmm. I'd need to add an entry to `$surface-states` in light and dark mode. Maybe I should. But this seems a bit too complex for me to handle quickly. I might come back to it if needed.

    Ok, so yes — I decide I should change the value of `--surface-inset-disabled`. I want to make it lighter in light mode. This will make the loading indicator lighter. But hopefully that's okay because it's pretty dark as-is. If not I'll fiddle with the loading colors separately.

1. Now, _how_ do I change the value of `--surface-inset-disabled`?

    In `light.scss`, I see `inset` within `$surface-colors`. I feel confident that I should _not_ change that. I don't want to affect the `--surface-inset` variable. Plus, as defined, I can't make it any lighter! Okay, so my only other option seems to be going into `$surface-states` > `disabled` > `l_delta` and tweaking _that_. This makes my stomach turn because I can immediately tell the blast radius is far too large. But let's see what happens.
    
1. I _can_ get my disabled cells looking decent by changing `l_delta` from `-20` to `-4`:

    Image

    (Ideally I might to have it a tad darker, but then I'd need to darken the cell border too.)
    
    But more importantly: how many _other places_ in the app will this change affect?? I don't necessarily want to change every disabled surface! Ugh. 😬

1. Let's go looking at all the disabled surfaces to see what the above change would actually affect. I search the codebase for `--surface-\w+-disabled` and get 7 hits.

    - In `App.svelte` is the occurrence that I care about changing. Good.

    - In `CellFabric.svelte`, I see that it basically wipes out the cell loading skeleton.

        - Before:

            Image

        - After:

            Image
        
        This is not ideal, but (as mentioned above) I can potentially work around this with other changes. Plus it might even look okay if I darken the cell border so that I can darken the disabled cell background too. Moving on...

    - In `TableCard.svelte`, I see that it affects the background of an unconfirmed import listed on the schema page.

        - Before:

            Image

        - After:

            Image

        The lighter background seems fine here. Ok.

    - In `Input.scss` and `LinkedRecordInput.scss` there are three occurrences affecting the background of disabled inputs. Let's see how those look:

        - Before:

            Image

        - After:

            Image
        
        Ugh, this is not good! The inputs don't look disabled anymore.

    - In `Checkbox.scss`, I see that it affects the background of a disabled checkbox.

        - Before: 

            Image

        - After:

            Image

        Again, not good! I don't want to change this background.


1. So to backtrack slightly, what if I still change `$surface-states` > `disabled` > `l_delta`, but I pick a value that seems to work for disabled inputs _and_ disabled cells. How about `-10`? (That's between `-20` and `-4`).

    - Disabled inputs:

        Image

    - Disabled cells:

        Image

        This seems okay, so long as I can darken the border a bit.

        How about this...

        Image

        That darker border seems good for cells!

1. But now I need to go trace down the effects of _that_ border color change to make sure it didn't cause any regressions elsewhere.

    I search for `--border-grid` and find 10 results.

    - The two occurrences in `App.svelte` are for cells — the ones I care about. Good.

    - There's one occurrence in `SingleSelectCell.svelte`, which seems like it needs to match the value used for cells. Good.

    - There are four occurrences in `DateTimePicker.scss`. The change actually improves the visibility of the up/down buttons to change time values. Good.

        - Before:

            Image

        - After:

            Image

    - One occurrence in `AccessControl.svelte` also seems like a win. Good.

        - Before:

            Image

        - After:

            Image

    - In `RecordSelectorCellWrapper.svelte`, this change also darkens the borders of record selector column header cells. This seems good too.

        - Before:

            Image

        - After:

            Image

    - In `TableInspector.svelte`, this change darkens the _right_ border of the table inspector. It's weird to me that the table inspector has different borders on different sides, but okay, this color change seems fine to me.

        - Before:

            image

        - After:

            Image

Okay, so that ought to do it! This was a three line change. But (along with this writeup) it took me two hours.

## User-facing issues I've fixed

### input_font_size

Fixed via 3207ca1f513b12548f69e362243ffdeff3c417d6

### border_on_hover

Fixed via be22581f7841a118f56f5c835a6195b3356bd626

### feedback_button_focus_border

Fixed via 0e3bf5a866ebbd7cef21e6dcc4dd2a27243b25ea

### parent_record_selector_fk_cell_bg

Fixed via 71bb66f5687c5e59ed5cf3e027f9179451ae0b9e

### disabled_checkbox

I fixed this via a2db492319f65431f48411f0b3366bda5030ed1e by adjusting the colors of the check marks when disabled and in light mode.

- Before my changes:

    Image Image

- After my changes:

    Image Image

### sort_condition_opacity

I fixed this via eb6330c0e29461f86379477229321046977b9c13 by slightly changing the design for the sort entries. 

| Before | After |
| -- | -- |
| Image | Image |

It's a subtle difference, but I don't think it looks weird anymore when dragging to rearrange sorting entries.

This seemed like the easiest way to fix the issue.

### active_cell_indicator

I fixed this via a9e714a9a5d0761d2a9e85cc4f4a9d29d347afb1.

### modal_and_overlay_bg

Fixed via 99057c1fa0c42329752750b26bcbad9f9ed5f8d8. Results:

Image

Image

### array_pill_bg

Fixed via 2d52db251a286953b0980a688b882112a2b3fffd. Results:

Image

### null_inside_array_pill

Fixed via 398786b61f038f08d80aa7367638ebf34793fd73.

### pk_cell_bg

Fixed via e22fa111c420f8afda5e523b99e6f9a262770e52. Results:

Image

## User-facing issues remaining

- [ ] panel_resize_indicator
- [ ] toast_timeout_indicator
- [ ] record_summary_pill_bg
- [ ] blue_green_color_combo
- [ ] db_page_schemas_loading_state
- [ ] import_preview_page_bg
- [ ] file_upload_hover_bg
- [ ] collapsible_fieldset_trigger_focus
- [ ] help_glow
- [ ] schema_card_bg_low_contrast
- [ ] schema_card_active_bg
- [ ] upgrade_cards
- [ ] login_page
- [ ] input_error_bg
- [ ] error_toast
- [ ] error_box_bg
- [ ] save_button_dot
- [ ] admin_menu_dot
- [ ] badge_count_bg

seancolsen:
Oh crap — I just realized... that long drawn-out fix I made for `pk_cell_bg`... it inadvertently ends up changing dark mode! And you had specifically said _"Yeah dark mode looks right here"_.

| Before my adjustment | After my adjustment |
| -- | -- |
| Image | Image |

# 🤦

I'll spend some more time ironing this out.

seancolsen:
@kgodey,

So @zackkrida tells me I'm handing this over to you now. Very well then. This comment serves to hand off everything I have so you can pick it up from here.

Regarding `pk_cell_bg`, I ended up making 6ada16bb894e0ca9e07fd92baaf91f8a5d6b4cb8 which I think addresses dark mode. The disabled cell backgrounds look the same as the did before any of my adjustments. The disabled _inputs_ did get a tad darker with this commit, but they seem to look fine that way to me. I'm considering this issue fixed.

Additionally, I fixed a couple more issues, and I've identified some new issues too...

## Additional user-facing issues fixed

### panel_resize_indicator

Fixed via 02774a3fda4739910e862894dadaad720a935029.

### toast_timeout_indicator

Fixed via f50e4b6a46f608fd9e8c2a419c418279abe517a6. Results:

Image

Image

Image

Image

## New user-facing issues reported

### password_reset_page

The password reset page is also missing style

Image

### schemas_loading_indicator

On the database page, there is no longer any visible indication that the list of schemas is loading:

Image

The `Skeleton` component seems to be the problem here. It's used in other contexts too, e.g. table loading state, etc. We should verify that whatever fix is made here also addresses the loading state in other contexts utilizing this component.

### missing_column_drop_indicator

When dragging to rearrange columns, the drop indicator is no longer as clearly visible.

- Before:

    https://github.com/user-attachments/assets/73d146bd-8f00-4f36-9b82-2b3941efd036

- After (light mode — less visible)

    https://github.com/user-attachments/assets/2d96fc0a-e726-4258-97a7-c4b360cf1705

- After (dark mode — not visible)

    https://github.com/user-attachments/assets/ed78a33c-6581-4c6d-9f25-82670e1696a2

### disabled_linked_record_input_low_contrast

When a linked record input is disabled, the record summary contrast is too low to be legible. 

Image

## Remaining user-facing issues

(This includes the new ones reported above)

- [ ] record_summary_pill_bg
- [ ] blue_green_color_combo
- [ ] db_page_schemas_loading_state
- [ ] import_preview_page_bg
- [ ] file_upload_hover_bg
- [ ] collapsible_fieldset_trigger_focus
- [ ] help_glow
- [ ] schema_card_bg_low_contrast
- [ ] schema_card_active_bg
- [ ] upgrade_cards
- [ ] login_page
- [ ] focus_indicators
- [ ] input_error_bg
- [ ] error_toast
- [ ] error_box_bg
- [ ] save_button_dot
- [ ] admin_menu_dot
- [ ] badge_count_bg
- [ ] password_reset_page
- [ ] schemas_loading_indicator
- [ ] missing_column_drop_indicator
- [ ] disabled_linked_record_input_low_contrast

seancolsen:
I wanted to flag another thing that we'll need to do here.

I see a bunch of new code in #4556 from Pavish which references CSS variables that this PR removes. These changes are in conflict conceptually but they won't cause git merge conflicts, so we need to be careful not to let them fall through the cracks. Once the row seeker PR is merged or this PR is merged, then someone will need to go through the row seeker code and fix the incompatible CSS. I assume the same thing probably applies to Pavish's work on the form builder.

kgodey:
I will not be able to get to this work for this release.

pavish:
I'm not going to be able to get this PR ready for 0.5.0

pavish:
@seancolsen  (cc: @zackkrida)

This PR is ready for another round of review.

As discussed, I'm expecting the comments from this review round to be focused on blockers. I'd like to wait for comments regarding color preferences until Kriti's had a chance to go through the product and take a final stab at the colors in a separate PR.

Since Sean's last commit, I've made a number of changes:
- The color system is now more granular.
- The light mode colors are modified to a large extent.
- Focus and active states are made more visible for all commonly used interactive elements.
- I've made some stylistic changes to handle cases like _blue_green_color_combo_
   - The record page and record fk utility colors are changed from asparagus to a fuchsia-like color.
       - This is to ensure that the fk tokens work well in all states (disabled, selected rows etc.,)
   - I've removed the green background from the Refresh button.
   - @kgodey I'm leaving it to you to take the call on modifying them or leaving them as-is.
- Fixed a lot of bugs I came across.

kgodey:
Awesome work, @pavish! I poked around the app while reviewing files, I think the new colors look amazing! I don't have any feedback or changes. Thank you for fixing them and for all the other changes. @seancolsen thanks also for fixing some of the bugs.

PR #4571, opened 2025-06-24
EXPERIMENTAL: Add read-only Django role creation

mathemancer:
This is related to an idea from a GTM meeting: What if we can set up a public-facing Mathesar instance that is _really_ read-only?

- No modification of user data, but also
- **No modification of Django state.**

This PR adds a management command that sets up a database role with read-only access to Django tables, with the exceptions:

- The role can write to the `last_login` column of the user table (otherwise you can't log in)
- The role can still write to the sessions table (otherwise you can't log in or out)

**Technical details**

In the dev setup, the management command can be called via:

```
docker exec -it mathesar_service_dev python manage.py add_readonly_django -v 3 mathesar_ro_django
```

Ugly parts:

- When you try to do something that would alter Django state (other than sessions), you get a permissions error
- The setup process is kinda arduous:
  - Install Mathesar in Docker
  - Point desired domain at it
  - Use typical setup flow to set up a Django superuser
  - Add desired public data
  - Create desired explorations, etc.
  - Create a DB role with limited access to the user database.
  - Create a Mathesar user, assign to that role.
  - Login as that user, do first-time password setup flow. (maybe we could skip this somehow)
  - Run the management command to create the read-only Django role.
  - Modify the docker-compose.yml variables so that the service uses the read-only role for Django.
  - Restart the service.

With that said, all of this works with no real modifications to our preexisting codebase. If we merge this, then later decide we hate it, we can just delete the directory.

To proceed: @zackkrida I'll assign this to you to look at for an initial pass, if you can. If you think it makes sense, we can try to actually set up a public-facing instance and publish that somewhere to see if we get pwned or whatever.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@mathemancer

Overall, it works quite well! The only unexpected behavior I noticed was that I was able to configure the default record summary for a table I had read-only access to. Aside from that, all actions I tested appropriately triggered permissions-related errors.

Specifically, the only places where the user encounters errors are:

- Selecting a base table when creating a new exploration
- Pasting into cells (which is a bug anyway, since we shouldn’t allow pasting into disabled cells)
- Creating relationships — this should probably already be disabled for read-only users (I'll go ahead and file a bug)
- Editing and creating Postgres roles on a schema (Mathesar's stored roles and collaborators are already correctly disabled)

In short, when combined with the existing read-only permissions for Schemas and Tables, this works great. That said, the necessary UI changes seem rather minimal (maybe @pavish or @seancolsen could weigh in here, let me know if either of you need more context). So, I'm happy to merge this, but also to move forward with the UI work to make this unnecessary.

zackkrida (approved):

seancolsen:
@zackkrida I've only just read your one comment, and I don't understand the broader context here. But those issues you mentioned do sound like easy things to address on the front end.

mathemancer:
> @mathemancer
> 
> Overall, it works quite well! The only unexpected behavior I noticed was that I was able to configure the default record summary for a table I had read-only access to. Aside from that, all actions I tested appropriately triggered permissions-related errors.
> 
> Specifically, the only places where the user encounters errors are:
> 
> * Selecting a base table when creating a new exploration
> * Pasting into cells (which is a bug anyway, since we shouldn’t allow pasting into disabled cells)
> * Creating relationships — this should probably already be disabled for read-only users (I'll go ahead and file a bug)
> * Editing and creating Postgres roles on a schema (Mathesar's stored roles and collaborators are already correctly disabled)
> 
> In short, when combined with the existing read-only permissions for Schemas and Tables, this works great. That said, the necessary UI changes seem rather minimal (maybe @pavish or @seancolsen could weigh in here, let me know if either of you need more context). So, I'm happy to merge this, but also to move forward with the UI work to make this unnecessary.

I can't seem to replicate the issue with configuring the default record summary. Were you able to actually save it? That shouldn't be possible since the Django role created by the command doesn't have write access to the relevant table. Are you sure you restarted Mathesar with that role set for the service?

mathemancer:
@seancolsen The broader context is this: This experiment is to try to figure out minimal changes we can make to the Mathesar codebase that would allow us to set up an instance with a publicly-known username and password. We'd load that instance with some public data sets, and create some example explorations of those data sets, and let people tinker with the data (through the Mathesar UI). They should be able to use all features of Mathesar that don't require any persistence. For example, they should be able to create an exploration, but shouldn't be able to save it. They should be able to look at tables and their data, but shouldn't be able to edit data or update table definitions or metadata. This specific PR just sets up a Django role that (should) make it possible to use Mathesar, but should make it impossible to persist anything in Mathesar's database (with minor caveats listed in the PR description).

mathemancer:
Okay, given your approval @zackkrida , I'm going to go ahead and merge this. Then, I'm going to set up a test box with some data, some explorations, and lock it down. We can kick the tires on that and see if anything unexpected arises.

zackkrida:
@mathemancer you're right, I missed this step: 

> Modify the docker-compose.yml variables so that the service uses the read-only role for Django.

It works correctly once I've done that.

PR #4582, opened 2025-06-26
[WIP] DigitalOcean App Platform one-click installation docs

zackkrida:
## TODO

- [ ] Finish, clean up, and restructure the Digital Ocean docs
  - Figure out balance between a how-to guide vs. "marketing"-style copy
- [ ] Apps in the DO app platform do not have persistent storage or volumes, which means Mathesar's internal DB _has to_ be in a paid, hosted database. This is fine (and good, from a backup and data integrity perspective!) but it needs to be clearly articulated to users to either:
  - Use and pay for a Digital Ocean managed Postgres DB cluster (default) for their Mathesar internal _and_ Data databases
  - If they have an existing Data database cluster, consider adding a new DB there for the Mathesar internal DB instead.
- [ ] Add information on how to scale up the Mathesar app instance

### DIFF
--- a/docs/docs/administration/install-one-click-digital-ocean.md
+++ b/docs/docs/administration/install-one-click-digital-ocean.md
@@ -1,0 +1,27 @@
+# One-click Install Mathesar on DigitalOcean App Platform
+
+Install Mathesar to DigitalOcean in just minutes using the convenient one-click deployment button below:
+
+[![Deploy to DO](https://www.deploytodo.com/do-btn-blue.svg)](https://cloud.digitalocean.com/apps/new?repo=https://github.com/mathesar-foundation/mathesar-digital-ocean/tree/main)
+
+Clicking the button will create a new, running Mathesar instance costing ~$12/month.
+
+## Why Choose DigitalOcean App Platform?
+
+[DigitalOcean App Platform](https://www.digitalocean.com/products/app-platform) simplifies deploying, scaling, and managing Mathesar without worrying about underlying infrastructure management. It's ideal for:
+
+- Beginners who need an easy, quick setup without delving into infrastructure specifics.
+- Small to medium-sized teams looking for reliable hosting with minimal overhead.
+- Developers who prefer easy scaling, clear pricing, and stability without needing to SSH into servers or manually manage configuration.
+
+## After succesfully installing Mathesar
+
+### Secure your installation with a `SECRET_KEY` (Strongly Recommended)
+
+### Connect your production database(s)
+
+### Add a custom domain name (optional)
+
+By default, Digital Ocean will give your Mathesar installation its own secured domain name, like `https://mathesar-*.ondigitalocean.app/`.
+
+You can easily add a custom domain name by following Digital Ocean's guide on [Adding a custom domain using their control panel](https://docs.digitalocean.com/products/app-platform/how-to/manage-domains/#custom-domain).
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -16,7 +16,7 @@ This is a quick way to play with Mathesar locally, but is not appropriate for sa
 
     ??? tip "Tips when trying Mathesar locally"
         - To open a [psql](https://www.postgresql.org/docs/current/app-psql.html) shell within the container, run:
-        
+
             ```
             docker exec -it mathesar sudo -u postgres psql
             ```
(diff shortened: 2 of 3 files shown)


zackkrida:
@mathemancer and @kgodey no need to look at this yet, but there are a few small "gotchas", caveats, or little technicalities to figure out here in terms of installing Mathesar with as little friction as possible. 

I'll figure out solutions, but I wanted to tag you in now in case you feel strongly about any of the "TODO"s I've referenced in the issue body, or any other concerns you'd like to raise now rather than later.

docs/docs/administration/install-one-click-digital-ocean.md
```diff
@@ -0,0 +1,52 @@
...
+- Beginners who need an easy, quick setup without delving into infrastructure specifics.
+- Small to medium-sized teams looking for reliable hosting with minimal overhead.
+- Developers who prefer easy scaling, clear pricing, and stability without needing to SSH into servers or manually manage configuration.
+
+## Instructions for installing Mathesar to DigitalOcean App Platform
+
+1. Use the DigitalOcean one-click install button to initialize a new Mathesar installation in DigitalOcean:
+    [![Deploy to DO](https://www.deploytodo.com/do-btn-blue.svg)](https://cloud.digitalocean.com/apps/new?repo=https://github.com/mathesar-foundation/mathesar-digital-ocean/tree/main)
+1. Create a Digital Ocean account and as payment information, if you do not have an existing account.
+1. **Important.** Secure your installation with a `SECRET_KEY` . Press the button below to generate a key, and automatically copy it. **Be sure to keep this key secret.**
```
mathemancer: Things broke between these steps for me. For whatever reason, by the time I'd gotten an account set up, worked through the captcha, and answered their survey about my role, etc., DO had forgotten to redirect me to the actual deployment.

mathemancer: Then I ended up here:

image

The problem was that after re-clicking the button (after getting an account and signing in), I got bogged down in the payment setup process (which wasn't part of the account setup), and neglected the environment variable setup.

I'm a dope, but so are many others...

docs/docs/administration/install-one-click-digital-ocean.md
```diff
@@ -0,0 +1,52 @@
...
+    {% include 'snippets/generate-secret-key.md' %}
+1. Add a `SECRET_KEY` environmemt variable. Click the "Edit" button next to the "Environment Variables" heading, then the "Add environment variable" button. When completed Digital Ocean should look something like this:
+  ![A screenshot of DigitalOcean's UI to configure environment variables](../assets/images/digital-ocean-app-platform-config.png)
+1. Press the "Create App" button.
+
+In a few minutes, you'll have a live Mathesar installation ready to use! Digital Ocean will display your domain name and a link to visit your new site. You’ll be prompted to set up an admin user account the first time you open Mathesar. Just follow the instructions on screen.
+
+## After installing Mathesar
+
+Congratulations on your new Mathesar install! After getting up in running, you may want to take some additional steps:
```
mathemancer: For some reason, after installing Mathesar, I ended up here:

image

So, it wasn't possible to set up an internal database. I understand that the instructions here are to "Connect your production database(s)", but I think "1-click" should let me set up a database on the internal DB cluster to start using Mathesar right away.

mathemancer (changes requested):
Overall, I'm excited about the concept. But, I ended up failing to get a working installation, and there were a few other rough edges.

First the good:

Love the actual docs, the explanation of "who is this for", and the secret key generator.

The bad:

The setup was far from painless for me, and ultimately failed due to a DB permissions issue.

I think we should take another look at the "who this is for" section. Specifically, I think if we're targeting "Beginners who need an easy, quick setup without delving into infrastructure specifics", we need to add more "troubleshooting" to the docs. Maybe give them a hint that the process will be smoothest if they've already signed up for DO, _and_ set up payment. Or, add more screenshots to follow though the "adding payment setup" part of things, so they don't have to leave the trail of the docs, then reacquire the spot they left off (I screwed up at that point).

Overall, I think we should really try hard to make the docs even more idiot-proof, or at least @mathemancer-proof, rather than aiming this deployment method at more advanced users. Obviously, getting the DB permission error worked out is a prerequisite.

I'm stuck at the creating a database step, but I'll try again later today and see if I can figure out how to find some relevant logs. Side note: we should document how to find relevant logs or output on DO if we want this deployment method to be truly first-class. The obviously-available logs on DO are just for the build and deployment, not the running app.

zackkrida:
Thanks, @mathemancer ! I think I can smooth some of these edges out. The DB thing is a pain that I actually forgot about in the current version of this PR. The type of DB DigitalOcean allows you to create doesn't allow CREATE permissions, you have to upgrade the DB. That's a single button click on their UI, but it's still friction and inconvenient for users without an existing database. Nonetheless I'll figure out how to best document it.

docs/docs/administration/install-one-click-digital-ocean.md
zackkrida: I've clarified this limitation in Digital Ocean with its own section of documentation. I think it should be clear enough now.

docs/docs/administration/install-one-click-digital-ocean.md
zackkrida: I've reworked things to prioritize creating an account and/or adding a payment method _before_ using the one-click button. I think that should be sufficient to help folks out.

zackkrida:
@kgodey and @mathemancer this is once again ready for review.

kgodey:
Will review tomorrow.

zackkrida:
@kgodey Really nice optimizations, thanks!

PR #4597, opened 2025-07-01
EXPERIMENTAL: Add example `SECRET_KEY` runtime generator

mathemancer:
This adds a bit of logic to generate and save a secret key if it doesn't exist. The logic is:

- If  there's a secret key in the environment, we use that,
- Else if there's a secret key persisted in config/settings/secrets/secret_key.py, we use that,
- Else we generate a new secret key and persist it in that file (from the Mathesar web service's perspective).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/config/settings/secrets/secret_key.py
+++ b/config/settings/secrets/secret_key.py
@@ -1,0 +1,1 @@
+secret_key = None  # This will be replaced on first run.
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ ENV MATHESAR_DOCKER_IMAGE='true'
 
 VOLUME /etc/postgresql/
 VOLUME /var/lib/postgresql/
+VOLUME /code/config/settings/secrets/
 
 EXPOSE 5432
 
@@ -95,6 +96,8 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     nodejs \
     && rm -rf /var/lib/apt/lists/*
+
+RUN echo -n "secret_key = None" > /code/config/settings/secrets/secret_key.py
 
 
 #=========== STAGE: DEVELOPMENT ==============================================#
--- a/config/settings/common_settings.py
+++ b/config/settings/common_settings.py
@@ -13,7 +13,10 @@ https://docs.djangoproject.com/en/3.1/ref/settings/
 import os
 from pathlib import Path
 
+from django.core.management.utils import get_random_secret_key
+
 from config.database_config import PostgresConfig, parse_port
+from config.settings.secrets import secret_key
 
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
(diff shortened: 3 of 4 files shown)


Anish9901 (approved):
Seems like the linter is failing but otherwise it looks good.
 
@mathemancer do you think we can make this the default way to set the `SECRET_KEY` and deprecate the need to set one for our prod docker compose?

mathemancer:
> Seems like the linter is failing but otherwise it looks good.
> 
> @mathemancer do you think we can make this the default way to set the `SECRET_KEY` and deprecate the need to set one for our prod docker compose?

I think we should. I want to double-check that this works in various deployment setups (for example, do we have any where the user running Mathesar would be unable to edit the required file due to permissions?

mathemancer:
Converted to draft so I can double-check some permissions issues.

zackkrida:
@mathemancer I've undrafted this and assigned it back to you for completion.

mathemancer:
@Anish9901 I completely changed the method of key persistence, so I'm re-requesting review from you.

mathemancer:
@pavish I've also assigned you to review, since I'm not sure how I should verify that this doesn't cause any problems for the build or scratch installation process. Please:

- Check that there are no problems introduced, and
- document or automate checking the build and scratch installation processes.

If there's already such documentation, please direct me to that, and I'll try the verification myself.

config/settings/secrets/secret_key.py
Anish9901, line 1: Is this needed anymore?

docker-compose.yml
```diff
@@ -31,10 +31,9 @@
 # See https://docs.mathesar.org/configuration/env-variables/ for more info.
 #
 x-config: &config
-  # (REQUIRED) Replace '?' with '-' followed by a 50 character random string.
-  # You can generate one at https://djecrety.ir/ or by running:
-  #   echo $(cat /dev/urandom | LC_CTYPE=C tr -dc 'a-zA-Z0-9' | head -c 50)
-  SECRET_KEY: ${SECRET_KEY:?}
+  # (Not recommended) To use a custom secret key, use an .env file, or add a
```
Anish9901: As an existing user (Not recommended) would make me think that there is something wrong with me having my own secret key configured and/or it is now deprecated.

I suggest moving `SECRET_KEY` from `x-config` to the `environment` section for `mathesar_service` container so that we can ensure backwards compatibility. New users don't need to worry about adding one at all.

We could move the documentation for secret key within the `environment` section as well. 

Also, the "HOW TO USE THIS FILE" section mentions the need to add a secret key before running docker compose which needs to be removed.

docker-compose.yml
```diff
@@ -176,6 +175,7 @@ services:
     volumes:
       - ./msar/static:/code/static
       - ./msar/media:/code/media
+      - ./msar/secrets:/code/.secrets
```
Anish9901, line 174: Please add documentation about the `secrets` volume in the "INFO ABOUT VOLUMES" section.

pavish:
@mathemancer 

> document or automate checking the build and scratch installation processes

There's a [GH workflow](https://github.com/mathesar-foundation/mathesar/blob/develop/.github/workflows/test-package-and-install.yml) that does this for every PR.

pavish (approved):
I reviewed the installation scripts specific parts and they look good to me.

config/settings/secrets/secret_key.py
mathemancer: Whoops, good catch.

docker-compose.yml
mathemancer: Okay, reorganized.

docker-compose.yml
mathemancer: Good catch.

mathemancer:
@Anish9901 Should be ready for re-review.

Anish9901 (approved):

PR #4634, opened 2025-07-09
Single Sign-on (SSO)

### LINKED ISSUE #4632: Configure django-allauth within mathesar to allow OIDC logins
https://docs.allauth.org/en/latest/installation/quickstart.html
https://docs.allauth.org/en/latest/socialaccount/providers/openid_connect.html

### LINKED ISSUE #4633: Allow mapping default postgres roles to OIDC logins
We'll need to
- write a custom adapter to map a postgres role to a django `User` created via OIDC.
- Implement email domain based restrictions for login via oidc

Relevant docs: https://docs.allauth.org/en/latest/account/adapter.html

### LINKED ISSUE #4635: Allow configuring OIDC settings and show appropriate buttons on login page

### LINKED ISSUE #4640: Add SSO to analytics

Anish9901:
Fixes #4632
Fixes #4633 
Fixes #4635
Fixes #4640 

**What works:**
- Configuring one or more OIDC providers within the `sso.yml` file.
- Connecting already existing Mathesar users with SSO accounts based on email.
- Automatic creation of Mathesar user on login.
- Auto provisioning of default postgres roles per database along with error message in case of failure.
- Email domain based login restrictions along with error message in case of failure.
- SSO analytics tracking
- Different .png in login buttons for different providers.

**What doesn't work(yet):**
- If a user whose Mathesar account was auto created by SSO tries to login without SSO i.e. by typing in username and password, login fails, as the password is randomly set on that account during user creation.
- Group based postgres role assignment.
- Need to figure out how to distibute the `sso.yml` file for different installation methods such that it doesn't get overwritten on update.

Screenshot 2025-07-15 at 7 13 47 PM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/build-scripts/package/package.py
+++ b/build-scripts/package/package.py
@@ -36,7 +36,6 @@ SOURCE_CONTENTS_TO_COPY = [
     "README.md",
     "requirements.txt",
     "setup",
-    "sso.yml",
     "THIRDPARTY",
     "translations",
 ]


zackkrida:
@Anish9901 this is looking great! I've created an issue to write docs for SSO here: https://github.com/mathesar-foundation/mathesar/issues/4639

I can help with that after I've reviewed this PR. Can you leave an update comment or revise the PR description to clarify what you're still working on and what _won't be_ worked on for this release? I'd like to review this but don't want to do so if you're still working.

Anish9901:
@zackkrida I am currently working on adding:
- .pngs for different providers
- SSO analytics tracking

These won't be worked on in this realese:
- Group based postgres role assignment.
- This edge case: If a user whose Mathesar account was auto created by SSO tries to login without SSO i.e. by typing in username and password, login fails, as the password is randomly set on that account during user creation.

I'll create a separate PR for:
- Figure out how to distibute the sso.yml file for different installation methods such that it doesn't get overwritten on update.

I'll let you know when this is ready to be reviewed should be ready in a couple hours.

zackkrida:
@Anish9901 great, that sounds perfect. Just curious, what's the plan for the logos? It crossed my mind that the simplest approach might be to hardcode a mapping of icons for popular providers and fall back to a generic "lock" icon for everything else, but I'm sure there's lots of good ways to do it.

Anish9901:
@zackkrida the PR is ready for a review, lmk if you run into any trouble or need help with configuration.

@mathemancer I've fixed the bug where the login was failing when a default role wasn't provided, turns out I was missing a null check. Additionally, I've wired up the analytics to track the SSO connected `User` accounts like we discussed in our 1:1.

zackkrida:
@Anish9901 I've done some preliminary testing and this worked great with Auth0 right away. It took me a second to figure out the callback URL but that's something I'm sure we'll include in our documentation.

I'll continue to do a more thorough review tomorrow. One thing that's a bit unfortunate is that the SSO user can see the controls to change their password but will never be able to use them because they don't know the randomly-generated password that was created for them. 

On the note of this problem: "Need to figure out how to distibute the sso.yml file for different installation methods such that it doesn't get overwritten on update." I think we _could_ try an approach with getting clever with dynamic ENV vars that wouldn't be too much trouble, at least for the basic config. For example, something like:

```env
# Example for Google
SSO_GOOGLE_CLIENT_ID=your_client_id_for_google
SSO_GOOGLE_SECRET=your_secret_key_for_google
SSO_GOOGLE_SERVER_URL=https://accounts.google.com
SSO_GOOGLE_ALLOWED_EMAIL_DOMAINS="example.com,xyz.org"

# Example for another provider (e.g., Okta)
SSO_OKTA_CLIENT_ID=your_client_id_for_okta
SSO_OKTA_SECRET=your_secret_key_for_okta
SSO_OKTA_SERVER_URL=https://okta.yourdomain.com
SSO_OKTA_ALLOWED_EMAIL_DOMAINS="example.com,xyz.org"
```

Of course the DB roles stuff would be the most cumbersome part to figure out. Those might mean we have to deal with some JSON strings or something anyway, so maybe @mathemancer's idea to stringify the whole config into a single ENV var is ideal.

Also, we _could_ add some frontend JS to the docs site to help the user generate the sso config, similarly to what I'm doing for secret keys in #4582. That might make whatever solution we choose even less burdensome.

mathemancer (commented):
Overall, very cool. 

Preliminary review: We need to mount an `sso.yml` in the prod docker environment, or the config doesn't work. Maybe this is a signal that we need to try harder with environment variables. I just think it's going to be pretty cumbersome for users to handle their own namespacing. Maybe I'm pessimistic.

You can mount a single file just like a directory, so I got this working in prod (mounted, at least; I haven't tried the superuser creation flow yet) by adding:
```
      - ./sso.yml:/code/sso.yml
```
to the volumes section for the service.

More later, but that's something to work on in the meantime. The other parts of the code are looking pretty good and mergeable to me. The big issue is the config (as you note in the description). We have to get that sorted for all current deployment methods to merge.

mathemancer:
Also: `sso.yml` should be in the `.gitignore`. I think we should do the same "move the example to the real location" trick as for `.env`.

docker-compose.yml
```diff
@@ -164,6 +164,7 @@ services:
     volumes:
       - ./msar/static:/code/static
       - ./msar/media:/code/media
+      - ./sso.yml:/code/sso.yml
```
Anish9901, line 167: @mathemancer I tried running docker compose without the `sso.yml` file and docker creates a new directory named `sso.yml/` if `sso.yml` is not present. This is not a breaking change but maybe we should consider keeping `sso.yml` within the image instead of the `.example` varient, that way:
- The newest `sso.yml` will be copied to users system on first run with the new volume changes.
- If the docker compose file and the sso.yml stays at the same location any changes made to sso.yml by us won't affect upgrades and won't overwrite the existing sso.yml, since the existing sso.yml will mount onto the container.

Issue #4555, opened 2025-06-17
"Paste X cell" modal confirmation window is not actually modal (and can be bypassed)


kha84:
## Description

![Image](https://github.com/user-attachments/assets/f394610c-d216-4639-8edc-d12a53bdb444)

## Expected behavior

1. Window should be truely modal - i.e. not allowing user to modify anything behind it
2. When the modal window is opened the focus should be set on one of the two buttons (probably realted #1886)
3. Ideally ESC key should be closing it

## To Reproduce

Have a value copied to a clipboard. Paste it into cell of a table in Mathesar. Without closing confirmation modal window continue to use keyboard to navigate through the table and modify other rows / columns. 

## Environment
 - OS: Windows 10
 - Browser: Firefox 
 - Browser Version: latest
 - Other info: Mathesar is the latest one

zackkrida:
Hi @kha84, I'm going to close this in favor of #1886, as it's essentially a more specific version of the same problem. Thanks for reporting it!

kha84:
Sure, np! Thank you!

Issue #1886, opened 2022-11-03
Implement a focus trap for modals


seancolsen:
When a modal is open, we should not be able to focus UI elements below it.

zackkrida:
@ghislaineguerin I've updated this description and have a UX question: do you have an opinion if focus should be "trapped" in Mathesar's dropdowns?

ghislaineguerin:
Do you mean a select input (native HTML) or a dropdown menu (custom component)?
If it's a select input, focus should not be trapped and pressing tab should close it and move to the next element in the form.
For a custom component, I think it makes sense to trap the tab key so that it is consistent with modals which are also custom.

zackkrida:
@ghislaineguerin I mean custom dropdown menus. I guess to elaborate a bit: do you think Mathesar's Filters, Group, and Sort dropdowns, for example, should behave more like:

- These aria specs, which don't trap focus:
  - [WAI-ARIA Menu Pattern⁠](https://www.w3.org/WAI/ARIA/apg/patterns/menu/) and the [WAI-ARIA Menu Button Pattern⁠](https://www.w3.org/WAI/ARIA/apg/patterns/menubutton/) (example component [here](https://ariakit.org/components/menu))
  - or the [WAI-ARIA Combobox Pattern⁠](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) (example component [here](https://ariakit.org/components/select))

- Or these ones, which do:
  - [WAI-ARIA Dialog Pattern⁠](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/) with example component [here](https://ariakit.org/components/dialog).

zackkrida:
This issue shows up in a copy/paste workflow using the keyboard. After copying and pasting rows or cell data via keyboard shortcut, the user _has to_ switch to the mouse to submit the paste via the confirmation modal.

seancolsen:
Fixed via #4659

PR #4659, opened 2025-07-29
Improve focus behavior for modals

seancolsen:
This PR _partially_ addresses the issues #1886 and #4555. My main goal here was to find the quickest path towards brining _some_ relief to the specific workflow described by a user in #4555.

## Before this PR

- As described in #4555, the focus behavior with the paste confirmation dialog was cumbersome because there was no way to confirm via the keyboard.

## After this PR

- When a modal opens, the first focusable element within the modal is automatically focused. The body, footer, and title bar are considered, in that order.
    - For modal confirmations (e.g. when pasting data into cells) the "Cancel" button is automatically focused. This seems good. The user can press <kbd>Tab</kbd>, <kbd>Enter</kbd> to confirm the paste operation.
    - For some modals (e.g. granted permissions on DB objects) the modal close button is focused because the content in the modal is not fully loaded when the modal opens.
- After the modal closes, the last-focused element is automatically re-focused. This is especially helpful with paste because in this case the cell is re-focused.
- All modal confirmations are closable via the Esc key.

## What this PR does _not_ do

- This PR does not implement a focus trap, even though it would be the most robust way to handle this problem. After some research, the best way to implement a focus trap seems be to be refactoring our modal system to use the `<dialog>` element (which wasn't widely supported when we originally implemented the modal system). That refactoring just seems a wee bit too intensive for now though. It wouldn't be terrible. But I'm aiming for a light touch right now.

- This PR does not change focus styling. Unfortunately our lack of focus styling makes it hard for the user to _discover_ that they can confirm the paste operation via <kbd>Tab</kbd>, <kbd>Enter</kbd>. But we can handle that separately. If we want to improve the focus styling, we'll need to do so in #4563 (or after it's merged).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@zackkrida I assigned you to review this PR since I think your head is in this problem pretty well, and I thought you might like taking a crack at fully reviewing some front end code. If not, feel free to assign to Pavish.

seancolsen:
Ugh these type errors in CI are annoying because they are the sort that my VS Code doesn't see due to TS version issues in docker vs host. I'll fix them tomorrow.

zackkrida:
@seancolsen I'll fully review this tomorrow, but on the subject of focus trapping: it still needs to be implemented manually in JS, even when using a `dialog` element (there are some cross browser inconsistencies with the "built in" implementation, as I recall). 

The aria specs I shared in a previous conversation only specify the need for focus trapping. The aria roles themselves don't implement focus trapping, as you noticed. 

All of that said—you've got a lot of the code already needed to trap focus! Typical implementations work by listening to tab and shift+tab keydown events; checking if the element in focus is the first or last in the modal's list of focusable elements; and if so, based on the tab direction, move focus to the first or last element. I see all the primitives necessary to implement that in the existing code.

seancolsen:
> there are some cross browser inconsistencies with the "built in" implementation, as I recall

Oh, dang. I tested `<dialog>` with Firefox and it worked for me, so I assumed that it was good to go. How unfortunate!

> Typical implementations work by listening to tab and shift+tab keydown events

Ok interesting. I figured it would be more elaborate than that. Cool that you know about this stuff! Hmm. This does make me wonder whether the focus trap is worth doing. One corner I cut was that `getFocusableElements` yields elements in DOM order, not `tabIndex` order. I should have documented that "gotcha" in the code. I might want to address that shortcoming if we were to leverage that function for more. I can see how a simple keyboard event listener + wrap-around logic wouldn't be too much extra work here. I'd be happy to do it! But I'll leave it to you to make the call based on time/priorities. Let me know.

zackkrida (approved):
@seancolsen the code here works exactly as described from a user perspective. The types and light code restructuring look good as well. Nice work! I think we should merge this as-is for immediate benefit. 

As for focus trapping, I _do_ think we should do it for this release. I'd be fine with a commit here, or a separate PR. I did look into a `dialog` based approach and I see that some of the previous issues with it I was aware of have been smoothed over in the last year or two, if you use the [corresponding JS apis](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/show) along with the dialog element.

That said, for our implementation, I think a simpler approach like the one I mentioned earlier, intercepting Tab and Shift+Tab via keydown listeners, makes more sense. Mathesar isn't fully a11y-compliant yet, so I see focus trapping here more as a keyboard usability enhancement for sighted "power users" than a full accessibility feature.

I _definitely_ want to get those button focus styles fixed, too!

seancolsen:
@zackkrida Cool. I'm bumping this back to you for re-review now.

I pushed some changes with the JS-based focus trap. I tested it a bit and it seems solid as far as I can tell.

Some notes:

- I changed the implementation details when auto-focusing the first input. I don't want to auto-focus the close button even though it's the first focusable element in a modal. Previously, I was ensuring that anything in the title bar got listed after the body and footer. Now I'm just setting a high tabIndex value on the close button. This has some tradeoffs.

    - Technically it's possible that consuming code could place focusable elements in the title. Such elements would be auto-focused. But I'd expect this to be highly unlikely.

    - The browser's focus navigation boundary seems to occur at the end of the DOM order, not the tabIndex order, at least in Firefox. This means that if you hit <kbd>Tab</kbd> from the last element within a modal you don't get to the close button. Instead, you get to the browser UI. In playing around with this, it actually seems okay though. Because it does mean that if you press <kbd>Shift</kbd>+<kbd>Tab</kbd> on the first element, you get to the close button.

seancolsen:
@zackkrida and I discussed this briefly on a call today and we decided that I'd push more changes to ensure that focus is also trapped within the document (i.e. that it doesn't move out into the browser UI).

seancolsen:
@zackkrida I pushed cbdf9be29dc3257ef524665ea17d4977fca12b28 to address the issue we discussed. This is ready for you to re-review now.

zackkrida (approved):
This is working great, @seancolsen. Nice work. I don't have any code suggestions either.

PR #4609, opened 2025-07-04
Forms frontend

pavish:
#### What's done structurally (visual work is still pending)
- The foundation for the form builder and form fillout views
- Adding, editing, deleting forms from the UI.
- Building a form by selecting fields, editing labels etc.,
- Form inspector
- Associated role selection, submission settings
- Form level validation
- Field appearance configuration
- Form sharing
- Public urls, anonymous form filling page
- Form submission UI

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #4665, opened 2025-07-30
Make certain components common, general improvements (Split from #4609)

pavish:
This PR is split from https://github.com/mathesar-foundation/mathesar/pull/4609

Note to reviewers:
- The changes in this PR are horizontal, however they are small enough to be part of the same PR.
- I'd recommend reviewing this PR commit by commit.
  - Each commit is self-contained.
  - The commit messages should provide context on the changes.

This PR targets the `develop` branch as it is self-contained and not specific to forms.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/dropdown-menu/DropdownMenu.svelte
+++ b/mathesar_ui/src/component-library/dropdown-menu/DropdownMenu.svelte
@@ -36,8 +36,12 @@
       {/if}
     </span>
   </slot>
-  <Menu slot="content" style="--Menu__min-width: 100%;{menuStyle}">
-    <slot />
+  <Menu
+    slot="content"
+    style="--Menu__min-width: 100%;{menuStyle}"
+    let:close={menuClose}
+  >
+    <slot close={menuClose} />
   </Menu>
 </Dropdown>
 
--- a/mathesar_ui/src/component-library/text-area/TextArea.svelte
+++ b/mathesar_ui/src/component-library/text-area/TextArea.svelte
@@ -28,7 +28,7 @@
    * Value of the input. Use bind tag for two-way binding.
    * Refer Svelte docs for more info on binding form input values.
    */
-  export let value: string | undefined | null = '';
+  export let value: string | undefined | null = undefined;
 
   export let element: $$Props['element'] = undefined;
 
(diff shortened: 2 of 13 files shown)


seancolsen (commented):
This all looks good, except that I'm a bit curious about your rationale for c48bbf4a11205b992eb3b927f798081e0d092c59. Can you say more about why you needed this change? It seems like the sort of low-level change that could end up causing regressions somewhere. I can't think of a specific bug in Mathesar that it would cause. But I haven't thought too hard. It just raises my eyebrows a bit, so I'd like to understand why it's there before I sign off on it.

pavish:
@seancolsen 

> Can you say more about why you needed this change? It seems like the sort of low-level change that could end up causing regressions somewhere. I can't think of a specific bug in Mathesar that it would cause.

Ideally, on such a low level component, there should never have been a default value. It hasn't caused any issues before and I guess that's one of the reasons we've not addressed it till now.

Consider this usage:

```svelte
<script>
  let valueInParent = undefined;
</script>

<TextInput bind:value={valueInParent} />
```

In this scenario, the `valueInParent` changes to `'' (empty string)` because it is `undefined` and there is a two-way binding. Here's [a small REPL](https://svelte.dev/playground/90ba69c7e15b4ae0985d382949f91c99?version=5.37.1) that demonstrates this behaviour.

A low level component like `TextInput` should not be making changes to the parent value until user actually types in the input.

This triggers listeners in cases where we're listening to changes in values.
For eg.,
- In forms, the default values are all `undefined`.
- The submit button is disabled and error messages are hidden when there are required fields that the user hasn't modified yet.
- However, once the user modifies a required field and then clears it, we show a validation error message.
- Since TextInput was setting a default value, the error messages were always being displayed, instead of only displaying when the user makes the change.

This would not be a concern when we reduce the use of two-way binding. Also, Svelte 5 makes these kind of scenarios more explicit, in fact, it does not allow setting a fallback value for a prop that is undefined and bound to the parent.

As for regressions, I went through the places we use these two components and the way we use them and also tested it. I focused more on the table rows and record page, and this change does not seem to cause any issues.

seancolsen:
Yeah, I understand the behavior here. And I agree that `undefined` would be a better initial value. I wouldn't be flagging this if it were a new component.

I'm just worried about the potential for regressions. I'm glad that you did some testing around this. I also gave it some more thought and poked around a bit. I'm reasonably confident we're okay. Cool.

PR #4674, opened 2025-08-05
Add modal record view

### LINKED ISSUE #4670: Modal record page
This is a feature to open the record page inside a modal from the table page.

UX is to be improvised during implementation.

seancolsen:
Fixes #4670

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
@zackkrida and @kgodey I've assigned you to review UX. Once UX is approved, we can do code review.

zackkrida (approved):
@seancolsen I'm really happy with this, it looks quite nice. I didn't anticipate all the complexities with nesting and quick open vs. opening the full page, but you've approached those problems in a manner that makes sense to me. 

I like the visual/organizational changes to the single record page/modal contents, it looks cleaner to me without reducing clarity.  

A few things irked me while working with these. I'm sure you're aware of these things as they're not specific to this PR, but they are pain points in the workflow of using these modals, which are ostensibly designed to streamline navigating into single records from the table view:

- There's no way to open the record modals via keyboard while navigating through the table, like by pressing "enter" while focusing the primary key cell
- Likewise, there's no way to exit the modal by pressing escape or clicking on the modal background overlay
- Opening a modal doesn't modify the page URL or create a history entry, so you can't navigate with the browser forward/back buttons in the same way you can if you were opening the single pages. This ends up feeling like a regression/net loss in functionality from the previous version, where I can, for example, click a single record link, use the back button on my mouse to return to the table view, and click a new single record.

None of those are blockers. If they're quick I'd love to see them added, but otherwise they can be future enhancements.

And finally, a future feature request: prev/next arrows next to the modal so you can navigate between the single records of all the rows in the current table view. That would be really slick.

zackkrida:
@seancolsen I'll let you know if/when @kgodey wants to review this, as well. Expect to hear from me tomorrow on that front.

seancolsen:
Great thoughts, @zackkrida. Each of the pain points you've described are things which are a bit tricky for various UX reasons. Easy implementation-wise. But tricky due to other UX consequences. I don't think this is the place to discuss them. Perhaps you could open a ticket for each of these?

And the prev/next buttons too! I kind of had to resist the urge to put that in this PR. I agree it would be super slick. It's not trivial to implement. But it's also not terrible. It would make for a good follow-up PR. I can work on that whenever you deem it to be a priority, including now if you like. I have some more thoughts to add, but again probably better to put those into a dedicated issue.

zackkrida:
Cool, sounds good @seancolsen. I'll create an issue for each item and reference this PR to connect them.

zackkrida:
@pavish this is ready for you to review the code; the UX is approved. @kgodey might give UX feedback in the future but she'll create new issue(s) as needed.

seancolsen:
Thanks for the critique @pavish. I think you've identified some good changes to consider.

And, as I did with Zack's critique, I'd like to bump the ball back to you to open specific issues for any that you deem important enough to address.

- You said:

    > Editing a related record in a nested modal does not reflect on the table in the parent modal.

    I'm not sure what this means. Can you walk through the reproduction steps in an issue?

- You said:

    > UX for editing a record in the related table is awkward, especially with text fields.

    Based on your screenshot, this seems unrelated to this PR. Here is a screenshot demonstrating the same issue without this PR:

    image

    While the issue is not present on the table page, it is present within the table widget, which appears on the record page (outside of the modal record view). A small amount of additional space at the bottom of the table widget would seem to do the trick. It's probably an easy fix. But it also doesn't seem particularly pressing to me. If it's important to you, maybe you could open an issue for it?

- You said:

    > I find it hard to mentally retain context when multiple nested fields are involved.

    Yeah I hear you on this. I'm not sure of the best way to solve the problem, design-wise. So I did the simplest possible thing I could do, just to get this feature shipped and move on to other priorities.

    Again, I think a new issue would be a good way to discuss various approaches and prioritize that work against other opportunities.

    Personally, I'd be inclined to see how our users end up using this new feature and whether the stacked modals actually end up being a pain point for them.
    
- You said:

    > Visually, I find the information density in the modal to be too high

    Interesting. I'd be curious what others think. Personally, I'd rather see _more_ information density in general, including in this modal record view. I'm not sure how to make the best call on this topic objectively.

- You said:

    > Copy pasting records within the related records end up in the fk values being null since that column is not displayed to the user.
    
    I don't understand how to reproduce this. Can you open an issue and walk me through the steps?

- You said:

    > There are some TODO items to reduce duplication, please create issues for them so that we can keep track of it.

    Yeah, I would have avoided creating this mess in the first place had it been easier to DRY this code up. I tried briefly. But (interestingly) it proved to be mildly tricky in this case. So I made the strategic call to take on a little tech debt here.

    Still, I disagree that this warrants a GitHub issue. It has no affect on users. And it's sufficiently low in priority that I don't imagine we'd ever intentionally choose to slate this work for a release (given the opportunity-cost when weighed against other work). So I don't see the utility of creating an issue to track it. To me, it's tech debt with a very narrow scope — two small components. So the best way to track it is in the code. The code comments serve to cue devs to look at the other component when making edits. Those devs might decide to opportunistically pay off that tech debt along with other refactoring in the future. Or not. To me it's even okay to keep this sort of minor code duplication indefinitely.

PR #4681, opened 2025-08-06
Forms management UI in Schema page (Split from #4609)

pavish:
This PR is split from #4609

- It displays the Forms section in the Schema Page, and makes form create, edit, and delete functional
- Adds RPC methods for forms
- Adds frontend models for forms
- Adds breadcrumb items for forms

Note to reviewers:
- The actual Form routes are not present in this PR.
- So, when a form is created, you'll notice a "Page not found" error since the route it not present.
- This is intentional to keep this PR small.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (commented):
Looking good!

Some thoughts:

- When creating a new form, I think we ought to follow the pattern established in #4089 where we _show_ the thing you created instead of _opening_ it. We can use the `highlightNewItems` action for this.

- In `AddEditDataFormModal` I suggest changing "Base table for form" to "Base Table".

- I think the base table should actually be the first field in the `AddEditDataFormModal` form. Mentally, when I think of creating a form, I think of the table before I think of what I want to name my new form. We could even auto-generate a name for the form based on the table that the user has chosen (though I don't want to add scope). It would be easy to rearrange the fields though, and I think it would make the modal slightly more intuitive.

- In `AddEditDataFormModal`, there is some logic I don't understand. You're disabling the `SelectTableWithinCurrentSchema` component when `sourceTableStructure` is loading. Why? The `sourceTableStructure` variable doesn't appear to be used anywhere. Is this ultimately needed for some future work? If so, why do we need to disable the UI when the table structure is loading. I noticed this in the UI first — not the code. It felt strange when filling the form out. Can we remove this logic?

- I do want to note that I had [argued against](https://github.com/mathesar-foundation/mathesar/issues/4509#issuecomment-2956352922) `RichTextJson`. We had some back and forth, and now I see that you've gone ahead with it anyway. I still have a slight preference for keeping this additional complexity out of the codebase until the time if/when it is needed. But I can see that you've scaled it back from the [slightly more complex interface](https://github.com/mathesar-foundation/mathesar/issues/4509#issuecomment-2966883219) you had originally put forth. So I'm okay with this as-is.

seancolsen:
Oh, and yesterday I forgot to mention — I made a couple very small commits adjusting the names of variables. These changes might need to be carried forward to other work based on this PR.

pavish:
@seancolsen 

> When creating a new form, I think we ought to follow the pattern established in https://github.com/mathesar-foundation/mathesar/pull/4089 where we show the thing you created instead of opening it. We can use the highlightNewItems action for this.

I've added https://github.com/mathesar-foundation/mathesar/pull/4681/commits/c89de5480298afd7e0c12237f597bd60be0a0afa to not open the form when its created.

<details>
<summary>My views on this pattern, for later discussion</summary>
I do not think striving for a uniform pattern for all objects make sense.

From my perspective, a database and a schema are akin to a "directory", and I would expect to remain on the same page and the newly created database/schema to be highlighted. However, a table, exploration, and a form are akin to "files", and whenever I create one, I would expect to edit them or view them.

Currently, creating an exploration follows a different UX where we open the data explorer, and creating a table follows different UX based on the method of creation. There is no established pattern for these objects yet, and we might want to think it through.
</details>

> In AddEditDataFormModal, there is some logic I don't understand. You're disabling the SelectTableWithinCurrentSchema component when sourceTableStructure is loading. Why? 

I've removed this. Previously, I had a pattern similar to data explorer, where this modal was used only for editing forms, and I used sourceTableStructure to load and display the name of the base table during editing (to reduce the use of top level tables store). However, I later decided to use this modal for both adding & editing forms, and this was left from the refactor.

I agree with your other points and I've added commits to resolve them.

seancolsen (approved):
Looks great!

PR #4683, opened 2025-08-06
Forms replace is patch

### LINKED ISSUE #4679: Replacing a form definition should not recreate the entire form
## Description
- Currently, when a form definition is replaced, we delete the form model and create a new one.
- This includes properties like form `token`, `publish_public`, `base_table_oid` etc., which should not change when a form definition is modified.
- The purpose of the `replace` method is to only modify the form definition and not to create an entirely new form.
- Only the following properties should be affected by the `form.replace` method. We might want to even rename the method to improve clarity.
   - `associated_role_id`
   - `header_title`
   - `header_subtitle`
   - `submit_message`
   - `submit_redirect_url`
   - `submit_button_label`
   - `fields`.
- The existing `Form` model should not be deleted on the database. Instead it should be updated.
- The `FormFields` models can be dropped and recreated.

Anish9901:
Fixes #4679

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


docs/docs/api/methods.md
```diff
@@ -146,14 +146,14 @@
       - add
       - delete
       - regenerate_token
-      - replace
+      - patch
```
mathemancer, line 149: Yeah, I think this makes more sense given the current functionality.

mathemancer (approved):
LGTM. As ever, please wait for validation from @pavish before merging.

pavish (approved):
@Anish9901 @mathemancer 

I noticed a security issue: The user might update the form without changing `associated_role` and we're not validating their access to the role in that scenario. This allows the user to modify a form that already has a role which they don't have access to.
  - I added a commit https://github.com/mathesar-foundation/mathesar/pull/4683/commits/9a9ae1a3ec946de1d3bbf5dbfe8d74c36a7a32b6 to fix this.
  - This commit also makes it optional so that the frontend can pass `null` for it (as it can do for create).

Please review the commit and merge the PR in, if you find it okay.

Also, I don't think `patch` is a good name for this method.
  - `patch` as a concept represents passing only properties that have changed and then reconciling them on the backend. If a property is not passed, `patch` does not modify that property.
  - This method however requires all properties that can be updated, and if they are not passed, they are considered `null`. This is the behaviour that the frontend also expects for this method.
  - I agree `replace` is not a good term, but I don't think `patch` should be used either.
  - I do not have a recommendation for the name, at the moment.
  - However, I don't think it needs to block this PR from being merged, I'll leave you both to take the call on whether it needs a different name.

mathesar/utils/forms.py
```diff
@@ -205,7 +209,25 @@ def submit_form(form_token, values, user):
...
+def patch_form(update_form_def, user):
+    form_model = Form.objects.get(id=update_form_def["id"])
+    associated_role = validate_and_get_associated_role(
+        user,
+        database_id=form_model.database.id,
+        associated_role_id=update_form_def.get("associated_role_id")
+    )
+    if fields_dict := update_form_def.get("fields"):
+        form_model.fields.all().delete()
+        create_form_fields(form_model, fields_dict)
```
pavish: Just noticed this.

If `fields` is `empty`, the existing fields should be deleted. `fields` will always be passed, but it can be empty.

pavish (changes requested):

mathesar/utils/forms.py
```diff
@@ -205,7 +209,25 @@ def submit_form(form_token, values, user):
...
-    Form.objects.get(id=form_def_with_id["id"]).delete()
-    form_model = create_form(form_def_with_id, user)
-    return form_model
+def patch_form(update_form_def, user):
+    form_model = Form.objects.get(id=update_form_def["id"])
+    associated_role = validate_and_get_associated_role(
+        user,
+        database_id=form_model.database.id,
+        associated_role_id=update_form_def.get("associated_role_id")
+    )
```
pavish, line 218: Okay, I spent some more time thinking about the permissions validation logic. I've been testing it in various contexts iteratively, which lead to these multiple separate review comments.

- Any user should be able to update `name`, `description`, `header_title`, `header_subtitle`.
  - This is because we allow them to modify these from outside the form builder UI.
  - For these changes, we **should not** run `validate_and_get_associated_role`.
- However, the properties `fields`, `submit_message`, `submit_redirect_url`, `submit_button_label`, and `version` can be modified only by a user with access to `associated_role` or the super user.
  - For these changes, we **should run** `validate_and_get_associated_role`.
- `associated_role` can be changed by anyone.
  - When it is modified, we **should run** `validate_and_get_associated_role`.

This permission logic was not made clear in the issue. These changes came up due to recent UX reviews on the frontend PR. Apologies for not updating earlier.

mathesar/utils/forms.py
Anish9901: Addressed in 74477e2

PR #4685, opened 2025-08-08
Display column descriptions in column header cells

### LINKED ISSUE #4612: Display column descriptions in the table header
## Problem

Column descriptions are only present in the inspector, and aren't accessible via the table headers on the table page.

For example, this Customers table has a description for the phone column, but you don't know that until opening the inspector:

Image

Image

## Proposed solution

Add an ℹ️ tooltip to column headers for columns with descriptions, or some other mechanism for accessing the description via the header cell.

seancolsen:
Fixes #4612

## Demo

[Screencast_20250808_100140.webm](https://github.com/user-attachments/assets/0a670d80-2070-4c7a-9e9f-a98708de1e16)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (approved):
@seancolsen this is great, but I find the icon to be _very_ heavy visually. Do we have an ℹ️ style icon avaliable, something like this:

Image

Especially as an outline, and aligned to the end of the cell, I think it's much less obtrusive than the dark "comment"/"text bubble" icon. 

Are either of these quick changes to make? If they are, please go ahead, if not it's fine to merge this as-is.

seancolsen:
@zackkrida this is the icon set that we're using: https://fontawesome.com/v5/search?ic=free&o=r

I don't immediately see something like that there. We could certainly make a custom icon for this. That's like ~20 minutes of work.

If we do change the icon, we'd need to decide the scope of the change. What I mean is this: we have a semantic system for our icons. This PR simply applies our own pre-defined Mathesar-specific `iconDescription` icon in the column header. This PR does not map `iconDescription` to a visual icon. So if we want to change the "description" icon then we'll need to decide whether that change should apply only to columns, or also to other places that it's used too. Where exactly is this icon used currently? I think the answer might actually be spread across multiple in-flight PRs. I'd need to do a little digging to fully answer this. But I know it's used in the list of explorations and list of forms on the schema page. Maybe other places too.

I don't have an opinion on the specific icon, visually-speaking.

Let me know how you'd like to proceed here. I can make the custom icon if you want. And let me know whether the new one should apply only within the column header, or everywhere we have something with a hoverable description.

seancolsen:
As a brief tangent there's something else that might be kind of interesting here...

You said:

> Do we have an ℹ️ style icon avaliable

And in the ticket you said:

> Add an ℹ️ tooltip

And I see you mockup above. Now I'm wondering if you made that mockup by inserting that emoji.

For me your mockup doesn't render the same as those emojis in your comments.

Here's what your comment looks like to me:

image

I know emojis can render differently on different platforms 😯. I'm not sure that's what's happening here. But I mention this for the sake of clarity 🎯 (and contemplation 🤔), so that you can see what I'm seeing 👀

zackkrida:
@seancolsen thanks for the details!

I do think an info-circle (sadly it's a paid pro icon in [font awesome 5](https://fontawesome.com/v5/icons/info-circle?f=classic&s=regular)) both looks better and has better semantic meaning for these kinds of descriptions throughout the app. Especially if we ever add any "comments" functionality in the future. Even concerning forms presently, the chat-like icon _could__ imply something related to form submissions rather than a description. 

Please go ahead and update this icon everywhere with a custom one. Thanks!

zackkrida:
Oh and regarding your latest comment @seancolsen in my mockup I pasted an SVG from font awesome seven into Mathesar with devtools 😊, and used an emoji in my messages.

seancolsen:
> update this icon everywhere with a custom one

Will do!

seancolsen:
@zackkrida this is ready for you to re-review now. Here is the result:

image

image

I'll note that in your mockup the icon is aligned to the right. I've retained the alignment to the left, not because I prefer that, but because I'm aiming for minimal design changes in this PR. Even before this PR we already have some icons like this that display in the column header cell, for example when sorting by a column. Since those icons display with the text on the left, I had this one display within that same group. If we want _this_ icon to display on right, then I think we probably ought to have _all_ modifier icons display on the right. I can make that change too if you want. It's very simple. But I'd rather not take responsibility for that design change (in case others take issue with the change, from a design perspective). Let me know if you want me to change that, or if you want me to make any other adjustments.

zackkrida:
@seancolsen ship it, looks great! The right side alignment for modifier icons isn't as important to me.

PR #4686, opened 2025-08-08
Fix UI hang issues

### LINKED ISSUE #4594: UI hangs when applying grouping then filtering
## Steps to reproduce

1. Load [birds.csv](https://github.com/user-attachments/files/20979702/birds.csv)

1. Group by `Flighted`

1. Begin adding a filter condition on `Common name` as follows:

    Image

1. Enter the character "e" as a filter value.

1. Expect the records to reload with the filter and grouping applied.

1. Instead, observe that the records do not appear to finish loading.

    Image

    At this point, some of the UI is usable (e.g. navigation), but some of the UI is broken (e.g. clearing the applied filter).

1. Refreshing the page will to return to a usable state. Interestingly, refreshing the page preserves the grouping and filtering (due to our mechanism for persisting filtering/grouping/sorting/pagination within the URL), and after the refresh the page is usable with the filtering and grouping applied.

## Possibly related

- #4648

### LINKED ISSUE #4648: UI hangs when changing page size of a grouped table
## Steps to reproduce

1. Use our movie rentals sample data.
1. Open the `movies_genres` table.
1. Group by the `movie` column.
1. Change the page size from 500 to 100.
1. Expect the records to reload successfully.
1. Observe that the API request to reload records succeeds but the new results are not displayed in the sheet. Instead the sheet appears as loading, and many UI actions have no effect.
1. A full page refresh will make Mathesar usable again.

## Demo

https://github.com/user-attachments/assets/f6fe4474-ee3a-4a76-996d-2e67060e438e

## Possibly related

- #4594

seancolsen:
Fixes #4648
Fixes #4594

## Analysis

In `combineRecordRowsWithGroupHeaders` we were throwing an error within a logic branch that _seemed_ impossible — but it wasn't!

Within that function, the `recordRows` and `grouping` parameters are tightly coupled. These values come from the same `records.list` API response. The `grouping` value is a data structure which references indices within `recordRows`. So it would seem that each of these index values should always point to a valid record row.

But the problem is that sometimes this function gets called with `recordRows` and `grouping` values _from different API responses_!

Here's how it happens:

1. The table page has loaded and all is well.

1. The user changes _something_ that causes the records to reload, e.g. filter, group, pagination, etc.

1. In `RecordsData.fetch` we get the records and grouping from the API.

1. Then (among other things) we call:

    1. `this.fetchedRecordRows.set`
    1. `this.grouping.set`

    These are _two separate calls_ to update _two separate stores_.

1. The `Display` class maintains a `displayRowDescriptors` store derived from (among other things) `recordsData.fetchedRecordRows` and `recordsData.grouping`.

    The derivation function calls `combineRecordRowsWithGroupHeaders`.

1. So we have a diamond problem:

    ```mermaid
    graph TD
      A[API response] --> B[records]
      A --> C[grouping]
      B --> D[displayRowDescriptors]
      C --> D
    ```

    This would be okay if Svelte were smart enough to batch the computation of derived stores. But it's not. 

1. So inside `RecordsData.fetch` when we call `this.fetchedRecordRows.set`, then `displayRowDescriptors` is immediately recomputed. And in doing so it uses the _latest_ value for `recordRows` but the _old_ value for `grouping`. Boom there's our problem. The interdependence between these two values is not enforced in our stores structure.

Under many circumstances, the `displayRowDescriptors` actually gets recomputed _four_ times!

To me the proper fix would seem to be a refactor that makes the granular stores more coarse by lumping them into an immutable data structure which enforces that interdependence, preventing the values from being decoupled with one another. We could potentially lump other similar stores into the same data structure too, but this will require some thought. In general, I think we ought to make an effort to avoid these diamond store structures, and this is an example of why.

For the time being though, I changed the error to a return. It's a crude way to fix the problem for now.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/display.ts
+++ b/mathesar_ui/src/stores/table-data/display.ts
@@ -108,9 +108,12 @@ function combineRecordRowsWithGroupHeaders({
       group.resultIndices.forEach((resultIndex) => {
         const recordRow = recordRows[resultIndex];
         if (!recordRow) {
-          throw new Error(
-            'Record from group result index is not found. This should never happen',
-          );
+          // ⚠️ This is a weird case. It _does_ actually occur. Returning is a
+          // bit of a hack. Fixing this properly will require some refactoring
+          // of stores. See comments in:
+          // https://github.com/mathesar-foundation/mathesar/issues/4594 and/or
+          // the related PR
+          return;
         }
         combinedRows.push({
           row: recordRow,


pavish (approved):
@seancolsen Nice work figuring out the cause! 

The bandaid fix works well, I think it's good until we refactor these stores.

PR #4727, opened 2025-08-29
Fix QA issues - 0.5.0

pavish:
Fixes the following issues from [this mail thread](https://groups.google.com/a/mathesar.org/g/mathesar-developers/c/fArihRw_v58/m/Eh3rZqPLDQAJ).
- Critical issue: The public share URLs for some forms will return a 500 error
- Disabled share links continue to display and work (data can be submitted) in the current logged-in session.

**Technical details**
- Backend:
  - https://github.com/mathesar-foundation/mathesar/commit/bcc624c7ec1fb4090878ddaf0c822e5db5354c95 accepts urls with and without backslash for shared forms.
    - This is because these urls can be copied, saved, and shared by the user, and we want it to be as flexible as possible.
  - https://github.com/mathesar-foundation/mathesar/commit/328f033c634f504c090b341cf97ebea3bb441cff only accepts submit requests when form is publicly shared.
    - The retrieval methods are common for both authenticated and anonymous RPC methods, so they aren't changed.
    - The submit method should only be accessible for a publicly shared form.
    - We removed the ability to submit forms for authenticated users on the UI following discussions from internal QA, and scheduled the feature for a later release.
- Frontend:
    - https://github.com/mathesar-foundation/mathesar/commit/25cf709e0c90cae70e344001d2ff2dc6f978de5e adds a trailing backslash to the shared form url for users to copy.
      - Though we allow urls with & without trailing slashes, this commit makes it consistent to the rest of the codebase where we always append a trailing backslash.
   - https://github.com/mathesar-foundation/mathesar/commit/2f0115170468f05540c66001518dc757d0dad442 commit hides the form and shows an error when an authenticated user accesses the public url of a form which is no longer shared.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/urls.py
+++ b/mathesar/urls.py
@@ -31,7 +31,7 @@ urlpatterns = [
     path('administration/settings/', views.admin_home, name='admin_settings'),
     path('i18n/', include('django.conf.urls.i18n')),
     path('info/analytics_sample_report/', views.analytics_sample_report, name='analytics_sample_report'),
-    path('shares/forms/<str:form_token>/', views.anonymous_route_home, name='shared_form'),
+    re_path(r'^shares/forms/(?P<form_token>[0-9a-zA-Z\-]+)/?', views.anonymous_route_home, name='shared_form'),
     re_path(
         r'^db/(?P<database_id>\d+)/schemas/(?P<schema_id>\d+)/',
         views.schema_route,
--- a/mathesar/utils/forms.py
+++ b/mathesar/utils/forms.py
@@ -204,9 +204,9 @@ def iterate_form_fields(fields, parent_field=None, depth=0, fields_to_pick=[]):
         )
 
 
-def submit_form(form_token, values, user):
+def submit_form(form_token, values):
     form_model = Form.objects.get(token=form_token)
-    assert has_permission_for_form(user, form_model), 'Insufficient permission to submit the form'
+    assert form_model.publish_public, 'This form does not accept submissions'
     fields_to_pick = [i for i in values.keys() if isinstance(values[i], dict) and values[i].get('type') == 'pick']
     field_info_list = [
         {
--- a/mathesar/rpc/forms.py
+++ b/mathesar/rpc/forms.py
@@ -414,5 +414,4 @@ def submit(*, form_token: str, values: dict, **kwargs) -> None:
         form_token: The unique token of the form.
         values: A dict describing the values to insert.
     """
-    user = kwargs.get(REQUEST_KEY).user
-    return submit_form(form_token, values, user)
+    return submit_form(form_token, values)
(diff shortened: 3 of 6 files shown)


mathesar/urls.py
```diff
@@ -31,7 +31,7 @@
     path('administration/settings/', views.admin_home, name='admin_settings'),
     path('i18n/', include('django.conf.urls.i18n')),
     path('info/analytics_sample_report/', views.analytics_sample_report, name='analytics_sample_report'),
-    path('shares/forms/<str:form_token>/', views.anonymous_route_home, name='shared_form'),
+    re_path(r'^shares/forms/(?P<form_token>[0-9a-zA-Z\-]+)/?', views.anonymous_route_home, name='shared_form'),
```
pavish, line 34: I've still allowed this url to accept any alpha-numeric string (plus hyphen), instead of restricting it to uuid.

This is fine since a uuid validation happens when retrieving the form, and this keeps the url flexible.

seancolsen (approved):
Front end changes look good.

mathesar/utils/forms.py
```diff
@@ -204,9 +204,9 @@ def iterate_form_fields(fields, parent_field=None, depth=0, fields_to_pick=[]):
         )
 
 
-def submit_form(form_token, values, user):
+def submit_form(form_token, values):
     form_model = Form.objects.get(token=form_token)
-    assert has_permission_for_form(user, form_model), 'Insufficient permission to submit the form'
+    assert form_model.publish_public, 'This form does not accept submissions'
```
Anish9901, line 209: So, this means that anyone in the world with the link to the public form on a public instance of mathesar can make submissions, and there is no way to restrict submissions for internal use? Is this really the behavior we want @pavish?

pavish: > So, this means that anyone in the world with the link to the public form on a public instance of mathesar can make submissions

Yes, that's the behaviour of public links as it was previously as well. I've only removed the internal usage.

> there is no way to restrict submissions for internal use

Users can have a form without public sharing enabled, but they won't be able to submit via that form, they can only edit it.

This is a feature we will probably have in the near future. We removed it for the first iteration due to disagreements on how permissions should work during the internal QA phase, and needed more time to think it through.

pavish:
@zackkrida I'd still like for you to confirm that the critical issue is resolved, and merge the PR in.

> However, the error in your mail mentions that the form does not exist, which is odd. I haven't been able to reproduce that exact error. Zack Krida Could you test this again and confirm that it works when the trailing slash is present?

As mentioned in [the mail thread](https://groups.google.com/a/mathesar.org/g/mathesar-developers/c/fArihRw_v58/m/Eh3rZqPLDQAJ), I **wasn't seeing** the exact error messages,
i.e., `modernrpc.exceptions.RPCException: [-28009] DoesNotExist: Form matching query does not exist`

My hunch is that these messages were thrown at some point when trying to access a non-existing form, and the actual 500 error related to the critical issue did not produce any console logs.

zackkrida:
@pavish yes, you're correct about the error messages, they weren't related to this issue. Apologies for the misdirection there. 

This is working well; LGTM!

Issue #4595, opened 2025-06-30
File attachments


zackkrida:
>  _This is a meta issue tracking the implementation of this project._

## Outcome

- Users can upload file(s) to rows in their Mathesar tables 

More detail:
- [Requirements](https://github.com/mathesar-foundation/mathesar-wiki/blob/update-file-requirements/docs/product/requirements/2025/file-attachments.md)
- [Project proposal](https://github.com/mathesar-foundation/mathesar-wiki/blob/file-attachments-proposal/docs/product/proposals/file-attachments.md)

## Tasks
*Issues to be created soon.*

## Related

- Conceptual [design prototype](https://wiki.mathesar.org/archive/product/design/exploration/#fall-2022) 
- Feature request: [#4737](https://github.com/mathesar-foundation/mathesar/issues/4737)

zackkrida:
This feature will release as part of Mathesar 0.7.0 🥳

PR #4728, opened 2025-08-29
File attachment loading

mathemancer:
Related to #4595 

This PR covers the backend for the following flow:

- list records, including file attachment info for columns so configured.
- receive mimetypes and links for each file attachment found:
  - thumbnails
  - download as attachment
  - directly load (for images, or other files we eventually handle this way)

**Technical details**

### Covered functions

Each of the following functions returns file attachment info when needed:

- `records.get`
- `records.list`
- `records.search`

### New Docker container for object storage

This PR adds a new container in the docker compose setup. This container acts as a local object store with the same API as S3. It also provides a web UI where one can upload and manage files (until we have the ability to do so through Mathesar). To use it, run

```
docker compose -f docker-compose.dev.yml up obj-store dev-service
```

You have to specify the `obj-store` service, since the `dev-service` doesn't depend on it. You can use the web UI to upload files or images to use.

### The storage format

For now, this supports JSON(B) columns with two keys:

```json
{
  "uri": "s3://path/to/an_image.jpeg",
  "mash": "md5hexdigestofhash"
}
```

To generate such a JSON, there's a function in the mathesar python library. So, to get data for testing:

1. Copy the file `file_storage.yml.example` to `file_storage.yml` in the root of the repo.
2. Start Mathesar using `docker compose -f docker-compose.dev.yml up obj-store dev-service`
3. Go to `localhost:9001` in your browser, login with user `mathesar` password `mathesar`
4. Create a bucket, and upload some files for experimentation. Note their paths.
5. Login to Mathesar, create a database, and create a table with a JSONB column (actually you'll need to use PSQL or another PostgreSQL client to create the JSONB column).
6. Use `docker exec -it mathesar_service_dev python manage.py shell` to get a python shell
7. In python, execute:
   ```python
   from mathesar.utils import download_links as dl
   dl.create_json_for_uri('s3://bucket-name/path/to/image.jpeg', 'local_minio')
   ```
   Note that you should replace the S3 path with the path of an object uploaed in step (3). Copy the generated JSON into the JSONB column of your table via Mathesar, or any PostgreSQL client.
8. Repeat creating and copying the JSON for each desired object.
9. Note the attnum of the JSON column, the database ID, and the table OID.
10. Back in your same python shell, execute:
   ```python
   from mathesar.utils import columns as colutils
   colutils.set_columns_metadata(["attnum": <attnum>, "file_backend": "local_minio"], <table_oid>, <database_id>)
   ```
   with appropriate `attnum`, `table_oid`, and `database_id` values.
11. List records from that table in Mathesar
   - You should see the JSON values in the column in the UI.
12. The response to the `records.list` call should have a new `download_links` blob, with links for thumbnails, direct loading and downloading of the files.

The keys are actually the `mash` values for the returned JSON for the column.

- This is to give a unique key that can be consistent for any pointer type
- For the foreseeable future, I'd like the backend to produce a `mash` value for _any_ file attachment and return it in the JSON returned to the front end (even if the actual data storage is not JSON(B) on the DB).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@Anish9901 You can look at this and try it out if you'd like, but no deep review is necessary till I add tests.

seancolsen:
@mathemancer and I had a quick pairing session where he got me up to speed on how this works. Although I'm out next week, at this point I have everything I need to begin work on the front end side of this feature. Nice work Brent! It looks great so far!

mathemancer:
@seancolsen @Anish9901 I applied a change that will affect any test data you've set up. You'll need to regenerate the `mash` wherever you're using it, since the hash algo changed.

mathesar/rpc/records.py
```diff
@@ -154,6 +155,7 @@ class RecordList(TypedDict):
     grouping: GroupingResponse
     linked_record_summaries: dict[str, dict[str, str]]
     record_summaries: dict[str, str]
+    download_links: Optional[dict]
```
Anish9901, line 161: Please document this in the docstring above.

mathesar/rpc/records.py
```diff
@@ -438,6 +457,7 @@ def search(
         limit: int = 10,
         offset: int = 0,
         return_record_summaries: bool = False,
+        download_link_columns: list[int] = [],
```
Anish9901: A bunch of args are missing from the docstring for this endpoint, can you add add those as well?

mathesar/rpc/records.py
```diff
@@ -236,6 +239,7 @@ def list_(
         filter: Filter = None,
         grouping: Grouping = None,
         return_record_summaries: bool = False,
+        return_download_links: list[int] = [],
```
Anish9901: This arg is named `return_download_links` here, but, it's named `download_link_columns` for other endpoints any reason for this discrepancy?

docker-compose.dev.yml
```diff
@@ -61,6 +74,10 @@ services:
       - POSTGRES_PASSWORD=mathesar
       - POSTGRES_HOST=${POSTGRES_HOST-mathesar_dev_db} # replace mathesar_dev_db with /var/dev_postgres_socket for connecting to mathesar_dev_db via unix socket instead
       - POSTGRES_PORT=5432
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+      - AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}
+      - AWS_ENDPOINT_URL_S3=${AWS_ENDPOINT_URL_S3}
```
Anish9901: These aren't read anywhere in the service are they?  It's not clear to me if we wanna support config via both yml and env vars. If we _do_ wanna support both ways, can an you comment these out or track somewhere that we need to read these?

mathesar/utils/download_links.py
```diff
@@ -0,0 +1,166 @@
...
+        img_byte_arr = io.BytesIO()
+        with of as f:
+            img = Image.open(f)
+            img.thumbnail(size)
+            img.save(img_byte_arr, format="JPEG")
+        thumbnail = img_byte_arr.getvalue()
+        link.thumbnail[key] = base64.b64encode(thumbnail).decode("utf-8")
+        link.save()
+    else:
+        thumbnail = base64.b64decode(bytes(thumb_64, "utf-8"))
```
Anish9901: Any reason why this is only restricted to jpegs? It should be more general IMO. The problem is that whenever a .png or .svg is uploaded, the link for a thumbnail is generated but accessing the link results in a 500. If we only want JPEGs(for now) I'd assume we'd want to avoid having the links to the thumbnails for other filetypes in any of the `records.*` endpoints.

.env.example
```diff
@@ -1,2 +1,5 @@
 ALLOWED_HOSTS=".localhost, 127.0.0.1, [::1]"
 OIDC_CONFIG_DICT="{\"version\": 1,\"oidc_providers\": {  \"provider1\": {    \"provider_name\": \"okta\",    \"client_id\": \"client-id\",    \"secret\": null,    \"server_url\": \"https://trial-2872264-admin.okta.com\"  }}}"
+AWS_ACCESS_KEY_ID="mathesar"
+AWS_SECRET_ACCESS_KEY="mathesar"
+AWS_ENDPOINT_URL_S3="http://mathesar-dev-store:9000"
```
Anish9901: These seem unintentional.

mathesar/utils/download_links.py
```diff
@@ -0,0 +1,166 @@
...
+
+def get_link_contents(session_key, download_link_mash):
+    link = get_object_or_404(
+        DownloadLink,
+        mash=download_link_mash,
+        sessions=session_key,
+    )
+    content_type = _mimetype(link.uri)
+    of = fsspec.open(link.uri, "rb", **link.fsspec_kwargs)
+    filename = posixpath.split(of.path)[-1]
```
Anish9901, line 30: Can we use `os.path` instead of `posixpath` here? From posixpath's docsting:
> Instead of importing this module directly, import os and refer to
this module as os.path.  The "os.path" name is an alias for this
module on Posix systems; on other systems (e.g. Windows),
os.path provides the same operations in a manner specific to that
platform, and is an alias to another module (e.g. ntpath).

mathesar/utils/download_links.py
```diff
@@ -0,0 +1,166 @@
...
+        settings.SECRET_KEY.encode('utf-8')
+        + backend_key.encode('utf-8')
+        + uri.encode('utf-8')
+    ).hexdigest()
+
+
+def create_json_for_uri(uri, backend_key):
+    return json.dumps(
+        {URI: uri, MASH: create_mash_for_uri(uri, backend_key)},
+        sort_keys=True
```
Anish9901, line 87: I'm guessing this sorting is used here to make these jsons look consistent in the data explorer.

mathesar/utils/download_links.py
```diff
@@ -0,0 +1,166 @@
...
+
+    if (thumb_64 := link.thumbnail.get(key)) is None:
+        of = fsspec.open(link.uri, "rb", **link.fsspec_kwargs)
+        img_byte_arr = io.BytesIO()
+        with of as f:
+            img = Image.open(f)
+            img.thumbnail(size)
+            img.save(img_byte_arr, format="JPEG")
+        thumbnail = img_byte_arr.getvalue()
+        link.thumbnail[key] = base64.b64encode(thumbnail).decode("utf-8")
```
Anish9901, line 53: I like the dynamic programming approach you've taken here, but, it has a downside of displaying an outdated thumbnail when an image is replaced with the same filename. We also don't have a way to regenerate these, unless a different size is requested. Hmm... I think this is fine for now though.

mathesar/utils/download_links.py
mathemancer: We specifically want posix path, regardless of the underlying system. URIs always use posixpath, regardless of the underlying system.

mathesar/utils/download_links.py
mathemancer: It's also to reduce duplication when saving and calculating download links.

mathesar/utils/download_links.py
mathemancer: Yeah, I was thinking we can add a "rebuild" (or similar) flag if it becomes a problem.

mathesar/rpc/records.py
mathemancer: Yep, I changed the name partway through. I'll fix that.

docker-compose.dev.yml
mathemancer: removed in https://github.com/mathesar-foundation/mathesar/pull/4728/commits/eaa21c7686ef0fbdab37cc06a5d1c0f3f7f92b4d

.env.example
mathemancer: You're correct. Removed!

mathesar/utils/download_links.py
mathemancer: This is a little more subtle of a problem. The intent here is to create thumbnails from any image filetype, but to save a unified image type for the actual thumbnail. The actual problem was that the input files couldn't be auto-converted to JPEG. As per @seancolsen 's recommendation, I've changed the thumbnail files to be AVIF type, and also set things up to convert SVG to PNG before saving a thumbnail (the original file is unmodified, though).

https://github.com/mathesar-foundation/mathesar/pull/4728/commits/bd7ac24c89b9d701dd71a12897d27126ce2d12e6

https://github.com/mathesar-foundation/mathesar/pull/4728/commits/14c30e96740bb70ca470aa8799c9cc9b1708c982

mathesar/rpc/records.py
mathemancer: Done in https://github.com/mathesar-foundation/mathesar/pull/4728/commits/8c6ba056c2d733b799cc835df193e741a104c308

seancolsen:
@mathemancer I'm just returning to this now after a week away. I'm trying to pick up from where we left off on our call on 2025-08-29. During that call I said I thought I had everything I needed to begin work on the front end. But now that I've spent more time with this, I realize that I need more stuff.

From the frontend's perspective, here's what I've identified that I need thus far:

- I need a way to determine which column is a File column. Specifically, I need `columns.list_with_metadata` to give me a way tell the difference between a File column and a jsonb column. I understand we'll use metadata for this purpose, but I don't understand what your vision is for exactly _how_ we'll employ that metadata.

- I need the `return_download_links` API parameter to go away. I recall discussing this with you on 2025-08-29. Since the front end requests `columns.list_with_metadata` in parallel with `records.list`, it can't rely on the columns metadata response to formulate the records listing request.

I'm basically blocked on further work until I get these.

mathemancer:
> * I need a way to determine which column is a File column. Specifically, I need `columns.list_with_metadata` to give me a way tell the difference between a File column and a jsonb column. I understand we'll use metadata for this purpose, but I don't understand what your vision is for exactly _how_ we'll employ that metadata.

Okay, done in https://github.com/mathesar-foundation/mathesar/pull/4728/commits/a596eee99c826b580b84129cd947e67f2dc9b0ef

For now, it just returns the top-level key for the backend. This will be in the `ColumnMetadataRecord` and `ColumnMetadataBlob` in the form:

```
    "file_backend": "local_minio"  # for example
```

If that's non-null, then there's a file backend configured for the column in question. I was hoping to avoid getting too deep into that part of things since it's not (or shouldn't be) relevant to the listing/getting/etc. But, my thinking is that there should be a separate endpoint for listing details of each backend, and so on that would be used for choosing a file backend, and then it'd be actually set for a column just by storing the appropriate key in the `file_backend` attribute. I'd like for everything involved in the file backend config to be handled in the YAML and then just mapped to a column via the key so we avoid confusion or duplication w.r.t. where a particular setting is stored.

> * I need the `return_download_links` API parameter to go away. I recall discussing this with you on 2025-08-29. Since the front end requests `columns.list_with_metadata` in parallel with `records.list`, it can't rely on the columns metadata response to formulate the records listing request.

Sorry, I couldn't recall the actual outcome there, so I left it as-was. Regardless, done: https://github.com/mathesar-foundation/mathesar/pull/4728/commits/990dcae63da705aa027c15437aa23368a09baa4e

mathemancer:
@seancolsen @Anish9901 I think I've covered everything, so this should be ready for re-review, and hopeful merging.

seancolsen (approved):
@mathemancer this works for me and (fingers crossed) gives me some stuff to build on top of for a little while. As far as I'm concerned, we can merge this.

Additionally, I wanted to point out some things I noticed which might be good to address at some point, perhaps in a separate PR.

In `mathesar/views/download_link.py`:

- For these three views, I was expecting to see a `Cache-Control` header to instruct clients use aggressive caching. In my original [proposal](https://github.com/mathesar-foundation/mathesar-wiki/blob/file-attachments-proposal/docs/product/proposals/2025/file-attachments.md) I had specified `public, max-age=31536000, immutable`, i.e. cached for one year.

    I think we've discussed this on a call at some point, but I can't remember.

    Now that I'm thinking about this client-side caching in the context of this files architecture we have, it raises some questions for me. My understanding is that the mash is derived from the S3 path + installation secret key. Right? What happens when the file at the S3 path is replaced with a new file (perhaps outside of Mathesar)? Like, what do we _expect_ to happen? And to what extent to we want to support scenarios like this? In the architecture I had originally proposed, files were immutable and content-addressable. That architectural approach enabled the aggressive browser caching I had proposed. But with files being mutable (potentially outside Mathesar), the client-side caching becomes trickier. I'd really like to avoid re-loading thumbnails for every time the sheet loads — I think would look janky.

    Some ideas I have:
    
    - With mutable files we could potentially use the (newer) stale-while-revalidate Cache-Control value. I've never used this before, so I don't know if there might be pitfalls. But in theory I think this might work.

    - When uploading a file _through Mathesar_, we could place the file in the S3 provider by using a path which contains a hash of the file content. This would effectively make Mathesar-managed files immutable. But then we'd need an unambiguous way of descending "Mathesar-managed" files from non-Mathesar-managed files.

    - We could incorporate the file content into the mash. I have a vague recollection of you expressing an opinion against this approach, but I don't recall why.

    This is an architectural question that I think would be wise for us to resolve early.

- I was also expecting to see a `Content-Security-Policy` header when serving the file directly.

    - I don't think we need this for thumbnails (because they'll always be benign images).
    - And I don't think we need it for downloads (because the Content-Disposition already protects us).
    - But for loading the file directly I think it's important that we add a CSP in order to protect against malicious uploads that could become a vector for XSS.

    I think a value of `default-src 'none'` or `sandbox` would be good. (My understanding is that there are some subtle tradeoffs here, but I'm not sure it matters too much for our use case.)

    This is non-urgent because we can easily apply this change without affecting other architecture or functionality. So we can chat more about it more later. But I do think we'll need to handle this before releasing this feature.

seancolsen:
@mathemancer and I had a 35 minute [call](https://tldv.io/app/meetings/68c037a33679700013bc14ea) where we decided:

- We'd add the following cache header to all three views:

    ```
    Cache-Control: public, max-age=31536000, immutable
    ```

    Later we'll work out the details of how we plan to enforce this immutability.
    
    - We discussed an approach that would leverage the etag provided by the S3 service. I won't spell out the whole thing here since it's in the video.
    - And in the future such file immutability (and hence caching) might be different per type of file storage backend.

- We'd add the following CSP to the file loading view:

    ```
    Content-Security-Policy: default-src 'none'
    ```

seancolsen:
@mathemancer I found a bug...

If a File cell is `NULL` the `records.list` API responds with the following error:

> TypeError: the JSON object must be str, bytes or bytearray, not NoneType

mathemancer:
@seancolsen @Anish9901 @zackkrida We'll use the `file_attachment_feature` branch for the... file attachment feature branch. Merging this into that.

PR #4746, opened 2025-09-08
Clear all statuses and errors when inserting a new row

### LINKED ISSUE #4608: Cell-level errors continue to display after INSERT from a draft record row that got data pasted into it
1. Create a table with two text columns, `a` and `b`, where `b` cannot be NULL.

    Image

1. Click "New Record" to begin adding a new row.

    Image

    Mathesar will not attempt to save the record yet, because it knows that we need a value for `b`.

1. Copy the following text content into your system clipboard:

    ```
    foo
    ```

1. Paste into the `a` cell.

    Image

1. Observe the following error:

    Image

    Tangentially, there are potentially two _other_ problems here:

    - The error is relevant to column `b` but displayed within the cell in column `a`. We have #4607 for this problem.

    - You could argue that perhaps Mathesar should not even attempt to create the record at this point because it still should understand that we're missing a value for `b`. This is a less-important problem which I don't think even warrants a GitHub issue right now.

    The state at this point is interesting. We have a "draft record row" (because it hasn't been saved yet). _And_ we have a cell-level error within that draft record row. We get cell-level errors in the following scenarios:
    
    - (A) When we fail to manually update a single cell within a _saved record_.
    - (B) When pasting into cells fails to UPDATE a record.
    - (C) When pasting into cells fails to INSERT a record.

    What's happening in this case is (C).

    What's interesting to notice is that we _don't_ get cell-level error when we fail to manually update a single cell within a _draft_ record row. So perhaps we should not set cell level errors in (C). **That might be one way to address this bug. I don't know if that's the best way though.**

1. Manually edit the `b` cell, setting its value to "bar" and saving.

    Image

1. Expect the record to save and the error in cell `a` to disappear.

1. Instead, observe that the record saves but an error still displays in the `a` cell.

    Image

    We should probably be clearing all cell-level errors any time we successfully insert a record. **This would be another (probably better) way to address this bug.**

1. Refreshing will clear the error and demonstrate that `a` is set to "foo". Good.

    Image

commonspider:
Fixes #4608

When a new row is inserted, it clears all statuses and errors from the row and its cells.

Additionally, it fixes a small bug in `Meta.clearAllStatusesAndErrorsForRows` by removing an unnecessary array destructuring. As far as I can tell, the only other usage for this function is garbage collecting after row deletion. Maybe this is why the bug has never been noticed.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/meta.ts
+++ b/mathesar_ui/src/stores/table-data/meta.ts
@@ -263,12 +263,12 @@ export class Meta {
     this.rowDeletionStatus.delete(rowKeys);
     const rowKeySet = new Set(rowKeys);
     this.cellClientSideErrors.delete(
-      [...this.cellClientSideErrors.getKeys()].filter(([cellkey]) =>
+      [...this.cellClientSideErrors.getKeys()].filter((cellkey) =>
         rowKeySet.has(extractRowKeyFromCellKey(cellkey)),
       ),
     );
     this.cellModificationStatus.delete(
-      [...this.cellModificationStatus.getKeys()].filter(([cellkey]) =>
+      [...this.cellModificationStatus.getKeys()].filter((cellkey) =>
         rowKeySet.has(extractRowKeyFromCellKey(cellkey)),
       ),
     );
--- a/mathesar_ui/src/stores/table-data/records.ts
+++ b/mathesar_ui/src/stores/table-data/records.ts
@@ -746,7 +746,7 @@ export class RecordsData {
       const record = response.results[0];
       const newRow = PersistedRecordRow.fromDraft(row.withRecord(record));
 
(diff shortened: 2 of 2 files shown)


seancolsen (approved):
This looks great and works great! Thanks @commonspider.

Good catch in `meta.ts`. (For type-checking purposes, I kind of hate how every string `string` is also an `Iterable<string>`. Otherwise TS would have caught this 🙄)

I can tell you know what you're doing, so it's really great seeing some code contributions from you! Please keep it up and feel free to reach out to me directly on Matrix if you ever have questions!

Issue #2724, opened 2023-03-20
StatementError when using the Record Selector to search by a partially-entered UUID


seancolsen:
## Steps to reproduce

1. Set up a table with a UUID

    ```sql
    CREATE TABLE example_table (
      id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
      column1 TEXT,
      column2 INTEGER,
      column3 BOOLEAN
    );
    INSERT INTO example_table (column1, column2, column3) VALUES
    ('foo', 1, TRUE),
    ('bar', 2, FALSE),
    ('baz', 3, TRUE);
    ```

1. Run reflection.

1. Open the record selector for that table

    ![image](https://user-images.githubusercontent.com/42411/226345996-96f8c651-4d38-4655-8c5d-9b2deb8f26b6.png)

1. Copy-paste a full UUID into the UUID search field. Observe that the record selector finds the correct record. Good!

    ![image](https://user-images.githubusercontent.com/42411/226347286-7121c747-cfb7-4fa0-9a5c-3f57a922f5fe.png)

1. Enter a _substring_ of a UUID into the UUID search field.

    **Expect** either: (A) no results (because no record has a UUID which exactly matches the search term), or (B) some results which partially match the search term (though I'm not sure how feasible that is with the functionality available in Postgres)

    Instead, **observe** that the front end hangs because it doesn't get a proper response from the back end

    ![image](https://user-images.githubusercontent.com/42411/226346819-c9a84700-d9da-4699-bd27-ed081315bee9.png)

    The records endpoint responds with the following 500 error

    <details>
    <summary>Traceback</summary>

    ```
    Environment:


    Request Method: GET
    Request URL: http://localhost:8000/api/db/v0/tables/395/records/?limit=10&offset=0&search_fuzzy=%5B%7B%22field%22%3A2048%2C%22literal%22%3A%22d%22%7D%5D

    Django Version: 3.1.14
    Python Version: 3.9.16
    Installed Applications:
    ['django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'whitenoise.runserver_nostatic',
    'django.contrib.staticfiles',
    'rest_framework',
    'django_filters',
    'django_property_filter',
    'mathesar']
    Installed Middleware:
    ['django.middleware.security.SecurityMiddleware',
    'whitenoise.middleware.WhiteNoiseMiddleware',
[...]

PR #4758, opened 2025-09-10
Files UI

seancolsen:
## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen:
This is coming along.

**WE HAVE CATS!**

image

seancolsen:
I'm dropping some notes here, mostly to help @pavish when he gets started working on this feature.

## How to work with this feature so far

These steps will get you working with the files feature using some sample data.

1. Check out my `files_ui` branch, and base your work on tot of that. That branch is in turn based on `file_attachment_feature`, where all merged work is fully reviewed.

1. Restart docker with:

    ```
    docker compose -f docker-compose.dev.yml up obj-store dev-service --build
    #                                           ^^^^^^^^^ THIS IS NEW
    ```

    This is a new service that runs [minio](https://www.min.io/) provide an S3 compatible file storage system locally.

1. Open the web interface to administer the minio service at http://localhost:9001

    1. Create a new bucket and upload a file. Then copy path of the file here:

        

        You'll need this file path on the next step.

1. Open a python shell in Mathesar via:

    ```
    docker exec -it mathesar_service_dev python manage.py shell
    ```

    Then customize the code below using the path to your file from the previous step, and run it within the python shell:

    ```py
    from mathesar.utils import download_links as dl
    dl.create_json_for_uri('s3://my_bucket/my_file.jpeg', 'local_minio')
    #   PUT YOUR FILE PATH HERE ^^^^^^^^^^^^^^^^^^^^^^^
    ```

    This should print out something like:

    > '{"mash": "a3306f9180bb61cd0c4240a35beeb1a9e8fde0d074dc40bbdbc0f75e7f181385", "uri": "s3://banana/Orange-colored-cat-yawns-displaying-teeth.webp"}'
    
    You'll need this JSON on the next step.

1. In Mathesar, create a new table with a JSONB column.

    Then Insert JSON into a cell, using the output from the previous step. That's our _reference_ to the file in S3.

    As an aside, you might be wondering what the heck the "mash" thing is, and why we need it. This was Brent's idea, but I'll try to explain it. The problem it's solving is this: we don't want to necessarily give a Mathesar user the capability to read arbitrary files from the S3 bucket configured with the file storage backend — it could be a security risk. Here's an example of the sort of problem we're protecting against with this "mash" system. Imagine there is no mash; only a raw S3 path. You have a table a File column, and you also have RLS set up on the table. I upload the file "s3://foo/health-records-2025.pdf" and give you access to read the row which contains it. You can see the cell which contains the file reference. And the Mathesar UI lets you download the file. Cool. But lets say you write a new record with a cell "s3://foo/health-records-2024.pdf" (just changing the year to _guess_ if there's a file that exists at that path.) Nobody ever gave you access _in PostgreSQL_ to view that file. That's a problem.

    Here's how the "mash" solves this access loophole: it's a hash of the installation secret key + the file path. Before serving a file to the front end, the service layer validates the mash. This prevents malicious users from guessing a path and writing it into Postgres just so they can read the file. When manually uploading a file, Mathesar will generate this mash value and return it. And if a user has a whole bunch of S3 paths _already_ stored in PostgreSQL then (at some point in a future release) we'll provide them with a tool to generate these mash values in bulk.

1. Now we need to get Mathesar to treat this plain old JSONB column like a proper ✨**File**✨ column. We do this via column-level metadata. Use the API to write the following column metadata for your column:

    ```json
    {
      "file_backend": "local_minio"
    }
    ```

At this point you should see a thumbnail of the image in the cell. Now you have basic sample data!

seancolsen:
These serve as a reference for @pavish, and anyone else who might have questions about some of the rationale for my approach.

## About my approach to handling File columns

- Originally I had wanted to use a dedicated UI type (aka "abstract type" as named in the code). But this seemed like it would require a decent amount of refactoring. You can see (in my previous comment) that a column becomes a File column when it gets _metadata_. That's structurally different from all our current UI types. Currently, we have an assumption that one database type maps to one UI type; and one UI type maps to multiple database types.

- The structure of "jsonb database type" + "mathesar metadata" = "file column" is more akin to "numeric database type" + "foreign key constraint" = "linked record column". So I've taken an approach that follows the patterns we've used thus far for FK columns. This is why the UI type for a File column is jsonb and not something like "File".

Zack and Brent and I [discussed](https://tldv.io/app/meetings/68c183c75430540013ec0d1c) this on a call.

seancolsen:
@pavish I have a question about the new colors system. It's somewhat tangential to this PR, but it relates to the UI here. I don't see this as very important to address, but I'm commenting in hopes that you can help me understand how to work within the new colors system correctly.

I have this modal image viewer:



The two buttons indicated in the screenshot both have their appearance set to "default". But the "Info" button is rendered as part of a Dropdown trigger whereas the "Remove" button is just a standalone Button. Why do these buttons end up with different background colors? I'd like them to have the same background, but I'm not sure how I'm supposed to do that cleanly.

I'm coming to you first before spending time troubleshooting it myself.

pavish:
> The two buttons indicated in the screenshot both have their appearance set to "default". But the "Info" button is rendered as part of a Dropdown trigger whereas the "Remove" button is just a standalone Button. Why do these buttons end up with different background colors?

@seancolsen There are distinctive styles for controls and inputs in the new system, and dropdowns fall under both. Eg., a dropdown within a Select menu is an **input**, whereas a dropdown menu (or usages like in the screenshot above) is a **control**.

The default appearance for Button is control, and Dropdown is input. Dropdown.scss has overrides for the default appearance which results in this difference. The distinction between control & input is not yet implemented in the Dropdown component. 

The best way to fix that is to have separate appearances for `control` and `input`. I had left this improvement as something to fix later.

> I'd like them to have the same background, but I'm not sure how I'm supposed to do that cleanly.

Please proceed with the PR. We can tackle this as part of fixing bugs with colors before the release. The bugs are tracked [here](https://github.com/mathesar-foundation/mathesar/issues/4769).

pavish:
Notes for anyone following along for setting up the files backend, as mentioned in [this comment](https://github.com/mathesar-foundation/mathesar/pull/4758#issuecomment-3280362685).

- Ensure to add a `file_storage.yml` file at the root of Mathesar. There is an example file that configures the local minio object store (`file_storage.yml.example`).
- The records endpoint would throw an error without this configuration if there's a column with the file backend metadata but there are no backends. This could occur in cases where the user has removed the backends. It seems to be fixed [here](https://github.com/mathesar-foundation/mathesar/blob/35b9da2a2916c5ea215009f66667434df9782c0a/mathesar/utils/download_links.py#L169-L178) in the [file attachment upload PR](https://github.com/mathesar-foundation/mathesar/pull/4762).

seancolsen:
@pavish heads up for you that @mathemancer changed somethings in #4762 that will eventually affect your development environment when you merge this `files_ui` branch into your `files_column_config` for #4770. The `file_storage.yml.example` has changed now. You'll need to change your `file_storage.yml` to match it. Specifically:

1. Change `local_minio` to `default`
1. Add a `prefix` property with a value set to the name of the bucket that you're using in minio.
1. Modify your column metadata to use `"default"` instead of `"local_minio"`
1. Re-create your sample data (because the mash values have now changed).

([Matrix discussion](https://matrix.to/#/!UZILDSNKobkelUYwBp:matrix.mathesar.org/$FfDSrjheDN6a1iwY4FdQsMo9mh_H5F4EZzdO8Td8SAY?via=matrix.mathesar.org&via=matrix.org))

seancolsen:
@pavish please review this now.

zackkrida:
Looks great, @seancolsen! I wanted to mention, because it came up in our quick call, that setting any value for an "equals" filter on a File column breaks the query and causes the page to error:

Screenshot From 2025-09-16 10-51-49

I can make a standalone issue for this, I don't see a need to address it now or in this PR.

seancolsen:
@zackkrida 

> make a standalone issue for this

Yeah, please do.

I'll point out, this is not specific to Files, but rather the json type and our current backend logic for filtering. This issue is similar to #2724 because the filter condition cannot be cast to the type in question. In this case you can't cast `any value` to JSON. You _can_ cast `7` to JSON. Filtering on that works. And if you enter the entire JSON of the file reference you can actually filter on files:



So what needs to happen is a backend fix for the json type or for filtering more broadly.

mathesar_ui/src/components/cell-fabric/CellFabric.svelte
```diff
@@ -4,12 +4,18 @@
...
   import type { HorizontalAlignment } from './data-types/components/typeDefinitions';
   import type { CellColumnFabric } from './types';
 
   export let columnFabric: CellColumnFabric;
   export let value: unknown;
   export let recordSummary: string | undefined = undefined;
+  export let fileManifest: FileManifest | undefined = undefined;
+  export let setFileManifest:
+    | ((mash: string, manifest: FileManifest) => void)
+    | undefined = undefined;
```
pavish, line 18: I don't quite like this approach of passing down `fileManifest` in this component, since by the current design, this prop doesn't apply for any other column type. It should ideally be a part of the `value` for file types.

mathesar_ui/src/components/file-attachments/file-uploader/ModalFileAttachmentUploader.svelte
```diff
@@ -0,0 +1,109 @@
...
+  // $: title = $_('upload_file');
+  // ```
+  //
+  // The error is:
+  //
+  // > "Cannot format a message without first setting the initial locale."
+  //
+  // I'm using a non-translated title for now.
+  //
+  // TODO_FILES_UI: troubleshoot and fix this error
```
pavish, line 57: This is because this component is used within AppContext which is set before the translations are loaded. Moving `AppContext` to within the if block with `isTranslationLoading`  should ideally fix this. Just making a note here for reference.

mathesar_ui/src/utils/OneWayMessageChannel.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }
+
+  function receive(listener: Listener<T>): ListenerCancellation {
+    listeners.add(listener);
+    return () => listeners.delete(listener);
+  }
+
+  return [send, receive];
+}
```
pavish, line 27: This is not being used. Is it used in a different PR?

mathesar_ui/src/utils/resizeObserver.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+  observer.observe(node);
+  const rect = node.getBoundingClientRect();
+  dimensionsStore.set({ width: rect.width, height: rect.height });
+  return {
+    destroy() {
+      observer?.disconnect();
+    },
+  };
+}
```
pavish, line 31: This isn't being used. Is this used in a different PR?

mathesar_ui/src/components/cell-fabric/CellFabric.svelte
pavish: This is not a blocker.

seancolsen: I'm following the same approach with `fileManifest` and `setFileManifest` as we have already been following with `recordSummary` and `setRecordSummary`. The file manifest is very closely analogous to the record summary — it's extra stuff that we want to have for the cell but that doesn't come _inside_ the cell. The API supplies this "extra stuff" in with basically the same structure for the record summary and the file manifest. So I think it makes sense to follow a consistent pattern for them. This is what I've done. The pattern in this PR is consistent with the record summary throughout the whole front end codebase, from the stores, to the cells. Personally I think it's a good pattern. But if we want to change it (e.g. passing the file manifest along with `value`), then I think it would be good to change it for record summaries and file manifests. That's not an easy change though. Because we use `value` in places like `records.patch`, so we'd need to find a place to strip the value down to only what the API needs. Overall, I feel pretty strongly about sticking with what we've got here.

mathesar_ui/src/components/file-attachments/file-uploader/ModalFileAttachmentUploader.svelte
seancolsen: Ah! Thanks for pointing this out! I've linked this comment in our [task](https://internal.mathesar.org/db/36/schemas/112012/tables/112014/29).

mathesar_ui/src/utils/resizeObserver.ts
seancolsen: Whoops! Thanks for flagging this. I'll take it out.

mathesar_ui/src/utils/OneWayMessageChannel.ts
seancolsen: Whoops! Thanks for flagging this I'll take it out.

PR #4778, opened 2025-09-16
Automate resetting mashes for files

### LINKED ISSUE #4767: Add admin RPC function to repair MASH fields
## Problem

If the key for the file attachment backend is changed, or the `mash` value is lost or corrupted, Mathesar will refuse to download a file store with that mash.

## Proposed solution

Create a function `columns.reset_mash(column_attnum, table_oid, database_id)` that can be called from the API, and which goes through a JSON(B) column and recomputes and stores the `mash` value for each `uri` in each row.

## Additional context

- No need to handle other types for now, just throw an error.
- No need to handle malformed JSON. Just get the URI key if one exists, compute the Mash, and store that in the JSON.
- Try to leave other keys unchanged if possible
- Don't throw an error if there isn't a URI or something goes wrong. Just fix as many rows as possible.

Anish9901:
Fixes #4767

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/columns.py
+++ b/db/columns.py
@@ -2,7 +2,9 @@ import json
 
 from db import connection as db_conn
 from db.deprecated.types.base import PostgresType
+from db.records import list_records_from_table
 
+from mathesar.utils.download_links import create_mash_for_uri
 
 DEFAULT = "default"
 DESCRIPTION = "description"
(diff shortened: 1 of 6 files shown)


db/columns.py
```diff
@@ -2,7 +2,9 @@
 
 from db import connection as db_conn
 from db.deprecated.types.base import PostgresType
+from db.records import list_records_from_table
 
+from mathesar.utils.download_links import create_mash_for_uri
```
mathemancer: This shouldn't be imported from the `db` layer. Imports go the other way.

So, I think this function should probably live in the `mathesar/utils/download_links.py` file.

db/sql/05_msar.sql
```diff
@@ -5763,6 +5763,50 @@ END;
...
+*/
+DECLARE
+  sch_name text := msar.get_relation_schema_name(tab_id);
+  tab_name text := msar.get_relation_name(tab_id);
+  col_name text := msar.get_column_name(tab_id, col_id);
+BEGIN
+  EXECUTE format(
+    $j$
+      UPDATE %1$I.%2$I
+      SET %3$I = jsonb_set(%2$I.%3$I::jsonb, '{mash}', mapping.mash, false)
```
mathemancer: I think we actually _should_ create that key if it doesn't exist.

mathemancer (changes requested):
Requesting a couple of changes. See specific comments.

db/columns.py
Anish9901: We, also don't allow `exec_msar_func` calls from `mathesar/`, so I've split the function.

db/sql/05_msar.sql
Anish9901: done in e45bfdf

mathemancer (approved):
Okay, I like it. Thanks.

Issue #3955, opened 2024-10-11
Text field doen't stay focused while changing column name in Data Explorer


Anish9901:
## Description

https://github.com/user-attachments/assets/5c7960f5-0e70-4b21-a346-ee1dd404ecee

## To Reproduce

- Open data explorer
- Add some columns to explore
- Try changing the column name from the inspector
- Notice that the text area looses stay focus while changing the column name.

PR #4802, opened 2025-09-21
Keep focus on selected placeholder cell on plane changes

### LINKED ISSUE #4751: Placeholder cell loses selection after data entry
## Steps to reproduce

1. Select a cell in the placeholder row
1. Enter data into it.
1. Observe that a new record is added with the data you entered.
1. Expect that the cell you typed into is still selected.
1. Instead, observe that no cell is selected.

This is annoying because it makes rapid data entry more laborious.

https://github.com/user-attachments/assets/1b7be1a6-2e39-4f62-9075-79d445d8b045

commonspider:
Fixes #4751 

When a new record is added from the placeholder, it keeps the focus on the currently selected cell.

**Technical details**

I have identified the source of the bug thanks to an helpful comment in `SheetSelectionStore.ts`. This is an explanation of the PR:

- `SheetSelectionStore.ts`: Added a `options` parameter to `SheetSelectionStore`'s constructor that is directly passed to `this.selection.update`. This allows each instance of the class to show a different behavior regarding which effects (currently just `'focus'`) to prevent.\
There could be a little problem here because before the PR the default behavior was to prevent focus, while after the PR it is not to prevent it. But since `SheetSelectionStore` is instantiated only twice in the code base, it should be fine.
- `QueryRunner.ts`: Auto-focus is prevented because of #3955. I've moved here the summary of the comment from `SheetSelectionStore.ts`
- `placeholderCellBasis.ts`: Function `adaptToModifiedPlane` is modified following to the description of the parent function. Honestly, I don't really understand the original code, maybe I've missed some edge cases.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/components/sheet/selection/SheetSelectionStore.ts
+++ b/mathesar_ui/src/components/sheet/selection/SheetSelectionStore.ts
@@ -16,22 +16,14 @@ export default class SheetSelectionStore
 
   private cleanupFunctions: (() => void)[] = [];
 
-  constructor(plane: Readable<Plane>) {
+  constructor(plane: Readable<Plane>, options: { prevent?: 'focus'[] } = {}) {
     super();
     this.selection = new PreventableEffectsStore(new SheetSelection(), {
       focus: () => this.focus(),
     });
     this.cleanupFunctions.push(
       plane.subscribe((p) =>
-        // Prevent auto-focusing the new active cell when the plane changes.
-        // Originally I had allowed the auto-focus to happen, but it was causing
-        // a [bug][1] in the Data Explorer. I think it makes sense to err on the
-        // side of caution here and prevent the auto-focus. There might be some
-        // cases where we want to auto-focus the new active cell, but we can
-        // handle those cases imperatively as needed.
-        //
-        // [1]: https://github.com/mathesar-foundation/mathesar/issues/3955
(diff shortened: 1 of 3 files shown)


seancolsen:
Hey @commonspider! I love that you took this on. And I can see that these changes _do_ fix the problem described in #4751. Thanks! And nice work figuring that out!

In looking at the code though, I can tell that this approach is going to cause some inadvertent regressions. I was able to specifically identify one regression it causes, and there could be others too. I'll spend some time describing that regression. And I'll also describe a new approach that I think we'll need to take in order to fix this without introducing regressions. But I'll need to do a little research first. And before I do that, I want to get the changes in this PR reverted since we just began our pre-release QA today. For the sake of QA, @zackkrida and I thought it would be best to err on the side of caution and let you take another stab at fixing this for 0.7.0 with some guidance from me. I've opened #4816 to revert these changes.

We hope that reverting your PR doesn't feel deflating to you! It's really just a consequence of timing here. I think @zackkrida was excited to squeeze this into 0.6.0. But it turns out we just actually need a bit more time for review and discussion, hence the reversion approach. We'd still very much like to involve you in the next steps of fixing this bug. We just need to take care of the 0.6.0 release first.

Going forward, I'll move discussion to #4751.

PR #4808, opened 2025-09-22
Improve filters

pavish:
Fixes #[issue number]

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


pavish:
I finally re-started work on this PR, after it being idle for a long time. It's functional, complete with drag and drop.

https://github.com/user-attachments/assets/e3fa43bc-a006-4136-bcb9-acb67b7096d8

However, I still feel like the UI doesn't quite look good. It might be because of the icons or the colors, or the lack of animation (I'm not sure what's bothering me).

@zackkrida @seancolsen @kgodey Please review the UX, play around with it, and let me know your thoughts.

If the UX looks good, I can cleanup the code and colors, and get the PR ready for code review.

seancolsen:
@pavish, here's another small UX suggestion. This one is less important, but I wanted to mention it because I think it would slightly improve usability, and it's possible it could make some of the UI implementation simpler.

This suggestion relates to the behavior when dragging an item. So it's somewhat difficult to mock up. I'll try to explain via text...

Currently, when dragging:

- The UI that follows my cursor is the entire thing I'm dragging. This can obscure the underlying tree of filter conditions, making it a little hard for me to see the place that I want to drop as I'm dragging.

- The filter entry that I'm dragging is removed from the tree. This makes it hard for me to understand where I'm moving it "from". It's a little harder to me to mentally retain my context.

- The drop target (the dotted outline) occupies space in the tree. This makes the tree layout change as I drag. That layout shift can makes dragging a little cumbersome sometimes. For example, if I drag 10px too far down, then I might need to go 100px back up to correct my mistake before dropping.

None of these issues are hugely problematic. After all, the patterns you're following here are quite prevalent across other apps. But with a complex drag-and-drop feature like this, I have a slight preference for the UI to basically "stay in place" as I drag.

I'd (slightly) rather it work like this:

- The UI that follows my cursor would be very small — just an icon, or even nothing at all. This way I could see the underlying tree more clearly.

- The filter entry that I'm dragging would stay in the tree but would get opacity 0.5 or something similar. This way I could see what I'm dragging _and where I'm dragging it from_.

- The drop target would be a blue line in between items, similar to the UX when dragging to rearrange columns. This way the layout would not shift around as I drag.

Not a big deal. But I wanted to suggest this. Who knows, perhaps you'd end up re-doing some of the DnD UI if change the layout to suite some revised aesthetic design choices, and perhaps the DnD behavior I'm proposing would be simpler to implement. I'm not sure. I've not looked at any of this code yet.

pavish:
Thanks @seancolsen. I implemented the most commonly used UX behaviour for drag and drop, and filter groups. I do like your suggestions, however, I'll wait until Zack and Kriti have weighed in, before I make a call.

pavish:
@kgodey This PR is ready for re-review following our last UX call.

### When no filters are added, the columns are displayed. (This change is for sort and group as well)
- Screenshot 2025-12-08 at 10 24 55 AM

### When the first filter is added
- Screenshot 2025-12-08 at 10 25 01 AM
- Note: I tried removing the **where any of the following is true** selector when there's only one filter, but adding more filters (especially filter groups) made the UI a bit jarring to use.
- The drag handles are re-positioned and do not have backgrounds
- The delete (trash) icon is changed to an 'x' icon

### When multiple filters are added
- Screenshot 2025-12-08 at 10 25 27 AM

### When a filter group is added
- Screenshot 2025-12-08 at 10 25 38 AM
- Screenshot 2025-12-08 at 10 25 48 AM
- Note: The filter group automatically switches the default operator (and/or) at each level. Eg., if the first group has 'or' and the user add a sub filter group, the default operator for the sub-group would be 'and'. 
- Lines are added to each level. They are solid/dashed based on the operators and/or respectively.

### When there are a lot of filters
Screenshot 2025-12-08 at 10 26 44 AM

pavish:
@seancolsen Please review code.

Note: I haven't changed the drag and drop behaviour. I think the behaviour in this PR is good. We can discuss/consider changing it if it becomes a concern but I'd like to avoid doing anything on that now.

pavish:
I made more changes w.r.t to the background colors of inner/sub-groups, and the text/appearance of buttons following a call with @kgodey:

Screenshot 2025-12-08 at 10 06 50 PM

seancolsen:
I'm merging this since I know we're trying to release soon.

@pavish I'd still be curious to hear your thoughts to the network request issue and the UI text issue. Maybe we need GitHub issues for those. Maybe not.

pavish:
@seancolsen 

> Prior to this PR, any change to the filter/sort/group/join leads to one records.list request.
> But in this PR, there seem to be four such requests for every change. I don't know why. I've not spent time troubleshooting this, but it would be nice to fix.

This is odd, I'm only seeing one request per change and I'm not able to reproduce this. We could try syncing up over a call to debug this.

> Group labels

> to separate the label and the input into distinct strings that can each stand on their own
> we'll run into difficulty translating

I understand the reasoning, however, when reading the filter dropdown visually, I find using "where all of the following are true" to be more friendlier and easier to understand. I think we can find a way to figure out issues with translation and use the RichText component more effectively, I'd like focus on that rather than change the text.

> in this context the word "any" is plural

We can make this change as part of QA.

PR #4836, opened 2025-09-30
Improve cell context menus

### LINKED ISSUE #1770: Closing the context menu should not pass the click through
## Steps to reproduce

1. Select one table cell.
1. Open the context menu within that table cell.
1. With the context menu still open, click on another table cell.
1. Observe the context menu to close (good).
1. Expect your original selection to remain unchanged because your click was intended only to close the menu. (In testing other products, clicking to close a context menu does not ever seem to pass the click through.)
1. Instead, observe that the click has been ass through modifying the cell selection.

### LINKED ISSUE #1771: Cell Context menu event should maintain the cell selection when triggered within the selection
## Steps to reproduce

1. Select a block of cells.
1. Trigger the context menu on a cell outside the selected block.
1. Observe the cell selection to change, selecting only the cell on which the context menu was triggered. (good)
1. Select a block of cells again.
1. Trigger the context menu on a cell _inside_ the selected block.
1. Expect the cell selection to remain unchanged when the context menu opens. This behavior will be important when/if we add actions within the context menu that operate on multiple cells/rows/columns.
1. Observe the cell selection to change, selecting only the cell on which the context menu was triggered. (bad)

### LINKED ISSUE #2234: Only show "Go to Record Page" context menu entry for rows that have saved records

### LINKED ISSUE #4544: Cell focus is lost after setting value to NULL
## Description

Currently the only way to set a cell's value to NULL is by using the context menu. After setting the value to null, focus isn't restored to the cell. 

Here's an example video (notice the grey cell border after selecting "Set to NULL" in the context menu:

https://github.com/user-attachments/assets/e18dbc36-c677-4e02-b9dd-fb3594630fa0

## Expected behavior

After selecting "Set to NULL", and potentially other context menu values, focus should be restored to the cell.

### LINKED ISSUE #4825: Cannot set filter on column via context menu in table widget
On the table page, you can set filters via the context menu from the column header cell.

In the table widget, the entry is present in the context menu, but it does nothing.

I noticed this while doing some refactoring on the context menu system. The refactoring I'm doing will fix this bug.

seancolsen:
This PR makes a lot of structural and usability improvements to the cell context menu.

## User-facing improvements

- Fixes #4825
- Fixes #4544
- Fixes #2234
- Fixes #1771
- Fixes #1770
- Adds the ability to **delete multiple rows** and set **multiple cells to null** from the context menu
- Traps the users focus and enables **keyboard navigation** of all "modal" menus across the app. This also affects menus like the site admin menu and "Add new sort condition" menu. There are lots of these!
- Adds "danger" color to menu option that deletes rows.
- Changes the cell context menu to use **sub-menus** instead of sections.
- Makes minor improvements to the two menu entries to view linked records, putting the record summary inside the menu label.

## Developer-facing improvements

- Refactors the context menu components to operate on JS primitives instead of Svelte primitives
- Refactors the table view context menu to operate at the sheet level instead of the cell level
- Adds a system for defining sub-menus
- Adds an improved system for defining divided sections within a menu

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/component-library/menu/Menu.svelte
+++ b/mathesar_ui/src/component-library/menu/Menu.svelte
@@ -5,6 +5,8 @@
   import {
     blurElement,
     focusElement,
+    getFirstFocusableAncestor,
+    getFocusableDescendants,
   } from '@mathesar-component-library-dir/common/utils';
   import {
     makeStyleStringFromCssVariables,
@@ -51,7 +53,7 @@
 
   function handleMouseMove(e: MouseEvent) {
     const target = e.target as HTMLElement;
-    const item = target.closest('[data-menu-item-focusable]');
+    const item = getFirstFocusableAncestor(target);
     if (!item) return;
     focusElement(item);
 
@@ -63,9 +65,7 @@
   }
 
   function moveSelectionByOffset(offset: number) {
-    const elements = [
-      ...menuElement.querySelectorAll('[data-menu-item-focusable]'),
-    ];
+    const elements = [...getFocusableDescendants(menuElement)];
 
     const targetIndex = (() => {
       const focusedElement = document.activeElement;
--- a/mathesar_ui/src/component-library/common/actions/focusTrap.ts
+++ b/mathesar_ui/src/component-library/common/actions/focusTrap.ts
@@ -1,82 +1,7 @@
 import { tick } from 'svelte';
 import type { ActionReturn } from 'svelte/action';
 
-import { focusElement, hasMethod, hasProperty } from '../utils';
-
-function getTabIndex(element: unknown): number | undefined {
-  return hasProperty(element, 'tabIndex') &&
-    typeof element.tabIndex === 'number'
-    ? element.tabIndex
-    : undefined;
-}
-
-/**
- * Yields focusable elements within a container. The elements are yielded in DOM
- * order (depth first, pre-order).
- */
-function* getFocusableElementsInDomOrder(
-  container: Element,
-): Generator<{ element: Element; tabIndex: number }> {
-  // Cast a wide net with selector for potentially focusable elements
-  const selectors = [
-    'input',
-    'button',
-    'select',
-    'textarea',
-    'a[href]',
-    'area[href]',
-    'iframe',
-    'object',
-    'embed',
-    '[tabindex]',
-    '[contenteditable="true"]',
-    '[contenteditable=""]',
-    'audio[controls]',
-    'video[controls]',
-    'details',
-    'summary',
-  ];
-  const potentiallyFocusable = container.querySelectorAll(selectors.join(', '));
-
-  for (const element of potentiallyFocusable) {
-    // Narrow the net by checking additional properties of each element...
-
-    const tabIndex = getTabIndex(element);
-
-    // Filter out elements which don't have a tabIndex property at all
-    if (tabIndex === undefined) continue;
-
-    // Filter out elements with negative tabIndex, e.g. div
-    if (tabIndex < 0) continue;
-
-    // Filter out disabled elements
-    if (hasProperty(element, 'disabled') && element.disabled) continue;
-
-    // Filter out elements that are not visible
-    const { display, visibility } = getComputedStyle(element);
-    if (display === 'none' || visibility === 'hidden') continue;
-
-    // Filter out hidden input elements
-    if (hasProperty(element, 'type') && element.type === 'hidden') continue;
-
-    // Yield most elements with a valid tabIndex
-    yield { element, tabIndex };
-  }
-}
-
-/**
- * Returns the focusable elements within a container. The elements are returned
- * sorted in tab order (i.e. the order of their tabIndex values primarily, and
- * then in DOM when their tabIndex values are equal).
- */
-function getFocusableElements(container: Element): Element[] {
-  const elements = Array.from(getFocusableElementsInDomOrder(container));
-
-  // Sort elements by tabIndex only, assuming stable sort keeps DOM order
-  elements.sort((a, b) => a.tabIndex - b.tabIndex);
-
-  return elements.map(({ element }) => element);
-}
+import { focusElement, getFocusableDescendants, hasMethod } from '../utils';
 
 interface FocusTrapOptions {
   /**
@@ -114,7 +39,7 @@ export default function focusTrap(
     // necessary because the list of focusable elements can change as state
     // changes (e.g. elements being added/removed from the DOM or state like
     // "disabled" being toggled).
-    const elements = getFocusableElements(container);
+    const elements = getFocusableDescendants(container);
     if (!elements.length) return;
 
     function wrap(index: number): number {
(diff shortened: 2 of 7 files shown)


mathesar_ui/src/component-library/menu/Menu.svelte
```diff
@@ -26,14 +48,103 @@
...
+    controller.openSubMenuAfterDelay(item);
+  }
+
+  function handleMouseLeave() {
+    blurElement(document.activeElement);
+  }
+
+  function moveSelectionByOffset(offset: number) {
+    const elements = [
+      ...menuElement.querySelectorAll('[data-menu-item-focusable]'),
```
pavish: I don't quite like this selector. Custom menu items (eg., The theme selector in the app header settings dropdown) are ignored if this data attribute is not present, and I don't think we should add this attribute to such menu items, even if we had a wrapper component for them.

I'd prefer a common utility which identifies focusable elements (ideally each menu item should have only one focusable element, and if there are multiple elements, I think it makes sense to focus each of them individually on arrow movements).

We're using a different check with known focusable elements in the `focusTrap` action. I'd like us to use the same logic here.

seancolsen: Fair point. I had considered doing what you suggest here. But I skipped it for the sake of time. I'll implement it.

seancolsen:
@pavish, to respond to your UX points...

## Re-triggering the context menu

You said:

> the browser's default context menu opens instead of our custom context menu

Right. Yeah, that does feel a bit weird. I'd like to adjust it if we can agree on the right way.

You said:

> I'd expect our custom context menu to open for the new cell and the existing context menu to close.

I can see how this would be a reasonable expectation. Unfortunately that's a bit difficult to implement, so I'd rather not spend time implementing this specific behavior unless we think it's important. I'm wondering if we can get by with something simpler.

I went looking at how other apps handle this case of re-trigger the context menu elsewhere when a context menu is already open. Here's what I found:

- These apps _close the original context menu without opening anything else_:
    - **LibreOffice Calc**
    - **Figma**
    - **Firefox**
    - **Chrome**
    - **GIMP** (GTK)
    - **Gwenview** (Qt)

    This seems like a common pattern. And it would be easy for us to implement.

- These apps have other behavior:
    - **Google Sheets** closes the original context menu and re-opens a new one based on where the user clicked. (This seems to be what you want.)

    - **Airtable** closes the original context menu and re-opens a new context menu. However, there is some strange behavior. If you right-click on a different _data cell_, then it does _not_ modify the cell selection to the new point where you clicked. This feels a bit weird to me. If you right-click on a row or column _header_ cell, then the selection _is_ modified when the new context menu opens.

    - **Baserow** is similar to Airtable but slightly worse because the selection is never modified, even when re-triggering the context menu on a header cell.

    - **Notion** closes the original context menu and then opens a _browser_ context menu where clicked. Of al the different behaviors I tried, this one felt the worst to me.

Based on the commonality of the "close original menu" behavior, I'd like to go with that. I pushed ed2d218a6ab60333ce102a5845700ad39bcde7e7 to implement it.

While I can appreciate that this might not be your _ideal_ behavior, I would like to know if you agree that this is "good enough".


## Nested menus

I really want the nested menus.

I think we have a fundamental divergence of thinking on this point:

> keep the context menu limited and spend time making sure every action in the context menu is valuable enough to the user.

I want all actions in the table inspector to be available from the context menu. And I even want some things in the context menu that I don't want in the inspector (e.g. "Copy", "Paste"). If we organize the menu well (using sub-menus), I think it will be quite usable. If we don't use sub-menus for "Row" and "Column" then we need to use sections (i.e. dividers) for those areas, and it makes it harder to split row-specific actions up with dividers. This is a bit hard to explain. It would be easier to show you what I'm talking about on a call.

I feel strongly about this "comprehensive context menu" approach because I see it as a simple solution that addresses multiple high-priority pain points that I've been experiencing while using Mathesar personally, which I'm doing almost every day now. From my phone and tablet, is no way to copy-paste. From my laptop, I prefer to work with the inspector closed. Leaving the inspector open consumes valuable screen space. Re-opening the inspector to perform actions requires much more hunting around than I think the context menu options would. I don't want to get rid of the inspector. But I want to bring actions closer to where the user needs them.

If we're going to continue debating this, I think it would be most productive to do it over a call where we can point at a screen share and have a tight feedback loop to weigh different approaches and nuances. For example, one idea that I would want to bounce off you is removing the row/column actions from the data cells entirely. There are some tradeoffs to that approach to consider.


## Placeholder context menu

You said:

> I really hate sites where I can't right click on something

Yeah, I agree.

Let me point out another place with this same problem: form a selection which spans multiple columns and then right-click on a column header cell. No menu. I agree this would be nice improve. We don't currently have anything (easy) to _do_ with multiple columns. Eventually I'd like to have "Delete Columns" (and more), but that will require more work.

If I'm understanding you correctly, your gripe is not necessarily that we lack a "Add new row" context menu option. But rather that you want to see _something_ open when clicking on the placeholder data cells (and presumably the column header cells in my example too).

If that's the case, then I think the better way to handle this is to fall back to displaying the browser context menu when we don't have any context menu options to show. I had considered doing this, but I chose to skip it, just for the sake of time. I've pushed d8a2b9a333c2c17edfb0c95c61b07f8bb0a5950c which does this now.

pavish:
> I'm wondering if we can get by with something simpler.

We discussed this in the internal engineering call, and we agreed to only handle closing the context menu when the user right-clicks elsewhere while an existing context menu is open.

> Nested menus

The rest of the team were happy with the nested menus, since the options that are nested are "secondary" actions from the context of a cell. I'd be curious to hear @kgodey's opinion as well, however, it doesn't have to block the PR.


Response to @seancolsen:

> Placeholder context menu

> If I'm understanding you correctly, your gripe is not necessarily that we lack a "Add new row" context menu option. But rather that you want to see something open when clicking on the placeholder data cells (and presumably the column header cells in my example too).
> If that's the case, then I think the better way to handle this is to fall back to displaying the browser context menu


I'm good with it opening the browser context menu.

We can merge the PR once https://github.com/mathesar-foundation/mathesar/pull/4836#discussion_r2417057350 is implemented.

mathesar_ui/src/component-library/menu/Menu.svelte
seancolsen: Ok I pushed b68476b395f56780a0e8fe5f54396ed7fece962a to address this.

seancolsen:
@pavish this is ready for re-review now.

PR #4841, opened 2025-10-01
Fix concurrency issues

### LINKED ISSUE #4831: Gunicorn concurrency issues
## Description

When running Mathesar under Gunicorn (both natively and in Docker), the application only starts with Gunicorn’s default configuration: 1 sync worker and 0 threads. This means all incoming requests are handled serially. Any long-running database query blocks all other requests, leading to severe concurrency issues for installs with multiple users. In Docker, the Compose healthcheck can also get stuck if the only worker is busy.

In both our Dockerfile and in our [direct install method](https://docs.mathesar.org/latest/administration/install-from-scratch/), the command to run Mathesar does not pass any worker/thread configuration.
 
## To Reproduce

1. Run Mathesar via Docker or our our direct installation method.
1. Issue one long-running SQL query from the UI.
1. Attempt to perform any other action in the app while the first query is still running.
1. Observe that all other requests are blocked until the first query finishes.

## Additional context

Suggested fixes:

- Patch bin/mathesar to include worker/thread settings, ideally parameterized via environment variables and configurable in both our Docker compose file and in the direct installation method
- Write documentation explaining concurrency to users and how to tune it properly for their installs based on expected user counts and so on

## Related issues

- #4765 
- #4807 

This may warrant its own patch release at maintainer discretion.

Anish9901:
Fixes #4831

**Technical details**

**Screenshots**

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/build-scripts/bin/mathesar
+++ b/build-scripts/bin/mathesar
@@ -1,12 +1,87 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-#< The location of this file after packaging will be `<install_root>/bin/mathesar`
+#=======> INLINING STARTS: ../common_utilities.sh
 
-#<======THIS SECTION IS UPDATED DYNAMICALLY DURING PACKAGING===================
+# Color definitions for output
+RED=$(tput setaf 1 2>/dev/null || echo "")
+GREEN=$(tput setaf 2 2>/dev/null || echo "")
+BLUE=$(tput setaf 4 2>/dev/null || echo "")
+RESET=$(tput sgr0 2>/dev/null || echo "")
 
-include_source "../common_utilities.sh"
+info() {
+  echo -e "${BLUE}==> $1${RESET}"
+}
 
+success() {
+  echo -e "${GREEN}==> $1${RESET}"
+}
+
+danger() {
+  echo -e "${RED}$1${RESET}" >&2
+}
+
+err() {
+  danger "ERROR: $1"
+  cat <<EOF
+
+Please refer to Mathesar's documentation or reach out to the Mathesar team for additional support:
+  * Documentation      : https://docs.mathesar.org
+  * Github repo        : https://github.com/mathesar-foundation/mathesar
+  * Community channels : https://mathesar.org/community
+
+EOF
+  exit 1
+}
+
+command_exists() {
+  command -v "$1" > /dev/null 2>&1
+  return $?
+}
+
+require_command() {
+  if ! command_exists "$1"; then
+    err "The script requires '$1' (command not found)"
+  fi
+}
+
+# This function runs a supplied command, intends the output and colors stderr
+run_cmd() {
+  if [[ $# -eq 0 ]]; then
+    # This should never happen in our code
+    err "run_cmd: no command specified"
+  fi
+
+  local status=0
+  local tmpfile
+  tmpfile=$(mktemp) || { err "run_cmd: mktemp failed"; }
+
+  # Why are we writing to a tempfile?
+  # - This is because some scripts re-route stdout to stderr & vice-versa.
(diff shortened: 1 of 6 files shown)


Anish9901:
- To get locust running: 
    - In your virtual env run(`pip install locust`) i.e outside of the docker container.

    And then,

    - `locust -f locust.py` from the root of the repo.
   
    Locust interface should start on http://localhost:8089

- For copying over the latest SEC dataset for creating huge explorations/tables:
`scp insider-trading.mathesar.org:/home/anish/backup .` (its a 7 GB file so might take a while to download)

cc @mathemancer

mathesar/rpc/explorations.py
```diff
@@ -193,7 +193,7 @@ def run_saved(*, exploration_id: int, limit: int = 100, offset: int = 0, **kwarg
     Returns:
         The result of the exploration run.
     """
-    user = kwargs.get(REQUEST_KEY).user
+    user = 1 or kwargs.get(REQUEST_KEY).user
```
Anish9901, line 196: change to your admin user(for testing using locust).

build-scripts/bin/mathesar
Anish9901, line 1: I copied this file from within the container. I think this file shouldn't be necessary and only changes in `build-scripts/bin/mathesar.sh` should suffice during the docker build process.

having a separate file helped me to mount and made it eaiser for me to make changes for the prod image.

build-scripts/bin/mathesar.sh
```diff
@@ -221,6 +221,7 @@ run_mathesar() {
   gunicorn_args=(
     config.wsgi
     -b "0.0.0.0:${MATHESAR_PORT}"
+    --workers $(($(nproc) + 1)) --threads $((2 * $(nproc)))
```
Anish9901, line 224: This is the only relevant change.

mathesar/views/__init__.py
```diff
@@ -22,6 +22,19 @@
...
 
+import time
+from django.http import JsonResponse
+
+from django.views.decorators.csrf import csrf_exempt
+@csrf_exempt
+def slow_view(request):
+    from mathesar.rpc.explorations import run_saved
+    t1 = time.perf_counter()
+    exp = run_saved(exploration_id=6, request=request)
```
Anish9901, line 33: Change the `exploration_id` to your big exploration that takes 2-3 secs to load(for testing using locust)

docker-compose.yml
```diff
@@ -203,6 +205,7 @@ services:
   #
   db:
     image: pgautoupgrade/pgautoupgrade:17-bookworm
+    shm_size: 5gb
```
Anish9901, line 208: This was necessary because postgres processes would run out of memory and result in the following error:
```
ERROR:  could not resize shared memory segment "/PostgreSQL.1684934532" to 4194304 bytes: No space left on device
```

build-scripts/bin/mathesar.sh
zackkrida: I'm curious what @mathemancer thinks but I strongly suspect we should expose this configuration to users, through env vars or yet another config file, or both, so they can tune these values for their particular setup.

mathemancer:
Closing in favor of #4844, since the method used here breaks the installation for MacOS.

build-scripts/bin/mathesar.sh
mathemancer: We can do it with an envvar, but only the number of workers is affected. This is fine, since the distinction between workers and threads doesn't seem that relevant to Mathesar. we're blocked on I/O, not CPU or Memory.

PR #4844, opened 2025-10-02
Config web concurrency

### LINKED ISSUE #4831: Gunicorn concurrency issues

mathemancer:
Fixes #4831 

This adds a `WEB_CONCURRENCY` environment variable to the docker compose setup. Appropriate settings are documented in the same file.

TODO

- [x] Document the variable for from scratch installs.
- [ ] Document the variable for 1-click deployments.

**Technical details**

This variable is used by Gunicorn to set the number of workers, and is the only variable supported for changing concurrency settings, other than the variable to set _all_ command-line arguments. The default I chose is "3", which is reasonable as long as Mathesar is running with at least 1 logical core available. This will improve the situation, but only a bit. Basically, with the `sync` worker type, we need one worker per concurrent request. So, this takes us from 1 to 3 :confetti_ball: .

Long-term, we should try to use the `gevent` worker type, since that enables async I/O, and Mathesar is extremely I/O, rather than CPU, bound. However, that doesn't work (easily) in Psycopg2, which means it doesn't affect Explorations. It _does_ improve (fix, really) performance in the non-exploration part of the codebase, but I didn't want to make the change in the current limited-QA situation. If we decide we are not going to try to squeeze this PR into 0.6.0, then I think we should make that change as well, to get the benefits in the parts of the codebase where we're using `psycopg3`.

This is yet another spot where our failure to deal with Explorations and the deprecated SQLAlchemy part of the codebase is biting us.

To test this, you can just load big tables in multiple browsers manually, or follow the instructions in #4841 .

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -169,7 +169,7 @@ services:
       SECRET_KEY: ${SECRET_KEY-}
 
       # If you are running Mathesar on a particularly powerful machine, you can
-      # improve performance by increasing this setting. Replace '5' below with
+      # improve performance by increasing this setting. Replace '3' below with
       # 2 * $(NUM_CORES) + 1. So if you have a VPS with 4 vCPUs you can use a
       # value of 2 * 4 + 1, or 9. The default is workable as long as you have at
       # least 1 logical core.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -168,6 +168,13 @@ services:
       # leave this as-is.
       SECRET_KEY: ${SECRET_KEY-}
 
+      # If you are running Mathesar on a particularly powerful machine, you can
+      # improve performance by increasing this setting. Replace '5' below with
+      # 2 * $(NUM_CORES) + 1. So if you have a VPS with 4 vCPUs you can use a
+      # value of 2 * 4 + 1, or 9. The default is workable as long as you have at
+      # least 1 logical core.
+      WEB_CONCURRENCY: ${WEB_CONCURRENCY-3}
+
     volumes:
       - ./msar/static:/code/static
       - ./msar/media:/code/media
@@ -208,8 +215,10 @@ services:
     environment: *config
     # Expose the internal port 5432 only to other containers and not
     # the underlying host.
-    expose:
-      - "5432"
+    # expose:
+    #   - "5432"
+    ports:
+      - "5432:5432"
     volumes:
       - ./msar/pgdata:/var/lib/postgresql/data
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -215,10 +215,8 @@ services:
     environment: *config
     # Expose the internal port 5432 only to other containers and not
     # the underlying host.
-    # expose:
-    #   - "5432"
-    ports:
-      - "5432:5432"
+    expose:
+      - "5432"
     volumes:
       - ./msar/pgdata:/var/lib/postgresql/data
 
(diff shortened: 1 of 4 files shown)


mathemancer:
@zackkrida This is mergable, and doesn't _harm_ the 1-click deployments, but I think it would make sense to document setting this variables for those scenarios as well. I'm not sure how the users should determine the number of vCPUs on various platforms, which is a prerequisite to setting this variable correctly.

docker-compose.yml
```diff
@@ -208,8 +215,10 @@ services:
     environment: *config
     # Expose the internal port 5432 only to other containers and not
     # the underlying host.
-    expose:
-      - "5432"
+    # expose:
+    #   - "5432"
+    ports:
+      - "5432:5432"
```
Anish9901: Is this change necessary, I rememeber making a conscious decision to not expose the postgres port to the host.

docker-compose.yml
```diff
@@ -168,6 +168,13 @@ services:
       # leave this as-is.
       SECRET_KEY: ${SECRET_KEY-}
 
+      # If you are running Mathesar on a particularly powerful machine, you can
+      # improve performance by increasing this setting. Replace '5' below with
```
Anish9901:
```suggestion
      # improve performance by increasing this setting. Replace '3' below with
```

pavish:
@mathemancer @zackkrida This commit https://github.com/mathesar-foundation/mathesar/pull/4844/commits/cb2504f1e9823d1d6b879fafda76d620c89455ac adds a default value for WEB_CONCURRENCY env variable for the from-scratch script based installation.

docker-compose.yml
zackkrida: @mathemancer was this for testing purposes or some other reason?

mathemancer: Oops! Great eye, @Anish9901 . That was for testing.

mathemancer: Okay, reverted in https://github.com/mathesar-foundation/mathesar/pull/4844/commits/1682771e1f0ce87ee2c441fe86f472b822548fa1

PR #4848, opened 2025-10-03
Add files to forms - backend

pavish:
Related to https://github.com/mathesar-foundation/mathesar/issues/4829

Updates the following endpoints to work anonymous for publicly shared forms:
- `files/` (POST - upload)
- `files/<slug:download_link_mash>/`
- `files/<slug:download_link_mash>/thumbnail/`
- `files/<slug:download_link_mash>/download/`

These endpoints are required by the frontend for being able to upload a file and view it in a FileInput field.

## Technical details
- For anonymous users, we check if the request contains a valid form token and a field key of a file type column.
  - This is to ensure that the download link endpoints are not accessible via publicly shared forms when there are no file type columns in the form.
  - This adds additional requests to the Django internal db and the user db. However, these requests are made only when the upload is anonymous. I think it's reasonable and essential to have these additional checks despite the additional requests.
- Security considerations:
   - Since the user session is part of the mash, we can say for certain that the user can only download a file that _they_ have uploaded, even when the session is anonymous. Note that anonymous sessions work similar to authenticated sessions.
   - I don't see any other potential attack patterns w.r.t to unauthorized access. I'd like reviewers to spend some time thinking of possible issues.
   - The concerns on public file uploads leading to DOS attacks and racking up server costs are still present. We need to document rate limiting patterns when deploying to production & dealing with public forms. This can be done at the proxy (Caddy) level.
- Size Limits:
  - I've added a size limit for anonymously uploaded files.
  - Note that this setting is an the env variable and not configured in `file_storage.yml`.
    - The file_storage yml file is currently specific to configuring storage backends and I didn't want to modify that to add this setting there, since once we add anything to the yml, we would have to support it (for a long time).
    - When we get more requirements to customize file upload behaviours, we can move the setting to the yml file. I think its reasonable to have it be an env variable for now.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/45_msar_type_casting.sql
+++ b/db/sql/45_msar_type_casting.sql
@@ -1076,6 +1076,16 @@ RETURNS character varying AS $$
   SELECT $1::text;
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
 
+CREATE OR REPLACE FUNCTION msar.cast_to_character_varying(integer)
+RETURNS character varying AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_character_varying(smallint)
+RETURNS character varying AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
 CREATE OR REPLACE FUNCTION msar.cast_to_character_varying(oid)
 RETURNS character varying AS $$
   SELECT $1::text;
@@ -1230,6 +1240,16 @@ RETURNS character AS $$
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
 
 CREATE OR REPLACE FUNCTION msar.cast_to_character(int8range)
+RETURNS character AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_character(integer)
+RETURNS character AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_character(smallint)
 RETURNS character AS $$
   SELECT $1::text;
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
@@ -1392,6 +1412,16 @@ RETURNS "char" AS $$
   SELECT $1::text;
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
 
+CREATE OR REPLACE FUNCTION msar.cast_to__double_quote_char_double_quote_(integer)
+RETURNS "char" AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to__double_quote_char_double_quote_(smallint)
+RETURNS "char" AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
 CREATE OR REPLACE FUNCTION msar.cast_to__double_quote_char_double_quote_(oid)
 RETURNS "char" AS $$
   SELECT $1::text;
@@ -1550,6 +1580,16 @@ RETURNS text AS $$
   SELECT $1::text;
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
 
+CREATE OR REPLACE FUNCTION msar.cast_to_text(integer)
+RETURNS text AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_text(smallint)
+RETURNS text AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
 CREATE OR REPLACE FUNCTION msar.cast_to_text(oid)
 RETURNS text AS $$
   SELECT $1::text;
@@ -1704,6 +1744,16 @@ RETURNS name AS $$
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
 
 CREATE OR REPLACE FUNCTION msar.cast_to_name(int8range)
+RETURNS name AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_name(integer)
+RETURNS name AS $$
+  SELECT $1::text;
+$$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
+
+CREATE OR REPLACE FUNCTION msar.cast_to_name(smallint)
 RETURNS name AS $$
   SELECT $1::text;
 $$ LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
--- a/mathesar_ui/src/component-library/common/utils/focusUtils.ts
+++ b/mathesar_ui/src/component-library/common/utils/focusUtils.ts
@@ -85,7 +85,10 @@ export function getFirstFocusableAncestor(
   element: Element,
 ): Element | undefined {
   if (getElementFocusCapability(element).focusable) return element;
-  const ancestor = element.closest(potentiallyFocusableElementsSelector);
+  if (!element.parentElement) return undefined;
+  const ancestor = element.parentElement.closest(
+    potentiallyFocusableElementsSelector,
+  );
   if (!ancestor) return undefined;
   return getFirstFocusableAncestor(ancestor);
 }


zackkrida (approved):
Works great.

mathesar/views/download_link.py
```diff
@@ -10,7 +15,37 @@
...
+                        if column_info and column_info.get("metadata") and column_info["metadata"].get("file_backend"):
+                            return f(request, *args, **kwargs)
+                except Form.DoesNotExist:
+                    pass
+
+            if request.headers.get("Content-Type", "").startswith("application/json") or response_type == "json":
+                return JsonResponse({"detail": "Unauthorized"}, status=401)
+
+            return HttpResponse("Unauthorized", status=401)
+        return wrapped
```
mathemancer: This decorator degrades functionality for non-logged-in users who now, instead of redirecting to the login URL to login and continue with their work, get a 401. It also has too much logic. Generally, the code here is also a little overwrought. For example, it has a vacuous `if` statement right before a `try/except` block that could just be skipped (with, if I'm reading correctly, one more exception caught).

For the former problem, see https://docs.djangoproject.com/en/4.2/_modules/django/contrib/auth/decorators/ for an example of how this should look.

For the latter problem, try to avoid nesting, and avoid validation checks when an exception will do.

mathesar/utils/forms.py
```diff
@@ -61,6 +61,33 @@ def get_field_tab_col_info_map(form_model):
     return tab_col_info_map
 
 
+def get_col_info_for_field(form_model, field_key):
```
mathemancer: Now that I think about it, why not wrap up the handling of what to do when the model or key are missing or wrong here? Just return an exception if anything goes wrong, and catch _that_ in the caller.

mathesar/utils/forms.py
```diff
@@ -61,6 +61,33 @@ def get_field_tab_col_info_map(form_model):
...
+    form_field = form_model.fields.get(key=field_key)
+    if not form_field.column_attnum:
+        return None
+
+    table_oid = form_field.parent_field.related_table_oid if form_field.parent_field else form_model.base_table_oid
+    column_attnum = form_field.column_attnum
+    table_oid_attnum_map = {table_oid: [column_attnum]}
+
+    with form_model.connection as conn:
```
mathemancer: Why call the DB here? Isn't the relevant part just in the Column Metadata? This is a performance issue. We should absolutely not be adding calls to the user DB in an auth decorator unless it's crucial.

mathesar/views/download_link.py
pavish: > degrades functionality for non-logged-in users who now, instead of redirecting to the login URL to login and continue with their work, get a 401

This was something I considered while implementing the 401 error message. The download_links depend on the user's session. These links are already invalid if the session has expired and the user has to login. So, redirecting to the login page made less sense.

I'm not strongly opinionated on this though. I'm fine with implementing the redirect.

> It also has too much logic. Generally, the code here is also a little overwrought. For example, it has a vacuous if statement right before a try/except block that could just be skipped

I disagree. The `if` block and other code in place is mainly to improve readability. It's clear to me just looking at the code that `form_token` and `form_field_key` are optional, and not entirely essential to the endpoint. Moving it into the exception makes that less clear.

mathesar/utils/forms.py
pavish: This was an additional check to ensure that the underlying column is a valid column.

I agree that its something we can probably remove.

FWIW, my reasoning was an edge case scenario:
- We do not modify forms when columns are removed. And any form rendering/submission failure is handled lazily. We also do not remove metadata when columns are removed outside of Mathesar.
- If a column is no longer present and the form is not manually updated, the form would still allow file uploads without this check.

pavish:
@mathemancer and I had a call to discuss the PR. Here's what we agreed upon:
- The 401 error message is kept as-is since the endpoints themselves are temporary in nature ([Source comment](https://github.com/mathesar-foundation/mathesar/pull/4848#discussion_r2413364515)) and they're not a traditional html page.
- The call to the user db to check if a column is valid can be removed. We're okay with removing this small check with little benefit, in favour of better performance ([Source comment](https://github.com/mathesar-foundation/mathesar/pull/4848#discussion_r2413379259)).
- We'll add a separate configuration in the yaml to disable public file uploads. The size check will also be moved to the yaml.
- The decorator needs to be simplified.

> It shouldn't be required for Mathesar to interpret a column as a file column for files to be accepted through a form. These should be detached.
> I don't think the backend should be attached to the underlying column, but rather the Form Field. 

We're deferring these discussions to later until we detach the underlying mechanism of identifying a UI type for a column. These are dependent on decisions and implementation already made for the files feature, and we do not want to revisit them until it becomes a priority.

zackkrida:
> We'll add a separate configuration in the yaml to disable public file uploads.

@pavish how will this work? Will it prevent users in the frontend from adding file fields to forms, with a descriptive message (something like "File uploads via form have been disabled by an administrator"), or will it just cause a backend error?

pavish:
@zackkrida 

> how will this work? Will it prevent users in the frontend from adding file fields to forms, with a descriptive message (something like "File uploads via form have been disabled by an administrator"), or will it just cause a backend error?

We'll show a message on the frontend when the user attempts to add a file type field to the form.

@mathemancer wanted the default setting for files in forms to be disabled. I would like the default setting to enable it. This does not affect code changes, but its a product level preference which I'd like you and @kgodey to weigh in. I think we can discuss this in our engineering call.

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+
+    `unauthorized_response` can be:
+        - `redirect_to_login` (default)
+        - `http_status`
+        - `json`
+    """
+    def decorator(view_func: Callable) -> Callable:
+        @wraps(view_func)
+        def wrapped(request: HttpRequest, *args: Any, **kwargs: Any):
+            if guard(request):
```
mathemancer: You need to use `is True` in situations like this. I think you're trying to lean on type annotations to guarantee this is a boolean, but it doesn't work that way in Python.

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+
+    if response_mode == "http_status":
+        return HttpResponse("Unauthorized", status=401)
+
+    return redirect_to_login(request.get_full_path())
+
+
+def user_is_logged_in(request: HttpRequest) -> bool:
+    """Checks if user is authenticated"""
+    user = getattr(request, "user", None)
```
mathemancer: Unless I'm missing something, this is overcomplicated. I don't think you can get a request without a user (perhaps `AnonymousUser`). Did you actually run into a problem that prompted this check?

For example, the [Django login_required decorator](https://docs.djangoproject.com/en/2.0/_modules/django/contrib/auth/decorators/) doesn't seem to perform this check. It just checks `is_authenticated`.

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+    return bool(user and user.is_authenticated)
```
mathemancer: Even if you're going to do this, you need to just use `user is not None` here.

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+    return _get_publicly_shared_form_from_request(request) is not None
+
+
+def user_has_file_backend_access(request: HttpRequest) -> bool:
+    if user_is_logged_in(request):
+        return True
+
+    if has_shared_form(request):
+        public_form_conf_for_file_backend = get_public_form_conf_for_file_backend()
+        return bool(public_form_conf_for_file_backend.get("enabled", False))
```
mathemancer: All this wrapping in type casts is a code smell in Python.

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+
+def _get_publicly_shared_form_from_request(request: HttpRequest) -> Optional["Form"]:
+    """
+    Returns a cached instance of the shared form, or retrieves it and caches
+    it in the request object.
+
+    The form is identified via the request query param `form_token`, and is
+    only cached/returned if it is valid and shared publicly.
+    """
+    SHARED_FORM_CACHE_KEY = "_shared_form_cached"
```
mathemancer: Why the ad-hoc caching?

mathesar/auth.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # `None` is a valid value here, we're only validating if the attr is present.
+    if hasattr(request, SHARED_FORM_CACHE_KEY):
```
mathemancer: You are heavily overusing these `-attr` functions. Please find a different way.

mathesar/auth.py
pavish: I'd like each of the guards/utilities to be able to work independently, however, I'd also like them to cache data that's used by several of them in the same request lifecycle, to avoid duplicate DB calls.

Eg., The `upload_file` request uses multiple guards and utilities:
- Guards:
  - `user_has_file_backend_access`,
  - `shared_form_field_column_has_file_backend`.
- Utilities:
  - `has_shared_form` - used to identify the max upload sizes for anonymous file uploads (Note that this can also be used as a guard).

All these functions need access to the form, but we don't want each of them to make duplicate calls to the DB to fetch the form, when used in the same request.

**Approach used to handle dependencies to make composability possible:**

Each of these guards can call other guards or utilities. The goal is to keep these components composable, so that they can be used freely for different requests without requiring us to manually specify all their dependencies at higher layers.

In the above example:
- `user_has_file_backend_access`: Uses `has_shared_form` internally to check if the request has a valid shared form. 
- `shared_form_field_column_has_file_backend` also needs to fetch the form from the DB to make its check.
- `has_shared_form` needs to fetch the form from the DB to make its check.

**How the ad-hoc caching works here:**

The ad-hoc caching is done in a single function `_get_publicly_shared_form_from_request` used by all these guards.
- The cache key is not exposed, and only that function is responsible for maintaining it.
- It checks if the form is already fetched at some point during the same request.
- If cached, returns the stored form, even if it is `None`, since a DB call could be made where there might not have been a form for the token provided.
- If a request was not made and the cache misses, it retrieves the form via a DB call and caches it.

**Why not fetch the form first?**

A different design could be to fetch the form at the start of the request and pass it to all guards and utilities. It was what was used previously in the simpler method. The downside of that approach is that it breaks composability.

Each request would need its own custom logic to prepare and pass around data for all relevant checks, effectively creating separate “auth-check” functions for every request.

The current design lets guards and utilities handle their own logic independently and makes them composable per request, and uses caching to avoid duplicate DB requests.

mathesar/auth.py
pavish: What is the harm in using `-attr` functions?

For my usecase, I want to make a DB call and cache the result in the request object. Note that the DB call can yield `None`, and its a valid value. 

My check is to see if the key is present in the request and return the value of the key, and if its not present, make the DB call.

I think the `attr` functions are the best suited ones for this usecase.

mathesar/auth.py
pavish: Handled in https://github.com/mathesar-foundation/mathesar/pull/4848/commits/9de3087e99731ed8d59e56454d4e6a06a9defb6b

mathesar/auth.py
pavish: Handled in https://github.com/mathesar-foundation/mathesar/pull/4848/commits/9de3087e99731ed8d59e56454d4e6a06a9defb6b

mathesar/auth.py
pavish: Handled in https://github.com/mathesar-foundation/mathesar/pull/4848/commits/9de3087e99731ed8d59e56454d4e6a06a9defb6b

mathesar/auth.py
pavish: Handled in https://github.com/mathesar-foundation/mathesar/pull/4848/commits/9de3087e99731ed8d59e56454d4e6a06a9defb6b

pavish:
@mathemancer I've removed the types and handled most of the comments in https://github.com/mathesar-foundation/mathesar/pull/4848/commits/9de3087e99731ed8d59e56454d4e6a06a9defb6b.

I added types to the Guards as a way to document how this was supposed to work, and less about introducing it as a pattern. I see your point in not having them.

I'm still not sure about [this comment](https://github.com/mathesar-foundation/mathesar/pull/4848#discussion_r2428122125) on not using the `-attr` functions, I think their usecase here is quite valid. I'd like to hear your opinions/suggestions on that.

mathesar/auth.py
mathemancer: I see. In that case, just memoize `get_publicly_shared_form_from_request` using `functools.cache`. This has the side-benefit of avoiding the whole `-attr` discussion.

pavish:
@mathemancer This comment is mainly to share differing point of views. I'm still fine with making the changes you suggested, however, I thought we ought to discuss this once.

> there's no point in checking if it has an attribute before getting it. Just get it, and handle the case where it's not there. Further, in Python it's generally preferred (certainly by me) to just use try/except as control flow in cases like this. Try, catch an AttributeError, and use that for branching. This avoids the hasattr and getattr

This is exactly what `hasattr` does internally. Here is the [python doc](https://docs.python.org/3/library/functions.html#hasattr). Why do that again when we could use the utility?

Also, `hasattr` being a problem was because of python 2 capturing all errors, and cases of it being misused where None was not an expected value. It's generally accepted that its not a concern with python 3, and we explicitly want None to be accepted as a valid value.

> As for setattr, in this case, you're setting an attribute of a mutable 3rd-party object (the HTTPRequest instance), which has the potential to override its out-of-the-box behavior, causing unexpected behavior anywhere else it's used (i.e., everywhere). I recognize that you have an attribute name that (hopefully) won't clash with anything, but that pattern is kind of a red flag to me.

This is the pattern used by Django and most middlewares. For eg.,
- The `user` attribute is added by the `AuthenticationMiddleware`
- The `session` attribute is added by the `SessionMiddleware`.

> I see. In that case, just memoize get_publicly_shared_form_from_request using functools.cache. This has the side-benefit of avoiding the whole -attr discussion.

I've only used memoization for pure functions (and primarily for rendering stuff on the frontend). I can see how it'd solve this usecase, but I'm more skeptical of using it as a pattern since the function here makes a DB call. It shouldn't be a concern if we continue to use `request` as the argument for the memoized function, but if someone refactors this code down the line and passes down the form token as the argument, this will lead to pesky bugs.

pavish:
@mathemancer This commit https://github.com/mathesar-foundation/mathesar/pull/4848/commits/061ec2b1ca69794354e1827a7cb475a0d2dc22de should resolve your concerns.

For others following along, @mathemancer and I had a call where we discussed this PR and decided to go with the memoization approach with some comments added. Eventually, when we have more authentication logic with other shared entities, we might end up implementing a custom middleware when we'd revisit this discussion.

PR #4854, opened 2025-10-07
Export explorations

### LINKED ISSUE #4619: Export from data explorer
>  _This is a meta issue tracking the implementation of this project._

## Outcome

Currently, Mathesar's UI only supports exporting data from tables, which limits utility. Users should also be able to export data from explorations. 

For our MVP, we'd like to keep this as simple as possible. 

## Related

- Multiple requests from users for this via user feedback form.
- Prior product work (private document) [Exporting Tables & Explorations](https://hackmd.io/FgFY_MZvRw6Og4QYyjNjCQ)
- https://github.com/mathesar-foundation/mathesar/discussions/2268

Anish9901:
Fixes #4619

Adds the ability to export saved explorations along with applied transformations as a CSV file.

The "Export" button on the data explorer page has the following hover text and state:
- When there are no columns in an unsaved exploration(button disabled and no help text):

    Screenshot 2025-10-08 at 1 12 35 AM

- When the exploration is unsaved(button disabled with help text):
 
    Screenshot 2025-10-08 at 12 41 00 AM

   
- When the exploration is saved(button enabled with help text):
     
   Screenshot 2025-10-08 at 12 33 35 AM

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/i18n/languages/en/dict.json
+++ b/mathesar_ui/src/i18n/languages/en/dict.json
@@ -272,7 +272,7 @@
   "explore_your_data": "Explore your Data",
   "exploring_from": "Exploring from",
   "export": "Export",
-  "export_exploration_50K_limit": "50K Export limit: Only the first 50,000 rows will be exported.",
+  "export_exploration_50K_limit": "Exports are limited to 50,000 rows. Since your dataset is larger, only the first 50,000 rows will be exported.",
   "export_exploration_as_csv_help": "Export the {explorationName} exploration as a CSV file. Your current transformations will be applied to the exported data.",
   "export_exploration_save_help": "Exploration has unsaved changes. Please save the changes before exporting.",
   "export_table_as_csv_help": "Export the {tableName} table as a CSV file. Your current filters and sorting will be applied to the exported data.",


mathesar/utils/explorations.py
```diff
@@ -154,6 +154,25 @@ def run_saved_exploration(exp_model, limit, offset, conn):
...
+    exp_model = get_exploration(exploration_id)
+    exp_results = run_saved_exploration(exp_model, limit, offset, conn)
+    yield exp_results["output_columns"]
+    records = exp_results["records"]
+    if records["count"] <= batch_size:
+        yield records["results"]
+    else:
+        chunks = records["count"] // batch_size
+        for i in range(chunks + 1):
+            yield records["results"][i * batch_size:(i + 1) * batch_size]
```
mathemancer:
```suggestion
    for i in range(0, records["count"], batch_size):
        yield records["results"][i:i + batch_size]

```

mathesar_ui/src/component-library/anchorButton/AnchorButton.svelte
```diff
@@ -28,6 +31,8 @@
   ].join(' ');
 </script>
 
-<a {...$$restProps} {href} class={allClasses}>
-  <slot />
+<a {...$$restProps} {href}>
+  <Button class={allClasses} {disabled}>
```
seancolsen: I just glanced at this PR. I expect @pavish will do a full review. But I wanted to point this out...

`<a>` elements cannot contain `<button>` elements.

[HTML standard](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) for the `<a>` element:

> there must be no interactive content descendant

I'm guessing you added the `<Button>` component in order to achieve the disabled styling/behavior. I would consider hiding the button instead. If we really want to continue displaying the button we'll need to make some other improvements to the `AnchorButton` component to support a `disabled` prop. (We'd need to remove the `href` value and implement some styling changes.) Hiding it is a tad easier.

Anish9901: I see, I'm fine with hiding the button, but, do we hide buttons elsewhere in our app?

Also, out of curiosity, what would be the implication of not conforming to the spec in this case? Do we risk compatibility issues with browsers or something else?

Edit: I could also switch the `AnchorButton` with `Button` when it needs to be disabled, if you prefer.

seancolsen: > do we hide buttons elsewhere in our app?

Yes. In fact, if you look [here](https://github.com/mathesar-foundation/mathesar/blob/19588f9da756022fc3b92c27504ac405e967e971/mathesar_ui/src/systems/table-view/actions-pane/ActionsPane.svelte#L75-L101), you'll see how this very same "Export" button is hidden on the table page under certain circumstances. Here's [another example](https://github.com/mathesar-foundation/mathesar/blob/19588f9da756022fc3b92c27504ac405e967e971/mathesar_ui/src/systems/table-view/table-inspector/column/ColumnActions.svelte#L76-L94).


> what would be the implication of not conforming to the spec

Yeah, it's a good question. I don't think I'm really going to have a satisfactory answer though. But I'll try. In the abstract sense... My understanding is that there is a substantial amount of behavior resting on the assumption that interactive elements are not nested. That "behavior" can exist at the browser level (e.g Firefox), or the framework level (e.g. Svelte), or the application level (e.g. Mathesar) — and even at other more-esoteric levels (e.g. assistive screen-reading technology like Jaws). So by invalidating that nesting assumption we're entering into _undefined_ behavior. Exactly _what_ undefined behavior would there be? How bad would it be, practically speaking? It really depends on the situation. In some cases nobody might ever notice. And in others, some functionality could be broken in subtle ways. What might be broken _in this case_? It's hard for me to tell immediately without spending some time to poke at it. But in my experience, these nesting rules are important to follow, especially for interactive elements.

I had a maddening bug at my last job due to the way that Vue/Nuxt handled invalid nesting in a slightly inconsistent manner across server-side rendering vs client-side. It took me _so_ many hours of troubleshooting to hone in on it. But the culprit was that we had a `<div>` inside a `<label>`. You'd think that would be allowed. It didn't _appear_ to break anything at first glance. But that nesting is indeed invalid, and it violated some invariant within Nuxt's SSR system. I changed the `<div>` to a `<span>` and the problem went away.

MDN has a good page on [Content Categories](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories). Generally the nesting rules are defined in terms of these content categories. So if you go to the page for `<a>` and look at the [Technical Summary](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#technical_summary) you can see that the "Permitted content" references "Transparent" content and "Interactive" content.

It's also worth mentioning that in a component-based framework like Svelte we seem to lack robust strategies for ensuring that HTML nesting is valid. I've gone looking for tooling that would statically validate element nesting, and I've not found it! So the responsibility rests on the engineers. That's hard though because components are opaque by design. If I use a `<Button>` component, I can pretty well guess that it'll contain a `<button>` element. What elements do you think `<NameWithIcon>` would contain? That's much less clear. Such is why I tend to use `<span>` in lower-level components instead of `<div>`. Otherwise, nesting `<NameWithIcon>` inside `<Label>` would be invalid, technically speaking.

mathesar/utils/explorations.py
```diff
@@ -154,6 +154,23 @@ def run_saved_exploration(exp_model, limit, offset, conn):
...
 
 
+def exploration_chunker(
+    conn,
+    exploration_id,
+    limit=None,
+    offset=None,
+    batch_size=2000
+):
+    limit = limit or 50000  # if no limit is provided we cap it to 50000
```
mathemancer:
```suggestion
    limit = limit or min(limit, 50000)  # We cap limit at 50000
```

mathesar_ui/src/component-library/anchorButton/AnchorButton.svelte
pavish: I agree with @seancolsen on this.

In addition to his comment, the `AnchorButton` is a low level component used throughout the app and its behaviour needs to be clear. It shouldn't have any other interactive elements. 

I assume you added the `Button` there to have a `disabled` behaviour, but you can note that it doesn't actually disable the anchor link, it only makes the button disabled.
- It still provides the link when right clicking on the anchor button
- The anchor link is still focusable
- Whether the link would work while clicking/keyboard-event is undefined behaviour and we cannot rely upon it, even though most browsers disable it in this scenario.

> I could also switch the AnchorButton with Button when it needs to be disabled, if you prefer

@Anish9901 I would be satisfied with this approach.

mathesar/utils/explorations.py
Anish9901: @mathemancer  `limit = limit or min(limit, 50000)` doesn't work when `limit=None` as `min(None, 50000)` is invalid, I changed it to `min(limit or 50000, 50000)` in 290f5d9.

Anish9901:
@pavish I've added the changes for:
- 50k limit within the tooltip 
- Using a `Button` instead of `AnchorButton` while disabling the "Export" button.

in this commit --> 4eb9a9b

Issue #4736, opened 2025-09-03
Misleading feedback when query is blocked by RLS


commonspider:
## Description

I have discovered this bug while deleting a row, but I believe that it could happen with other kinds of queries as well.
When a user tries to delete a row, and their action is blocked by row level security, Mathesar shows a "1 record deleted successfully!" green popup regardless of the actual outcome of the query (which is that no rows have been deleted).

## Expected behavior

A red popup (or something similar) should appear instead, notifying the user that their action wasn't successful.

## To Reproduce

- Create a table `table_test` and a user `user_test` with read and write permissions on it
- Populate `table_test` with a few rows
- Execute the following queries:
```sql
ALTER TABLE table_test ENABLE ROW LEVEL SECURITY;
CREATE POLICY allow_read ON table_test FOR SELECT USING ( true );
```
- Log in as `user_test`, select a record, click "Delete Record" and confirm
- The green popup will appear but the record will still be there (as it should be)

## Environment
 - Mathesar version: 0.4.0
 - OS: Debian 12 (bookworm)
 - Browser: Firefox
 - Browser Version: 141.0.3 (64-bit)

commonspider:
May I be assigned to this issue?

zackkrida:
@commonspider, sure, you can work on this now! Please default to just surfacing up the actual DB-level error into the UI; during PR review we can decide if we want to use a more tailored, user-friendly error message instead.

commonspider:
Thank you @zackkrida 
Unfortunately RLS raises a db error only on `INSERT`, and the user can already see this error. The queries `DELETE` and `UPDATE` fail silently. I will open 2 PRs, one for each kind of query, with a possible solution.

zackkrida:
ah I didn't realize that about update, in particular, @commonspider

PR #4859, opened 2025-10-09
More accurate user feedback regarding row deletion

commonspider:
Related to: #4736 

Gives a more accurate feedback to the user about which rows have failed to be deleted. For example, Mathesar  currently says that all rows have been deleted even when RLS blocks the deletion of some of them.

**Old Feedback**

https://github.com/user-attachments/assets/e43a4d4a-e06f-4024-948b-35e7bebe46be

**New Feedback**

https://github.com/user-attachments/assets/beee72a4-6754-4c8e-8097-6c21df2acd35

**Questions for the reviewer**
- Can I safely change the return type of `msar.delete_records_from_table`?
- Is the edit of the description of `records.delete` acceptable?
- Which error message should be shown instead of "Unable to delete row ${rowId}" in  `RecordsData.deleteSelected`?

**Technical details**
- `msar.delete_records_from_table` returns a `jsonb` instead of an `integer`
- `RecordsData.deleteSelected` populates `rowsSuccessfullyDeleted` and `rowsFailedToDelete` according to the results from the database
- SQL tests have been updated
- Description of RPC function `records.delete` has been updated

**How to test this PR**
- Create `table_test` and give write access to `user_test`
- Add a boolean column named `allow_delete`
- Run the following queries:
```sql
ALTER TABLE table_test ENABLE ROW LEVEL SECURITY;
CREATE POLICY allow_read ON table_test FOR SELECT USING ( true );
CREATE POLICY allow_delete ON table_test FOR DELETE USING ( allow_delete );
```
- Create rows as admin and set which ones are allowed to be deleted
- Delete rows as `user_test`

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
Thanks for this PR @commonspider! Neat find!

I've reviewed the front end changes here. I added a few commits.


- In 8d41dc1182bfccb2ebdd92303e6abf478e311c8c I changed an Array to a Set for faster repeated lookup when we're deleting lots of rows at once.


- In f10477e04590b6c6fcd7bff29be07b3112583ddf I added a small code comment explaining why were handling this case, since the idea of RLS might not occur to future devs.

- In 1f163a0986d0dd989dab46f42b790688a89078ad I improved the toast error message.

    - Before:

        > Unable to delete row. Error: Unable to delete row 2

    - After:

        > Unable to delete row. Error: Row 2 was silently ignored by PostgreSQL.

---

The front end changes are good-to-go from my perspective. But I think @mathemancer ought to review the backend changes and the overall approach. In particular, I kind of wonder whether the API should be responsible for actually returning an error in this case. My understanding is that this case is indistinguishable from a situation where we're deleting a row that has already been deleted by another user. That's tricky for us. My gut would be to have the API return an error and do all the row deleting in one transaction that either works 100% or fails without any changes explaining why. But I don't have a strong opinion on this. This PR is definitely an improvement over the current behavior! So I don't want to hold it up for the sake of perfection.

mathemancer (approved):
Backend code looks good, and I agree with @seancolsen that this is an improvement over the current state, so we should merge it.

I'm going to create an issue to work through some of the concerns @seancolsen noted in his comments.

Issue #4618, opened 2025-07-07
Import into existing tables


kgodey:
>  _This is a meta issue tracking the implementation of this project._

## Outcome

Mathesar only supports importing data when creating new tables. Users should be able to upload data into existing tables as well.

More detail:
- Requirements TBD
- Project proposal TBD

## Tasks
*Issues to be created soon.*

mathemancer:
Completed in #4870 and #4866

PR #4866, opened 2025-10-14
UI for importing into existing tables

seancolsen:
image

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #4870, opened 2025-10-15
Implement Importing into existing tables

Anish9901:
Related to #4618

https://github.com/user-attachments/assets/d8db4708-ab7d-4f1a-8929-e08fff086246

To test this PR:
```py
import requests

s = requests.Session()

cookie = {'Cookie': 'display_language=en; csrftoken=<your_csrftoken>; sessionid=<your_sessionid>;',
          'X-csrftoken': '<your_csrftoken>'}

s.cookies.update(cookie)

headers = {"X-CSRFToken": s.cookies.get("X-csrftoken"), "Referer": "http://localhost:8000/"}


payload = {
 'header': True,
 'target_table_oid': 929665,
 'database_id': 2,
 'mappings': '[{"csv_column": {"index": 1, "name": "col_1"}, "table_column": 2}, {"csv_column": {"index": 3, "name": "col_3"}, "table_column": 4} , {...}]'}

with open("your_file.csv", "rb") as f:
     files = {"file": ("your_file.csv", f, "text/csv"),}
     response = s.post('http://localhost:8000/bulk_insert/', data=data, files=files, headers=headers)
```
## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/imports/datafile.py
+++ b/mathesar/imports/datafile.py
@@ -78,7 +78,11 @@ def _process_column_names(column_names):
 
 
 def insert_into_existing_table(user, data_file_id, target_table_oid, mappings, conn):
-    header_to_validate = sorted([(i["csv_column"]["index"], i["csv_column"]["name"]) for i in mappings], key=lambda x: x[0])
+    header_to_validate = sorted([
+        (
+            i["csv_column"]["index"], i["csv_column"].get("name")
+        ) for i in mappings
+    ], key=lambda x: x[0])  # sometimes we don't have "name" when there is no header.
     temp_table = copy_datafile_to_table(
         user,
         data_file_id,


Anish9901:
@seancolsen Please let me know if this meets the frontend's requirements.

mathesar/imports/datafile.py
```diff
@@ -3,12 +3,13 @@
...
 from db.identifiers import truncate_if_necessary
 from db.tables import create_and_import_from_rows
+from db.records import insert_from_select
 
 from mathesar.models.base import DataFile
 
 
 def copy_datafile_to_table(
-        user, data_file_id, table_name, schema_oid, conn, comment=None
+        user, data_file_id, table_name, schema_oid, conn, comment=None, import_into_temp_table=False, header_to_validate=[]
```
mathemancer: These would be better on separate lines. Plus, that will make future diffs easier to validate.

Anish9901:
> SQL looks good, but I'm curious why you went with objects rather than arrays for the src -> dest mappings.

I wanted to use the mappings to create a table in SQL with `jsonb_to_recordset` hence the use of objects instead of arrays for mappings.

> I see that you've used more DRF code for this PR

Refactored to regular views instead.

mathemancer (approved):
I think this looks pretty good. FE should verify that the error message and API setup generally meet expectations.

seancolsen:
@Anish9901 I'm encountering some difficulty working with this API, and I'd like your help.

Here are the steps to reproduce:

1. Use my code from #4866

1. Create a table from scratch. Add one text column named "foo".

1. Create a CSV from scratch using the following content:

    ```
    foo
    bar
    ```

1. Use the UI to import the CSV into the table.

1. Observe a request like this:

    ```
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f
    Content-Disposition: form-data; name="header"

    true
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f
    Content-Disposition: form-data; name="database_id"

    1
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f
    Content-Disposition: form-data; name="target_table_oid"

    110269
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f
    Content-Disposition: form-data; name="file"; filename="foobar.csv"
    Content-Type: text/csv

    foo
    bar
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f
    Content-Disposition: form-data; name="mappings"

    [{"csv_column":{"index":0,"name":"foo"},"table_column":2}]
    ------geckoformboundary4cca7277a9f2d50294844310227ae09f--
    ```

1. Expect the request to succeed.

1. Instead, observe a 400 error with the following response

    ```json
    {"errors": "Unable to import data into existing table. Parsing mismatch"}
    ```

Honestly I'm not sure if something needs to change on the front end or the back end. But the request seems right to me, so my next step in troubleshooting this would be to dig into the backend code to understand _why_ this is failing. I spent a little time with this, but my work day is drawing to a close, so I want to try bumping this over to you. Can you try troubleshooting this?

Anish9901:
Oh, I had made it such that the backend expected the csv columns to be indexed starting from 1 not 0. I've pushed a fix, should be good now @seancolsen.

seancolsen:
Ok, now I'm getting a different error with the same repro steps:

```
Unable to import data into existing table. INSERT has more target columns than expressions
LINE 1: INSERT INTO public.foobar(foo) SELECT FROM pg_temp_71.fooba...
 ^
QUERY: INSERT INTO public.foobar(foo) SELECT FROM pg_temp_71.foobar
CONTEXT: PL/pgSQL function msar.insert_from_select(regclass,regclass,jsonb) line 45 at EXECUTE
```

Anish9901:
Ok, I remember the reason for 1-based indexing now, the attnums of the temp table that is created based on the csv columns starts from 1 instead of 0. I'm handling the handling the mismatch b/w csv and temp attnum indexing by adding 1 to the csv indicies before sending it to the SQL function.

seancolsen:
@Anish9901 for the happy path, this is working well now!

However, in the case where the CSV doesn't have a header row, I'm getting this error:

```json
{"errors": "Unable to import data into existing table. 'name'"}
```

- As a side note: since this API is plain HTTP and not JSON-RPC, I think it would be better to return an error response body as plain text, not JSON. So I'd rather see the response body be:

    ```
    Unable to import data into existing table. 'name'
    ```

- But as for the cause of this error in particular, I do wonder exactly what you expect to receive for the `header` field in the case where the CSV doesn't have a header row. Currently I'm sending an empty string because it appeared to me that the backend is just checking the truthiness of the `header` value. So, for example, I didn't imagine that sending `false`, would work because (as a string) `false` is not truthy, and all form fields are strings at the HTTP level. If you want to fiddle with the value that the front end is sending, you can do it [here](https://github.com/mathesar-foundation/mathesar/blob/910db27aa7470a21e847c7f97b5ad3f132e4c9c3/mathesar_ui/src/api/rest/bulkInsert.ts#L27).

seancolsen (approved):
Looks good now! Thanks @Anish9901!

zackkrida:
@seancolsen I'll fix the merge issue here, I was testing branch protections (here https://github.com/mathesar-foundation/mathesar/issues/4876#issuecomment-3415106131) and I think it caused the issue.

PR #4874, opened 2025-10-16
Add PG 18 to CI

Anish9901:
Adds PG 18 to CI

#### Context about test failures:

Previously `pg_constraint` only used to store not null entry for domains, Starting from v18, this not null entry is also stored for table columns.

#### Before version 18
Screenshot 2025-10-17 at 1 50 53 AM

```
mathesar=# SELECT oid, conname, contype FROM pg_constraint where conrelid='add_pkeytest'::regclass::oid AND conkey='{1}'; 

   oid   |  conname   | contype 
---------+------------+---------
 1262157 | mysuperkey | p
(1 row)
```

#### Version 18
Screenshot 2025-10-17 at 1 51 08 AM

```
postgres=# SELECT oid, conname, contype FROM pg_constraint where conrelid='add_pkeytest'::regclass::oid AND conkey='{1}';

  oid  |          conname           | contype 
-------+----------------------------+---------
 16987 | add_pkeytest_col1_not_null | n
 16992 | mysuperkey                 | p
(2 rows)

```

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ frozendict==2.4.6
 fsspec==2025.7.0
 gunicorn[gevent]==23.0.0
 pillow==11.3.0
-psycopg[binary]==3.2.6
+psycopg[binary]==3.2.10
 psycopg2-binary==2.9.10
 pyyaml==6.0.2
 requests==2.32.4


db/sql/05_msar.sql
```diff
@@ -2906,13 +2906,15 @@ Args:
...
+  added_contype char[];
 BEGIN
   con_create_defs := __msar.process_con_def_jsonb(tab_id, con_defs);
   PERFORM __msar.add_constraints(
     __msar.get_qualified_relation_name(tab_id),
     variadic con_create_defs
   );
-  RETURN array_agg(oid) FROM pg_constraint WHERE conrelid=tab_id;
+  added_contype := array_agg(x->>'type') FROM jsonb_array_elements(con_defs) AS x;
+  RETURN array_agg(oid) FROM pg_constraint WHERE conrelid=tab_id AND contype=ANY(added_contype);
```
mathemancer: This changes the semantics of the return value, which changes the response at the API. We used to return all constraints for the table via the API, but this now only returns constraints which happen to have one of the same constraint types as one of the added constraints. I don't think that change makes sense.

mathemancer (changes requested):
I think you should change the return value of the constraint adder to be the same as before, unless there's a reason I'm not understanding to make the change.

db/sql/05_msar.sql
Anish9901: Reverted, I thought we were returning only the constraints we added.

PR #4880, opened 2025-10-17
Use `NULL` instead of 'id' for columns named 'id'

### LINKED ISSUE #4878: Column named "id" causes problems during CSV import
## Steps to reproduce

1. Create a table from scratch in Mathesar. In addition to the id column, give the table a text column named `foo`.

1. Import the following CSV with "header" set to "true"

    ```
    foo
    bar
    ```

    This works ✅

1. In the CSV, change the column name from `foo` to `id`, and attempt to import it again.

    ```
    id
    bar
    ```

    This fails ❌

    ```
    Unable to import data into existing table. syntax error at or near \")\"
    
    LINE 1: COPY pg_temp_17.\"foobar-id\"() FROM STDIN\n ^
    ```

1. Here's another example which fails in a similar way:

    ```
    id,foo
    1,bar
    ```

    ```
    Unable to import data into existing table. extra data after last expected column
    CONTEXT: COPY foobar-id, line 1: \"1\tbar\"
    ```

Anish9901:
Fixes #4878

Using `id` as column names gets filtered out by the function `__msar.process_col_def_jsonb` which resulted in the generation of a malformed temp table and hence a malformed `COPY` statement.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/05_msar.sql
+++ b/db/sql/05_msar.sql
@@ -3386,7 +3386,9 @@ DECLARE
   col_names_sql text;
   copy_sql text;
 BEGIN
-  col_defs := jsonb_agg(jsonb_build_object('name', n)) FROM unnest(col_names) AS x(n);
+  -- Passing 'id' as column name gets filtered out by __msar.process_col_def_jsonb,
+  -- pass NULL when 'id' is encountered.
+  col_defs := jsonb_agg(jsonb_build_object('name', NULLIF(n, 'id'))) FROM unnest(col_names) AS x(n);
 
   IF NULLIF(tab_name, '') IS NOT NULL AND NOT EXISTS(
       SELECT oid FROM pg_catalog.pg_class WHERE relname = tab_name AND relpersistence = 't'


zackkrida:
@pavish this appears to be working. @seancolsen and @mathemancer, please review this though before @mathemancer cuts the release next week. Thanks!

mathemancer:
@seancolsen I agree with you in general. To double-check my understanding, this change isn't _adding_ the bugs you noted; it's just not solving them (when a more comprehensive solution would), correct?

If so, I think we should still merge this, and try to figure out what (if any) ramifications there would be to removing the concept of a "Mathesar default ID column". I've never really been a fan of that concept, but it was convenient way back when. Given our focus on working with preexisting DBs now, I think we should:

- remove any dependence (if any still exist) on the concept of a "Mathesar ID column", and
- Then remove any machinery based on that concept.

seancolsen:
@mathemancer 

> this change isn't _adding_ the bugs

Correct.

This change adds more code which follows a problematic pattern instead of fixing the pattern.

I'll merge.

Just to be clear on the "problematic pattern", this isn't something big. Your two bullet points in your comment suggest something more far reaching that what I have in mind. I'm just suggesting removing `OR col_def_obj ->> 'name' <> 'id'` [from within __msar.process_col_def_jsonb](https://github.com/mathesar-foundation/mathesar/blob/e07b39d333570ab2991a6becafca8956560057fc/db/sql/05_msar.sql#L2569) and changing the call site [inside msar.add_mathesar_table](https://github.com/mathesar-foundation/mathesar/blob/e07b39d333570ab2991a6becafca8956560057fc/db/sql/05_msar.sql#L3194) as necessary. That "add_mathesar_table" function _is_ dependent on the concept of a "Mathesar ID column", and I think that's okay. When you create a new table via Mathesar, you get a PK column named `id`. That's fine. What's not fine is the level at which the hard-coded "id" column name is implemented. The logic is present at too low a level, which is why we ended up with this CSV import bug.

PR #4931, opened 2025-10-28
issue 4890 - Allow renaming of primary key (Mathesar ID) columns

wbonicki:
Fixes #[4890]

This PR fixes an issue that previously prevented renaming a table’s primary key (PK) column.
Update: Added logic to raise an exception when attempting to rename a primary key column.

**Technical details**

One function in 05_msar.sql was changed:

`alter_columns`: when renaming a primary key columns (Mathesar ID columns) is detected an exception is raised.
One test in `test_sql_functions.sql` was added.
**Screenshots**

image

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/05_msar.sql
+++ b/db/sql/05_msar.sql
@@ -3853,7 +3853,6 @@ BEGIN
     FROM jsonb_array_elements(col_alters) AS x(col_alter_obj)
       INNER JOIN pg_catalog.pg_attribute AS pga ON pga.attnum=(x.col_alter_obj ->> 'attnum')::smallint AND pga.attrelid=tab_id
       LEFT JOIN pg_catalog.pg_attrdef AS pgat ON pgat.adnum=(x.col_alter_obj ->> 'attnum')::smallint AND pgat.adrelid=tab_id
-    -- WHERE NOT msar.is_mathesar_id_column(tab_id, (x.col_alter_obj ->> 'attnum')::integer)
   LOOP
     IF col.new_name IS NOT NULL AND msar.is_mathesar_id_column(tab_id, col.attnum) THEN
       RAISE EXCEPTION USING


mathemancer (changes requested):
First of all, thank you for your interest in this issue!

- This PR modifies the `is_mathesar_id_column` function, but that function should simply be removed if the `id` name is no longer relevant.
- This PR doesn't solve the core problem where some alterations of the primary key column (or Mathesar ID column) will fail _silently_, i.e., without throwing any error.
- In fact, this makes behavior more confusing, since some alterations are allowed, but others aren't, of Mathesar ID columns.

Figuring out what to do about modification of Primary keys is a broader product question that I'm not ready to answer at the moment.

Your options for getting this merged:

- Simply throw an error whenever _any_ modification of a Mathesar ID column is attempted, or
- Modify your current solution to throw an error if any modification other than those you've greenlit is attempted.

I prefer the former, since it's simpler and we'll need to get rid of the `mathesar_id_column` concept in the near future anyway.

wbonicki:
Okay, I got it! I will change the solution, the first proposition seems fine - throw an error whenever any modification of a Mathesar ID column is attempted

wbonicki:
I have a question about throwing an exception — did you mean a SQL exception?”

zackkrida:
@mathemancer tagging you here for follow-up on @wbonicki's [question](https://github.com/mathesar-foundation/mathesar/pull/4931#issuecomment-3511607675).

mathemancer:
> I have a question about throwing an exception — did you mean a SQL exception?”

Yes, exactly. Throwing an SQL exception will get picked up by the rest of the machinery and result in the error message for that exception being shown in the front end.

wbonicki:
First of all, thank you all for the quick responses.
I’ve implemented the suggestion — you can see the result in the attached screenshot (PR description).
I have one additional question: do we assume that the Mathesar ID column is always named "id"?
The current logic makes this assumption, but there may be cases where someone has already renamed it.
In such a scenario, it would still be possible to rename it again.
Please let me know if this should also be fixed (i.e., by removing the attname='id' condition in the is_mathesar_id_column function).

mathemancer:
> First of all, thank you all for the quick responses. I’ve implemented the suggestion — you can see the result in the attached screenshot (PR description). I have one additional question: do we assume that the Mathesar ID column is always named "id"? The current logic makes this assumption, but there may be cases where someone has already renamed it. In such a scenario, it would still be possible to rename it again. Please let me know if this should also be fixed (i.e., by removing the attname='id' condition in the is_mathesar_id_column function).

I'd rather not change that logic in this PR. If the column is not named "id", it's not a Mathesar ID column by definition. The "Mathesar ID column" concept is rather old and not really useful anymore, but I'm not sure what assumptions we make in the rest of the code base about it. We'll likely replace the concept at some point.

PR #4948, opened 2025-11-03
Add feature to select multiple cells on touch-only devices

seancolsen:
This PR fixes a limitation that I've run into when using Mathesar from a phone or tablet where I'm unable to select multiple cells.

## Before this PR

- Shift+Click is the only way to select multiple cells.
- If you don't have a physical keyboard, you can't select multiple cells.

## After this PR

1. We have a low-level utility that watches events to determine information about the user's device as follows:

    - `hasPhysicalMouse` becomes `true` whenever we see a `mousemove` event.
    - `hasPhysicalKeyboard` is also set to `true` when `hasPhysicalMouse` becomes true.
    
        I did some research trying to find a way to detect a physical keyboard, and I couldn't find a good way to detect it directly. But I think the presence of a physical mouse is a very good proxy. I chose to separate these properties to improve the readability of the logic on the consuming side and to leave the door open for a future where we might be able to find a way to directly detect the presence of a physical keyboard.
    
    - `hasTouchCapability` becomes `true` whenever we see a touch event (`pointerdown` event with `eventType` of `'touch'`).

1. On a device that has a physical keyboard and does not have touch capability, this PR provides no user-facing changes. (That way we avoid cluttering the UI with additional options the user doesn't need.)

1. If the device has no physical keyboard (i.e. we've seen a `mousemove` event) OR if the device has touch capability (i.e. we've seen a touch event), then the cell context menu gets this new entry:

    image

1. After triggering the context menu option, the sheet goes into a special cell-region-selection mode, and the user gets this visual feedback:

    Screencast_20251103_112656

1. Clicking another cell will then select a region of cells between the two. And clicking _anywhere_ in the app exits cell-region-selection mode.

## Notes

- The `deviceInfo` utility might seem like overkill here. But I have other plans for it! There are a number of small UX optimizations that I'd like to make for mobile devices, similar to this PR. One of the main motivations for this PR was really to provide a testbed for this new `deviceInfo` utility.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/packages/svelte-device-info/index.ts
+++ b/mathesar_ui/src/packages/svelte-device-info/index.ts
@@ -52,7 +52,7 @@ export const deviceInfo = {
 
   // We don't (yet?) have a good way to detect a physical keyboard, so we use
   // a physical mouse as a reasonably good proxy.
-  hasPhysicalKeyboard: readOnly(hasPhysicalMouse),
+  mayHavePhysicalKeyboard: readOnly(hasPhysicalMouse),
 
   hasTouchCapability: readOnly(hasTouchCapability),
 };
--- a/mathesar_ui/src/systems/table-view/context-menu/entries/selectCellRange.ts
+++ b/mathesar_ui/src/systems/table-view/context-menu/entries/selectCellRange.ts
@@ -6,9 +6,9 @@ import { deviceInfo } from '@mathesar/packages/svelte-device-info';
 import { buttonMenuEntry } from '@mathesar-component-library';
 
 export function* selectCellRange(p: { beginSelectingCellRange: () => void }) {
-  const hasPhysicalKeyboard = get(deviceInfo.hasPhysicalKeyboard);
+  const mayHavePhysicalKeyboard = get(deviceInfo.mayHavePhysicalKeyboard);
   const hasTouchCapability = get(deviceInfo.hasTouchCapability);
-  if (hasTouchCapability || !hasPhysicalKeyboard) {
+  if (hasTouchCapability || !mayHavePhysicalKeyboard) {
     yield buttonMenuEntry({
       icon: iconSelectMultipleCells,
       label: get(_)('select_cell_range'),


mathesar_ui/src/packages/svelte-device-info/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  const cleanupFunctions = [listenForMouseEvents(), listenForTouchEvents()];
+  return () => cleanupFunctions.forEach((f) => f());
+}
+
+export const deviceInfo = {
+  hasPhysicalMouse: readOnly(hasPhysicalMouse),
+
+  // We don't (yet?) have a good way to detect a physical keyboard, so we use
+  // a physical mouse as a reasonably good proxy.
+  hasPhysicalKeyboard: readOnly(hasPhysicalMouse),
```
pavish: I don't think identifying this is possible. I'd consider renaming this to `mayHavePhysicalKeyboard` and adding a userAgent string check along with the physical mouse check.

mathesar_ui/src/packages/svelte-device-info/index.ts
```diff
@@ -0,0 +1,58 @@
+import { type Readable, type Writable, writable } from 'svelte/store';
+
+function readOnly<T>(store: Writable<T>): Readable<T> {
+  return store;
+}
+
+const hasPhysicalMouse = writable(false);
+
+const hasTouchCapability = writable(false);
```
pavish, line 9: The default value can be set by checking if window has `ontouchstart` (and additional userAgent checks to detect mobile devices) rather than `false`.

pavish (approved):
@seancolsen This works great! I had a couple suggestions. I'll leave it upto you to decide if you want to add them to this PR (or not).

mathesar_ui/src/packages/svelte-device-info/index.ts
seancolsen: I changed the variable name in 181b727a5d33ea3e4c0c21935e191c7d469d4c65.

seancolsen:
Thanks for the suggestions, @pavish. I did rename the variable. But I'm not going to do the other ones now. I'd rather not mess with user agent detection. It just seems like more work than it's worth since this works for what we need. I also don't have a good way to test logic that checks if window has `ontouchstart` since all my devices have touch capability. The logic as-is seems to work for what we need, so I'm inclined to avoid spending more time polishing this until we need improved logic.

PR #4956, opened 2025-11-05
Fix/4895 record summary template references deleted column

### LINKED ISSUE #4895: Error when listing records if record summary template references a deleted column
## Steps to reproduce

1. Begin with the following sample data:

    ```sql
    create table issue_4895 (
      id int primary key generated always as identity,
      a text,
      b text,
      parent int
    );
    insert into issue_4895 (a, b, parent) values
    ('foo', 'bar', NULL),
    ('baz', 'bat', 1);
    alter table issue_4895 add constraint issue_4895_parent_fkey
    foreign key (parent) references issue_4895 (id);
    ```

1. Open the table page for `issue_4895`.

1. Select the "foo" FK cell, and configure the Linked Record Summary to display both `a` and `b` columns.

    Image

    Save.

1. Now delete column `b` from the table.

1. Expect the sheet to load, displaying "foo" as the record summary for the linked record. Because the "bar" value from column `b` is no longer available, it should be silently ignored.

1. Notice that when the sheet reloads, there is an error.

    Image

    Although this error is not shown to the user, you can see it in the network tab of your dev tools if you find the request to the `records.list` API:

    > UndefinedColumn: column base.........pg.dropped.3........ does not exist
    > LINE 15: || COALESCE(msar.format_data(base."........pg.dropped.3....
    >  ^
    > QUERY: 
    >  WITH
    >  count_cte AS ( SELECT count(1) AS count FROM public.issue_4895 ),
    >  enriched_results_cte AS ( SELECT msar.format_data(id) AS "1", msar.format_data(a) AS "2", msar.format_data(parent) AS "4", NULL AS __mathesar_gid, NULL AS __mathesar_gcount FROM public.issue_4895 ORDER BY "1" ASC LIMIT '5… jsonb_build_array()
    >  )
    >  ,
    >  'count', coalesce(max(count_cte.count), 0),
    >  'grouping', NULL
    >  ) AS rj
    >  FROM enriched_results_cte
    >  LEFT JOIN groups_cte ON enriched_results_cte.__mathesar_gid = groups_cte.id
    >  CROSS JOIN count_cte
    >  )
    >  SELECT records_json_cte.rj || summaries_json_cte.sj
    >  FROM records_json_cte, summaries_json_cte;
    >  
    > CONTEXT: PL/pgSQL function msar.list_records_from_table(oid,integer,integer,jsonb,jsonb,jsonb,boolean,jsonb) line 31 at EXECUTE

wbonicki:
Fixes #4895

The PR fixes bug with listing records if record summary template references a deleted colum.
**Technical details**

Two sql functions were changed:
1. get_column_name - added condition that checks if att is dropped
2. build_record_summary_query_from_template - added checks to skip dropped or non-selectable columns and fixed incorrect privilege validation on the final referenced column (ref_col_id) to prevent invalid SQL generation
**Screenshots**

image
image

## Checklist
 
  index.md`).
- [ ] I added or updated documentation (if applicable). (not applicable in this case)
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/05_msar.sql
+++ b/db/sql/05_msar.sql
@@ -4938,7 +4938,6 @@ BEGIN
             ref_tab_name text;
             alias text;
             join_clause text;
-            ref_col_name text;
           BEGIN
             IF NOT pg_catalog.has_column_privilege(contextual_tab_id, fk_col_id, 'SELECT') THEN
               -- Silently ignore FK columns that we don't have permissions to select.


db/sql/05_msar.sql
```diff
@@ -397,7 +397,7 @@ Args:
   rel_id:  The OID of the relation.
   col_id:  The attnum of the column in the relation.
 */
-SELECT attname::text FROM pg_attribute WHERE attrelid=rel_id AND attnum=col_id;
+SELECT attname::text FROM pg_attribute WHERE attrelid=rel_id AND attnum=col_id AND NOT attisdropped;
```
Anish9901, line 400: Looks good! We also have an overload of this function `msar.get_column_name(rel_id oid, col_name text)` which we don't use as much. But, It would be good to have the behavior consistent for the overload as well.

Also, in the docstring for both the function please mention that we would only return the name of attributes that aren't dropped.

db/sql/05_msar.sql
```diff
@@ -4979,6 +4978,10 @@ BEGIN
         END LOOP;
 
         ref_col_id := ref_chain[ref_chain_length];
+        ref_col_name := msar.get_column_name(contextual_tab_id, ref_col_id);
+        IF ref_col_name IS NULL THEN
+          CONTINUE template_parts_loop;
+        END IF;
```
Anish9901: Seems redundant, we check whether `ref_col_name IS NOT NULL` after a couple of lines. Please revert this.

db/sql/05_msar.sql
```diff
@@ -4926,17 +4926,14 @@ BEGIN
...
             END IF;
 
-            fk_col_name := msar.get_column_name(contextual_tab_id, fk_col_id);
-
-            IF fk_col_name IS NULL THEN
-              -- Silently ignore references to non-existing FK columns. This can happen if a column
-              -- has been deleted.
+            IF NOT pg_catalog.has_column_privilege(contextual_tab_id, fk_col_id, 'SELECT') THEN
               CONTINUE template_parts_loop;
             END IF;
```
Anish9901: What does this change achieve? Also, those comments that you removed are useful to know why the checks were added in the first place, why remove it?

db/sql/05_msar.sql
```diff
@@ -4945,8 +4942,11 @@ BEGIN
...
             IF ref_tab_id IS NULL THEN
-              -- Silently ignore references to non-FK columns. This can happen if the constraint
-              -- has been dropped.
+              CONTINUE template_parts_loop;
+            END IF;
+
+            ref_col_name := msar.get_column_name(ref_tab_id, ref_col_id);
+            IF ref_col_name IS NULL THEN
               CONTINUE template_parts_loop;
             END IF;
```
Anish9901: Since `fk_col` is dependent on `ref_col` via a foreign key constraint, it is not possible to drop `ref_col`:
- Without dropping the foreign key constraint. _or_
- Without dropping the dependent cols, `fk_col` in this case.

So, this check is redundant as we will never encounter this scenario. Please revert this.

db/sql/test_sql_functions.sql
```diff
@@ -7323,3 +7323,50 @@ BEGIN
   );
 END;
 $$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION test_build_record_summary_query_skips_dropped_column()
```
Anish9901: Please see `test_custom_record_summary_template()` or other record_summary tests and write a similar test which checks that the record summaries are returned without errors and the returned value is as expected, even after dropping the column.

Anish9901 (changes requested):

db/sql/05_msar.sql
wbonicki: Fixed! Overloaded function as well as docstrings

db/sql/05_msar.sql
wbonicki: Reverted to the original version

db/sql/05_msar.sql
wbonicki: I cannot remove it, because without this part tests fail (Failed tests:  90, 165, 167, 169 including my test).

db/sql/test_sql_functions.sql
wbonicki: Fixed - please check if now it has correct structure. I also moved the test from the end of the file.

db/sql/05_msar.sql
wbonicki: removed

db/sql/05_msar.sql
Anish9901: Tests failed because `ref_col_name` was removed. Please revert all the changes made in this function(`msar.build_record_summary_query_from_template`). Only the changes you made in `msar.get_column_name` are sufficient to fix #4895.

db/sql/test_sql_functions.sql
```diff
@@ -5414,6 +5414,30 @@ END;
 $$ LANGUAGE plpgsql;
 
 
+CREATE OR REPLACE FUNCTION test_record_summary_after_dropped_referenced_column()
```
Anish9901, line 5449: Nice test!

db/sql/05_msar.sql
wbonicki: Ahh thank you for pointing this out! Fixed!

Anish9901 (approved):
Looks great after the changes, Thanks for your work on this @wbonicki!

PR #4963, opened 2025-11-07
Make table page status pane respond well to narrow viewports

seancolsen:
This PR fixes an issue I've been experiencing while using Mathesar on mobile devices.

## Before

The status pane UI responded very poorly to narrow viewports, especially for tables with lots of records.

image

## After

- The whole status pane responds fluidly to narrowing viewport sizes.

    [Screencast_20251107_121733.webm](https://github.com/user-attachments/assets/7bf8b9cf-0d80-41ae-ba26-605fb977f2c8)

    Different UI hides at different points, with an emphasis on displaying the most important information and hiding what is not strictly necessary.

- The table page uses our newer MiniPagination component instead of the older (wider) `PaginationGroup` component.

- I put the page size configuration _inside the MiniPagination component_. This means that it takes one more click to change the page size. I think that's okay though, given the improvement in visual simplicity.

    image

## Tangentially

- Before this PR, the page size options were 100, 200, 500. I changed this list to 10, 50, 100, 500 because I've been finding that it's helpful to work with smaller page sizes due to the scrolling slowness and due to the workflow problems that I described [here](https://github.com/mathesar-foundation/mathesar/issues/4545#issuecomment-3433370973).

- I also improved the rendering of large numbers, adding locale-aware grouping separators for better readability.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/imports/datafile.py
+++ b/mathesar/imports/datafile.py
@@ -40,8 +40,9 @@ def copy_datafile_to_table(
                 f"{COLUMN_NAME_TEMPLATE}{i}" for i in range(len(next(reader)))
             ]
             f.seek(0)
+        processed_rows = ([None if val == '' else val for val in row] for row in reader)
         import_info = create_and_import_from_rows(
-            reader,
+            processed_rows,
             table_name,
             schema_oid,
             column_names,
--- a/mathesar_ui/src/components/sheet/clipboard/SheetClipboardHandler.ts
+++ b/mathesar_ui/src/components/sheet/clipboard/SheetClipboardHandler.ts
@@ -6,7 +6,7 @@ import { getErrorMessage } from '@mathesar/utils/errors';
 
 import type SheetSelection from '../selection/SheetSelection';
 
-import { MIME_MATHESAR_SHEET_CLIPBOARD, MIME_PLAIN_TEXT } from './constants';
+import { MIME_PLAIN_TEXT } from './constants';
 import { type CopyingContext, getCopyContent } from './copy';
 import { type PastingContext, paste } from './paste';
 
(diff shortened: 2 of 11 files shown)


seancolsen:
@pavish I've assigned you to review code/UX.

@zackkrida I've assigned you for additional UX review, and to loop in anyone else you think is necessary to approve the UX changes here.

pavish:
@seancolsen I don't quite like the MiniPagination for the table page. I believe it reduces usability in desktop browsers.

One of the things I instinctively do is to visit the last page in a table while adding data. Entering a page in a text box is something I hate to do, simply because I find it annoying, and when it comes to UX, I'd choose something difficult over annoying. These preferences are not specific to Mathesar, I tend to do this with any application I use where there are tables with pagination.

I'd be satisfied if we can show the mini pagination (or reduce the number of page buttons displayed) only when the screen width is small.

These views are subjective, I'll wait for others to weigh in to see if they find the MiniPagination satisfactory.

seancolsen:
@pavish, it sounds like your main problem with the MiniPagination UI is that it makes it too hard to jump to the last page. Makes sense!

Instead of showing the old pagination UI, I'm wondering how you would feel about keeping the MiniPagination UI and using some buttons to jump to the first/last pages.

I just implemented this. Here are the results:



Would this address your concerns?

seancolsen:
@pavish and @zackkrida I'd really love it if we could get this change into 0.8.0.

Without this change, viewing a large table on a mobile device is pretty horrendous. That would pair poorly with our public data set initiative, because I would guess that we'd get a decent amount of traffic from mobile devices.

seancolsen:
@zackkrida yes, it's easy to conditionally render different components at different breakpoints. Indeed, I even have the necessary logic in place in this PR to show/hide various UI elements as the status bar gets more narrow. I'm okay with showing the full pagination UI for larger viewports, but the problem is that it can be so incredibly wide that the breakpoint would thus need to be pretty wide if we handle it statically. I have more nuanced and complex thoughts on this matter. But overall it sounds like we need some clarity around the finer points of this design and behavior. I've added it to the agenda for our upcoming engineering team meeting.

pavish:
Here's a suggestion that we discussed at the engineering call today:

Screenshot 2025-11-20 at 8 16 41 PM

This UX shows the current page and the total pages, and clicking the middle button would open the dropdown.

When the screen size is large enough (i.e. desktop), we would also display the current page size selector.

seancolsen:
@pavish and @zackkrida this is ready for re-review now.

Here are the different UI elements, and the order in which they hide as the available space gets smaller:

image

So for a very narrow viewport we have:

image

I think this ought to address everything we discussed in the last meeting about this.

pavish (approved):
Looks great to me!

PR #4975, opened 2025-11-12
Refactor table view system to use string-based column ids

seancolsen:
As discussed in yesterday's [meeting](https://tldv.io/app/meetings/691492dd8e34c70013a29202)

@zackkrida and I did some pairing on this, using Cursor. Then I did extensive review, cleanup, and QA.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/stores/table-data/filtering.ts
+++ b/mathesar_ui/src/stores/table-data/filtering.ts
@@ -42,7 +42,10 @@ export interface FilterEntry {
  * for this compatibility layer.
  */
 function filterEntryToSqlExpr(filterEntry: FilterEntry): SqlExpr {
-  const column: SqlColumn = { type: 'attnum', value: Number(filterEntry.columnId) };
+  const column: SqlColumn = {
+    type: 'attnum',
+    value: Number(filterEntry.columnId),
+  };
 
   /** Generate an SqlLiteral value */
   function value(v = filterEntry.value): SqlLiteral {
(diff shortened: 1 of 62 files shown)


seancolsen:
Call me a Cursor fan now! This would have taken me a lot longer without it!

mathesar_ui/src/stores/table-data/filtering.ts
```diff
@@ -42,7 +42,10 @@ export interface FilterEntry {
  * for this compatibility layer.
  */
 function filterEntryToSqlExpr(filterEntry: FilterEntry): SqlExpr {
-  const column: SqlColumn = { type: 'attnum', value: filterEntry.columnId };
+  const column: SqlColumn = {
+    type: 'attnum',
+    value: Number(filterEntry.columnId),
```
pavish, line 47: I don't quite like the `Number` constructor:
- `Number(null)` and `Number("")` return `0`.
- `Number("randomstring")` is `NaN`

I'd be much more comfortable with a utility method that handles null and empty string cases and throws an error if the string is not a valid number.

Alternatively, the API already accepts numbers as strings, atleast in this case for filters, we could try updating everything to strings.

pavish (changes requested):
It look me a bit longer to review this PR, I tested this quite extensively and haven't run into errors.

My major concern with this PR is the way we cast strings to numbers and vice-versa. At the moment, I didn't find cases where this would cause issues, but I think using the `Number` constructor would come back to hurt us as we move forward.

Additionally, I had expected that we would convert `Column.id` to string when we receive it from the response at the API layer, and supply it as a number when making requests. The changes in this PR handles stringification and parsing at points of usage. I don't have any concerns with this of course, I think both are reasonable approaches. I however, would like to note that the APIs seem to accept column.id as string in requests.

mathesar_ui/src/stores/table-data/filtering.ts
seancolsen: Can we skip this for the sake of moving fast?

pavish: I assumed this would be a quick change to make, nevertheless, I've created https://github.com/mathesar-foundation/mathesar/issues/4982 to keep track to this.

pavish (approved):
https://github.com/mathesar-foundation/mathesar/issues/4982 tracks handling the review comment in this PR. It's not a high priority to block the PR.

PR #4977, opened 2025-11-14
Build UI to select joined columns

seancolsen:
Fixes task: [Build "Related Columns" picker UI](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/2)

## Screenshots

image

image

## Notes

- The checkboxes don't do anything in this PR.
- The list of available checkboxes is rendered after one call to `tables.list_joinable`. I did a tiny amount of backend, adding some additional info to the response in order to prevent the need for more API calls.
- While the API call is resolving, the dropdown renders a loading spinner.
- I added the help text to help avoid user confusion about why _other_ relationships and join paths are not shown here.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
+++ b/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
@@ -1,0 +1,4 @@
+<script lang="ts">
+</script>
+
+<div>TODO</div>
--- a/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
+++ b/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
@@ -1,4 +1,31 @@
 <script lang="ts">
+  import { api } from '@mathesar/api/rpc';
+  import { Spinner } from '@mathesar/component-library';
+  import ErrorBox from '@mathesar/components/message-boxes/ErrorBox.svelte';
+  import type { Table } from '@mathesar/models/Table';
+  import { getErrorMessage } from '@mathesar/utils/errors';
+
+  import JoinConfig from './JoinConfig.svelte';
+  import { getSimpleManyToManyRelationships } from './joinConfigUtils';
+
+  export let table: Table;
+
+  $: joinableTablesPromise = api.tables
+    .list_joinable({
+      database_id: table.schema.database.id,
+      table_oid: table.oid,
+      max_depth: 2,
+    })
+    .run();
 </script>
 
-<div>TODO</div>
+{#await joinableTablesPromise}
+  <Spinner />
+{:then r}
+  {@const simpleManyToManyRelationships = [
+    ...getSimpleManyToManyRelationships(r),
+  ]}
+  <JoinConfig {simpleManyToManyRelationships} />
+{:catch error}
+  <ErrorBox>{getErrorMessage(error)}</ErrorBox>
+{/await}
--- a/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
+++ b/mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
@@ -22,9 +22,7 @@
 {#await joinableTablesPromise}
   <Spinner />
 {:then r}
-  {@const simpleManyToManyRelationships = [
-    ...getSimpleManyToManyRelationships(r),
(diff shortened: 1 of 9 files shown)


seancolsen:
@pavish I've assigned you to review front end code

@Anish9901 I've assigned you to review backend changes

@zackkrida I've assigned you to review UX

mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/joinConfigUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  const targetInfo = targetTableInfo[joinableTable.target];
+
+  const intermediateOid = joinableTable.join_path[0][1][0];
+  const intermediateInfo = targetTableInfo[intermediateOid];
+  const countIntermediateNonPkColumns = Object.values(
+    intermediateInfo.columns,
+  ).filter((c) => !c.primary_key).length;
+  // Intermediate table must have exactly two non-PK columns. We already know
+  // these will be FK columns based on the previous validations of fkey_path.
+  if (countIntermediateNonPkColumns !== 2) return undefined;
```
pavish: Any of the FKs could themselves be PKs, individually or as a composite key. This logic would not hold in such scenarios.

I'm fine with not handling the composite case at the moment, but I think we can handle the individual case and make this a bit more robust by checking the FK columns by their attnum values first, removing them, and then removing any remaining PK columns, and ensure that there are no remaining columns.

mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
```diff
@@ -0,0 +1,38 @@
...
+</script>
+
+{#await joinableTablesPromise}
+  <div class="loading"><Spinner /></div>
+{:then r}
+  {@const simpleManyToManyRelationships = getSimpleManyToManyRelationships(r)}
+  <JoinConfig {simpleManyToManyRelationships} />
+{:catch error}
+  <ErrorBox>{getErrorMessage(error)}</ErrorBox>
+{/await}
```
pavish, line 29: The empty state scenario needs to be handled.

Image

pavish (changes requested):

mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/joinConfigUtils.ts
seancolsen: > Any of the FKs could themselves be PKs, individually or as a composite key

Yeah, good point. I wasn't thinking about composite PKs.

But my understanding from Brent is that the various APIs for this new feature will not handle composite PKs anyway. So I think we actually _don't_ want a mapping table with a composite PK to show up in this list (because it wouldn't be compatible with the other layers of our implementation).

In 0f4ee12f4f5fd57137258829ba89073b26c94ba3 I adjusted the logic slightly, making it more strict. Now it needs the intermediate table to have exactly three columns. And it needs to see exactly one PK column. This is intentionally _not_ future-proof. It's just designed (narrowly) to cover the assumptions we're making for our targeted use case.

mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/Join.svelte
seancolsen: Thanks. Handled in 0cdb7f034abd291a1098c8319d3ea417dc8ed805

seancolsen:
@pavish this is ready for you to re-review now.

Anish9901 (approved):
Backend code looks good!

mathesar_ui/src/systems/table-view/actions-pane/record-operations/join/joinConfigUtils.ts
pavish: Cool, the strict scenario works for the exact usecase we need to solve. We can handle additional usecases, when we receive such requirements.

pavish (approved):

pavish:
I've approved the PR, I'll wait for @zackkrida take a look at the UX before merging this in.

zackkrida (approved):
Looks great @seancolsen, go ahead and merge when you like.

PR #4980, opened 2025-11-17
fixed issue #4753 allowing column reordering in table widget

radheshpai87:
Fixes #4753

This PR ensures that column re-ordering on the table page is immediately reflected in the Table widget on record pages. Previously, the updated `table.metadata.column_order` persisted correctly, but the widget did not update until a full page reload.

## Summary of Fix
-   The parent component (`Widgets.svelte`) now explicitly depends on `$currentTablesData.tablesMap`, ensuring widget inputs recompute when table metadata updates.

-   The `{#each}` key for widget instances now includes the serialized `table.metadata.column_order`. This forces Svelte to remount the `TableWidget` when the column order changes.

-   `TableWidget.svelte` now creates `TabularData` once per widget instance, avoiding stale subscriptions or broken child stores.

## Files Updated
-   **Widgets.svelte**

    -   Added dependency on `$currentTablesData.tablesMap`

    -   Updated widget key to:

        ``${table.oid}-${fkColumn.id}-${JSON.stringify(table.metadata?.column_order ?? [])}``

-   **TableWidget.svelte**

    -   Restored stable initialization of `TabularData` (created once per instance)

## Rationale
Letting Svelte remount the widget on metadata changes is the simplest, safest fix. It ensures fresh `TabularData` with the correct `column_order` without modifying internal data-store behavior.

## Videos
**Before:** <https://github.com/user-attachments/assets/bd2f6ad2-980a-4828-882e-2bb6c2b7bb7c>\
**After:** <https://github.com/user-attachments/assets/0743515f-58ae-4dcf-a49b-3d4dc1d1e3ca>

## Checklist
- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I tried running the project locally and verified that there are no visible errors.

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

[...]

PR #4985, opened 2025-11-18
allow re-ordering columns within table widget

### LINKED ISSUE #4753: Re-ordering columns within table widget does not work
On the table page, when I drag to re-order columns I see the result right away. This is good ✅

https://github.com/user-attachments/assets/49813b6f-8132-44bf-a139-33a81d5fb15f

But in the table widget, when I drag to re-order columns, it appears as though nothing has changed. This is bad ❌

https://github.com/user-attachments/assets/2bcbc850-e117-43a4-b3b2-ddbacfd2a6c0

Interestingly, the column order _does_ get saved though. Refreshing the entire page will show the columns in the new order.

I'm not positive this will be easy to fix, but it might be. I've labeled it "good first issue", but if you're new please understand that it's possible there will be deeper changes required. It's hard to tell without getting into it myself.

radheshpai87:
Fixes #4753 

**Technical details**
Based on previous closed PR #4980 

Previously, when columns were reordered in the main table view, the
table widget on the record page would not reflect the new column order.

This fix adds explicit reactivity tracking for table.metadata.column_order
within the TableWidget component. When the column order changes, the
TabularData instance is recreated with the updated table metadata,
ensuring the widget displays columns in the correct order.

The solution maintains component encapsulation by handling all reactivity
logic internally within TableWidget, without requiring changes to the
parent Widgets component.

## Checklist
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


radheshpai87:
@seancolsen I have tried a different approach which ensures component encapsulation.

radheshpai87:
Hey @seancolsen, just checking in — are my changes okay, or should I make any additional updates?

seancolsen:
@radheshpai87

Sorry, but I'm having trouble deciding how to proceed here.

This does seem to work. But the code has some strange patterns that I'm hesitant to merge. I'd rather not take the time to explain why right now.

I did some work myself towards implementing this fix "the right way", but I ran into a bug in a different part of the codebase that is proving tricky to track down. Overall, I need to make sure I don't spend too much time on this since it's not a high priority for maintainers right now.

I'd like to keep the open and potentially come back to it later.

For what it's worth, the direction I wanted to head was to use this code in `Widgets.svelte`:

```ts
function buildWidgetInput(
  joinableTable: JoinableTable,
  tablesData: TablesData,
) {
  const table = tablesData.tablesMap.get(joinableTable.target);
  if (!table) return undefined;
  const fkColumnId = joinableTable.join_path[0].slice(-1)[0][1];
  const { name } =
    joinableTablesResult.target_table_info[table.oid].columns[fkColumnId];
  return { table, fkColumn: { id: fkColumnId, name, metadata: null } };
}

$: tableWidgetInputs = joinableTablesResult.joinable_tables
  .filter((joinableTable) => joinableTable.multiple_results)
  .map((joinableTable) => buildWidgetInput(joinableTable, $currentTablesData))
  .filter(isDefinedNonNullable)
  .sort((a, b) => a.table.name.localeCompare(b.table.name));
```

Along with sorting out the import for `TablesData`, this is a very small change that I was expecting to fully address the problem. But it's surfacing what seems to be a different bug. I'm not sure why. I'm not even 100% positive this is the best direction to head. Just leaving this comment here so that I can move on. If you choose to follow this lead, please keep in mind that I need to minimize my time spent on this issue, so please resist the urge to ask follow-up questions. If you'd like to push something else that you think will work, then I can take a quick look.

radheshpai87:
Hi @seancolsen 

I've implemented a solution based on your suggestion. 

Changes made:

- Widgets.svelte: Updated buildWidgetInput to accept tablesData as a parameter, making the reactivity explicit as you suggested.

- TableWidget.svelte: Instead of recreating TabularData on every update, I now create it once and override the processedColumns derived store to include currentTablesData as a dependency. This way, when column order changes, only processedColumns re-computes using the updated table metadata, while the rest of TabularData remains intact.

- tables.ts: Exported the TablesData interface.

This approach combines your suggestion with a minimal override to handle column reordering reactively, avoiding the component recreation issues from the previous implementation.

Sorry for the late response :(

seancolsen (approved):
@radheshpai87

Interesting story here...

Your changes still didn't look right to me. Still too much of a hacky band-aid solution. In `TableWidget.svelte` you had basically duplicated a substantial block of code from `tabularData.ts`. And we also don't want to be reaching into a TabularData instance and mutating the `processedColumns` property like that. The TabularData class should "own" that property and have control over how it's computed. Ideally we should probably have `readonly` set on that property (and many more), but that's also true for most properties and basically we usually don't indicate it mostly because we're lazy and it takes up a fair bit of space in the code. (Personally I wish we could tell TS that _all_ class/object properties would be read-only by default, and then we'd mark ones as mutable selectively, but I digress.)

I took a stab at tracing down the "different bug" that the minimal change was surfacing, as I described in my last comment.

It was some code like this in `Body.svelte`:

```
bind:row={$displayRowDescriptors[item.index].row}
```

This code is bad because we binding to a svelte _readable_ store. Yuck! So Svelte was complaining that `displayRowDescriptors` had no `set` method. Makes sense.

Coincidentally, I just recently refactored that bad code out as part of #5097.

So with the merge you just made into this PR, the minimal fix now seems to work quite well!

I reverted your changes to TableWidget.svelte via b2b3c9219f9d4fb33a06272bdd39e82a29332830.

Now this PR is very small, and I'm going to merge it.

PR #4997, opened 2025-11-21
Implement "User" data type

zackkrida:
This PR adds a user data type to Mathesar. This type can be used to track the user who last edited a column via the Mathesar UI. This editing is not tracked outside of the Mathesar UI, so is not currently meant to be used as a robust form of auditing. 

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@mathemancer please review the backend implementation in [`6c34b8c` (#4997)](https://github.com/mathesar-foundation/mathesar/pull/4997/commits/6c34b8ceff4475d759847ab820e97ff61bbaacc8). 

Regarding the frontend, I need to compare this to some overlapping work in cross-table editing to validate the approach, and make this a bit easier to review for the frontend team before getting frontend review.

mathesar/models/base.py
```diff
@@ -239,6 +239,13 @@ class ColumnMetaData(BaseModel):
     duration_max = models.CharField(max_length=255, null=True)
     display_width = models.PositiveIntegerField(null=True)
     file_backend = models.CharField(max_length=255, null=True)
+    user_type = models.BooleanField(default=False, null=True)
+    user_display_field = models.CharField(
+        choices=[("full_name", "full_name"), ("email", "email"), ("username", "username")],
+        max_length=50,
+        null=True
+    )
+    user_last_edited_by = models.BooleanField(default=False, null=True)
```
mathemancer: I find this name really confusing. What about `track_editing_user`?

mathesar/rpc/columns/metadata.py
```diff
@@ -50,6 +53,10 @@ class ColumnMetaDataRecord(TypedDict):
     duration_min: Optional[str]
     duration_max: Optional[str]
     display_width: Optional[int]
+    file_backend: Optional[str]
```
mathemancer, line 56: Oops, nice catch

mathesar/rpc/columns/metadata.py
```diff
@@ -50,6 +53,10 @@ class ColumnMetaDataRecord(TypedDict):
...
+    user_type: Optional[bool]
+    user_display_field: Optional[Literal["full_name", "email", "username"]]
```
mathemancer, line 57: If `user_display_field` is optional, I guess I don't see why `user_type` is needed. If this is non-null, then display the column as specified, otherwise, just display as an integer.

If you really really want to keep `user_type`, I _don't_ think that should be optional since it's always either true or false. Just use `false` for the default.

mathesar/rpc/columns/metadata.py
```diff
@@ -72,6 +79,8 @@ def from_model(cls, model):
             duration_max=model.duration_max,
             display_width=model.display_width,
             file_backend=model.file_backend,
+            user_type=model.user_type,
+            user_display_field=model.user_display_field or "username",
         )
```
mathemancer, line 81: I think `user_last_edited_by` is missing here.

mathesar/rpc/forms.py
```diff
@@ -414,4 +414,7 @@ def submit(*, form_token: str, values: dict, **kwargs) -> None:
         form_token: The unique token of the form.
         values: A dict describing the values to insert.
     """
-    return submit_form(form_token, values)
+    from modernrpc.core import REQUEST_KEY
+    request = kwargs.get(REQUEST_KEY)
+    user = request.user if request else None
```
mathemancer: - Why are you reimporting `REQUEST_KEY`?
- This isn't going to be called without a request, right? If not, there's way too much logic here.

mathesar/rpc/records.py
```diff
@@ -233,16 +342,16 @@ def from_dict(cls, d):
...
+    *,
+    table_oid: int,
+    database_id: int,
+    limit: int = None,
+    offset: int = None,
+    order: list[OrderBy] = None,
+    filter: Filter = None,
+    grouping: Grouping = None,
+    return_record_summaries: bool = False,
+    **kwargs,
```
mathemancer: It seems like your editor really hates the double-indent function args style, but I prefer it. This comment is applicable to many of your PRs (maybe all)

mathesar/rpc/records.py
```diff
@@ -365,26 +487,33 @@ def add(
         The created record, along with some metadata.
     """
     user = kwargs.get(REQUEST_KEY).user
+    # Automatically set user_last_edited_by columns to current user ID
+    _set_last_edited_by_columns(record_def, table_oid, database_id, user.id)
```
mathemancer: This should be in a separate module, and should not modify records in place.

mathesar/rpc/records.py
```diff
@@ -273,27 +382,34 @@ def list_(
...
+    record_info["download_links"] = (
+        get_download_links(
+            kwargs.get(REQUEST_KEY),
+            record_info["results"],
+            download_link_columns,
+        )
+        or None
+    )
+
+    _add_user_display_values_to_record_info(record_info, table_oid, database_id)
```
mathemancer: This should be in a separate module, and should not modify in place.

mathemancer (changes requested):
Partial review so you have stuff to think about:

- Please avoid hiding the mutation of python objects in function calls, as you do in `_add_user_display_values_to_record_info` et al. I'd really prefer the modifications be inline where they're clearly visible; see the addition of `download_links` for example.
- You need to refactor this so that a given RPC call never generates multiple requests for the same Django model object. For example, `records.add` calls `get_columns_meta_data` in both `_set_last_edited_by_columns` and `_add_user_display_values_to_record_info`. The problem is also present in `records.list` since both `get_download_link_columns` and `_add_user_display_values_to_record_info` call the `get_columns_meta_data` function. We either need to cache that metadata getting function for the duration of a request, or (easier) just call it in the main `rpc` function body then use it where needed.

More to come later; late for the team event!

mathesar/rpc/records.py
zackkrida: Yeah I'll disable automatic formatting in vscode for Mathesar. Preferences aside it clutters the diffs too much anyway.

zackkrida: Yeah I'll disable automatic formatting in vscode for Mathesar. Preferences aside it clutters the diffs too much anyway.

mathesar/rpc/columns/metadata.py
zackkrida: Got rid of user_type 👍

mathesar/rpc/columns/metadata.py
zackkrida: Covered by `track_editing_user` instead.

zackkrida:
@mathemancer I addressed your initial feedback some time ago but never requested re-review. This should be good for you to look at once more. 

@pavish, likewise for the frontend. I haven't revisited that work in some time so I imagine there are some aspects that aren't ideal. I'll be available to make changes as you see fit and happy to do any major refactoring. 

@mathemancer one thing I wanted to mention. @kgodey and I discussed that we'd like to hold off on making the user edit tracking and auditing more robust or enforced at the DB level. I could get into reasons why, but the main one is "scope" and "we can do it later if requested". Feel free to reach out to either of us for more details though if you'd like to push back on that.

zackkrida:
@pavish and @mathemancer I finally had some time to give this a deeper review, it's not ready for your reviews yet. There's an obvious issue with migrations and some frontend behavior I'd like to fix ASAP. I'll undraft this when it's ready for review once more.

mathesar/rpc/columns/metadata.py
```diff
@@ -96,6 +104,9 @@ class ColumnMetaDataBlob(TypedDict):
         duration_max: The largest unit for displaying durations.
         display_width: The pixel width of the column.
         file_backend: The name of a backend for storing file attachments.
+        user_display_field: Which user field to display for user columns (full_name, email, or username).
+            If non-null, the column is treated as a user column.
+        track_editing_user: Whether this column automatically tracks the last user who edited the record.
```
mathemancer: This is a bit fishy: What happens if the columns _isn't_ a `user_display_field`, or the underlying type is inappropriate? What happens to `track_editing_user`? Should that logic run, or is it predicated on both being true?

mathesar/migrations/0011_columnmetadata_user_type.py
```diff
@@ -0,0 +1,31 @@
...
+                    ("username", "username"),
+                ],
+                max_length=50,
+                null=True,
+            ),
+        ),
+        migrations.AddField(
+            model_name="columnmetadata",
+            name="track_editing_user",
+            field=models.BooleanField(default=False),
```
mathemancer: Seems like it should not be possible (logically) to set `track_editing_user` to true unless `user_display_field` is true. You can accomplish this with

```sql
-- track_editing_user IS true => user_display_field IS NOT NULL
CHECK (NOT track_editing_user OR user_display_field IS NOT NULL)
```

This would make the data model a bit more robust.

mathesar/rpc/records.py
```diff
@@ -316,6 +333,7 @@ def list_(
         The requested records, along with some metadata.
     """
     user = kwargs.get(REQUEST_KEY).user
+    columns_meta_data = list(get_columns_meta_data(table_oid, database_id))
```
mathemancer, line 336: Good idea combining these DB hits.

mathesar/rpc/records.py
```diff
@@ -327,15 +345,28 @@ def list_(
...
     record_info["download_links"] = get_download_links(
         kwargs.get(REQUEST_KEY),
         record_info["results"],
         download_link_columns,
     ) or None
 
+    # Add user display values for user columns
+    user_summaries = get_user_linked_record_summaries(
+        columns_meta_data, record_info.get("results", [])
+    )
```
mathemancer, line 361: The inconsistency between this and the `get_download_links` call is a bit offputting: In one case we only pass along the columns for which we need the download links, but in the other we pass along all metadata and hope (from this perspective) that the function called is smart enough to only generate summaries for appropriate columns. Given my concerns about this function knowing too much about the definition of `download_link_columns`, I think we should do the latter, so the `get_download_links` function should accept the entire metadata blob.

mathesar/rpc/records.py
```diff
@@ -327,15 +345,28 @@ def list_(
...
             return_record_summaries=return_record_summaries,
-            table_record_summary_templates=get_table_record_summary_templates(database_id),
+            table_record_summary_templates=get_table_record_summary_templates(
+                database_id
+            ),
         )
-    download_link_columns = get_download_link_columns(table_oid, database_id)
+
+    # Add download links for file columns
+    download_link_columns = [c.attnum for c in columns_meta_data if c.file_backend]
```
mathemancer: I don't like this function knowing which attribute constitutes a "download_link_column".

mathesar/rpc/records.py
```diff
@@ -385,12 +416,24 @@ def get(
...
         kwargs.get(REQUEST_KEY),
         record_info["results"],
         download_link_columns,
     ) or None
+
+    # Add user display values for user columns
+    user_summaries = get_user_linked_record_summaries(
+        columns_meta_data, record_info.get("results", [])
+    )
+    if user_summaries:
```
mathemancer: Same problem here.

mathesar/rpc/records.py
```diff
@@ -327,15 +345,28 @@ def list_(
...
         kwargs.get(REQUEST_KEY),
         record_info["results"],
         download_link_columns,
     ) or None
 
+    # Add user display values for user columns
+    user_summaries = get_user_linked_record_summaries(
+        columns_meta_data, record_info.get("results", [])
+    )
+    if user_summaries:
```
mathemancer:
```suggestion
    if user_summaries is not None:
```
protects against falsy but non-iterable outputs of `get_user_linked_record_summaries`.

mathesar/rpc/records.py
```diff
@@ -327,15 +345,28 @@ def list_(
...
+        if record_info.get("linked_record_summaries") is None:
+            record_info["linked_record_summaries"] = {}
+        record_info["linked_record_summaries"].update(user_summaries)
```
mathemancer:
```suggestion
        record_info.setdefault("linked_record_summaries", {}).update(user_summaries)

```
Avoid nested ifs for maximum pythonicity

mathesar/rpc/records.py
```diff
@@ -385,12 +416,24 @@ def get(
...
     ) or None
+
+    # Add user display values for user columns
+    user_summaries = get_user_linked_record_summaries(
+        columns_meta_data, record_info.get("results", [])
+    )
+    if user_summaries:
+        if record_info.get("linked_record_summaries") is None:
+            record_info["linked_record_summaries"] = {}
+        record_info["linked_record_summaries"].update(user_summaries)
```
mathemancer: as above.

mathesar/rpc/records.py
```diff
@@ -424,14 +467,29 @@ def add(
...
         )
+
+    # Add user display values for user columns
+    user_summaries = get_user_linked_record_summaries(
+        columns_meta_data, record_info.get("results", [])
+    )
+    if user_summaries:
+        if record_info.get("linked_record_summaries") is None:
+            record_info["linked_record_summaries"] = {}
+        record_info["linked_record_summaries"].update(user_summaries)
```
mathemancer: As above.

mathesar/utils/user_display.py
```diff
@@ -0,0 +1,99 @@
...
+        if c.user_display_field
+    ]
+    if not user_columns:
+        return None
+
+    linked_record_summaries = {}
+    for column_attnum, display_field in user_columns:
+        user_ids = set()
+        for record in results:
+            user_id = record.get(str(column_attnum)) or record.get(column_attnum)
```
mathemancer: Is this check necessary? If it is, there's a bug in `apply_track_editing_user`: If the column reference is an integer, you'll add a new column in that record with a string key (instead of integer).

mathesar/utils/user_display.py
```diff
@@ -0,0 +1,99 @@
...
+        columns_meta_data: List of column metadata objects (already fetched)
+        user_id: The ID of the current user
+
+    Returns:
+        A new dict with track_editing_user columns set to user_id
+    """
+    result = dict(record_def)
+    for col in columns_meta_data:
+        if col.user_display_field and col.track_editing_user:
+            result[str(col.attnum)] = user_id
```
mathemancer: Potential bug if it's possible for the result keys to be integers.

mathesar/utils/forms.py
```diff
@@ -220,6 +232,66 @@ def submit_form(form_token, values):
             fields_to_pick=fields_to_pick
         )
     ]
+
+    # Set track_editing_user columns for authenticated users
+    if user and user.is_authenticated and hasattr(user, 'id'):
```
mathemancer: All of this logic should be extracted to other functions.

mathesar/utils/forms.py
```diff
@@ -204,9 +205,20 @@ def iterate_form_fields(fields, parent_field=None, depth=0, fields_to_pick=[]):
...
+    Args:
+        form_token: The unique token of the form.
+        values: A dict describing the values to insert.
+        user: Optional Django user object. If provided and authenticated,
+            track_editing_user columns will be set to the user's ID.
+            If None or anonymous, these columns will remain unset (NULL).
+    """
     form_model = Form.objects.get(token=form_token)
     assert form_model.publish_public, 'This form does not accept submissions'
+    values = dict(values)
```
mathemancer: What is this protecting against? Did you encounter errors without it?

mathesar/utils/forms.py
```diff
@@ -220,6 +232,66 @@ def submit_form(form_token, values):
...
+                field_info['table_oid'],
+                field_info['parent_key'],
+                field_info['depth'],
+            )
+            if context not in seen_contexts:
+                insertion_contexts.append({
+                    "table_oid": field_info['table_oid'],
+                    "parent_key": field_info['parent_key'],
+                    "depth": field_info['depth'],
+                })
```
mathemancer: What's the point of both of these objects? Seems like only one or the other should exist. Why not just the set of tuples? What is the `dict` bringing to the table?

mathesar/utils/forms.py
```diff
@@ -220,6 +232,66 @@ def submit_form(form_token, values):
...
+                synthetic_count += 1
+                key = (
+                    f"__track_editing_user_{table_oid}_{column_attnum}_{synthetic_count}"
+                )
+                field_info_list.append({
+                    "key": key,
+                    "parent_key": parent_key,
+                    "column_attnum": column_attnum,
+                    "table_oid": table_oid,
+                    "depth": depth
```
mathemancer: This part makes me realize that `track_editing_user` should be a table property, not a column one. In particular, there should be a choice at the table level of which column we should auto-insert user info into, and the column metadata should purely refer to display style. This has benefits for the user, but also makes most of this logic extraneous.

mathemancer (dismissed):
High-level things I noticed:

- When setting the display from Username to, e.g., email, the UI doesn't update to show the new display. This is contrast to other display styles, and breaks user expectations driven by those other interactions without warning.
- Handles inappropriate values in user-display integer column (e.g., entered manually via psql) gracefully by degrading to integer display. Consider some kind of "broken link" display similar to how we handle file data type columns when the thumbnail is unfetchable. 
- I probably already pestered you about this but please avoid massive formatting overhauls intermingled with logic changes. It's a pain to review. I'd be open to choosing a formatter and running it on the whole python codebase so we can avoid the problem moving forward.
- It occurred to me over the course of reviewing this that the user flow is upside down, and discoverability is reduced by hiding "track editing users" deep in column metadata. Better to have that be table metadata, and have a flow for creating a column to store the info if needed. 

That last point makes it obvious (to me) that this feature was conceived of "feature-first" rather than "use-case-first". If the desired use-case is "track editors of table", then that should be a table setting, and where the editor ID is stored is sort of an implementation detail. The current setup asks a user to stumble across an implementation detail, choose to use that (the User column) without any indication that it's for tracking editors, then discover that they can leverage the type they chose for editor tracking and auditing.

mathesar/utils/forms.py
zackkrida: I really appreciate this suggestion, it dramatically improved the PR's code and user experience.

PR #5007, opened 2025-11-22
Fix: Prevent record selector opening on Shift+click of FK cell chevron (fixes #4913)

### LINKED ISSUE #4913: Shift+click to select multiple cells should not open the record selector when the final click lands on the chevron
## Steps to reproduce

1. Find a table with multiple foreign key cells.
1. Select one FK cell.
1. Use Shift+Click to select a second FK cell. Target your click on the record summary in the second cell.

    Observe that your cell selection is extended from the first cell to the second cell. Nothing else happens. Good. ✅

1. Now Shift+Click on a third FK cell. But this time target your click on the downward-facing chevron at the right of the FK cell.

    Expect the same behavior as in the previous step.

    Instead, observe that the record selector launches. Bad. ❌

    This is annoying because it's almost certainly not what the user intended to do if they're holding down Shift.

We should fix this by _not_ opening the record selector if the chevron click is performed with the Shift key down.

garvitsingh171:
## Description
This pull request fixes a bug where Shift+clicking the chevron icon in a foreign key cell incorrectly opened the record selector modal during multi-cell selection (see issue #4913). With this change, Shift+click now correctly extends the cell selection without showing the modal, making the selection flow consistent with user expectations. Normal (non-Shift) clicks still open the record selector.

## Changes Made
- Updated the click handler for the chevron button in `LinkedRecordCell.svelte` to check for `event.shiftKey` and suppress the record selector if Shift is held.
- Similarly, added the check to `toggleRecordSelector` in `LinkedRecordInput.svelte` so that the input does not trigger the selector on Shift+click.

## Testing Done
- Manual testing: Verified that Shift+clicking on chevron and FK cell does not launch the selector and extends selection instead.
- Regular click still opens the record selector as expected.
- Ran frontend tests (`npm test`) to confirm regression-free behavior.
- Linting passed with `./lint.sh`.

## Impact
This update improves user experience for table cell multi-selection by matching typical spreadsheet behavior, without affecting other data entry or selection workflows. No backend or data logic was affected.

## Related Issue
Fixes #4913

image
image


pavish (approved):
@garvitsingh171 The changes look good.

As mentioned in the issue description, the bug is related to sheet range selection and is only relevant to the LinkedRecordCell component. The LinkedRecordInput component is not used within the sheet, it is used in other places in the app like forms, filters etc.,

I reverted the changes to LinkedRecordInput in https://github.com/mathesar-foundation/mathesar/pull/5007/commits/afeca93bd5b0bf338988f2304639a857f9fd5d5e.

PR #5023, opened 2025-11-23
Mathesar Replace Number() constructors with utility method when parsing column IDs (Fixes #4982)

### LINKED ISSUE #4982: Replace `Number()` constructors with a utility method when parsing column ids
## Description
- Prior to https://github.com/mathesar-foundation/mathesar/pull/4975, column ids were both numbers and strings based on the context. Eg., the table page used numbers and data explorer used strings.
- In https://github.com/mathesar-foundation/mathesar/pull/4975, we decided to move all usage to string-based.
   - We now use the `Number` constructor in several places where columnId is expected to be a number.
- The Number constructor has this behaviour:
  - `Number(null)` and `Number("")` return `0`.
  - `Number("randomstring")` is `NaN`
- While this doesn't cause any visible issues, this might come back to affect us in the future, as it opens a number of possible edge cases, where `null` needs to be supported.

## Changes required
- The `Number` constructor usages for `column.id` should be replaced with a utility which handles the `null`, empty string, and `NaN` scenarios.

navyads138-star:
Fixes #4982

This PR replaces direct usage of `Number()` with the shared utility method for parsing column IDs.  
The goal is to maintain consistent behavior across the codebase and avoid potential parsing issues.

**What was changed**
- Updated all instances where `Number()` was used for column id parsing.
- Replaced them with the project's utility method (e.g., `parseColumnId()` / `toNumber()` — whichever applies in this file).
- Ensures consistent column id parsing logic.

**Why this change is needed**
Using a single utility method avoids:
- Unexpected NaN behavior from Number()
- Code duplication
- Inconsistent parsing logic

**Testing done**
- Verified the app starts without errors
- Checked actions that read or parse column IDs
- Confirmed no type or runtime errors in the updated areas


zackkrida:
Hi @navyads138-star this PR has a number of serious issues that prevent it from being even _considered_ for review by a Mathesar maintainer:

- targeting the master branch instead of develop
- is not using our standard pull request template
- the described changes to not appear to be present at all

Please look at some existing PRs made by maintainers and read our contributor guide to get a better understanding of the requirements for contributing to Mathesar before attempting to contribute again. Thank you.

PR #5026, opened 2025-11-23
Mathesar Replace Number() constructors with utility method when parsing column IDs  (Fixes 4982)

navyads138-star:
This PR replaces all direct usages of Number() for parsing column IDs with the shared utility parsing method used across the Mathesar codebase.

✔ What was changed

Searched for all instances where Number() was used specifically for parsing column IDs.

Replaced them with the project’s standardized utility method:

parseColumnId() (or the appropriate method for that file)

Ensures consistent ID parsing behavior throughout the project.

✔ Why this change is needed

Using Number() directly can lead to:

Unexpected NaN values

Inconsistent behavior across files

Repeated logic that the utility method already handles

Switching to the shared utility function keeps the logic centralized and reliable.

✔ Testing Done

Confirmed the app builds and starts with no errors.

Tested UI interactions that involve reading or parsing column IDs.

Verified that no type or runtime issues occurred after the changes.


zackkrida:
Hi @navyads138-star, this PR changes permissions on nearly every file in Mathesar from 100644 => 100755, making it impossible to review the diff of changes here. 

This also appears to be a duplicate of #5023.

PR #4995, opened 2025-11-21
Simple many-to-many joined columns

pavish:
⚠️  Stacked on top of https://github.com/mathesar-foundation/mathesar/pull/4989 ⚠️ 

Tasks:
- [Add joined column to tabular data](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/7)
- [Implement a cell component for the simple many-many type](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/9)
- [Show selected record/column/cell details in inspector when a joined column cell is selected](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/12)
- [Get context menu options for Rows working in joined columns](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/14)

**Technical details**
- Adds a separate joinedColumns store to tabularData
- Stores joinableTable in tabularData since its used in several places (the join column dropdown, the links section in inspector, the joined columns).
- Handles selection logic for joined columns.
   - Specifies a separate type called 'range-restricted' columns.
   - These columns, when selected, is restricted to one cell per selection.
   - Range selections from outside these columns do not select these columns, similarly range selection cannot start from these columns.
- Displays basic join information in Columns and Cell tab in table inspector.
- Allows resizing of joined columns, but the size is not persisted anywhere.

**Note**
- When editing the cell, the record selector currently opens. This should be replaced with the multi-tagger dropdown.

**Screenshots**
- Screenshot 2025-11-24 at 2 52 54 AM
- Screenshot 2025-11-24 at 2 53 18 AM

## Checklist
 
- [x] My pull request has a descriptive title (not a vague title like `Update
  index.md`).
- [x] My pull request targets the `develop` branch of the repository
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no
  visible errors.

[best_practices]:https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


[...]

PR #5033, opened 2025-11-25
Multi-tagger UI

seancolsen:
⚠️ Stacked on #4995

This PR completes the following tasks:

- [Build multi-select UI with records.list_summaries](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/4)
- [Integrate multi-select summary dropdown with simple-many-to-many cell](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/13)

## Screenshots

image

(@zackkrida I thought this ☝️ make a good pic for the release notes. I made this by rearranging columns and scrolling right.)

## What works

- Integration into the cell
- Data modification via checking/unchecking boxes
- Filtering via search box at top
- Pagination
- Keyboard navigation
- Focus trap
- Loading indicator for checking/unchecking a box
- Layout for short viewports (only the results scroll — search/pagination stays visible)
- Layout for narrow viewport (record summaries are truncated with ellipses)
- Handling duplicate mappings (all mappings are removed when un-checking the box)

## What doesn't (yet?) work

- [Refresh cell when multi-tagger checkboxes change](https://internal.mathesar.org/db/36/schemas/126064/tables/126066/16) (I think we should handle this in a subsequent PR)
- A way for the user to add a new genre and map it to the movie. I think users will want this. It's a little tricky, given that we're searching/listing record summaries, not individual fields. We should chat about how best to implement this if we decide to take it up at some point.
- A way for the user to un-check all boxes or check all boxes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathesar_ui/src/systems/multi-tagger/MultiTagger.svelte
```diff
@@ -0,0 +1,184 @@
...
+
+  function handleKeyDown(e: KeyboardEvent) {
+    const action = new Map([
+      ['Escape', close],
+      ['ArrowUp', selectPrevious],
+      ['ArrowDown', selectNext],
+      ['Enter', toggleSelected],
+    ]).get(e.key);
+    if (!action) return;
+    e.preventDefault();
```
pavish, line 101: I'd have preferred to leverage `ListBox` and update it to support this sort of toggle functionality. We use it for almost every other list like use-case. Even it not, perhaps we could move the common parts of this logic into its own utility.

While its not a priority for this PR, I do think that we shouldn't duplicate this.

mathesar_ui/src/systems/multi-tagger/AttachableMultiTagger.svelte
```diff
@@ -0,0 +1,30 @@
...
+
+  $: ({ triggerElement, multiTagger } = controller);
+</script>
+
+{#if $multiTagger}
+  <AttachableDropdown
+    trigger={$triggerElement}
+    isOpen
+    autoReposition
+    trapFocus
```
pavish, line 17: Pressing `tab` after opening the multi tagger dropdown, leads to a race condition where the focus shifts to the table itself, and the dropdown is still open.

- Open the dropdown.
- Notice that the search bar is focused.
- Press tab.
- The focus is now present in the table (not the active cell), and the dropdown is still open.

mathesar_ui/src/systems/multi-tagger/MultiTaggerController.ts
```diff
@@ -0,0 +1,116 @@
...
+  currentRecordPk: ResultValue;
+  intermediateTable: {
+    oid: number;
+    attnumOfFkToCurrentTable: number;
+    attnumOfFkToTargetTable: number;
+  };
+  targetTable: {
+    oid: number;
+    pkColumnAttnum: number;
+  };
```
pavish, line 40: Since we're reconstructing joinPath for the request, it might be better to pass down the joinPath directly from the cell and calculate intermediateTable and targetTable in the controller.

mathesar_ui/src/systems/multi-tagger/MultiTagger.svelte
```diff
@@ -0,0 +1,184 @@
...
+
+<style lang="scss">
+  div[data-multi-tagger] {
+    max-width: min(30rem, 90vw);
+    max-height: inherit;
+    overflow: hidden;
+    position: relative;
+    display: grid;
+    grid-template-rows: auto 1fr auto;
+  }
```
pavish, line 154: I find the dropdown height extending till the end of the view port frustrating since the dropdown always ends up displaying on the right/left of the cell instead of below it.

Image

I prefer it appearing below the cell, similar to the row seeker.

Image

While this is subjective and I'm willing to update my views based on your reasoning, I do think that both these components need to present a similar UX, whichever we choose.

pavish (approved):
@seancolsen 

The code looks good and this works pretty well, nice job!

I had a few comments, none of them are blocking merging this PR in, and they can be discussed/handled in separate PRs.

pavish:
I'm merging this PR in.

@seancolsen I've added tasks in our internal checklist for the items from my review.

@zackkrida Please feel free to review post-merge. I think it might be more effective to review the full UX in the feature PR: https://github.com/mathesar-foundation/mathesar/pull/5035, once all tasks are complete.

PR #5036, opened 2025-11-26
Table-view: Enable right-click context menu on rows and correct rendering structure

mosesacknals139:
File-Mathsar
This PR implements the complete client-side behavior required for activating and properly rendering the row context menu within Row.svelte. The goal is to bring the Table View’s row-level interactions in line with standard data-grid UX patterns while enabling new actions such as Duplicate Record.

Key Additions & Logic

Added an on:contextmenu event listener to each row to enable right-click support.

Ensures that when a user right-clicks a row, that row becomes selected before the menu opens.

This guarantees consistent action behavior, especially when multiple rows are already selected.

Implemented logic to prevent the default browser context menu and show the custom context menu instead.

Added support for opening the menu at the mouse cursor position for better spatial clarity.

Structural Fixes

Moved the <ContextMenuWrapper> outside the main sheet container.

This change avoids layout clipping issues caused by overflow rules within the sheet.

Ensures that absolute positioning behaves correctly and the menu is never cut off.

Allows the context menu to be freely positioned anywhere in the viewport.

Addresses several subtle UI conflicts related to stacking context and containment that were preventing the menu from appearing consistently.

Feature Enablement

This refactor exposes the new “Duplicate Record” action directly from each row.

Also supports the full range of row actions (edit, delete, etc.) now and in future enhancements.

Provides consistent behavior whether acting on a single row or multiple selected rows.

UX Improvements

Right-clicking now behaves exactly as users expect in a table-based interface.

Multi-select support is fully respected: right-clicking any selected row opens the menu scoped to the selection rather than breaking it.

The interaction model is now aligned with common spreadsheet, database, and file-manager conventions.

Why This Matters

These changes significantly improve the usability of the Table View, fix previously inconsistent context menu behavior, and establish a stable foundation for more advanced row actions. This also ensures Mathesar feels familiar and intuitive to users coming from other database tools.
File-Mathsar


Anish9901:
Hey @mosesacknals139, please follow our PR template while writing PR descriptions, you can look at how the descriptions should look from one of the previously merged PRs.

Some things that should be fixed before we can review your PR:
- Resolve conflicts
- Fix PR description(make it match our PR template)
- Fix failing test(if any)

Also, where are these screenshots from? These screenshots don't look like they are taken from Mathesar, and why are they duplicated?

PR #5039, opened 2025-11-27
Fix/3784 send null on empty

### LINKED ISSUE #3784: Send `null` for string-like columns that don't allow empty strings when cell is cleared

dheena017:
Fixes #3784

### Summary
This PR fixes a backend validation error (`CheckViolation`) that occurred when clearing cells for string-like column types that do not accept empty strings (Date, DateTime, Time, Email, URI).

Previously, clearing these cells sent an empty string `""` to the backend. The database rejected this because these types require `NULL` instead of an empty string. This change updates the frontend formatters to correctly convert empty input into `null` before the patch is sent to the API, matching the existing behavior of the `Duration` type.

**Technical details**
* **New Formatters:** Created `EmailFormatter.ts` and `UriFormatter.ts` in `src/utils/` to handle the specific normalization logic for these types.
* **Updated Formatters:** Modified `DateFormatter`, `TimeFormatter`, and `DateTimeFormatter` to explicitly check for empty strings and return `{ value: null }` in the `parse` method.
* **Component Wiring:** Updated the component factories in `src/components/cell-fabric/data-types/` (specifically `email.ts` and `uri.ts`) to use `FormattedInput` with the new formatters, replacing the generic text input.
* **Configuration:** Updated type configurations for Email and URI to point to the correct abstract types (`type: 'email'`, `type: 'uri'`) instead of `string`.
* **Typing:** Updated TypeScript definitions to ensure `InputFormatter` correctly handles `null` return types without errors.

**Screenshots**
###Berfore:
Screenshot from 2025-11-27 23-07-53

###After:
Screenshot from 2025-11-27 23-14-41

## Checklist

<details>

Developer Certificate of Origin Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors. 1 Letterman Drive Suite D4700 San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.

(a) The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or

(b) The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or

(c) The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.

(d) I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.

</details>


mathesar_ui/src/components/cell-fabric/data-types/uri.ts
```diff
@@ -1,14 +1,26 @@
...
 
 const uriType: CellComponentFactory = {
-  initialInputValue: '',
-  get: (): ComponentAndProps => ({ component: UriCell }),
-  getInput: (): ComponentAndProps => ({ component: TextInput }),
-  getDisplayFormatter: () => String,
+  get: (): ComponentAndProps => ({
+    component: TextBoxCell,
+    props: {},
+  }),
```
pavish: The existing cell component should not be changed from UriCell to TextBoxCell. Why was this change made?

mathesar_ui/src/utils/email/EmailFormatter.ts
```diff
@@ -0,0 +1,24 @@
...
+  ParseResult,
+} from '@mathesar-component-library/types';
+
+// We use <string | null> to allow null values without red lines
+export class EmailFormatter implements InputFormatter<string | null> {
+  parse(input: string): ParseResult<string | null> {
+    const cleanedInput = input.trim();
+
+    // THE FIX: If empty, return NULL
+    if (!input || cleanedInput === '') {
```
pavish: This check is superfluous. Just having `!cleanedInput` should suffice.

mathesar_ui/src/components/cell-fabric/data-types/email.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// Standard Display Component
+import { EmailFormatter } from '../../../utils/email/EmailFormatter';
+
+import TextBoxCell from './components/textbox/TextBoxCell.svelte';
+import type { CellComponentFactory } from './typeDefinitions';
+
+// Import your fixed formatter logic
+
+// --- THE FIX: We define the object without importing the broken type file ---
```
pavish: What broken type file?

mathesar_ui/src/components/cell-fabric/data-types/components/typeDefinitions.ts
```diff
@@ -207,3 +210,19 @@ export interface ArrayCellProps
...
+  ) => ComponentAndProps;
+  getInput: (
+    column: RawColumnWithMetadata,
+    config?: Record<string, unknown>,
+  ) => ComponentAndProps;
+  getSimpleInput?: (column: RawColumnWithMetadata) => ComponentAndProps;
+  getDisplayFormatter: (
+    column: RawColumnWithMetadata,
+  ) => CellValueFormatter<unknown>;
+}
```
pavish, line 228: What is the purpose of adding this type? Why not use the same type definitions used in other cell type files?

mathesar_ui/src/components/cell-fabric/data-types/email.ts
```diff
@@ -0,0 +1,33 @@
...
+    component: TextBoxCell,
+    props: {},
+  }),
+
+  // EDIT MODE: Use FormattedInput (from library) + EmailFormatter
+  getInput: (): ComponentAndProps => ({
+    component: FormattedInput,
+    props: {
+      formatter: new EmailFormatter(),
+      placeholder: 'Enter email address',
```
pavish, line 15: Placeholders should not be specified for this input.

mathesar_ui/src/components/cell-fabric/data-types/email.ts
```diff
@@ -0,0 +1,33 @@
...
+// Import your fixed formatter logic
+
+// --- THE FIX: We define the object without importing the broken type file ---
+
+const emailType: CellComponentFactory = {
+  // VIEW MODE: Show standard text display
+  get: (): ComponentAndProps => ({
+    component: TextBoxCell,
+    props: {},
+  }),
```
pavish: TextBoxCell doesn't fix the issue. The cell would still send `''` instead of `null` when editing.

mathesar_ui/src/components/cell-fabric/data-types/uri.ts
pavish: Also TextBoxCell does not fix the issue when editing in the sheet, it should still send an empty string instead of null.

mathesar_ui/src/utils/email/EmailFormatter.ts
```diff
@@ -0,0 +1,24 @@
+import type {
+  InputFormatter,
+  ParseResult,
+} from '@mathesar-component-library/types';
+
+// We use <string | null> to allow null values without red lines
+export class EmailFormatter implements InputFormatter<string | null> {
```
pavish, line 7: This should be a common formatter within `formattedInputUtils` and be just an object (instead of class) since there is no state.

mathesar_ui/src/utils/uri/UriFormatter.ts
```diff
@@ -0,0 +1,21 @@
+import type {
+  InputFormatter,
+  ParseResult,
+} from '@mathesar-component-library/types';
+
+export class UriFormatter implements InputFormatter<string | null> {
```
pavish, line 6: This can be a common formatter within `formattedInputUtils`.

pavish (changes requested):
@dheena017 The fix here is incomplete and does not address the issue entirely, it also introduces regressions for the cell types such as UriCell.

I'm marking this PR as a draft. Please re-request review after all the concerns mentioned here are addressed.

mathesar_ui/src/components/cell-fabric/data-types/uri.ts
mosesacknals139: Thanks for catching that. The temporary switch to TextBoxCell was an exploratory attempt to route empty strings through a formatter, but it was the wrong layer and didn’t fix sheet editing. I’ve kept the component as UriCell and applied the fix in the formatter instead: the Uri formatter now trims and returns null on empty input, so sheet edits send null, not an empty string. The same empty→null handling exists for email/date/time/datetime. I’ll run Prettier to satisfy CI.

dheena017:
Closing this PR for now.

Issue #4514, opened 2025-05-28
Uninformative error when deleting a record fails due to FK


kgodey:
## Description

When you try deleting a record that has a record from another table linking to it, the operation fails with no usable information.

<img width="351" alt="Image" src="https://github.com/user-attachments/assets/05816f3a-d6a4-415a-94fe-454fa128b576" />

## Expected behavior

The error should explain what failed.

## To Reproduce

- Create Table A with a record X
- Create Table B with a column with FK to Table A
- Create record in Table A and record linking to it in Table B
- Attempt to delete record in Table A

## Additional context

The API response does include the type of error, so this seems like a frontend issue.

```json
{
	"id": 0,
	"jsonrpc": "2.0",
	"error": {
		"code": -30087,
		"message": "ForeignKeyViolation: update or delete on table \"Ideas\" violates foreign key constraint \"Idea Category Map_Idea_fkey\" on table \"Idea Category Map\"\nDETAIL:  Key (id)=(3) is still referenced from table \"Idea Category Map\".\nCONTEXT:  SQL statement \"\n    WITH delete_cte AS (DELETE FROM \"Product Planning\".\"Ideas\" WHERE (msar.format_data(id))::text IN (SELECT jsonb_array_elements_text('[3]')) RETURNING *)\n    SELECT count(1) FROM delete_cte\n    \"\nPL/pgSQL function msar.delete_records_from_table(oid,jsonb) line 13 at EXECUTE"
	}
}

```

## Note

Eventually we should allow people to actually delete any dependent records too, but for now, making the error more informative would help.

## Similar

- #2202
- #4177

kgodey:
@zackkrida assigned to you to figure out what to do with it.

kgodey:
This seems like a duplicate of #4177, I did not realize I'd already opened this issue.

Anish9901:
I think the next step here is to just show the error message sent by the backend instead of "Error".

PR #5049, opened 2025-11-30
Fix: handle FK violation error in row deletion (issue #4514)

kamalakiruthi8:
Summary

This PR improves error handling in the RowActions component by showing a clear, user-friendly message when a record cannot be deleted because it is referenced by another table (foreign key constraint).

Previously, if a user attempted to delete such a record, Mathesar displayed a generic error message that didn’t explain the actual cause. This fix provides better clarity and improves UX.

Problem

When deleting a row that is referenced by another table, the backend raises a PostgreSQL ForeignKeyViolation error.
But the UI was only showing the generic toast.fromError(e) message, which was confusing for users.

What This PR Does

✔ Detects FK violation errors using:

if (e?.error?.code === -30087 || e?.error?.message?.includes("ForeignKeyViolation"))

✔ Shows a custom, descriptive error toast:

Title: “Cannot delete due to FK”

Message: “Record is referenced by other table”

✔ Falls back to the previous error handler for other error types.

Code Changes

In RowActions.svelte:

Replaced the old onError: (e) => toast.fromError(e) with custom FK error handling logic.

Added localized messages using $_ translation function.

Ensured the rest of the delete flow (onProceed, onSuccess) remains unchanged.

How to Test

Create two tables in Mathesar:

Parent table with primary key.

Child table with a foreign key referencing parent.

Add a row in both.

Try deleting the parent row.

Expected result:
✔ A toast appears:
Cannot delete due to FK
Record is referenced by other table

Try deleting a normal row without FK constraint — should still delete normally.

Result

Users now get a clear, correct error message when FK constraints prevent deletion. This improves clarity and helps users understand why the deletion is blocked.


Anish9901:
This is a duplicate of #5050, @kamalakiruthi8 please don't open duplicate PRs.

PR #5050, opened 2025-11-30
Show informative error when row deletion fails due to foreign key violation (#4514)

kamalakiruthi8:
Summary

This PR improves the error feedback shown to users when attempting to delete a record that is referenced by another table.
Previously, Mathesar displayed a generic "Unable to delete row. Error" toast with no actionable information.
Now, when the backend reports a ForeignKeyViolation error (code -30087), the UI displays a clear message explaining why the row cannot be deleted.

Problem / Root Cause

When deleting a record, the backend correctly returned an error payload indicating a foreign key violation, but the frontend was not interpreting this specific error type.
As a result:

The UI fell back to a generic error handler

No helpful information was shown to the user

Users could not understand why the delete action failed

This was tracked in issue #4514.

What This PR Changes
✔ Added explicit handling for FK violation errors in RowActions.svelte

Detects backend error via either:

err.error.code === -30087, or

err.error.message.includes('ForeignKeyViolation')

Shows a meaningful and localized error toast:

Title: cannot_delete_due_to_fk

Message: record_is_referenced_by_other_table

✔ Replaced incorrect usage of $_() inside object with proper subscription

The i18n store _ must be dereferenced ($_) before passing into toast; it cannot be called as a store.

✔ Cleaned up the error handler:

Added proper TypeScript type for onError: (e: unknown)

Removed unused variables

Ensured no trailing spaces and proper Prettier formatting

✔ Updated code to meet:

ESLint rules

TypeScript type-checking

Svelte formatting guidelines

How to Test

Create Table A with some records

Create Table B with a column referencing Table A (FK)

Insert a row in Table B that references a row in Table A

Attempt to delete that referenced row from Table A

Observe:

The new, descriptive FK error toast is displayed

No generic error appears


Anish9901:
Hey @kamalakiruthi8, Can you add in-app screenshots of how the error looks with your changes?

Issue #4847, opened 2025-10-03
Support `sslmode` and `sslrootcert`


salim-b:
## Description

By default, Mathesar tries an ***unencrypted*** connection to external Postgres DBs, which of course fails if the Postgres server mandates TLS connections:

<img width="2220" height="268" alt="Image" src="https://github.com/user-attachments/assets/960b8e27-7b17-4cff-8713-202b64de8315" />

A user has to manually set the environment variable `PGSSLMODE=require` on the Mathesar server in order to secure Postgres connections. There is no field in the UI to configure [`sslmode`](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE).

Furthermore, there is no way to make Mathesar validate the full certificate chain up to the CA, i.e. use `sslmode=verify-full` (env var `PGSSLMODE=verify-full`) and set [`sslrootcert`](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) accordingly (e.g. via env var `PGSSLROOTCERT=/my_ca.pem`).

## Expected behavior

- Mathesar should properly respect the environment variable [`PGSSLROOTCERT`](https://www.postgresql.org/docs/current/libpq-envars.html).
- Mathesar should support setting the Postgres connection options `sslmode` and `sslrootcert` via its UI.
- Mathesar should default to `sslmode=require` and only try an unencrypted connection as a fallback.

## To Reproduce

Connect Mathesar to any Postgres server that mandates TLS encryption, for example the managed offering by [aiven.io](https://aiven.io/docs/products/postgresql).

## Environment
 - Mathesar version: 0.6.0
 - OS: Debian 12.12 (official Mathesar OCI image)
 - Browser: *irrelevant*
 - Browser Version: *irrelevant*
 - Other info:

PR #5062, opened 2025-12-02
Make SSL mode configurable for internal and external databases

zackkrida:
Related to #4847

**Technical details**

This PR makes the ssl mode for Postgres connections configurable via environment variable (for internal db connections) and via the UI (for connecting to existing data databases). 

This also allows the ssl mode to be edited for existing DBs.

This PR does **not** support configuring the `PGSSLROOTCERT` or the various `verify-` ssl modes. That's a more advanced setting that adds complexity for typical users, so we should only add it based on user requests for the missing functionality. 

**Screenshots**

Screenshot From 2025-12-02 15-14-25
Screenshot From 2025-12-02 15-14-09

Preview of new docs:

image

image

## Testing Instructions

To make this easier to test, I set up a gist for running a DB with SSL required locally. You should be able to try to connect to this in the Mathesar UI and see an error when you set SSL mode to disabled:

https://gist.github.com/zackkrida/dda14152cb5182982542085690d64373

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@mathemancer here's an alternative to #4981 that I think will be better for users as it allows them to set the ssl mode per server and edit the ssl mode, rather than hardcode a single ssl mode for all connections. 

@seancolsen, the UI changes here are pretty minimal but please review them nonetheless.

mathesar/utils/permissions.py
```diff
@@ -34,21 +34,25 @@ def set_up_new_database_for_user_on_internal_server(
         raise BadInstallationTarget(
             "Mathesar can't be installed in the internal database."
         )
+    # Use sslmode from internal database config if available
+    sslmode = getattr(conn_info, 'sslmode', None) or 'prefer'
```
mathemancer: I'd prefer `try/except` here. Also, did you run into a case where `sslmode` _isn't_ an attribute of `conn_info`? I don't see how that's possible. Finally, why not just use `'prefer'` as the default?

mathesar/utils/permissions.py
```diff
@@ -62,7 +66,8 @@ def set_up_new_database_for_user_on_internal_server(
 
 @transaction.atomic
 def set_up_preexisting_database_for_user(
-        host, port, database_name, nickname, role_name, password, user, sample_data=[]
+        host, port, database_name, nickname, role_name, password, user,
```
mathemancer: We typically aren't wrapping in this style in this project. Either one line for the whole args list, or separate lines for each name (my preference), please.

mathesar/utils/permissions.py
```diff
@@ -87,9 +93,17 @@ def set_up_preexisting_database_for_user(
...
-        host, port, database_name, nickname, role_name, password, user
+        host, port, database_name, nickname, role_name, password, user,
+        sslmode='prefer'
 ):
-    server, _ = Server.objects.get_or_create(host=host, port=port)
+    server, _ = Server.objects.get_or_create(
+        host=host, port=port,
+        defaults={"sslmode": sslmode}
+    )
+    # Update sslmode if the server already exists but sslmode differs
```
mathemancer: Remove this comment, please.

mathesar/utils/permissions.py
```diff
@@ -34,21 +34,25 @@ def set_up_new_database_for_user_on_internal_server(
         raise BadInstallationTarget(
             "Mathesar can't be installed in the internal database."
         )
+    # Use sslmode from internal database config if available
```
mathemancer: Remove this comment please.

mathesar/utils/permissions.py
```diff
@@ -87,9 +93,17 @@ def set_up_preexisting_database_for_user(
 
 @transaction.atomic
 def _setup_connection_models(
-        host, port, database_name, nickname, role_name, password, user
+        host, port, database_name, nickname, role_name, password, user,
+        sslmode='prefer'
 ):
-    server, _ = Server.objects.get_or_create(host=host, port=port)
+    server, _ = Server.objects.get_or_create(
```
mathemancer: Couldn't this just be an `update_or_create` ? The pattern below is a bit unwieldy.

mathemancer (changes requested):
I made some specific comments. Mostly around style, but I'd much rather use the `update_or_create` function where I noted.

Also, there's a lot of comments explaining simple logic. Please remove those to reduce clutter.

zackkrida:
Thanks for the review @mathemancer. 

I've switched to `update_or_create`, removed all the redundant comments--a habit I abhor from others but still do myself at times when writing python--, fixed the formatting, and made 'prefer' the default in `PostgresConfig` so we can just access `conn_info.sslmode` directly.

seancolsen (approved):
Docs changes and front end changes look good!

- I pushed df2167b7b0e515f0580b434a94a64bee2395df2d to address a small user-facing issue.

    I like the markdown documentation you wrote about this new setting! So I changed the in-app help text to link out to the user guide. I think this means we don't need quite as much content in the help text because readers can get that in the guide by following the link if they need it.

    Before:

    

    After:

    


- I pushed 300c4e894e047d6c984bf1758e69e5ffe57dacc5 to address a small developer-facing issue.

    In `SelectSslMode.svelte` you had:

    ```ts
    $_(option.labelKey)
    ```

    We've been actively avoiding using dynamic i18n keys like this. And (to my knowledge) all i18n function arguments across the codebase are currently static strings. Why? We have some tooling that assumes static strings. If you look at `check_unused_strings.py` you'll see that we're searching the codebase for static strings inside translation functions. This helps us identify orphaned strings in `en/dict.json`.

    This is essentially an undocumented code standard. I really wanted us to have an i18n developer guide immediately after implementing the i18n system, but we've not chosen to prioritize work on that. Oh well.

    While fixing this code standard, I also refactored the way the options were defined. The code is now a bit simpler.

I didn't review backend changes.

zackkrida:
Those changes look great; thanks @seancolsen!

mathemancer (approved):
Okay, looks good to go from my end.

PR #5068, opened 2025-12-03
Improve datetime parsing for loose timestamp formats

### LINKED ISSUE #5060: Try more date and time formats when parsing temporal values on the front end
## Steps to reproduce

1. Set up a table like this:

    Image

    Verify all UI matches with the arrows.

1. Open your browser dev tools network tab.

1. Copy the string:

    ```
    2000-01-01 09:00:00
    ```

1. Edit the `date_time` cell. While in edit mode, paste the string and press <kbd>Enter</kbd> to save.

1. In your dev tools, observe that Mathesar sends something like the following:

    ```
    2000-01-01 09:00:00 -05:00
    ```

    That `-05:00` bit at the end is the timezone, as determined by your browser. It might be different for you depending on what timezone you're in.

    This is good. The front end doesn't just tack this on at the end. The front end actually parses your entire date entry and then reformats it in a format that PostgreSQL can interpret unambiguously.

1. Add a new record and edit the `date_time` cell, this time pasting the following value:

    ```
    2000-01-01 9:00:00
    ```

    Notice that the only thing we've change here is that we've removed the leading 0 before the 9.

1. This time, observe that Mathesar sends exactly the string you entered.

1. And (unless you're in UTC timezone) PostgreSQL will actually send back will be a different time, making the displayed cell different from you you entered. This is bad.

## What's going on (and how to fix it!)

1. On the front end, find the `DateTimeFormatter` class, and look at the `parse` method. You can see some code comments describing how this logic works (and why).

1. We need to make the `parseWithSpec` free function more robust so as to handle input like the format in the above repro steps. The bulk of your changes will probably end up being in `DateTimeSpecification.ts`.

1. You can tighten your feedback loop on this parsing function by adding so console logging in `DateTimeFormatter.parse` and observing the output as you type into the input on the record page:

    Image

1. It would be good to try to think of additional date/time formats that we can parse in addition to the one listed above.

## Additional context

- This bug was originally described in #4911

Ahamed1846:
Fixes #5060

This PR enhances date/time parsing so that Mathesar correctly handles valid timestamp inputs that omit leading zeros or seconds. Previously, entries such as "2000-01-01 9:00:00" or "2000-01-01 09:00" were not parsed and were sent to PostgreSQL unchanged, causing unexpected timezone shifts.

### Problem

`parseWithSpec` matched only strict timestamp formats and rejected variants without leading zeros or without seconds. When parsing failed, the unparsed string was sent directly to PostgreSQL, resulting in incorrect values for timestamptz columns.

### Fix

Additional timestamp format variants were included in
`DateTimeSpecification` so that:

- Hours without leading zero (`H:mm:ss`, `H:mm`)
- Timestamps without seconds (`YYYY-MM-DD HH:mm`)
- Timestamps using single-digit hour (`YYYY-MM-DD H:mm:ss`)

are now correctly parsed and converted to canonical form.

### Behavior Changes

- Valid timestamp inputs with missing leading zeros now parse correctly.
- Values are consistently normalized before being sent to the backend.
- No changes were made to date-only parsing or time-only behavior.

### Files touched

- `mathesar_ui/src/utils/date-time/DateTimeSpecification.ts`
- `mathesar_ui/src/utils/date-time/DateTimeFormatter.ts`

---

## Checklist
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


seancolsen (approved):
This seems to work well.

Ideally I'd like the code to be cleaned up a bit and have some tests now that this logic is getting more complex. And ideally I'd like to also parse times like `5pm`. But we can handle that separately if/when needed. This is a net improvement and doesn't seem to introduce any regressions.

Thanks for your work on this @Ahamed1846!

PR #5076, opened 2025-12-06
Fix 1758 constraint validation

### LINKED ISSUE #1758: Fix erroneous behavior of the constraint endpoint while invalid column ids are passed
## Description

The following results in a 500 status code.
- Attempting to set a unique constraint by supplying the id of a column that exists in a different table from the table whose id is supplied in the URL.
- Non-existent column id, e.g.:

    ```json
    {"type":"unique","columns":[9999]}
    ```

## Expected behavior

The above should result in a 400(Bad Request) response.

## Additional context

See #889, #1406

mosesacknals139:
Summary

Fixes #1758
This PR adds validation to prevent invalid column IDs from reaching create_constraint() which previously caused a 500 Internal Server Error.
Now, when the user sends an invalid column id (ex: column does not exist in table), the RPC returns a 400 Bad Request with IntegrityAPIException, instead of crashing.

What was changed

Valid column IDs are collected using get_column_info_for_table(table_oid)

Before calling create_constraint(), each column is checked

If an invalid ID is found, IntegrityAPIException is raised and request stops early

Prevents database failure and the bug mentioned in reviewer feedback

Files Modified

✔ mathesar/rpc/constraints.py

✔ mathesar/api/exceptions/database_exceptions/base_exceptions.py

(Attach your screenshot here)

Checklist

 Changes tested locally

 Only required files modified

 No breaking behavior for valid constraints

 Fix prevents 500 error and returns meaningful response


Anish9901:
Hey @mosesacknals139, thanks for your interest in #1758, but this issue is no longer relevant. Were you able to successfully able to reproduce the behavior described in the issue before creating a PR? I'm quite sure 500s aren't possible with the way our RPC endpoints work.

PR #5097, opened 2025-12-10
Refactor cell updating logic to fix bugs

### LINKED ISSUE #4889: Weird bugs with cell value state when editing
Here are two (very likely related) bugs:

- A

    1. Begin with a unique column
    1. Edit cell, setting value to an existing value in the column.
    1. Observe error. Good.
    1. Refresh sheet via bottom right button. Notice error disappear. Good.
    1. Edit same cell again with same value.
    1. Expect to see the error as before.
    1. Instead, observe no error. Bad ❌
    1. Refresh via bottom button again. Expect the value to reset. Observe it to still display in the cell, even when selecting other cells. This is _really_ bad! ❌
    1. Refreshing the whole page resets the value.

- B

    1. Make or find a "Time" cell.
    1. Verify that the "Time Format" for the column is set to "Short time 24 hr"
    1. Double-click the cell to enter edit mode.
    1. Manually enter the text "7:00 PM" and press <kbd>Enter</kbd>.
    1. Observe that the cell saves and displays as "19:00". Good.
    1. Double-click to enter edit mode again.
    1. Observe that the cell still displays as "19:00". Good.
    1. Press <kbd>Escape</kbd> to exit edit mode.
    1. Expect the cell to still display as "19:00".
    1. Instead, observe that the cell now displays as "7:00 PM". Bad ❌
    1. Select another cell and observe that the original cell returns to displaying "19:00". Good. But weird!!

To any community contributors considering picking this up: I expect this issue to be tricky. A fix will most likely involve a fair bit of refactoring and require a good understanding of the specific ways in which Svelte handles state.

### LINKED ISSUE #4892: Exiting cell edit mode by clicking on another cell should select the second cell
## Steps to reproduce

1. Enter edit mode on a table cell
1. Click on another cell in the table
1. Expect the second cell to be selected.
1. Instead, observe that the first cell is selected.

This appears to be a regression introduced by #4822. So in fixing this bug, please ensure that we don't end up with #4751 again.

seancolsen:
Fixes #4892
Fixes #4889

## Overview of changes

- **Two-way binding removed**

    **Previously**, we were using Svelte's two-way data binding to pass cell changes up from low-level cell components _way_ up to higher level sheet components like `Body.svelte`. To me, two-way data binding seems appropriate for low-level, synchronous operations like form inputs. But it seems ill-suited for high-level asynchronous operations like saving cell values to a sheet. I think this was a key factor contributing to the bugs.

    **Now** a large swath of components is refactored to use `value` and `setValue` props instead of `bind:value`. This makes it a bit easier to reason about state and avoid bugs.

- **Code path consolidation**

    **Previously**, in RecordsData, we had substantially separate code paths for handling updates via paste vs updates made to single cells via edit mode. This was tech debt leftover from when we implemented paste.

    **Now** we use `RecordsData.bulkUpdate` for everything.

- **Optimistic UI improved**

    **Previously**, we were applying optimistic UI updates to cells, but only by virtue of the two-way data binding. This meant that the optimistic UI didn't apply to paste operations.

    **Now** the optimistic UI is implemented within `RecordsData.bulkUpdate` — a much more centralized location for this sort of logic.

## Notes

- @pavish [said](https://github.com/mathesar-foundation/mathesar/issues/4889#issuecomment-3593383089): 

    > Since the stores held the entire record, updating the store re-renders the entire record. This prevented that re-render and allowed a better UX, when user is updating multiple cells one after the other, where a request completing in the previous cell did not affect the current cell that's being edited.

    I've tested this scenario (with heavy throttling) and it seems to work well. You can see how I'm handling it by looking at the `shouldStoreNewValue` variable inside `postProcessRecordRow`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

    maintained indefinitely and may be redistribued consistent with
```

</details>


mathesar_ui/src/stores/table-data/records.ts
```diff
@@ -692,10 +696,31 @@ export class RecordsData {
...
+          if (blueprint.cells.some((r) => r.columnId === columnId)) return true;
+
+          // Ignore any _other_ loading cells. This is to handle the case where
+          // a user is updating multiple cells in one record faster than the
+          // network can respond. For example if the user updates cell A, then
+          // updates B in the same row, and _then_ we receive the response for
+          // cell A, we don't want to mutate cell B with the response from
+          // update A because that would wipe out the optimistic UI updates that
+          // we did at the start of operation B.
+          if (cellsLoading.has(cellId)) return false;
```
pavish: This doesn't work because `cellsLoading` contains `cellKey` which is different from `cellId`. We need to update `cellModificationStatus` to use `cellId`.

Notice the `t2` column in the video below:

https://github.com/user-attachments/assets/78a84b13-3f2c-46bb-ba19-c893f27cd9c3

mathesar_ui/src/stores/table-data/records.ts
pavish: This erases all the statuses for every column in the row. We need to only update the statuses for cells that are modified and not ones that have a request in process.

mathesar_ui/src/stores/table-data/records.ts
```diff
@@ -715,124 +740,43 @@ export class RecordsData {
...
-      // eslint-disable-next-line no-console
-      console.error('Error refetching row', err);
-      return row;
-    }
+    const updatedRecord = result.results[0];
+    this.fetchedRecordRows.update((rows) =>
+      rows.map((r) =>
+        r.identifier === rowId ? r.withRecord(updatedRecord) : r,
+      ),
+    );
```
pavish: This method should also use the same logic from the `bulkUpdate` to only update the cells which do not have a request in process.

mathesar_ui/src/systems/table-view/row/RowCell.svelte
```diff
@@ -85,31 +90,21 @@
...
+        : { modificationRecipes: [{ row, cells }], additionRecipes: [] },
+    );
+    if (isWithinPlaceholderRow) {
+      // Re-focus the cell just edited. The placeholder row requires this extra
+      // logic because it gets a new id value after it's saved, and that ends up
+      // wiping out the fact that a cell in that row was selected.
+      const newRowId = $newRecords[$newRecords.length - 1].identifier;
+      selection.update((s) =>
+        s.ofRowColumnIntersection([newRowId], [columnId]),
+      );
```
pavish, line 107: This causes the selection to jump a bit, but the selection not persisting bug is present in the previous release as well.

I remember implementing a logic where when a placeholder row is saved, the identifier in the new PersistedRow or DraftRow is the same, which helped the selection to remain in the same place.

It looks like that behaviour was changed when we modified the selection handling for the PlaceholderRow where we retain the same identifier for the placeholder row (`placeholderRowId.set(placeholderRow.identifier);` within `display.ts`).

We ought to take this case into account when we refactor the records.ts file.

pavish (changes requested):
@seancolsen I like the direction in this PR.

Please find my comments on scenarios where multiple requests are fired in parallel (eg., when a new request for refreshRow or bulkDml happens before a previous request resolves).

We handled these kind of scenarios by cancelling previous requests when a new request is made on the same row. This ensured that the outdated promise didn't resolved and doesn't trigger any updates.

Now, we have a different logic in place where we check the values and partially update stuff after a request resolves and I understand this is the only approach we can take for bulk updates. We would have to have a common block where we handle this scenario for any kind of request that updates a record.

mathesar_ui/src/systems/table-view/row/RowCell.svelte
pavish: I don't think this needs to be handled in this PR.

pavish:
A GH action seems to have triggered this commit:
[chore(docs): update TOC](https://github.com/mathesar-foundation/mathesar/pull/5097/commits/917df751863966502a8fcfdfc48321cc958c492a) https://github.com/mathesar-foundation/mathesar/pull/5097/changes/917df751863966502a8fcfdfc48321cc958c492a, when merging develop into this branch

pavish:
@seancolsen I pushed a commit that fixes the bugs I've mentioned in my review w.r.t multiple requests updating the same cell: https://github.com/mathesar-foundation/mathesar/pull/5097/commits/bf19d5a2cb754fb4ec50aac2658510ae1578adee

- I'm now using a combination of an id based check and the cellStatus for the case we previously used promises.cancel to handle (the case is specified in the PR description notes).
- The status indicator is fixed, which was resetting for all cells of a row when any request for that row succeeds.
- The refreshRow method does not update/merge any cell that's being processed.

With this, I'm inclined to merge this PR in. We still need to create a separate issue [for this](https://github.com/mathesar-foundation/mathesar/pull/5097#discussion_r2619881685).

I'll let you take a look at the PR before merging.

mathesar_ui/src/stores/table-data/records.ts
pavish: Resolved in https://github.com/mathesar-foundation/mathesar/commit/bf19d5a2cb754fb4ec50aac2658510ae1578adee

mathesar_ui/src/stores/table-data/records.ts
pavish: Resolved in https://github.com/mathesar-foundation/mathesar/commit/bf19d5a2cb754fb4ec50aac2658510ae1578adee

mathesar_ui/src/stores/table-data/records.ts
pavish: Resolved in https://github.com/mathesar-foundation/mathesar/commit/bf19d5a2cb754fb4ec50aac2658510ae1578adee

pavish (approved):

seancolsen:
@pavish and I did some pairing on this now and we're both approved.

Issue #4890, opened 2025-10-21
Unable to rename PK column named `id`


seancolsen:
## Steps to reproduce

1. Create a new table from scratch.
1. Use the table inspector to rename the `id` column to `pk`.
1. Expect the column to be renamed.
1. Observe that the column is not renamed and no error message is displayed.

This appears to be a backend bug somewhere below the `columns.patch` API. That API gives a success response in this case when in fact the column was not renamed.

mathemancer:
Correction: This only affects columns which are both primary keys, _and_ named `id`. I've updated the issue title accordingly. For example, if one creates a table on the DB:

```sql
CREATE TABLE testable2 (
  mypk integer PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY, 
  col1 text
);
```
You can change that primary key column to be named `pk`, or even `id`. But, if you change it to be named `id`, you can't change it anymore after that.

As with a bug in a recent PR (#4880), this goes back to the concept of a "default Mathesar ID column". See that issue for related discussion.

I still think it's probably worth taking a comprehensive look at whether we really need that concept anymore. 

As a short-term solution, we should just issue an error when a user attempts to `ALTER` a Mathesar default ID column. I think in the long term, they should be allowed to modify it as they wish, but that may have other effects.

wbonicki:
Hi, can I be assigned to it? I managed to run the app locally and I would like to help!

seancolsen:
@wbonicki we currently do not assign issues to community contributors, but we still encourage you to work on this and submit a PR!

wbonicki:
Excellent, therefore I will take care of it.

UPDATE: I am working on it, I am very close to fix it, but I need some more time.

wbonicki:
Here is the PR:
https://github.com/mathesar-foundation/mathesar/pull/4931

Two notes:
1. SQL functions need to be reinstalled on running instance.
2. No specific code formatting style was found, so the SQL code was not formatted according to any particular convention.

If anything else is needed please let me know!

wbonicki:
@mathemancer there is an additional issue with the PK column. During table creation you can assign a different name, so shouldn’t this be addressed as well?

<img width="502" height="495" alt="Image" src="https://github.com/user-attachments/assets/4b3b509e-44fd-4810-8a8d-a219fe68a835" />

[...]

PR #5167, opened 2026-01-08
Revert "issue 4890 - Allow renaming of primary key (Mathesar ID) columns"

### LINKED ISSUE #5158: Unable to create table from data import
## Description

While creating a table through data import, clicking on the "Confirm & create table" fails to create the table and gives the following error:
```
CheckViolation: Mathesar ID column cannot be renamed
SQLSTATE: 23514
```

This happens when the Primary Key Column strategy is set to `Add a new primary key column` and the generated `id` column is the first column in the Table Preview (default behavior).

Image

## Expected behavior

Clicking on the `Confirm & create table` button on the Import Confirmation page should not produce the above error, and should create a table with the generated `id` column as the Primary Key column and the rest of the columns according to the imported data.

## To Reproduce

1. Navigate to any schema page within Mathesar
2. Click on `New Table` and then `From Data Import`
3. Either upload a CSV file, or simply paste the following into the `Copy and Paste Text` section (or any data for that matter):
```
number
123
```
4. Click on `Proceed`
5. On the Import Confirmation page, click on `Confirm & create table`

## Environment
 - Mathesar version: 0.8.0
 - OS: Arch Linux x86_64 (Linux 6.17.7)
 - Browser: Zen Browser (based on Firefox)
 - Browser Version: 1.17.6b (Firefox 144.0.2)

## Additional context

I believe this is related to #4890 

I have also discovered a workaround for this issue. After clicking on `Confirm & create table`, refreshing the confimation page anytime afterwards (even after going out of the page and returning to it) creates the table as originally expected from clicking on the button and redirects to the table page. This behavior does seem weird, however, and might be related to the problem behind this issue.

shamblonaut:
Fixes #5158 

Reverts mathesar-foundation/mathesar#4931 as suggested by @Anish9901

- The Mathesar schemas need to be reinstalled in the respective databases for the fix to be observed.
- Reverting #4931 also reintroduces the problems mentioned in #4890 (primarily the silent failure).

### DIFF
--- a/db/sql/05_msar.sql
+++ b/db/sql/05_msar.sql
@@ -3774,8 +3774,6 @@ CREATE OR REPLACE FUNCTION
 msar.alter_columns(tab_id oid, col_alters jsonb) RETURNS integer[] AS $$/*
 Alter columns of the given table in bulk, returning the IDs of the columns so altered.
 
-Exception is raised when mathesar ID column is tried to be renamed.
-
 Args:
   tab_id: The OID of the table whose columns we'll alter.
   col_alters: a JSONB describing the alterations to make.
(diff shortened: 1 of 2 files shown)


mathemancer (approved):
LGTM. ~~Separately, I'll re-open the issue that the troublesome PR fixed.~~

Actually that issue's already still open!

PR #5185, opened 2026-01-20
Fixes issue #5184

### LINKED ISSUE #5184: Disconnect/uninstall fails when msar schema owner role name requires quoting (e.g. mathesar-user on Azure PG) — connect_admin() uses regrole::text and tries to login as "role"
Hi Mathesar team

I’m running Mathesar on Kubernetes against Azure Database for PostgreSQL (Flexible Server) and I’m hitting a disconnect/uninstall failure that appears to be caused by how connect_admin() resolves the “admin role” from the msar schema owner.

What happens

When I try to disconnect a database (even with “Remove Mathesar’s internal schemas” unchecked), the request ends up calling Database.uninstall_sql() and then Database.connect_admin(). connect_admin() determines the admin role by running:

```
SELECT nspowner::regrole::text
FROM pg_catalog.pg_namespace
WHERE nspname='msar';
```

On my instance, the role is named mathesar-user (hyphenated), so Postgres returns the identifier with quotes:

- admin_role_name (print): "mathesar-user"
- admin_role_name (repr): '"mathesar-user"'

That value is then used both to look up the role in ConfiguredRole and as the login username. This causes two failure modes:

1. ConfiguredRole lookup mismatch (if roles are stored unquoted):

- ConfiguredRole.DoesNotExist → NoAdminConnectionAvailable

1. If I create a ConfiguredRole entry named "mathesar-user" to satisfy the lookup, Mathesar then attempts to connect via psycopg using the literal username including quotes, which fails:

- FATAL: password authentication failed for user ""mathesar-user""

So the root cause seems to be that regrole::text returns a quoted identifier for role names that require it, and that quoted form is not suitable for (a) matching ConfiguredRole.name as stored and/or (b) passing into psycopg as the username.

Environment
    •   Deployment: Kubernetes (official Docker image)
    •   Backend DB: Azure Database for PostgreSQL Flexible Server
    •   Role used by Mathesar: mathesar-user (DB owner)
    •   Schemas ownership (confirmed while connected as mathesar-user):
    •   msar, __msar, mathesar_types owned by mathesar-user
    •   public owned by azure_pg_admin (expected on Azure)

Logs / error

Disconnect triggers:
    •   NoAdminConnectionAvailable (when ConfiguredRole lookup fails)
    •   or psycopg.OperationalError: password authentication failed for user ""mathesar-user"" (when storing the quoted role name)

Reproduction steps
    1.  Create a Postgres role with a hyphenated name (e.g. mathesar-user) and use it to install Mathesar SQL objects so it owns schema msar
    2.  Ensure Mathesar stores a ConfiguredRole with name mathesar-user (no quotes)
    3.  Attempt to disconnect database / uninstall schemas
    4.  Observe connect_admin() resolves admin role as "mathesar-user" and subsequent failures

Expected behavior

Disconnect/uninstall should work regardless of whether the installing role name requires identifier quoting (hyphens, uppercase, etc.). If the role is stored as mathesar-user, Mathesar should connect using username mathesar-user (no quotes).

Suggested fixes

Any of these would solve it cleanly:

1. Avoid regrole::text quoted identifier output by joining pg_roles:

```
SELECT r.rolname
FROM pg_catalog.pg_namespace n
JOIN pg_catalog.pg_roles r ON r.oid = n.nspowner
WHERE n.nspname = 'msar';
```

2. Or normalize the string returned by regrole::text before using it:

Strip wrapping quotes and unescape "" to "

Happy to provide more logs / details if needed. Thanks!

paschembri:
Fixes https://github.com/mathesar-foundation/mathesar/issues/5184


pavish (approved):
@paschembri Thank you for identifying and fixing this bug! The fix works great.

PR #5191, opened 2026-01-24
Replace deprecated functions in Url64 methods

### LINKED ISSUE #5183: Usage of deprecated functions in Url64 util methods
## Description

In the [`Url64`](https://github.com/mathesar-foundation/mathesar/blob/develop/mathesar_ui/src/utils/Url64.ts#L16) class, the `encode` and `decode` methods use the deprecated `unescape` and `escape` functions. This was done to support encoding unicode strings (See #2680). 

I believe those functions were used according to the [recommendations in the MDN docs at the time](https://web.archive.org/web/20230302121832/https://developer.mozilla.org/en-US/docs/Glossary/Base64#solution_1_%E2%80%93_escaping_the_string_before_encoding_it). But the docs have since been [updated with a new approach](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings) which recommends using `TextEncoder.encode` and `TextDecoder.decode` methods.

### Affects users

**None**

### Affects developers

**Low**

Although the `escape` and `unescape` functions are deprecated, as mentioned in #2680, it is unlikely that they are removed because of backward-compatibility of browsers.

## Suggested change

Using the new approach suggested in MDN, we can replace the usage of the deprecated functions by encoding the string with `TextEncoder`, converting the bytes into a binary string, and then passing it to `btoa()`. The process can be reversed for decoding (get the binary string from `atob()`, convert to bytes, and then decode using `TextDecoder`).

Code from MDN docs:
```js
function base64ToBytes(base64) {
  const binString = atob(base64);
  return Uint8Array.from(binString, (m) => m.codePointAt(0));
}

function bytesToBase64(bytes) {
  const binString = Array.from(bytes, (byte) =>
    String.fromCodePoint(byte),
  ).join("");
  return btoa(binString);
}

// Usage
bytesToBase64(new TextEncoder().encode("a Ā 𐀀 文 🦄")); // "YSDEgCDwkICAIOaWhyDwn6aE"
new TextDecoder().decode(base64ToBytes("YSDEgCDwkICAIOaWhyDwn6aE")); // "a Ā 𐀀 文 🦄"
```

shamblonaut:
Fixes #5183 

This PR changes the implementation of the `encode` and `decode` methods in the `Url64` util class, to use native encoding and decoding using `TextEncoder` and `TextDecoder` respectively with byte arrays (according to the [recommendation in MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings)), instead of the previous implementation which used deprecated functions like `escape` and `unescape`.

**Technical details**

The new implementation encodes the string using [TextEncoder's `encode()` function](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode), whose byte array result is converted into a binary string to be passed into `btoa()`. The decoding mechanisms follows the same principle, taking the binary string from `atob()`, converting it into a byte array, and then getting the decoded string with [TextDecoder's `decode()` function](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode).

This implementation deviates a bit from the MDN docs as *char codes* were used instead of *code points* during the byte array conversions. This was done for the code to work well with typescript rules, and also since char codes were more suitable for byte values.

**Demo**

The following link shows a table comparing the resultant values of the previous and new implementations (V1 and V2 respectively, to show that their results are identical): https://jsfiddle.net/yn6d1tpw/2/

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar_ui/src/utils/Url64.ts
+++ b/mathesar_ui/src/utils/Url64.ts
@@ -9,7 +9,7 @@ export default class Url64 {
   static encode(s: string): string {
     const bytes = new TextEncoder().encode(s);
     const binString = Array.from(bytes, (byte) =>
-      String.fromCodePoint(byte),
+      String.fromCharCode(byte),
     ).join('');
     const base64 = btoa(binString);
 
@@ -20,7 +20,7 @@ export default class Url64 {
     const base64 = s.replace(/-/g, '+').replace(/_/g, '/');
 
     const binString = atob(base64);
-    const bytes = Uint8Array.from(binString, (m) => m.codePointAt(0));
+    const bytes = Uint8Array.from(binString, (m) => m.charCodeAt(0));
     return new TextDecoder().decode(bytes);
   }
 }
--- a/mathesar_ui/src/utils/Url64.ts
+++ b/mathesar_ui/src/utils/Url64.ts
@@ -4,23 +4,23 @@
  * The resulting string won't have any characters which require URL encoding. We
  * change `+` to `-` and `/` to `_`. We remove the padding characters `=`
  * altogether because atob only needs them when inputs are concatenated.
- *
- * TODO:
- *
- * - Refactor this code to remove use of deprecated function `escape` and
- *   `unescape`.
- * - See: https://developer.mozilla.org/en-US/docs/Glossary/Base64
- * - Consider using a 3rd party library. Maybe `base64-js`, `base64url` or
- *   `utf8`.
  */
 export default class Url64 {
   static encode(s: string): string {
-    const base64 = btoa(unescape(encodeURIComponent(s)));
+    const bytes = new TextEncoder().encode(s);
+    const binString = Array.from(bytes, (byte) =>
+      String.fromCodePoint(byte),
+    ).join('');
+    const base64 = btoa(binString);
+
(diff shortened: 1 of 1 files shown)


shamblonaut:
I had also noticed that the documentation pages for [`atob()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) and [`btoa()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa) recommended using [`Uint8Array.fromBase64()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) and [`Uint8Array.prototype.toBase64()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64) respectively in case `Uint8Array`s are used.

These functions additionally allow options to be passed in which include support for URLs and removing padding (which is implemented separately in this case). I decided to not go ahead with that implementation as the baseline implementation of those functions were only widely available since 2025.

I have included that implementation in the [comparison demo](https://jsfiddle.net/yn6d1tpw/2/) along with links to the MDN docs and CanIUse pages.

PR #5266, opened 2026-05-12
[Experimental] Auto provisioned role db

mathemancer:
This PR adds functionality to automatically create a DB and role for a new user when created via a call to `users.add` or via a new SSO signin.

**Technical details**

If the environment variable `AUTO_HOME_DB_AND_ROLE` variable is set to `true`, then the following happens when `users.add` is called or a new SSO signin occurs:

1. A new user is added to the Django model table.
2. A role with the same name as the username of the new user is created on the default internal DB server.
3. A database is created on the same server with the same name.
4. The user is automatically mapped to use that role on that database in Mathesar.

This is in addition to any other DBs and mappings created for new users. 

CAVEAT: The `User` model had to be modified to reduce the allowed length from 150 to 63 characters.

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


mathemancer:
@pavish Marking this as "Draft" so you can build off of it if desired. Otherwise, I can mark it for review.

@Anish9901 take a look if you have time.

pavish:
@mathemancer I have a few concerns to note:
- Users can currently update their username via the UI.
- When logging in with SSO, the username is assigned automatically by django-allauth, which might resort to using the email as the username, which could cause bugs with the current implementation, if it truncates.
 
I think we should be expanding the username field to 254 characters (as per the RFC) to support emails as usernames. It might be best to generate a unique hash or a short name as the db name and the role name, and have a mapping to it. As of now, identifying the role for the user is purely incidental.

Even if we don't have a mapping table (for now), we can only store in `user_database_role`. The only additional work required would be to generate a unique hash/short-name/slug like https://pypi.org/project/unique-namer/, but it would still require some collision avoidance mechanism. I'd even be comfortable with an UUID or an integer id with a custom sequence.

pavish:
@mathemancer @Anish9901 I've added one additional commit to this PR, that's not specific to this one.

https://github.com/mathesar-foundation/mathesar/pull/5266/commits/768fd2ced23b93caefd007d3510bb384d0742210: Fixes a regression.
- After our security fixes for login endpoints, where we verify user db role map using database_id, admin users are no longer able to access those endpoints without being a collaborator.
- This prevents them from being able to add themselves as a collaborator on a db, which was a behaviour we explicitly allowed in previous Mathesar versions.
- This becomes prominent with auto created dbs where only the user is the collaborator and admins cannot add themselves as collaborators via their admin UI.
- The commit fixes that by allowing superusers access endpoints without the user db role map check.

pavish:
@Anish9901 @kgodey I'm going to merge this PR in. Please feel free to review retroactively.

PR #5284, opened 2026-05-22
Fix public schema permissions

Anish9901:
I decided to add all the catalog table calls like `pg_attribute`, `pg_constraints` etc, to be name-spaced under `pg_catalog` since someone can create a table with the name `pg_attribute` or `pg_constraints` in the `public` schema which would cause the non name-spaced catalog calls from `msar` funcs to use the tables in `public` schema instead of `pg_catalog` which would break the app.

Additional context: https://www.enterprisedb.com/blog/new-public-schema-permissions-postgresql-15

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/db/sql/50_msar_permissions.sql
+++ b/db/sql/50_msar_permissions.sql
@@ -30,7 +30,6 @@ $$ LANGUAGE plpgsql;
 
 
 GRANT USAGE ON SCHEMA __msar, msar, mathesar_types TO PUBLIC;
-GRANT CREATE ON SCHEMA public TO PUBLIC;
 GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA msar, __msar, mathesar_types TO PUBLIC;
 GRANT SELECT ON ALL TABLES IN SCHEMA msar, __msar, mathesar_types TO PUBLIC;
 SELECT msar.grant_usage_on_custom_mathesar_types_to_public();
--- a/db/sql/50_msar_permissions.sql
+++ b/db/sql/50_msar_permissions.sql
@@ -30,6 +30,7 @@ $$ LANGUAGE plpgsql;
 
 
 GRANT USAGE ON SCHEMA __msar, msar, mathesar_types TO PUBLIC;
+GRANT CREATE ON SCHEMA public TO PUBLIC;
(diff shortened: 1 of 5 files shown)


db/sql/50_msar_permissions.sql
```diff
@@ -30,6 +30,7 @@ $$ LANGUAGE plpgsql;
 
 
 GRANT USAGE ON SCHEMA __msar, msar, mathesar_types TO PUBLIC;
+GRANT CREATE ON SCHEMA public TO PUBLIC;
```
pavish: @Anish9901 Mathesar should not be automatically modifying privileges on objects it doesn't own. This will affect our self-hosted users with pre-existing databases. Also, depending on the role Mathesar uses, it might not have the access required to do this, which will cause this to fail.

We should only grant access to public specifically for new databases created within Mathesar, when PER_USER_DATABASES_ENABLED is true. This is the only case where we create the db as a different user and automatically change ownership without an explicit user action.

pavish (changes requested):

db/sql/50_msar_permissions.sql
Anish9901: > Mathesar should not be automatically modifying privileges on objects it doesn't own. This will affect users with pre-existing databases. 

Yeah, I thought about that but it was the default anyway before PG < 15 which also persists through upgrades if not manually changed. So, I decided to go ahead with it.

> depending on the role Mathesar uses, it might not have the access required to do this, which will cause this to fail.

That's real and annoying, changing trigger only when PER_USER_DATABASES_ENABLED is set.

Anish9901:
It turns out that managed postgres in Azure has different settings than postgres default settings for public schema ownership, and that's why we were running into this issue.

More context here: https://learn.microsoft.com/en-us/azure/postgresql/security/security-access-control#public-schema-ownership-changes-in-azure-database-for-postgresql

pavish (approved):
@Anish9901 Looks good!

pavish:
This was accidentally merged into master, since the base changed from the release branch to master. It doesn't affect any flows. Here's the PR that merges master into develop, after this PR got merged in: https://github.com/mathesar-foundation/mathesar/pull/5296

PR #5311, opened 2026-06-09
Handle anonymous users in Common Data getter

### LINKED ISSUE #5310: Non-logged-in users get 500 rather than 404 for nonexistent paths.
## Description

When `DEBUG=false` (but not when it's `true`), Mathesar returns a 500 when a non-logged-in user requests a nonexistent path.

## Expected behavior

It should return 404. The 500 errors are triggering maddening numbers of alerts.

## To Reproduce

Set `DEBUG=false` in your `.env` and start the dev environment. Log out if necessary. Paste

`http://localhost:8000/aslkjbasdlfkj` 

into your url bar.

## Additional context

This is swamping our alerts for 500 error rates on the cloud.

mathemancer:
Fixes #5310 

Moves logic for determining which common data to return to its proper place within `get_common_data`, so it's consistently applied.

**Technical details**

The underlying problem causing #5310 was that the 404 handler was calling `get_common_data` rather than `get_anonymous_common_data`. The design of the common data getters meant the caller had to know which to call. After this change, the getter dispatches based on whether the user is logged in or not. This solves the issue, and probably other random bugs that may have arisen from trying to get a common data blob for which some info was privileged.

NOTE TO REVIEWERS:

The issue #5310 only occurred in prod, or at least when `DEBUG=false`. So, verification that this fixes the problem needs to be done in those situations. However, when `DEBUG=false`, but you're running in the dev docker environment, the static files don't really work (unrelated to these changes), so the best way to test this is to build a prod image with

```
docker build -t mathesar/mathesar:latest .
```

and then run the app in prod mode with `docker compose up`.

Then, navigate to `http://localhost/aslkfjasflkj`

Note that after this fix, the back end returns 404, but the front end does not render anything because it _never_ renders `index.html`. I think this is a net improvement, and it does solve the linked issue, but we should consider figuring out what we want to actually render in cases where non-logged-in users navigate to a non-existent path and implement that (someday).

## Checklist
 
  index.md`).
  visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/mathesar/tests/views/test_page_not_found_view.py
+++ b/mathesar/tests/views/test_page_not_found_view.py
@@ -1,0 +1,69 @@
+"""
+Tests for the custom 404 handler (page_not_found_view).
+
+Verifies that:
+- Non-logged-in users get a 404 (not 500) when requesting a nonexistent path.
+- Authenticated users also get a 404.
+- Anonymous users receive only minimal context (no sensitive data).
+- Authenticated users receive full context.
+"""
+from unittest.mock import MagicMock
+
+import pytest
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from django.test.utils import override_settings
+
+from mathesar.views import get_common_data, page_not_found_view
+
+NONEXISTENT_PATH = '/nonexistent-path-xyz/'
+TEST_STORAGES = {
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
+    },
+}
+
+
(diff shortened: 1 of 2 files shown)


mathemancer:
@pavish I requested review from you so you can think about whether we want to add some 404 page for non-logged-in users to this PR, or just merge it as-is.

pavish (approved):
Changes look good!

